PR #275, opened 2014-02-23
PSX: Support cue files

gizmo98:
http://en.wikipedia.org/wiki/Cue_sheet_(computing)


Diskutant:
Showing .cue files in the frontend isn't a good idea imho.
Those files contain information about the tracklist of mixed CDs (Data track + Audio tracks), those are read by the emulator, but the frontend shouldn't show those files as you need at least a .bin file next to the .cue meaning you show those games twice in the frontend.

joolswills:
This is an old pull request - if you have a bug open a new ticket - or better still for this please discuss on the forum.

PR #547, opened 2015-01-02
Configuration for Logitech F710 wireless gamepad

ProxyCell:
This is a popular gamepad that is sold in game stores and at PC shops.
I believe this is the "best" controller from Logitech at this time. Here is the official page for it: http://gaming.logitech.com/en-us/product/f710-wireless-gamepad

Here it is for sale online: http://www.amazon.com/Logitech-940-000117-Gamepad-F710/dp/B0041RR0TW

The controller is wireless (2.4GHz) and needs two AA batteries. It is styled to look a lot like a Playstation Dual Shock controller and has a vibration function. This controller even has a MODE button that switches the d-pad and analogue sticks.

It also has a switch for XInput/DirectInput modes. You can play this on most of the latest games and also some of the very oldest with your computer.

For windows OS it has software that lets you create custom profiles for the controller map.


mzeyrek:
Just had a fresh install and i can see  supplementary/RetroArchConfigs/LogitechGamepadF710.cfg located but not recognized with my F710 as it has been recognized as Logitech_Logitech_RumblePad_2_USB.cfg So wont be an issue i believe device ID's are different as my hesitation was to override my settings. Case is solved :)

PR #549, opened 2015-01-03
reworked ini/config setting to reduce copy/paste and make the calling simpler

joolswills:
Since the most common usage is to change a few configuration values at once, we now have a iniConfig function that sets up the delimiter type (such as "=" or " = "), the value quote character (often "" nothing or "\"" a single quote) and the filename. you can then just call "iniSet" or "iniUnset" with two parameters for key and value. A third parameter can be used in case you wish to use the current ini settings but on a different file. We should now easily be able to handle any key/value configs whilst keeping the code simple/compact

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -25,6 +25,18 @@
 #  Many, many thanks go to all people that provide the individual packages!!!
 #
 
+function printMsg()
+{
+    echo -e "\n= = = = = = = = = = = = = = = = = = = = =\n$1\n= = = = = = = = = = = = = = = = = = = = =\n"
+}
+
+function fatalError()
+{
+    printMsg "Fatal Error"
+    echo "$1"
+    exit 1
+}
+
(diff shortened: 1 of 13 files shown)


joolswills:
Doing this one via a pull request as it is a larger change and affects some core functionality. I have tested every file changes to make sure they work with the changes.

scriptmodules/helpers.sh
```diff
@@ -44,15 +56,27 @@ function addLineToFile()
     echo "Added $1 to file $2"
 }
 
-# arg 1: set/unset, arg 2: delimiter, arg 3: quote character, arg 4: key, arg 5: value, arg 6: file
-function iniSet()
+# arg 1: set/unset, arg 2: delimiter, arg 3: quote character
```
petrockblog: Is this comment still up to date?

joolswills: thanks. c&p fail. fixed.

PR #551, opened 2015-01-03
switch scummvm to build from source

joolswills:
enabling:
 * virtual / on screen keyboard
 * keymapper gui for mapping keyboard etc
 * newer build than debian repository (1.7.0)

rework the scummvm emulationstation config. we launch it from a script now (in roms/scummvm/) that will show up in emulation station too (and act
as a option to launch scummvm gui). Once launched, games can be added to scummvm. on exit, these are then added as files to rom/scummvm with the game id
as the name (.svm) which then show up in emulation station. this makes adding/launching games automatic.

idea for launching came from http://blog.petrockblock.com/forums/topic/guidescript-add-all-your-scummvm-games-to-emulationstation/#post-4814

### DIFF
--- a/scriptmodules/emulators/scummvm.sh
+++ b/scriptmodules/emulators/scummvm.sh
@@ -22,7 +22,7 @@ function build_scummvm() {
 function install_scummvm() {
     make install
     mkdir -p "$md_inst/extra"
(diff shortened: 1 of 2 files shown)


joolswills:
Please don't merge this yet - forgot a couple of config changes needed.

joolswills:
ok this is ready now.

PR #558, opened 2015-01-05
Add userland.sh

gizmo98:
If threaded_video runs slower or audio stutters after rpi-update rebuild userland libs with this script.

It is some kind of hocus pocus script, because i have the impression self build userland libs run better. There is a CFLAG difference between cross compiled and nativ compiled builds. A native build uses add_definitions("-mcpu=arm1176jzf-s -mfpu=vfp -mfloat-abi=hard"). A cross compiled build uses ADD_DEFINITIONS("-march=armv6").

### DIFF
--- a/scriptmodules/supplementary/userland.sh
+++ b/scriptmodules/supplementary/userland.sh
@@ -1,0 +1,21 @@
+# If threaded_video runs slower or audio stutters after rpi-update rebuild userland libs.
+# Retroarch seems to run slower with prebuild libs.
+rp_module_id="userland"
+rp_module_desc="Raspbian userland libs"
+rp_module_menus="4+"
+
+function depends_userland() {
+	checkNeededPackages gcc-4.8 g++-4.8
+}
+
+function sources_userland() {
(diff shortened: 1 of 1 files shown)


joolswills:
CC / CXX works  cmake honours them. either export it beforehand saving any previous value/restoring it or do something like

```
sed -i 's/-DCMAKE_BUILD_TYPE=Release/-DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=gcc-4.6 -DCMAKE_CXX_COMPILER=gcc-4.6/' buildme
```

I noticed a couple of other issues - your path for your sed is wrong - you should remove /userland/ from the front (you will be in the correct folder as the script does a pushd for the build functions)

Also you have a "sync" at the end which shouldn't be included.

I would just open a ticket upstream querying the compiler flags and your observations - if they reject it perhaps we can gather some more evidence then. They may well go "oh yeh" and change the flags.

gizmo98:
I will do further tests. If the script has any improvements i will reopen the pr.

Issue #553, opened 2015-01-03
multiple binary archives idea


joolswills:
It takes a long time to build the binaries archive - even with distcc and parallel builds etc

When stuff changes that adds new files, or changes the location (such as a the move from dosbox via deb to dosbox via source), people end up with breakage (recoverable by manually building missing stuff from source but still).

I was thinking it might be a nice idea to have multiple binary archives, one for each module (we can have a module flag for the ones that should produce a binary). We can then build individual archives as needed to upload and when something changes, rather than having to rebuild everything everytime. We would still want to rebuild everything sometimes to get up to date packages of course.

I'm pretty sure I can automate this in a nice way too. What do you think ?

petrockblog:
We could use this as an opportunity and start using the Raspbian packaging system. In a first step we can start generating packages for the individual modules, e.g., as described at http://ubuntuforums.org/showthread.php?t=910717. We should write a script that automates the creation and update of those packages as much as possible. These packages could be downloaded and installed via "dpkg -i PACKAGENAME". In a second step we could come up with a package repository that manages the packages such that the APT system of Raspbian could be used. But, again, as a first step the creation of the .deb packages without using a package repository would already be a huge step forward, I think.

joolswills:
I'd rather look at that after. Switching the system into split packages is an hours work or less. Creating debian packages is a few hours work per emulator at least. So happy to consider this as a start, but I have little motivation right now to build debian packages of everything :)

joolswills:
my 3 hours per module is probably way underestimated too. building debian packages takes serious time and effort.

joolswills:
checked that link - that is not a really sufficient example imho. Having built myself a variety of packages for ubuntu for Ubuntu on the O2 Joggler (http://joggler.exotica.org.uk/ubuntu/) we need to handle proper dependencies, build rules, and so on, and we need to be able to build source packages that can be signed for uploading to a repository etc. Just making some quick packages of some already made binaries isn't going to help with that really.

I'm not even sure if it is worth the effort the more I think about it (or at least I'm not particularly interested in spending time on it). the source and binary separation is a reasonable compromise imho. I'd rather spend time on the user interface / ease of use - (such as a the recent stuff to switch screenmodes, or the module improvements etc)

petrockblog:
I do not see why we need source packages. Sure, to use the full set of functionalities of the package system we would also have to consider that. But we could concentrate on the binaries and, hence, skip taking care for the other functionalities. 
[...]

PR #561, opened 2015-01-07
Split binaries

joolswills:
Here's the split binaries functionality from #553 - should be ready to go. I have uploaded all the binaries.

additional flag for modules is "nobin". This is for modules that do not have a standard binary archive (they can still have one, but they must handle it themselves). So things like modules that just change settings etc. I did it like this as most modules are emulators etc - so it is simplest to include it for others.

if a module doesn't have a nobin flag, and it is called with "./retropie/setup module install_bin" the binary will be automatically downloaded and unpacked. If the module also has a function "install_bin" that will be called afterwards.

modules can have a function install_bin_modid even if they have a nobin flag - in the case where they handle the binary download/install themselves - this is the case with sdl2 now - there is no sdl2binaries module - the sdl2 module has a install_bin function instead.

a module can be called with "create_bin" also. in this case a tar.gz is created from its folder and left in /tmp/archives/module_type/module_id.tar.gz

there is a new admin module called "createbinaries" which will archive up all the emulators/ports that don't have a "nobin" flags.

Cheers!


joolswills:
we could make an additional menu now, allowing users to install individual emulators from binary (each option would do depends / install_bin / configure). would be useful for those wanting to make a custom set up quickly with emulation station and a few emulators.

PR #565, opened 2015-01-08
Retroarch: Add aspect ratios for mobile consoles.

gizmo98:


joolswills:
Is this required ? Can we not use the core provided aspect ratio in most cases ?

gambette for example sets the base_width/base_height which is used for the aspect ratio - the problem is we have a hardcoded aspect in the main retroarch.cfg which I think should be removed or set to use libretro core values.

joolswills:
If we can remove the main config, and then add it just for cores which really don't get it right (are there any or is the default value set for another reason? - I removed it on my set up as it didn't seem correct - but haven't done detailed testing)

gizmo98:
Could be even better. I don't know why there is a fixed aspect ratio. Let the core decide. Did you enable video_aspect_ratio_auto = false?

# If this is true and video_aspect_ratio is not set,
# aspect ratio is decided by libretro implementation.
# If this is false, 1:1 PAR will always be assumed if video_aspect_ratio is not set.
# video_aspect_ratio_auto = false

joolswills:
Neither do I know why it is forced but it is - it was added in f81404731f0dea9a091b88e791fc43167811802b

I think we should go with the core provided values, which in most cases seems correct to me.

joolswills:
so if @petrockblog doesn't mind - can we set it to auto, and then fix up manually any emulators that need this changed ? Which should hopefully mean we don't need to set it for many cores, if any ? and any it is broken on, we should report upstream.

gizmo98:
Sounds good.

petrockblog:
All right - let's set it to auto.

joolswills:
@gizmo98 do you want to do a new pr for that or for me to change it ?

gizmo98:
Will open a new pr.

PR #560, opened 2015-01-09
New UAE Emulator

njt1982:
I started playing with http://www.raspberrypi.org/forums/viewtopic.php?f=78&t=80602 last night and it seems like a much newer/better/actively maintained UAE emulator...

It also comes with a free kickstart so will at least boot SOMETHING (if not everything) without having to find your own kickstarts (beit legally or not).

I also started to rewrite the shell script that starts the UAE emulator so that it loads all the disks in a folder for a game and puts them as params to the emulator, eg `-s floppy0=CannonFodder1.adf -s floppy1=CannonFodder2.adf....`

I'll open a PR for that once I've had a chance to get all this working.

### DIFF
--- a/scriptmodules/emulators/uae4all.sh
+++ b/scriptmodules/emulators/uae4all.sh
@@ -36,8 +36,10 @@ file="\$1"
 [ -z "\$file" ] && exit 1
 
 pushd "$md_inst"
-rm -rf df0.adf
-ln -s "\$file" df0.adf
+
+ROMDIR=\$(dirname "\$file")
+echo "path=\"\${ROMDIR}\"" > ./conf/adfdir.conf
+
 $rootdir/supplementary/runcommand/runcommand.sh 1 ./uae4all "$md_id"
 popd
 _EOF_
--- a/scriptmodules/emulators/uae4all.sh
+++ b/scriptmodules/emulators/uae4all.sh
@@ -4,25 +4,24 @@ rp_module_menus="2+"
 rp_module_flags="dispmanx"
 
 function depends_uae4all() {
-    checkNeededPackages libsdl1.2-dev libsdl-mixer1.2-dev libasound2-dev
+    checkNeededPackages libsdl1.2-dev libsdl-mixer1.2-dev libasound2-dev libguichan-sdl-0.8.1-1 libguichan-0.8.1-1
 }
 
 function sources_uae4all() {
-    wget -O- -q http://downloads.petrockblock.com/retropiearchives/uae4rpi.tar.bz2 | tar -xvj --strip-components=1
-    sed -i "s/-lstdc++$/-lstdc++ -lm -lz/" Makefile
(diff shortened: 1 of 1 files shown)


njt1982:
@joolswills This is my PR... My only concern is the URL that you download from changes. I wonder if we could request the author put a symlink to the latest tarball called "latest" or something?!

njt1982:
Or would we pull a version back to the petrockblog asset server?

joolswills:
we would either self host the file, or better still pull from the github repository mentioned on the uae thread on the raspberry pi forum.

I would rather not replace our source build with a binary only though - we should build the new uae from source, and install the resulting binaries

joolswills:
also if there is no way to insert a disk via commandline on the new version, I would seek to amend that first.

njt1982:
Ah ok - the repo is here: https://github.com/sphaero/uae4all-rpi

njt1982:
I think this is the repo https://github.com/sphaero/uae4all-rpi - but its being maintained by someone else...

The tarball has source in it. Maybe we mirror it when it changes?!

Why would we compile it on the Pi when it comes with a working compiled binary? (out of interest)...

I cannot find a way to get disks working from command line - I think it has been removed from this version.... But there is an adfconf for setting the rom directory....

joolswills:
Because that's how it is designed to work - build from source, and creating binaries - we can fix stuff then - and have control over which changes go in and which don't. We also don't have to trust some binary built by someone else. Retropie is also expanding to other platforms that may need alternative build options (our build options may also be difference from the person who built that binary).

If binaries cannot be loaded, then we need to change it so they can be ideally. It's probably a relatively simple change. unfortunately the management of this is a mess - no development with source control just a bunch of people handing sources back and forth. first step would be to diff with the previous version I guess and go from there.

njt1982:
Ahhh cool. Ok I'll l ok into making those changes.

njt1982:
Btw: the emulator works very well!!

image

joolswills:
From the thread btw it seems as though the github repo you mentioned will be kept up to date. Although the person making some pi related control changes will not be contributing to it directly but passing source over to the person I think - possibly they are not familiar with source control - not sure.

Appreciate your efforts here - if you want I could take a look when I get some free time - but I'm happy for you to do it if you prefer :)

nickuccan:
Sweeet

BrianGilbert:
May be helpful.. https://github.com/lubomyr/uae4all2

joolswills:
https://github.com/petrockblog/RetroPie-Setup/commit/a2962b9f54731e46366c9a6ec692f97cacf43b72

joolswills:
to update an existing installation you can install it from menu "5" in retropie_setup (from binary or source).

joolswills:
Closing as the emulator is implemented now - we are running off a fork of the lubomyr repository. Cheers all.

PR #580, opened 2015-01-12
Retronet

joolswills:
cleanup of retronetplay …

 * moved config to $rootdir/configs/all/retronetplay.cfg
 * reduce complexity of parameters passed to setESSystem
 * code duplication / cleanup etc
 * use runcommand.sh to handle netplay parameters - which is better than having to regenerate emulationstation configs all the time and simplifies things
 * instead of having to go into retronetplay configuration to enable/disable it (and then back to emulationstation etc), add the option to launch with netplay enabled into the runcommand.sh menu

### DIFF
--- a/configs/retronetplay.cfg
+++ b/configs/retronetplay.cfg
@@ -1,6 +1,0 @@
-__netplayenable="D"
-__netplaymode="H"
-__netplayport="55435"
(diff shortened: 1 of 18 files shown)


joolswills:
Should be ready to go.

PR #581, opened 2015-01-12
Add experimental vectrex libretrocore

gizmo98:
Compiles and runs.

### DIFF
--- a/scriptmodules/libretrocores/libretro-vecx.sh
+++ b/scriptmodules/libretrocores/libretro-vecx.sh
@@ -13,15 +13,18 @@ function build_libretro-vecx() {
 }
 
 function install_libretro-vecx() {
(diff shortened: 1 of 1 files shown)


gizmo98:
Should be better now.

joolswills:
almost - but I would either just include the .bin files in the md_ret_files (they would then install without the bios subfolder on dest), or copy manually, they dont need to be installed as they are the header files used to create the .bin files.

joolswills:
(so only the .bin files should be copied to the install folder during source install)

gizmo98:
Should be. But i have only tested with all files and my pi is not working atm.

joolswills:
looks good, but ill tweak it after merge to remove the installing of the .h to the /opt/retropie-setup also (to keep it clean from source).

PR #582, opened 2015-01-12
Add experimental 3DO libretrocore

gizmo98:
Compiles and runs but is much slower than mupen64plus-libretro. Needs optimization.


joolswills:
thanks.

PR #587, opened 2015-01-13
fmsx-libretro: Change system file distribution

gizmo98:


gizmo98:
The files are missing in 2.4.1 Beta.

joolswills:
yeh, the image will need an update. the binary package is built though, so when I next update the image it will get updated too.

joolswills:
aah sorry you mean the rom files  I'll rebuild the binary and then update the image

PR #592, opened 2015-01-14
add a menu to allow individual installation of emulators/ports from bina...

joolswills:
...ry or source - if there are no binary packages available for the platform

it just allows installing from source
change update script / reboot keys to u and b to avoid having to re-number when we add stuff.

note I have intentionally left out logging. we currently have logging configured in multiple places, and my next plan is to clean that up a a bit, so it is more maintainable and less duplicated.


joolswills:
the main reasoning for this additional menu was that people were requesting a way to individually install /rebuild emulators - this has been available from the commandline of course, but easier with the gui. This is ready for merging if you are happy with it

PR #607, opened 2015-01-22
LogitechLogitechCordlessRumblePad2

ProxyCell:
This is the Logitecg F710... but in DirectInput mode (there is a switch on the back) it is identified as this controller.


joolswills:
thanks.

PR #609, opened 2015-01-22
mupen64plus: minor improvements

gizmo98:
-Align render resolution to display resolution (VGA)
-Force bilinear filtering (rice plugin)


joolswills:
thanks.

PR #610, opened 2015-01-22
mupen64plus: Revert some useless changes

gizmo98:
mupen64plus.cfg is build at the first run of mupen64plus. So we cannot change any mupen64plus.cfg related setting here. Mupen64plus runs without --osd and --windowed.


joolswills:
Maybe you can a similar thing as we do with dosbox / advmame where we actually launch the binary so it can generate the config before we modify it (if it's possible to launch it a compatible way ?).

gizmo98:
Or create a optimized mupen64plus.cfg with cat? I will play around to get the best performance/visual quality settings. I could add the resulting config file to the module.

joolswills:
might be worth to see if it will work with a partial config too - saves a lot of bloat in the module - with vice for example we feed a partial config - everything else will use the default which is fine (and on saving settings the complete config will be built).

PR #613, opened 2015-01-24
mupen64plus: Align gles2n64 render resolution to screen resolution

gizmo98:


gizmo98:
Everything should be there now.
-Start mupen64plus without rom name creates mupen64plus.cfg
-Use iniset to set default video-gles2n64 and audio-omx plugins
-Set VGA resolution in gles2n64.conf because 400x240 was too small.
-Added some extra option in gles2n64.cfg (Hacks, frame skip options and so on)
-gles2n64 can be used now. RICRPI merged a fix for flickering textures (https://github.com/ricrpi/mupen64plus-video-gles2n64/commit/0cd825e01eb08477a075e09bb457c7a8bcc79709).

joolswills:
thanks

PR #617, opened 2015-01-26
Add videopac libretrocore O2EM

gizmo98:
Compiles and runs if o2rom.bin BIOS file is present.


joolswills:
actually. ill merge it and then fix it up, as I see there are some other mkRomDirs to fix also.

joolswills:
thanks

gizmo98:
Yeah. Many modules use the old style mkdir:
armsnes, catsfc, cavestory, doom, fbalibretro, tyrquake, hatari, apple2

joolswills:
They should all be sorted now.

PR #627, opened 2015-01-31
allow parameters to be passed to modules - this is for possible future c...

joolswills:
...hanges where we can have both gui configuration and commandline configuration for modules. It also simplifies the retropie_packages argument code


petrockblog:
I like that idea. Could you add some comments to those modified lines that give an idea about the logic behind the parameter shift? At least for me it would make things easier in the future :-)

joolswills:
Sure thing. Done.

PR #629, opened 2015-02-04
added initial platform config for rpi2

joolswills:


joolswills:
this is useful for anyone who wants to build from source / optimised for the cortex-a8 cpu for the pi 2. standard retropie binaries will still work. I don't have a pi 2 for testing, but am currently building a binary set. some makefiles will override our CFLAGS, but we can sort those out as we find them.

Of course, if it turns out there is no noticable speedup from compiling for armv7, then we may not need to distribute binaries at all, but it's worth looking at I guess.

scriptmodules/system.sh
```diff
@@ -74,6 +74,17 @@ function platform_rpi() {
     __binary_url="http://downloads.petrockblock.com/retropiebinaries/"
 }
 
+function platform_rpi2() {
+    __default_cflags="-O2 -mfpu=neon -march=armv7-a -mfloat-abi=hard"
+    __default_asflags=""
+    __default_makeflags=""
+    __default_gcc_version="4.7"
+    __qemu_cpu=cortex-a8
```
petrockblog: Can you give me a (brief) explanation how to use this parameter best in a cross-compile environment? I still did not get a complete overview about that topic.

joolswills: Well, to compile for the rpi2 from a pi (or in a chroot), you would just do 

```
sudo __platform=rpi2 ./retropie_setup.sh
```

I have built a binary set for the rpi2, which i can upload and users can try and see if there is any noticable performance improvement.

if cross compiling with distcc, you shouldn't have to do any more that the adjusted commandline above, and add any additional makefile options etc such as pathing in distcc binaries.

1st8:
Hey guys, will this config be picked up automatically when running the setup on a Pi2 + current raspbian?

I am currently running the install with precompiled binaries on a Pi2, curious if it'll work.

joolswills:
it will work. but you can build with

sudo __platform="rpi2" ./retropie_setup.sh 

to get most binaries built for armv7. this code is merged so is available if you update retropie-setup

scriptmodules/system.sh
petrockblog: Thanks for that info, I got it.

PR #636, opened 2015-02-09
Add experimental mupen64plus (testing) mainline module.

gizmo98:
Ricrpi has done a really nice job. But his fork is not updated as frequently as mainline repo. PRs and upstream fixes are not always merged. Mainline repo is working now with rpi. There is slight performance hit, but it is still much faster as mupen64plus-libretro. Moreover the mainline repo is not stripped down for ARMv6 and could run better with PI2. 

Rice is the default video plugin. I added a texture filtering fix to get rid of pixelated textures (shadows, water and so on) and enabled default 2x multisampling antialiasing.

### DIFF
--- a/scriptmodules/emulators/mupen64plus-testing.sh
+++ b/scriptmodules/emulators/mupen64plus-testing.sh
@@ -1,0 +1,251 @@
+rp_module_id="mupen64plus-testing"
+rp_module_desc="N64 emulator MUPEN64Plus (Testing)"
+rp_module_menus="4+"
+rp_module_flags="!odroid"
+
+function depends_mupen64plus-testing() {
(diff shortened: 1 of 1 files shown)


gizmo98:
Next step: Get Naranns highly reworked video rice plugin running.

joolswills:
I wanted to hold off on this just until the existing module is working, so we avoid having more modules for people to say don't work on the pi2 :) Is that ok ?

gizmo98:
Should be working now without using "sed".

joolswills:
Will merge and give it a test. Over at the retropie forum, someone has commented that they thought the pi2 build is slower/less smooth than the previous armv6 optimised build.

gizmo98:
Gles2n64 seems to run slower with all opt flags enabled. Please tell me if you see a Performance regression.

PR #646, opened 2015-02-13
mupen64plus-libretro: Don't use rpi2 target for now.

gizmo98:
It always crashes with a segfault. If -mfpu=neon-vfp4 is set the libretrocore will crash.

### DIFF
--- a/scriptmodules/libretrocores/mupen64libretro.sh
+++ b/scriptmodules/libretrocores/mupen64libretro.sh
@@ -13,11 +13,11 @@ function sources_mupen64plus-libretro() {
 function build_mupen64plus-libretro() {
     rpSwap on 750
     make clean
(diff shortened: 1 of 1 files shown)


joolswills:
I'd prefer the lines removed rather than commented out (and possibly a comment put in). I know I'm fussy - sorry :)

gizmo98:
Yup. Just wait. I will test something else.

joolswills:
we want to be able to control the platform flags ourselves really from the script - but that depends on the makefile being made to allow this - so many don't (I just did fixed this with the new uae4all for example).

joolswills:
Should this be merged or you need to do some more testing still ? thanks for all the changesets.

gizmo98:
It is working now with on a fixed fork. I will close it.

PR #647, opened 2015-02-14
remove binaries update - this option is no longer useful and users are c...

joolswills:
...onstantly using it and then reporting stuff as not working due to missing dependencies etc from updated binaries.

one way forward would be to make it check dates of archives and work on that, but option 1 is the best way to update everything, and it doesn't take very long. Also users
can update individual components from menu 5 which will run depends/configure etc.


joolswills:
BTW my idea is to remove this until we have a better mechanism, not necessarily permanently.

joolswills:
added some additional safety to avoid errors when updating - this time with the script. see above commit.

this is ready for merging. thanks.

PR #648, opened 2015-02-15
Add experimental nes libretrocore nestopia

gizmo98:
Builds and runs.


gizmo98:
https://github.com/petrockblog/RetroPie-Setup/issues/621

PR #649, opened 2015-02-15
Add experimental gba libretrocore vba-next

gizmo98:
Builds and runs. But is even a little bit slow on a PI2 with default clock.


joolswills:
is it worth having then ? I'm getting the feeling we are moving towards quantity rather than quality - and it ends up being more to maintain.

gizmo98:
I try to get an working gba libretrocore atm. If gpsp runs better we should use gpsp. There should be at least one libretrocore if possible.

joolswills:
okey dokey. Let's see how they compare and get some user feedback perhaps.

PR #652, opened 2015-02-15
Add experimental gba libretrocore gpsp

gizmo98:
Builds and runs.


frthery:
Hi, just for information since 1 Feb gpsp core has poor performances (PI1):
https://github.com/libretro/gpsp/commit/fb984fcaeffe8fd99c801738d4b238f1e99e131c

because dynarec has been disabled :/

joolswills:
Was it working for you before ? if so please report it to https://github.com/libretro/gpsp/issues/11

gizmo98:
Did you test with dynarec enabled? Could tune the makefile if it is running with dynarec enabled.

frthery:
Before this modification core was running good, that why i am using a version compiled before.

PR #656, opened 2015-02-15
Messages and info

joolswills:
A whole load of cleanups relating to error & info messages, and error logging/reporting.

also fixes a bug with duplicate errors being created.


joolswills:
this is all ready to go if you are happy with it.

PR #664, opened 2015-02-17
mupen64plus-testing: Cleanup and rework

gizmo98:
It is no longer necessary to create config files. The files were merged in the appropriate plugin repository. I created rom dirs for gles2n64 and gles2rice. There will be always games running only with one of these plugins.

Other changes:
-Added two dependencies for audio plugins.
-Added make option -j2 for PI2. Compiles much faster with two threads. 
-Added experimental glide64 plugin. Runs but is really slow.
-Copy *.conf files as well.


joolswills:
-jX should not be in the modules. see https://github.com/petrockblog/RetroPie-Setup/issues/663

gizmo98:
Good to know ;-) 
I removed it.

PR #668, opened 2015-02-17
-Now asplashscreen script will parse each line of the splashscreen.list ...

free5ty1e:
...instead of just looking at the first line to decide what to do.  Previously, if the first line was an image, then fbi would just be called with the list (and if there was more than one image in the list, they would all be displayed slideshow style).  If the first line was a movie, the grep did not actually detect it.  Now each line is handled separately.  This does not affect normal operation with an unmodified splashscreen.list, but if a power user wants to manually add more images and / or videos to splashscreen.list then they will now play back in the background during startup as expected.  Emulation station will interrupt the playback / display if startup finishes before the splash sequence.

### DIFF
--- a/supplementary/asplashscreen/asplashscreen
+++ b/supplementary/asplashscreen/asplashscreen
@@ -11,19 +11,22 @@
 
 do_start () {
 
-    line=$(head -n 1 /etc/splashscreen.list)
-    isMovie=$(echo $line | grep -o "*.mpg")
-    if [ -z "$isMovie" ]; then
-      /usr/bin/fbi -T 2 -once -t 20 -noverbose -a -l /etc/splashscreen.list &
-    else
-      mplayer $line &
-    fi
+    while read splashline; do
+        echo Playing splash video or image $splashline
+        isMovie=$(echo $splashline | grep -o ".mov\|.mp4\|.mkv\|.3gp\|.mpg")
+        if [ -z "$isMovie" ]; then
+            /usr/bin/fbi -T 1 -once -t 10 -noverbose -a $splashline
+            sleep 12
+        else
+            mplayer $splashline
+        fi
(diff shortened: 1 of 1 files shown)


free5ty1e:
There we go.  Had to delete my fork, refork, and create new branch.  Perhaps next time I try to sync forks it will fast-forward instead of creating ugly merge commits, I must have done something to my fork that screwed that up before.

free5ty1e:
Two other changes of note that may or may not be helpful: 

* fbi now displays on virtual console TTY1 instead of TTY2 (this appeared to work well for me but may not be necessary)

* Timeout for splash screen down from 20 seconds to 10 seconds to provide more time for startup videos to play (perhaps this should be an optional component in the splashscreen.list, time to display each image... future enhancement?)

Let me know if these need to be changed, or anything else I've neglected here

joolswills:
Just my point on the other pull request  that our configuration tool allows the choice of 1 image currently, so I think that needs sorting at the same time, otherwise I'm not convinced of the benefit of this change. sorry.

free5ty1e:
No worries, if and when that happens, this pull request might become helpful or relevant.  I figured it out for me and wanted to share.

joolswills:
all feedback and support is appreciated. thanks.

HerbFargus:
@ free5tyl1e 

I found your pull request incredibly useful and exactly what I needed (as I've been beating my head against the wall to figure out how to make videos work as a splash screen while still having the option to revert back to static images if needed). The only thing that I modified was I installed omxplayer on my system so as a result in the code I changed mplayer to omxplayer and it worked beautifully. Thank you so much!

free5ty1e:
Awesome, glad it came in handy elsewhere!

joolswills:
This is now included. Thanks.

PR #671, opened 2015-02-19
RetroArch input autoconfig rework

gizmo98:
It makes no sense to have two modules for retroarch autoconfig files (retroarchjoypadautoconf.sh and retroarchautoconf.sh). It also makes no sense having autoconfig files with different hotkey combos for exit, saving, loading, open menu and so on. The user will be confused if start+select works with one controller but not the other. I have done four changes to overcome this problem:
-I created a python script which adds the same hotkey combos to all *.cfg files under /opt/retropie/emulators/retroarch/configs. It's called setAutoconfHotkeys.py and it is located under /supplementary. I don't know if this is the right location. Please correct me if it should be stored in another location.
-I merged retroarchjoypadautoconf.sh and retroarchautoconf.sh so we have only one module and can use autoconfig files from libretro repo. (there are even more config files now)
-The new retroarchautoconf.sh runs setAutoconfHotkeys.py at the end. All config files get the same hotkey combos.
-I also added setAutoconfHotkeys.py to retroarchjoyconfig.sh so you have preset hotkeys if you register a controller.


free5ty1e:
Awesome work here -- Looking forward to using this on my RetroPie!

scriptmodules/supplementary/retroarchautoconf.sh
```diff
@@ -3,7 +3,17 @@ rp_module_desc="RetroArch-AutoConfigs"
...
 
+function sources_retroarchautoconf() {
+    gitPullOrClone "$md_build" https://github.com/libretro/retroarch-joypad-autoconfig
+}
+
 function install_retroarchautoconf() {
     mkdir -p "$emudir/retroarch/configs/"
     cp "$scriptdir/supplementary/RetroArchConfigs/"*.cfg "$emudir/retroarch/configs/"
+    cp -r $md_build/udev/* $emudir/retroarch/configs/
+    sudo chown -R $user:$user $emudir/retroarch/configs/
```
joolswills: the sudo is a mistake ? (script is run as root already)

scriptmodules/supplementary/retroarchautoconf.sh
```diff
@@ -3,7 +3,17 @@ rp_module_desc="RetroArch-AutoConfigs"
...
 rp_module_flags="nobin"
 
+function sources_retroarchautoconf() {
+    gitPullOrClone "$md_build" https://github.com/libretro/retroarch-joypad-autoconfig
+}
+
 function install_retroarchautoconf() {
     mkdir -p "$emudir/retroarch/configs/"
     cp "$scriptdir/supplementary/RetroArchConfigs/"*.cfg "$emudir/retroarch/configs/"
+    cp -r $md_build/udev/* $emudir/retroarch/configs/
```
joolswills: doesn't need the -r

joolswills:
although bash is not always great for config file parsing/changing, I think I would prefer if it could be done in the module file in bash rather than a separate python script. Not sure the python is doing anything that couldn't be done in bash either is it ?

gizmo98:
Would be better. But i don't know how to do it (now) and pythons configparser and try/except error handling are much familier to me.

joolswills:
I'm planning on converting/removing some of the other pythons to simpler bash scripts too (for a config menu in es), hence not wanting more. I will merge now, but then rewrite this in bash if that's ok.

gizmo98:
No problem. I removed the problematic parts. 
I will try to write some retroarch autoconfig files to mupen64plus autoconfig file converter.

PR #682, opened 2015-02-23
Cleanup

gizmo98:
-Remove all autoconfigs. I contributed our autoconfigs to https://github.com/libretro/retroarch-joypad-autoconfig. 
-Replace setAutoconfHotkeys.py in retroarchjoyconfig.sh with Jools "retroarchautoconf.sh configure" method.
-Add RetroArch overlays and assets: http://filthypants.blogspot.de/2012/09/background-shader-for-retroarch.html
-Add "#include /opt/retropie/configs/all/retroarch.cfg" to ensureSystemretroconfig. With #include appendconfig is no longer necessary. 
-Set savesate and savefile directory in every libretro module. There was a annoying "cannot save SRAM" error every time a closed a libretrocore. RetroArch stored a error dump file every time in my home directory.


joolswills:
thanks.

PR #683, opened 2015-02-23
Mupen64plus testing update

gizmo98:
-Added vanilla glide64mk2 from mupen64plus
-Use vanilla rice from mupen64plus
-Added default settings for rice and glide64mk2
-Added own video settings repo so it is easier to modify RiceVideoLinux.ini and gles2n64rom.conf.
-Use updated audio-omx and video-gles2n64 plugins from ricrpi.
-Move to source based installations, because all plugins are running.

### DIFF
--- a/scriptmodules/emulators/mupen64plus-testing.sh
+++ b/scriptmodules/emulators/mupen64plus-testing.sh
@@ -1,13 +1,10 @@
 rp_module_id="mupen64plus-testing"
 rp_module_desc="N64 emulator MUPEN64Plus (Testing)"
-rp_module_menus="2+"
+rp_module_menus="4+"
 rp_module_flags="!odroid"
 
 function depends_mupen64plus-testing() {
-    if ! hasPackage libsdl2-dev && isPlatform "rpi"; then
-        rp_callModule sdl2 install_bin
-    fi
-    getDepends libsamplerate0-dev libspeexdsp-dev
+    getDepends libsamplerate0-dev libspeexdsp-dev libsdl2-dev
 }
 
 function sources_mupen64plus-testing() {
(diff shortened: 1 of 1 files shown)


joolswills:
if you plan to move this to the standard build, what's the current difference between testing/non testing. I think we should have a single mupen64plus - whichever is working the best? if we really need both, then I think we need to differentiate between them in more detail.

scriptmodules/emulators/mupen64plus-testing.sh
```diff
@@ -1,23 +1,29 @@
...
 rp_module_desc="N64 emulator MUPEN64Plus (Testing)"
-rp_module_menus="4+"
+rp_module_menus="2+"
 rp_module_flags="!odroid"
 
 function depends_mupen64plus-testing() {
-    getDepends libsamplerate0-dev libspeexdsp-dev libsdl2-dev
+    if ! hasPackage libsdl2-dev && isPlatform "rpi"; then
+        rp_callModule sdl2 install_bin
+    fi
```
joolswills: this code is not needed - it reverted my earlier change to just use libsdl2-dev with getDepends.

scriptmodules/emulators/mupen64plus-testing.sh
```diff
@@ -64,25 +71,52 @@ function install_mupen64plus-testing() {
...
+    cp -v "$md_build/mupen64plus-video-settings/"{*.ini,*.conf} "$md_inst/share/mupen64plus/"
 }
 
 function configure_mupen64plus-testing() {
     # to solve startup problems delete old config file
-    rm -f "$configdir/n64/mupen64plus.cfg"
-    mkdir -p "$configdir/n64/"
+    rm -f "$rootdir/configs/n64/mupen64plus.cfg"
+    mkdir -p "$rootdir/configs/n64/"
     # Copy config files
```
joolswills: why not configdir ?

scriptmodules/emulators/mupen64plus-testing.sh
gizmo98: I will modify this section and repost.

gizmo98:
I would like to use vanilla mupen64plus always if possible. I have not tested if vanilla mupen64plus runs as good as ricrpis mupen64plus on a PI1. I will reset it to experimental for now.

scriptmodules/emulators/mupen64plus-testing.sh
gizmo98: I should update my testing branch more often...

gizmo98:
Opened up a cleaner pull request.

PR #685, opened 2015-02-24
mupen64plus-testing: Add glide64 and default settings

gizmo98:
-Added vanilla glide64mk2 from mupen64plus
-Use vanilla rice from mupen64plus
-Added default settings for rice and glide64mk2
-Added own video settings repo so it is easier to modify RiceVideoLinux.ini and gles2n64rom.conf.
-Use updated audio-omx and video-gles2n64 plugins from ricrpi.


joolswills:
you still seem to have some rootdir/configs rather than the newer configdir variable usage btw.

gizmo98:
You are right. I should take a nap or something else ;-)

joolswills:
sorry :)

PR #703, opened 2015-02-27
Snes9x and genesis plus gx

gizmo98:
Add snes9x-next and use genesis-plus-gx for megadrive/genesis emulation.


joolswills:
please can we use the older naming for now? maybe just call it genesisplusgx for module name (and filename) - until we have all agreed on what we will do about the naming.

gizmo98:
snes9x-next as well?

joolswills:
yes please (can just call it snes9xnext.sh / snex9xnext for module name for now)

joolswills:
(without my typos ;-) )

joolswills:
Cheers.

PR #710, opened 2015-03-02
retropie-setup as a module

joolswills:
Seemed logical since everything else is a module apart from init stuff, to make the retropie-setup a module too.

did a bit of code cleanup too.

no major differences doing it like this, apart from you can easily trigger any of the menus manually from retropie_packages - such as `./retropie_packages.sh setup binaries` to do a binary install (and later could be called directly from the retropie menu in emulationstation for doing an update etc)


petrockblog:
Good idea!

PR #712, opened 2015-03-03
Add bluez 5.x module for PS3 controller

gizmo98:
The newest bluetooth daemon with six axis support. Replaces the current
bluetooth daemon. Should also replace QTSixA. At the end of the
installation it is necessary to manually set trust for the PS3
controller.

Originates from here: http://blog.petrockblock.com/forums/topic/using-the-latest-bluez-for-ps3-bluetooth-how-to/

I use this script since sd card image 2.4.1. No blinking LEDs, no disconnection problems, no QTSixA requirement. Just wanted to share before i forget to add it ;-).

The instruction message box in bluez.sh needs some loving because the text is a little bit displaced. The bluetooth daemon is running but could need a second look.


joolswills:
thanks - it will need to be adjusted though to be either packaged, or installed so it doesn't overwrite the system stuff or we are taking a step backward with the work done for SDL / dispmanx SDL / SDL2 etc. It should be possible to build a debian package - maybe just backporting the jessie package if it is new enough.

joolswills:
Appreciate you are just sharing this though - those who want to try it can of course. Would be nice to have an updated packaged bluetooth stack though. If jessie is new enough, we won't have to wait too long as I believe it will be released soon(ish)

free5ty1e:
The last required part "manually trust the controller" is what killed bluez 5.0 for me.  I have set up a couple RetroPies for friends that are not Linux-savvy and do not want to mess with any command line.  For these setups, I opted to keep the `qtsixad` and `sixpair` method in use, since I can have an Emulationstation-launched script run `sudo sixpair` to make repairing after PS3 console usage simple enough.  

I'd like to see this new bluez 5.0 added as an option and not to completely replace the qtsixad / sixpair solution, since both have their merits.  

Or, if someone can come up with a totally automated trust script that actually automates the process, that would also work.

At least on my friends' units for the time being, I'll be sticking with the qtsixad / sixpair method.

gizmo98:
There is a ini file for every bluetooth controller under /ver/lib/bluetooth. Every ini file has a "trusted" flag. I have not tested if it is only necessary to set this flag. If this is all we can use iniSet on every controller file to trust all paired controllers.

gizmo98:
@free5ty1e 
It is really as easy as i told you before. There is a info file under /ver/lib/bluetooth/%macbtstick%/%macps3controller/. If you set trusted=true your ps3 controller connects after restart. ;-)
So it should be no problem to add a udev rule or use bluez5.x without the instructions message box.

gizmo98:
Personally i think it is a really cool idea to add a udev rule for ps3 controllers. The new bluez already sets up your controller if you plug it in. It is only necessary to set trust. So if a udev rule sets trust for any ps3 connector at plug in our rpis would act like real ps3. ;-)
Plug in your ps3 controller and everything is set up! 

@joolswills 
Do we have any example code editing only certain files in subdirs?

joolswills:
@gizmo98 please can you give me a more specific example ?

gizmo98:
I have two PS3 controllers. Bluez has created two diretories:
/var/lib/bluetooth/00:1A:7D:DA:71:15/00:26:43:0C:09:98/info 
/var/lib/bluetooth/00:1A:7D:DA:71:15/00:26:43:A1:B6:14/info

As you can see file paths contain MAC address of my bluetooth dongle and my PS3 controllers (system specific).

"info" contains:
[General]
Name=PLAYSTATION(R)3 Controller
Class=0x000508
SupportedTechnologies=BR/EDR;
Trusted=true
Blocked=false
Services=00001124-0000-1000-8000-00805f9b34fb;00001200-0000-1000-8000-00805f9b34fb;

[DeviceID]
Source=2
Vendor=1356
Product=616
Version=256

So if info has "Name=PLAYSTATION(R)3 Controller" it is necessary to set "Trusted=true".

free5ty1e:
Oh, awesome!  I was trying to echo -e commands to the bluetoothctl to manually trust the connected controller from a script but had no luck there.  Your solution appears to be possible to automate, I definitely want to try that out :D

joolswills:
recursively grep (grep -rl) for the string, and then use iniSet on the resulting files.

joolswills:
don't need -r if they are the same level, just 

```
for file in $(grep -rl "/var/lib/bluetooth/*/info"); do
  do something
done
```

etc

gizmo98:
Thank you very much!

joolswills:
I'm tired - so I updated my not very good/non working example code :)

joolswills:
you probably want to check the files exist first too, or just redirect stderr or something.

gizmo98:
This works for all info files:
find /var/lib/bluetooth -name "info" -exec sed -i -e 's/Trusted=false/Trusted=true/g' {} \;
I need to add the "Name=PLAYSTATION(R)3 Controller" test. I don't now if it is necessary to exclude other bt devices. But i only want to trust PS3 controllers atm.

joolswills:
What was the problem with grepping for files with the correct Name= ?

gizmo98:
It did not find info files. Is it necessary to modify /var/lib/bluetooth/*/info because i have two subdirectories? /var/lib/bluetooth/*/*/info did not work. I have only started to play around. So please stay tuned.

joolswills:
you said the info files are in a subdirectory so that should work.

joolswills:
eg

```
for file in $(grep -l "Name=PLAYSTATION(R)3 Controller" /var/lib/bluetooth/*/info); do
    sed -i "s/Trusted=false/Trusted=true/ig" $file
done
```

gizmo98:
grep: /var/lib/bluetooth/*/info: No such file or directory

joolswills:
then the files don't exist - I'm only going on the information you posted earlier with the paths.

joolswills:
aah ok. so i missed a slash in the opath - my eyesight. just stick another wildcard in then.

```
for file in $(grep -l "Name=PLAYSTATION(R)3 Controller" /var/lib/bluetooth/*/*/info); do
    sed -i "s/Trusted=false/Trusted=true/ig" $file
done
```

gizmo98:
Nice!
pi@retropie ~ $ sudo ./foldercheck.sh
/var/lib/bluetooth/00:1A:7D:DA:71:15/00:26:43:0C:09:98/info
/var/lib/bluetooth/00:1A:7D:DA:71:15/00:26:43:A1:B6:14/info

gizmo98:
I updated the module. Bluez needs some time to add controller directories. So i added "sleep 5". Would be better to check directory update with inotifywait or incrond.

free5ty1e:
awesome!!  gonna try it out...

free5ty1e:
Does it only attempt to auto-trust during the actual module installation?  Or is that bluezps3.sh module going to be run in the background every time a controller is connected via USB?  (is the auto pair not active yet?)

gizmo98:
Only turing actual module installation "at the moment".

gizmo98:
@free5ty1e 
PS3 controllers work after restart at the moment. If i restart the bluetooth daemon settings will not be taken over. I could test something like stop bluetooth daemon, hciconfig hci0 down, hciconfig hci0 up, start bluetooth daemon.

free5ty1e:
OK, I've been experimenting with this Bluez5 implementation for a couple days now.  I have had several issues with REconnecting a controller after it has connected the first time via bluetooth.  Pair up, can control emulationstation wirelessly, and run jstest to see joystick inputs changing with my button presses.  

Then, hold PS3 button for 12 seconds to disconnect (or restart system) and can pair wirelessly again, yet all inputs from the controller are now ignored!  Once this happens I can only seem to fix it by reimaging the Pi and starting fresh. 

Have you run into any reliability issues or other problems like what I described above?  

Also, are you using Bluez-5.27 for any particular reason over v5.28 or the freshly-released 5.29? 

All 3 of these versions appear to have the same behavior for me, but maybe I'm missing something.  I'm about ready to go back to qtsixad at this point....

gizmo98:
I haven not experienced such behavior. But i haven't disconnected with PS button. My controllers connect every time after startup if i press PS buttons. Different bluetooth dongle?

I haven't added bluez 5.28 or 5.29 because both versions were not released at module creation.

gizmo98:
I have startup problems if bluetooth dongle and ps3 controller(s) are connected at startup, even with a 2A power supply.

free5ty1e:
Might have something to do with the bluetooth dongle, its the standard cheapie one here:

http://www.miniinthebox.com/mini-bluetooth-2-0-adapter-dongle_p79211.html?litb_from=sysmail

Never had these issues with qtsixa.  This isn't caused by turning the controllers off via the PS3 button, this just appears to happen after I've had a controller paired and working ONCE.  The next time, whether thats after a reboot or a manual disconnect... i get the paired controller that is showing up as /dev/input/js0 but no input is getting through to jstest or emulationstation anymore.

If you haven't seen this issue, then it might just be something to do with this adapter, which sucks... but I guess I can go back to sixad if its just me :(

free5ty1e:
Also, to note my setup:  I have a Raspberry Pi B+ with that bluetooth adapter, ethernet, the PS3 controller (i've tried having it both connected and disconnected during startup on USB, no difference), and running just on a class 4 SD card, with an official CanaKit 2.5A power supply.  Even with no overclock, this is still an issue for me.  

I did remove the sixad service from starting automatically, and purged all bluez 4.99 stuffs with `sudo apt-get -y purge bluez` before installing bluez 5.x.  I wonder if I'm missing something else that needs to happen...?

Anyway, awesome progress on this, even if I can't use it right away for my setup!   Thanks!

gizmo98:
If my ps3 controller is connected over usb and i have not pressed PS button /dev/input/js0 appears but there is no button response. It is necessary to push PS button at least once. 

I try to reproduce your issue. I should use a fresh install for testing but i dislike to overwrite my current sd card ;-)

free5ty1e:
Yeah, I thought it might be that too -- but when this happens to me, the PS3 button also has no effect after the pairing is complete.  Controller just lit up on player 1 and the Pi is ignoring all my button presses / axis movements.  Yet if I reimage and freshly build / install Bluez 5 again, it works again.  

Definitely works over USB; must press PS3 button first, then USB controls are activated.  

Thanks for looking into it, I'm probably going to keep searching for potential solutions in the meantime too and will post anything I find here!   I am quite stubborn, lol

free5ty1e:
Oh!  Forgot to mention, this is using a RetroPie 2.5.0 beta as the base image.  

I've tried using RetroPie 2.6.0 rpi1 image, but that image has some problems... I am not even able to compile things on it!  (checking whether gcc works.... no:  lol)

gizmo98:
@free5ty1e and @m42i 
Should we keep this or close it for now? In my opinion the old sixad module runs as good or better now and it is even not necessary to build a new bluetooth daemon.

free5ty1e:
I have been testing the latest sixad module with the new rule in place, and it works flawlessly!  And if I want to use USB PS3 controllers instead of bluetooth, I simply remove my bluetooth adapter from the Pi before connecting the controllers.  Brilliant!

No need for bluez 5.x with the sixad stuff as it is now :D

m42i:
I tested it with a fresh RP 3 Beta 2 image on a RPi 1. 

The auto-detection and all works fine, but the limitations still apply:
- Only two controllers at a time
- Both have LED 1 lid up

I still suspect my BT dongle (ID 0a12:0001 Cambridge Silicon Radio, Ltd Bluetooth Dongle (HCI mode)).

gizmo98:
I have only two controllers but i see no reason why only two should work at a time. LED lid can be setup with sixad default profile config file. There is already a PR for this issue https://github.com/petrockblog/RetroPie-Setup/pull/806. @free5ty1e the "no controller function if bluetooth dongle and ps3 controller is connected" issue could be solved with an updated sixpair source code. Sixpair detaches the joypad kernel driver and transfers the MAC-address of a bluetooth dongle to the ps3 controller. There is also a reattach function which is not implemented at the moment.

Tele42:
So, I did some testing with my Raspberry Pi 2 B, with a generic bluetooth BR/EDR adaptor and 4 dualshock 3 controllers, two controllers work fine, the third takes a long time for /usr/sbin/sixad-sixaxis to spawn, but it connects and works, the fourth does not connect and any of the other controllers will not reconnect if turned off and back on. Resetting the bluetooth adaptor allows for the controllers to reconnect. All controllers were paired ahead of time, this test was run multiple times with a random sequence between the controllers.

PR #722, opened 2015-03-07
Add converted libretro/common-shaders repo

gizmo98:
Converted all cg/cgp shaders to glsl/glslp shaders.
https://github.com/libretro/common-shaders

This commit will blow up the whole setup script. It would be better to provide and install an own common-shaders repo. @Florian and @joolswills could we add a extra repo for glsl common-shaders?


joolswills:
either that perhaps or we include converting from the upstream repository (if it can be automated).

gizmo98:
It is not possible because there is no nvidia-cg-toolkit for arm. Converted all shaders on a x86 machine.

joolswills:
aah ok. then maybe an external repo is best or just include in ours. Thinking about repositories, we currently have stuff under various accounts (mine / yours etc). Was thinking perhaps we should have retropie under a github organisation perhaps and have all our repositories there? that way for example, if I needed to fork some emulator, I can fork it under retropie - keeps everything together.

gizmo98:
I made a own repo now. I totally agree we should think about one location for forks and own repos in the future.

petrockblog:
Forehandedly, I have reserved "retropie" as Github organization. We can discuss about a possible usage in the issue #725 if you like.

Issue #726, opened 2015-03-10
[Feature Reqest] USB copy Gamelist, Downloaded_Images, and Configurations


robertybob:
I like the USB romservice feature but I don't understand why we don't do the same thing for gamelist.xmls and the scraped/downloaded images/box arts we have. I create all of these on my PC and rather than keep using SSH to transfer them over, I'd like to be able to just plug my USB stick into my Pi, my PC, then back into my Pi and have it copy over all of my ROMs, artwork and custom gamelist.xml files in one go.

petrockblog:
In addition it would make sense to also sync configuration files with the USB stick. In this way the USB copy service could become a tool for a simple backup and restore of most of the customized things of the users.

joolswills:
how would you envisage the configurations to be synced ? I'm just thinking whether it would be confusing if roms are synced one way, but configurations the other.

petrockblog:
Good point.
Maybe it would be a good compromise, if there were the folders `configs_fromRetroPie` and `configs_toRetroPie`.
After mounting the USB service would create a new backup of the configuration files (maybe in an archive, keeping older revisions of configurations in separate files).
Per default the folder `configs_toRetroPie` contains only a `read.txt` telling the user that if he puts configuration files in that folder they will be copied over to the RPi. 
A backup and restore use case would then consist of
1. There user hooks up the USB stick to the RPi, the USB service copies all configuration files to the folder `configs_fromRetroPie`.
2. The user copies the configuration archive to the folder `configs_toRetroPie`.
3. The user hooks up the USB stick to the RPi, the USB service sees that there are configurations to import and extracts/copies these configurations to their corresponding folders.

These are brainstorming thoughts, though. I might have overlooked something.

petrockblog:
I have committed what I have basically described above (see commit d0c798417264426394ce68c334a8de537865fd14). However, before we close this issue we need to discuss, which set of data actually belongs to the configurations that we want to have a backup of. The [copy script](https://github.com/petrockblog/RetroPie-Setup/blob/master/supplementary/01_retropie_copyroms) needs to be adapted accordingly then.

joolswills:
I think if doing this we may want to make the changes discussed in https://github.com/petrockblog/RetroPie-Setup/issues/619 .

I also think we should have three usbromservice options - install with copy script, just install service, and remove (we have install/remove currently afair). Possibly making the config backups optional.

[...]

PR #727, opened 2015-03-10
Lr module and emulator rework

joolswills:
emulator choices are configured in /opt/retropie/configs/emulators.cfg

There is a new file in supplementary with mapping from system name to "full name" and supported extensions. this is used for the es configuration - users can copy this to opt/retropie/configs/all/ and add their own extensions etc if they want. I split it out for systems where there were multiple emulators, and duplicated extensions etc.

Not all emulators have been switched to the new addSystem call - and it doesn't matter for those where there is a single emulator. with mupen64plus-testing we call this 3 times, so you can switch gpu core, so it can be used for some emulator options too (but isn't specifically designed for that)

feel free to ask any questions - this is not ready to go yet - I'm still testing etc.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -142,8 +142,6 @@ function updatescript_setup()
         return
     fi
     popd
-    printHeading "Updating ESConfigEdit script."
-    updateESConfigEdit
     printMsgs "dialog" "Fetched the latest version of the RetroPie Setup script. You need to restart the script."
 }
 
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -42,5 +42,6 @@ function configure_advmame() {
 
     setDispmanx "$md_id" 1
 
+    addSystem 0 "$md_id" "mame-advmame" "$md_inst/libpocketsnes.so"
     setESSystem "MAME" "mame-advmame" "~/RetroPie/roms/mame-advmame" ".zip .ZIP" "$rootdir/supplementary/runcommand/runcommand.sh 0 \"$md_inst/bin/advmame %BASENAME%\" \"$md_id\"" "arcade" "mame"
 }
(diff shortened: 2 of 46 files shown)


gizmo98:
Looks good. There are still some setSystem functions left. Do you keep them because there is only one system emulator like libretro-handy or 4do atm?

joolswills:
wasn't a bad question but the answer was yes - because we had only a single emulator. we can migrate them as needed (if we migrate them, people can overrride the file extensions which can be useful perhaps).

joolswills:
(but there is no need to do it right away)

gizmo98:
Next dumb question! 
addsystem x is used to set the emulator priority? addsystem 0 is the default emulator? So we can set different default emulators for rpi1/2?

joolswills:
1 means default. so yeh, you can put logic in the modules depending - https://github.com/petrockblog/RetroPie-Setup/pull/727/files#diff-a262c7bd2c764f87a07733f8c2f19a18R44

joolswills:
there isn't any priority, it's just 0 or 1 for the default.

gizmo98:
Nice.

joolswills:
this is pretty much ready now. has had some testing and some fixes applied.

one thing to note is that the new code now runs the emulators with no video mode switching by default. From reading of the forum etc, it seems most users do not want this - performance can be affected, but the screenmode can be changed as needed - as this menu is now a major part of configuring set up, I expect it will be used more.

After this pull request I will add the ability to configure the retroarch render resolution also from the menu. This should then completely solve any problems with performance at 1080p etc.

joolswills:
ok. Please merge if happy. I have some others fixes etc to do on trunk, unrelated to this, but to avoid having to rebase/fix merge conflicts I'll wait until this is sorted.

petrockblog:
With the configuration becoming even more important now I like the idea of enhancing the USB copy service script such that all configuration files are also synced with the USB stick. I have added this thought to the issue #726.

PR #730, opened 2015-03-13
Cleanup

gizmo98:
-Move nestopia, armsnes, snes9x-next, gpsp-libretro and catsfc to source based installations.
-Cleanup mupen64plus module
-Mupen64plus/Mupen64plus-testing: Use vfp optimized code id platform=rpi1


joolswills:
I have built binaries for the emulators as moving them to 2+ also includes them in the binary installation.

BTW regarding the menu config, I am thinking  of changing this, so we have some flags which specify if the module is core / optional / experimental etc rather than just some id. Need to think about it some more though before submitting any code.

joolswills:
and thanks for the changes :)

Issue #531, opened 2014-12-26
runcommand and mode selection


joolswills:
I have some changes for runcommand.sh that seem logical and solve an issue in my case, but wanted to go over some details.

a) fbset -depth 8 && fbset -depth 16

Why is this needed? It first tries to set 8 bit depth, and if that succeeds it tries a 16 bit depth ? Are there cases where 16 bit will fail ?

b) starttype==1: set video mode to VGA ONLY IF tvservice is in HDMI mode, and run command

This mode is 4:3 ratio even on 16:9 monitors I believe ? Currently this mode is used for a bunch of emulators - but if on a 19:6 monitor mode 3 (mode 3: 720x480 @ 60Hz 16:9, clock:27MHz progressive) would seem a better choice to keep the correct aspect ratio ?

c) starttype 2 / 3 / 6 / 7 / 8 are not used by anything. Do they serve any purpose ?

--

Basically the changes I want to make are to take into account the preferred mode of the monitor, and record the aspect and group of that and use that as a basis for mode choice. My picade monitor is 1024x768, and the preferred mode is "state 0x12001a [HDMI DMT (16) RGB full 4:3], 1024x768 @ 60.00Hz, progressive". Some of the current emulators try to use starttype 4 (set video mode to 720p60 ONLY IF tvservice is in HDMI mode and run command) - which is unsuitable for a 4:3 monitor like mine. My changes would only allow this mode to be selected if the monitor preferred mode is the same aspect. In the case of my monitor, the preferred mode is a better choice.

If you could help me with some of these questions, I can make sure I don't break any existing functionality in use, but make runcommand work better for 4:3 and 16:19

gizmo98:
We need runcommand.sh to lower resolution because some emus (mostly libretrocores) slow down at 1080p. I have pushed another way to lower render resolution of retroarch without changing the display/framebuffer resolution. The pi can upscale any source to the frambuffer resolution for free --> https://github.com/libretro/RetroArch/pull/1308

All libretrocores can be run with "runcommand.sh 0". The render resolution can be set in retroarch.cfg with video_fullscreen_x and video_fullscreen_y. If both values are zero retroarch uses the display resolution (old behavior).

Upscaling can be tuned with boot arguments:
https://github.com/raspberrypi/firmware/issues/104

joolswills:
A step forward definitely - thanks - but I think this adds some more configuration problems unless I have misunderstood? 

[...]

PR #735, opened 2015-03-14
added options to launch menu to switch render resolution of retroarch #531

joolswills:
@gizmo98

The final piece ? As discussed on ticket #531 this now includes the ability to switch retroarch render resolution.

Options are

"320x240"
"640x480"
"960x720"
"1280x960"
"Use video output resolution"
"Use config file resolution"

it defaults to 640x480 - I thought about having a different default for pi2, but then I thought maybe best to err on the side of caution and go with best performance at a res that should be enough as a default. And of course it is easily switchable. But let me know what you think and if there is enough choice.

### DIFF
--- a/supplementary/runcommand.sh
+++ b/supplementary/runcommand.sh
@@ -66,6 +66,7 @@ function get_save_vars() {
     # convert emulator name / binary to a names usable as variables in our config file
     emusave=${emulator//\//_}
     emusave=${emusave//[^a-Z0-9_]/}
+    rendersave="${emusave}_render"
     romsave=r$(echo "$command" | md5sum | cut -d" " -f1)
 }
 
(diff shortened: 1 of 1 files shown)


joolswills:
I forgot about overlays of course, which could look better at 960x720. let me know if 640x480 is ok as a default anyway or if you definitely think pi2 would be better at higher default and I can put some additional logic in.

gizmo98:
Should be ok. I will check it tomorrow.

gizmo98:
Please update the retroarch binary. If i do a binary update aspect ratio will not be respected if 640x480 is set. All libretrocores are stretched.

joolswills:
I didn't notice that - but maybe because I was testing on a 4:3 screen.

Binaries are updated (quite a size difference now due to all the overlays / assets etc)

PR #737, opened 2015-03-16
Add repo for missing system themes

gizmo98:
Added missing themes for retropie and videopac. Can be removed if someone adds these systems to simple_latest.zip.

### DIFF
--- a/scriptmodules/supplementary/emulationstation_theme_simple.sh
+++ b/scriptmodules/supplementary/emulationstation_theme_simple.sh
@@ -24,5 +24,4 @@ function install_esthemesimple() {
     
     # copy simple-theme-plus themes
     cp -a "$md_build/"* "/etc/emulationstation/themes/simple/"
-    chown $user:$user -R "/etc/emulationstation/themes/simple/"
 }
--- a/scriptmodules/supplementary/emulationstation_theme_simple.sh
+++ b/scriptmodules/supplementary/emulationstation_theme_simple.sh
@@ -3,6 +3,10 @@ rp_module_desc="EmulationStation Theme Simple"
 rp_module_menus="2+"
 rp_module_flags="nobin"
 
+function sources_esthemesimple() {
+    gitPullOrClone "$md_build" https://github.com/gizmo98/simple-theme-plus.git
+}
+
 function install_esthemesimple() {
     # download themes archive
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - sounds good.

I wrote to @petrockblog regarding a theme for retropie - has this been implemented upstream or it's something custom ?

It might be better to handle it all from a git repository and so changes can be pulled in upstream - your's would be fine imho until we move perhaps to the central org etc - but I'll leave this one for @petrockblog to merge. Can you give more details on the retropie theme ?

probably don't need to chown anything as the themes live in /etc/ which is ok to be owned by root imho. but a minor thing ;-) I don't think we chown the current theme - not sure.

gizmo98:
It's something custom. It's oldschoolretropie asplashscreen with a retropie logo (bauhaus font). The videopac logo uses the original videopac font. Don't know if you are working on a new image but our new retropie menu should have at least a theme :-)

Feel free to remove chown.

gizmo98:
I.e. to customize logo and background replace retropie.png and retropie_art_blur.png. Theme.xml is already set up. Feel free to replace these to files with official logos and backgrounds.

petrockblog:
Thanks a lot for your additions!
In fact, I have contacted Nils from EmulationStation and asked if he would have an idea for a RetroPie theme. I am still waiting for a reply, though.
In either way, it would be great if your work would be added to the "simple theme" or another theme - hopefully there will be additional themes soon.

gizmo98:
I removed chown. In my opinion you can merge it if simple theme will not be updated in the next time.

joolswills:
I believe Nils is working on something for us - so I guess we should wait a bit more - if you don't mind. Unless you think we can merge this now, and then replace later ?

joolswills:
(Not that your theme isn't appreciated though :) )

petrockblog:
I was told by Nils that he wants to wait until we have a logo (I plan a logo contest, but some things still have to be waited for). I suggest that we use this theme for the RetroPie configuration. When we have an official logo, we can still adapt the theme accordingly.
Again, thanks a lot for your work!

joolswills:
Competition sounds like a nice idea!
merged :) thanks @gizmo98

gizmo98:
Yeah. Competition sounds good.

PR #750, opened 2015-03-22
Update autostartemustat.sh

bircoe:
2 changes:
option 2, if you have emulation station closed logging in via ssh will start es, adding a check for display and which tty the /etc/profile file is being run from prevents this.
option 3, forces emulation station to restart upon closing, as I couldn't find another way to refresh the games list after adding extra roms this is my current work around.

### DIFF
--- a/scriptmodules/supplementary/autostartemustat.sh
+++ b/scriptmodules/supplementary/autostartemustat.sh
@@ -8,6 +8,7 @@ function configure_autostartemustat() {
     options=(
         1 "Original boot behaviour"
         2 "Start Emulation Station at boot."
+        3 "Start Emulation Station at boot, closing Emulation Station will restart it."
     )
     choices=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
     if [[ -n "$choices" ]]; then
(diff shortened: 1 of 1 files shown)


joolswills:
thanks

it shouldn't have started when logging in via SSH - are you saying the previous check didn't work - I wasn't able to reproduce the problem here.

I like the patch, but it needs to probably remove and re-add if switching between options 2 and 3 - currently if you do option 2, you would have to choose option 1 first before you could change it to 3.

joolswills:
it could be de-duplicated too, as they are basically the same with a minor change to the launching so the other code should be included once perhaps with a parameter.

joolswills:
Having another think about this, instead of a 3rd option, we could have a prompt to ask the user to press a key, and wait a second. if they press a key they drop to shell, else emulationstation is loaded again. what do you think ?

bircoe:
Hi, yes, the previous check definitely didn't work for me, it's a pretty specific use case though, ES has to be closed and the Pi sitting at a console, logging in via SSH causes ES to start on the monitor attached to the Pi, the problem was there across multiple reboots, granted it is a pretty minor problem but still annoying if you wanted ES to stay closed.

I agree, that it should be removed first, I hadn't thought of that scenario when a user wanted to switch between 2-3 and vice versa.

Instead of the 3rd option another prompt is fine by me, I won't have time to have another go at it till later tonight but I like that idea.

Cheers,
Martin

joolswills:
I can implement the other idea if you like - I would probably do it by changing the launch script, and making the loop by default perhaps.

I can't reproduce the ssh issue. the environment variable should be set on any ssh session which it checks.

bircoe:
PS also note that the 3rd option could be desired default as it would save users having to find a keyboard if they accidentally closed ES or it crashed.

joolswills:
it would default to restarting es, but press a key to drop to console.

bircoe:
The environment variable was definitely set, so I don't understand why it happened... that was the first thing I checked for once I located where emulationstation was started from.

bircoe:
If you like the idea of having ES restart by default with a key press to drop to shell then perhaps drop option 3 and do the loop and key check in /usr/bin/emulationstation?

First start could be a normal start entering the loop after ES starts preventing the press any key to drop to shell message on boot.

joolswills:
yeh - the loop in the launcher sounds like the best option.

bircoe:
Well let's close this pull and I'll create 2 new pulls?

joolswills:
https://github.com/petrockblog/RetroPie-Setup/pull/756

PR #756, opened 2015-03-24
restart emulation station on exit (unless a key is pressed). Since emula...

joolswills:
...tion station doesn't yet have a mechanism to pick up on

any changed roms, it is quite annoying for users to have to keep restarting / plug in a keyboard etc. With this change afetr 5 seconds
emulationstation will restart, unless a key is pressed and the user wants to drop to terminal.


joolswills:
https://github.com/petrockblog/RetroPie-Setup/pull/750

PR #760, opened 2015-03-26
rename retroarchjoyconfig to retroarchinput, and add keyboard configuration to it

joolswills:


HerbFargus:
Random potentially unrelated question to this merge- will there ever be a point when on first boot into emulationstation that the configuration box that shows up- will configure controllers for both emulationstation and retroarch?

joolswills:
well, it might happen in the future, would need some mechanism to convert from the es_input.cfg to something usable by retropie. It's not something I personally have planned though.

joolswills:
or something to run first which would generate the es config and the retroarch configs.

HerbFargus:
well in the meantime I think retroarch-config in the new RetroPie menu was a really good step in the right direction.

petrockblog:
> Random potentially unrelated question to this merge- will there ever be a point when on first boot into emulationstation that the configuration box that shows up- will configure controllers for both emulationstation and retroarch?

I have planned to work on that. I already have quite a clear idea about that and fits very well into the modular organization of the RetroPie-Setup Script. 
The retroarch-config in the RetroPie menu definitely was a good step.

joolswills:
recallbox does something like this where they have some additional control set up stuff in es, then generate configs for retroarch and a few other emulators based on this. It's something that would be nice to have for sure, but requires time to implement.

Issue #766, opened 2015-03-27
Game specific config files for retroarch


janezkranjc:
Hello!

Would it be possible to implement a separate config file for certain roms?

http://blog.petrockblock.com/forums/topic/game-specific-config/ <- this guy proposed a pretty good solution but it doesn't work because he can't put %ROM% twice in the command.

joolswills:
Can you give me an example use case ? it could be done via runcommand, but what would you need to change on a per rom basis ?

janezkranjc:
I am using lr-genesis-plus-gx which allows me to play 4 player games such as general chaos and fifa with 4 controllers if I add this setting to the megadrive retroarch.cfg:

input_libretro_device_p2="1025"

However, if I have this setting enabled all other games that do not support multitap do not work as controllers are no longer detected by the games. Whatever I press even on controller 1 doesn't work.

So I would need to have a separate config for multi-tap games (general chaos, fifa, nba jam) and regular one or two-player games

joolswills:
ill see what I can do.

janezkranjc:
Thanks for merging!

joolswills:
no probs. I've made a couple of changes - to use the dialog editbox rather than nano, and I removed the additional regular expression, as it can just be appended to the conf variable..

janezkranjc:
It looks really cool now! And it deletes the config file if it's empty, very clever :)

PR #770, opened 2015-03-28
enabled the option in the menu to edit custom config for each rom, if config exists it is added to appendconfig (only for retroarch)

janezkranjc:
This pull request addresses the issue #766.

I am not very proficient in bash programming, but this is the solution I've been searching for. Even if this pull request is not merged it may provide a simple way to solve this problem for somebody.


joolswills:
thanks :)

fillipefeitosa:
Thanks. This is perfect for multitap games in SNES. Cause some games does not work with the multitap set on.

joolswills:
Please use the RetroPie forum for help.

PR #771, opened 2015-03-28
Retroarch cleanup

gizmo98:
Remove all shaders and remove assets. All old shaders are moved to
common-shaders repo.
https://github.com/petrockblog/RetroPie-Setup/issues/762


joolswills:
thanks :)

PR #773, opened 2015-03-29
mupen64plus: Cleanup and new plugins

gizmo98:


gizmo98:
We can test it and move it over later.
https://github.com/petrockblog/RetroPie-Setup/issues/772

gizmo98:
Ok. Should be better now.

PR #786, opened 2015-04-04
reicast: fix startup from emulationstation

gizmo98:
Add a reicast start script

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -42,7 +42,15 @@ function configure_reicast() {
 
     # create bios dir. Copy dc_boot.bin and dc_flash.bin there.
     mkdir $md_inst/data
+
+    cat > $md_inst/reicast.sh << _EOF_
+#!/bin/bash
+cd /opt/retropie/emulators/reicast
+sudo ./reicast.elf -config config:image=$1
+_EOF_
+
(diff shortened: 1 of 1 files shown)


petrockblog:
Thanks!

scriptmodules/emulators/reicast.sh
```diff
@@ -42,7 +42,15 @@ function configure_reicast() {
 
     # create bios dir. Copy dc_boot.bin and dc_flash.bin there.
     mkdir $md_inst/data
+
+    cat > $md_inst/reicast.sh << _EOF_
+#!/bin/bash
+cd /opt/retropie/emulators/reicast
+sudo ./reicast.elf -config config:image="$1"
```
gizmo98: @joolswills how do i write "$1" to a file. I don't want to write the value of variable $1 of module reicast into reicast.sh.

joolswills: you can quote it

```
\$1
```

instead of using cd btw you could use pushd / popd so if someone launches the .sh they stay in their current folder on exit. also you can use $md_inst instead of the /opt/retropie/emulators/reicast path

if you don't need any variables to be used in a script you are writing out using 

```
<< _EOF_
```

you can add a backslash

```
<< \_EOF
```

which we use for example in bashwelcometweak

gizmo98:
Thank you very much!

PR #787, opened 2015-04-04
Add some sixad optimizations

gizmo98:
https://github.com/petrockblog/RetroPie-Setup/issues/654
-Add an own pairing module.
-Pair PS3 controller if controller is connected without retropie-setup. Use udev rules.
-Add hciconfig pscan to pair more than one controller.

### DIFF
--- a/scriptmodules/supplementary/ps3controller.sh
+++ b/scriptmodules/supplementary/ps3controller.sh
@@ -48,7 +48,8 @@ function install_ps3controller() {
     update-rc.d sixad defaults
 
 # If a bluetooth dongle is connected set state up and enable pscan
-    cat > $md_inst/bluetooth.sh << _EOF_  
+    cat > $md_inst/bluetooth.sh << _EOF_
+#!/bin/bash
 /usr/bin/hciconfig hci0 up
 if ! hciconfig | grep -q "BR/EDR"; then
     /usr/bin/hciconfig hci0 pscan
@@ -59,6 +60,7 @@ chmod +x "$md_inst/bluetooth.sh"
 
 # If a PS3 controller is connected over usb check if bluetooth dongle exits and start sixpair
     cat > $md_inst/ps3pair.sh << _EOF_  
+#!/bin/bash
 if ! hciconfig | grep -q "BR/EDR"; then
     /usr/bin/hciconfig hci0 pscan
     $md_inst/sixpair
--- a/scriptmodules/supplementary/ps3controller.sh
+++ b/scriptmodules/supplementary/ps3controller.sh
@@ -51,8 +51,8 @@ function install_ps3controller() {
     cat > $md_inst/bluetooth.sh << _EOF_
 #!/bin/bash
 /usr/bin/hciconfig hci0 up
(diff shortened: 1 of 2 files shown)


free5ty1e:
Awesome!  I will test this out this weekend and see if it is still possible to use a controller on USB while a BT dongle is present.  (sixpair still probably kills the USB connection but at least it looks like I can disconnect the BT dongle to disable the rule...)

scriptmodules/supplementary/ps3controller.sh
```diff
@@ -47,6 +47,43 @@ function install_ps3controller() {
...
+# udev rule for ps3 controller usb connection
+    cat > /etc/udev/rules.d/99-sixpair.rules << _EOF_
+# Pair if PS3 controller is connected
+DRIVER=="usb", SUBSYSTEM=="usb", ATTR{idVendor}=="054c", ATTR{idProduct}=="0268", RUN+="$md_inst/ps3pair.sh"
+_EOF_
+
+# Start sixad daemon
+/etc/init.d/sixad start
+
     md_ret_files=(
```
joolswills: indentation

scriptmodules/supplementary/ps3controllerpairing.sh
```diff
@@ -0,0 +1,24 @@
...
+rp_module_flags="nobin"
+
+function configure_ps3controllerpairing() {
+    if [[ -f "$rootdir/supplementary/ps3controller/sixpair" ]]; then
+        # Only start sixpair. Don't waste time with driver compilation.
+        rp_callModule ps3controller configure
+    else
+        # Install PS3 controller driver
+        rp_callModule ps3controller	
+	  fi
```
joolswills: tab indented.

joolswills:
This would be best done in a single script I think - with a menu to choose to setup and then pairing.

gizmo98:
I need to test if manual pairing is really necessary anymore. udev rules should do everything if a controller is connected. So if udev rules run reliable pairing script and configure section can be removed.

gizmo98:
Tested! 
1. Paired one controller with my PS3.
2. Installed RPIE 3.0B2.
3. Installed PS3 driver.
4. Reboot
6. Connected PS3 controller over USB and pressed PS button.
6. Remove USB cable and press PS button.
7. Paired!

taalas:
Hi @gizmo98,

is the current version sixad version functional for you?

I am using an updated RetroPie installation (binary installation plus ps3 controller driver from today). After pairing the controller, removing the cable and pushing the PS button the controller will start blinking, then rumble for about 5-6 seconds while alternating the LEDs from 1 to 4 and back. I have never seen a PS3 controller do that before...any idea what could be going on?

Thanks!

gizmo98:
If the controller rumbles sixad is working. If it does not rumble the default linux driver should be activ. After some seconds only one led should light up and the controller should work.

taalas:
Oh ok, thanks for your quick reply. The controller rumbles, LEDs cycle...during this process I can see the controller being recognized when checking dmesg. After a while it just stops and no LEDs light up, the device is gone at this point. No idea what is going on but I imagine there must be something wrong on my end then.

PR #792, opened 2015-04-05
Rewrite limelight setup script to a module

gizmo98:
Based on this script:
https://github.com/stsfin/RetropieLimelightInstaller/blob/master/limelightconfig.sh

I cannot test this module finally because i have no geforce to connect to. Is it necessary to have an own section for limelight? Could put it under ports or somewhere else.

Also see: petrockblog#721

### DIFF
--- a/scriptmodules/supplementary/limelight.sh
+++ b/scriptmodules/supplementary/limelight.sh
@@ -20,13 +20,13 @@ function sources_limelight() {
     gitPullOrClone "$md_build" "https://github.com/stsfin/RetropieLimelightInstaller.git"
     
     wget https://github.com/irtimmer/limelight-embedded/releases/download/v1.2.2/libopus.so
-	wget https://github.com/irtimmer/limelight-embedded/releases/download/v1.2.2/limelight.jar	
+    wget https://github.com/irtimmer/limelight-embedded/releases/download/v1.2.2/limelight.jar	
     
     # Download limelight simple theme
-	wget https://github.com/stsfin/RetropieLimelightInstaller/releases/download/1.3.1/theme.xml
+    wget https://github.com/stsfin/RetropieLimelightInstaller/releases/download/1.3.1/theme.xml
     wget https://github.com/stsfin/RetropieLimelightInstaller/releases/download/1.3.1/limelight.png
-	wget https://github.com/stsfin/RetropieLimelightInstaller/releases/download/1.3.1/limelight_art.png
-	wget https://github.com/stsfin/RetropieLimelightInstaller/releases/download/1.3.1/limelight_art_blur.png
+    wget https://github.com/stsfin/RetropieLimelightInstaller/releases/download/1.3.1/limelight_art.png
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/limelight.sh
```diff
@@ -0,0 +1,119 @@
...
+    #Install limelight simple theme
+    mkdir /etc/emulationstation/themes/simple/limelight
+    mkdir /etc/emulationstation/themes/simple/limelight/art
+    cp $md_inst/theme.xml /etc/emulationstation/themes/simple/limelight/
+    cp $md_inst/limelight.png /etc/emulationstation/themes/simple/limelight/art/
+    cp $md_inst/limelight_art.png /etc/emulationstation/themes/simple/limelight/art/
+    cp $md_inst/limelight_art_blur.png /etc/emulationstation/themes/simple/limelight/art/
+
+    #Run limelight configuration
+pushd $md_inst
```
joolswills: indentation

scriptmodules/supplementary/limelight.sh
```diff
@@ -0,0 +1,119 @@
...
+    clear
+    echo -e "\nDiscovering GeForce PC:s, when found you can press ctrl+c to stop the search, or it will take a long time \n"
+    #discover IP-addresses of Geforce pc:s
+    java -jar limelight.jar discover
+    echo -e "\n"
```
joolswills: wouldn't just echo suffice ?

first echo can probably lost the first \n and just be a normal echo.

joolswills:
ill merge the two requests and fix up styling, but please double check indentation etc

stsfin:
Hi just to answer gizmos question, the original version of the script installed limelight under ports. The reason why I made the installer to make it as its own console was because it created all the different resolution settings and the configurator and thus the ports became cluttered.  This not really an issue if there is only a configurator script and a "run limelight" script, and you set the resolution during the configuration. This would anyhow be smarter since the 1080p 60FPS wont run without overclockin the h.264 freg atleast to the ~350 mark. (im using 500 for a higher bitrate).

taalas:
Hi and thanks for creating the supplementary module for limelight!

Some minor things I discovered while testing this:

- line 115 in limelight.sh points to the "ports" instead of the "limelight" directory in roms. I am guessing this is just a copy/paste error...

- the emulationstation script for configuring limelight calls "retropie_packages.sh limelight configure" without sudo and won't work (at least on my system)

Apart from that it works really well!

stsfin:
Have you used the newest version?

stsfin:
If possible Limelight should be build on retropie during install, this would ensure the newest working version.

taalas:
I agree, the points I mentioned are not connected to Limelight itself but rather to the way it is currently embedded into RetroPie though.

gizmo98:
I have seen taalas has already done a PR for the wrong file path. I will add sudo as well.

taalas:
Yes, I wasn't sure if you wanted the sudo in there or rather another approach for the settings so I left that out in my PR ;)

PR #796, opened 2015-04-09
Add some supported formats to platforms.cfg

diogomsantos:
I have some roms with the .bin extension that work in the listed platforms. I tested with all emulators, for the respective platform.
I have also some .sg roms that i think they are Mega Drive roms, but not sure, could be SG-1000. They work with the 2 Mega Drive emulators so i added too.


joolswills:
thanks.

PR #801, opened 2015-04-15
Fix sixad controller autoconf detection issue

gizmo98:
Sixad adds the MAC address of a controller to the device name. Autoconfig detections like retroarch autoconfig files or mupen64plus rely on the exact device names. If there is any addition the detection is not working. This "fix" modifies sixad source code so MAC address will not be added anymore.

P.S. VID/PID detection does not work with sixad. Retroarch's udev driver does not check VID/PID for virtual or bluetooth devices. Furthermore according to udevadm the sixad joypad jsX has no VID/PID.

### DIFF
--- a/scriptmodules/supplementary/ps3controller.sh
+++ b/scriptmodules/supplementary/ps3controller.sh
@@ -84,7 +84,29 @@ _EOF_
 # Pair if PS3 controller is connected
 DRIVER=="usb", SUBSYSTEM=="usb", ATTR{idVendor}=="054c", ATTR{idProduct}=="0268", RUN+="$md_inst/ps3pair.sh"
 _EOF_
-
+    
+    # add default sixad settings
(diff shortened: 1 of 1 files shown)


gizmo98:
@joolswills and @petrockblog .
Works just fine for me. All wireless PS3 controllers have the same name now. Retroarch autoconfig works. All PS3 controller changes of the last week should fix the following issues:
https://github.com/petrockblog/RetroPie-Setup/issues/776
https://github.com/petrockblog/RetroPie-Setup/issues/719 
https://github.com/petrockblog/RetroPie-Setup/issues/654

Rumble support could produce lag (http://steamcommunity.com/app/237930/discussions/2/624075036520525118/?insideModal=1). I haven't experienced any lag. If necessary we can disable rumble support easily.

joolswills:
Cheers.

PR #817, opened 2015-05-01
Update retropiemenu.sh

HerbFargus:
Added wicd-curses to the RetroPie menu as a graphical interface to connect to wireless networks.


joolswills:
Thanks.

PR #818, opened 2015-05-01
Add famicom disk system as an own system

gizmo98:
Because it is a own system like a SEGA CD or SEGA 32.
http://blog.petrockblock.com/forums/topic/specify-an-emulator-in-runcommand-sh/


joolswills:
thanks.

PR #819, opened 2015-05-01
-Removed hardcoded hciconfig binary location.  PS3 controller pairing on...

free5ty1e:
... my Pi2 RetroPie fails because my hciconfig binary is actually located in /usr/sbin/ and not /usr/bin.  This is far more compatible and safer; as well as used this way everywhere else in this script (this was probably an oversight).


joolswills:
thanks.

free5ty1e:
no prob!  :D

PR #820, opened 2015-05-04
Update packagecleanup.sh

ProxyCell:
pulseaudio & cups are not in the latest raspbian image
could more things be added for removal?

### DIFF
--- a/scriptmodules/supplementary/packagecleanup.sh
+++ b/scriptmodules/supplementary/packagecleanup.sh
@@ -14,7 +14,6 @@ rp_module_menus="3+"
 rp_module_flags="nobin"
 
 function configure_packagecleanup() {
-    # remove PulseAudio since this is slowing down the whole system significantly. Cups is also not needed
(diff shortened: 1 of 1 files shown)


joolswills:
I believe those are removed by another retropie module - as pulseaudio directly interferes (or has been said to).however, it could be argued that it was all moved to this module depending on if pulseaudio is still an issue for basic functionality.

joolswills:
ignore my previous comment - I misread the patch. are you saying that raspbian doesn’t come with cups and pulseausio installed by default ?

ProxyCell:
yeah, i tried removing them manually as they are written in that script file
only the wolfram and sonic-pi were possible for removal. the others were not present in the latest raspbian image (from feb 2015)

joolswills:
it's possible someone might install them later however, and so it's still useful perhaps to have the option to remove them, but this "module" isn't too useful at present - should probably have a menu and some options.

ProxyCell:
thanks for that @joolswills

PR #821, opened 2015-05-04
Update platforms.cfg

Pacolo:
Add CUE to have sound in Sega CD


Pacolo:
The comment that need to be merged always is the cue addition, the other is a proposal. (remove some extensions)

joolswills:
manually added as a single commit. thanks.

PR #827, opened 2015-05-10
sixad: Use newer sixpair sources

gizmo98:
Use libusb 1.0 instead of libusb 0.1 so ps3 controller kernel driver can be reattached after sixpair has done the job.


free5ty1e:
YES thank you for figuring this out!  xD

PR #828, opened 2015-05-11
Add DXX-Rebirth Port

MattWatt:
Add script to build DXX-Rebirth from current source.  Add Descent and
Descent 2 to the Ports system.  The build method compiles both d1x-rebirth and d2x-rebirth, therefore, they'll both be installed to a single dxx-rebirth folder, with separate execution scripts.


free5ty1e:
Awesome!  Glad to see this in retropie module format.

joolswills:
thanks.

PR #840, opened 2015-05-24
Mupen64plus-testing: Audio buffer underun fix

gizmo98:
AUDIO OMX: Enable extended options to increase audio buffer


joolswills:
cheers.

PR #842, opened 2015-05-24
Add experimental NDS emu Desmume

gizmo98:
It is really slow and needs much more processing power than a pi2 has.


joolswills:
Please can you correct the indentation, and resubmit the above changes as a single commit.

joolswills:
I manually merged it as a single commit.

PR #846, opened 2015-05-30
Add uae4arm (RPI2 - dispmanx)

amigagamer:
Adds Uae4Arm installer to experimental menu (raspberry pi 2 only)

### DIFF
--- a/scriptmodules/emulators/uae4arm.sh
+++ b/scriptmodules/emulators/uae4arm.sh
@@ -26,8 +26,8 @@ function build_uae4arm() {
 
 }
 function install_uae4arm() {
- 	cp data "$md_inst" -r
-	md_ret_files=(
+    cp data "$md_inst" -r
+    md_ret_files=(
         'uae4arm'
   #      'uae4all'
   #      'Readme.txt'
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Please see above comments regarding formatting etc.

scriptmodules/emulators/uae4arm.sh
```diff
@@ -0,0 +1,65 @@
...
+function configure_uae4arm() {
+    mkRomDir "amiga"
+
+    mkdir -p "$md_inst/conf"
+   #   echo "path=$romdir/amiga" >"$md_inst/conf/adfdir.conf"
+    chown -R $user:$user "$md_inst/conf"
+
+    mkdir -p "$md_inst/kickstarts"
+    chown -R $user:$user "md_inst/conf"
+
```
joolswills: This should be $md_inst btw. Don't worry though, rather than merging the pull request, I will apply it manually with a couple of changes. thanks.

joolswills:
I committed the code with some additional changes. the md_inst line was also changing the permissions on the wrong folder - but since we symlink the kickstarts, I left the kickstarts folder non-writable by the pi user. it's still in the experimental section anyway. thanks.

PR #847, opened 2015-05-31
Update kodi.sh

HerbFargus:
This fixes the black screen on exit to emulationstation

see post here: 
http://blog.petrockblock.com/forums/topic/exiting-kodi-to-emulationstation/

### DIFF
--- a/scriptmodules/ports/kodi.sh
+++ b/scriptmodules/ports/kodi.sh
@@ -30,26 +30,11 @@ function configure_kodi() {
 
     cat > "$romdir/ports/Kodi.sh" << _EOF_
 #!/bin/bash
-LOG_FILE=$HOME/.kodi/temp/kodi.log
- 
-rm $LOG_FILE 2> /dev/null
- 
-/usr/lib/kodi/kodi.bin --standalone &
- 
-while [[ ! -f $LOG_FILE ]] ; do
-  sleep 1s
-done
- 
-while read line ; do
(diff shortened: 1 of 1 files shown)


joolswills:
Does kodi not exit cleanly then - the script has a busy loop watching the log, then tries to kill kodi - if kodi exits ok, we can leave that out and just reset the framebuffer. I'm not keen on adding this in the current form as it seems a rather heavy handed workaround.

HerbFargus:
@joolswills It doesn't exit cleanly for quite a few people I've seen as I've trolled the forums in the past few months (including myself) it would just freeze up on the black screen when you exit to emulationstation. 

You were correct about only needing the framebuffer- I took all the log rubbish out and just left the framebuffer reset and it works just fine.

joolswills:
Yeh so it's exiting ok, but leaves the framebuffer in an unusable state. Probably running it via runcommand may fix it too (as that does a framebuffer reset).

Please could you submit an additional pull request with just the change as a single commit - alternatively i could apply the change manually (avoids having two commits adding them removing code).

HerbFargus:
Ok, sorry for my inexperience with github, I'm still figuring it out. I've added another pull request with runcommand and it seems to work with initial testing.

joolswills:
np.  thanks for the contributions :)

Nothing wrong with multiple commits sometimes, but it's just tidier to have a single commit for a small change.

PR #851, opened 2015-05-31
Update kodi.sh

HerbFargus:
Added runcommand to kodi


joolswills:
thanks.

HerbFargus:
Seems I spoke too soon on the matter and the black screen freeze issue still happens after watching a video for some time in kodi (perhaps it is because emulationstation  goes to sleep?) Thoughts on a more concise solution than the posts here?:
http://blog.petrockblock.com/forums/topic/exiting-kodi-to-emulationstation/

joolswills:
I will take a look when I get a free moment. thanks for the info.

PR #856, opened 2015-06-08
Input configuration

joolswills:
This is still a WIP, but wanted to do the request so we don't collide. 

De-duplication and cleanups - the config file locations were duplicated in every function.

I'm thinking we can actually do this without so many functions - and have it so the submodules manage the logic themselves - should de-duplicate even more.

### DIFF
--- a/supplementary/moduledata/supplementary/emulationstation/configscripts/emulationstation.sh
+++ b/supplementary/moduledata/supplementary/emulationstation/configscripts/emulationstation.sh
@@ -8,57 +8,6 @@
 # at https://raw.githubusercontent.com/petrockblog/RetroPie-Setup/master/LICENSE.md.
 #
 
-###### input configuration interface functions ######
-
-#######################################
-# Interface functions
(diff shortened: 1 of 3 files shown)


joolswills:
@petrockblog  - in the modules in the first functions you use named parameters then shorten it later - was this to show an "example" of the parameter ordering? if so, I think one example is enough perhaps ? I think it's actually more confusing to a new set of eyes that the functions differ.

joolswills:
Now I'll check I didn't break anything :)

joolswills:
ok this seems to be ok now.

I have some usability things that I think need doing on the emulationstation side, but the scripts are working even after ive messed around with them (they were working before also ;-) )

petrockblog:
> in the modules in the first functions you use named parameters then shorten it later - was this to show an "example" of the parameter ordering? if so, I think one example is enough perhaps ? I think it's actually more confusing to a new set of eyes that the functions differ.
Yes, they are supposed to be examples. But a single example should also be enough I think.

PR #859, opened 2015-06-11
-Updated reicast.sh installation script to pull from my latest branch…

free5ty1e:
…, which has been rebased from the latest upstream master and has additional improvements and features: (see post here for details: http://blog.petrockblock.com/forums/topic/configuring-controllers-in-reicast/page/2/#post-99514 )

* Launch script now runs emulator and the fallback file allocation for rpi2 from tmpfs in memory instead of performing lengthy SD writes mid-emulation (which was causing long freezes in gameplay and general overall slowdown)
* Controller mapping via emu.cfg is now a reality, individually for different controllers for different players
* Basic multiplayer is now implemented!
* 8 VMUs are now connected!
* VMUs and emu.cfg and BIOS files now all reside in home folder /home/pi/.reicast and .reicast/data/ for the BIOS files
* More improvements but I forget ATM

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -18,7 +18,7 @@ function depends_reicast() {
 }
 
 function sources_reicast() {
-    gitPullOrClone "$md_build" https://github.com/gizmo98/reicast-emulator.git skmp/rapi2-audiofix
+    gitPullOrClone "$md_build" https://github.com/free5ty1e/reicast-emulator.git free5ty1e/rpi2/retropie-reicast-updated
 }
 
 function build_reicast() {
@@ -40,13 +40,25 @@ function install_reicast() {
 function configure_reicast() {
     mkRomDir "dreamcast"
 
-    # create bios dir. Copy dc_boot.bin and dc_flash.bin there.
-    mkdir $md_inst/data
+    # Create home VMU, cfg, and data folders. Copy dc_boot.bin and dc_flash.bin to the ~/.reicast/data/ folder.
+    mkdir -p "$home/.reicast/data"
 
     cat > $md_inst/reicast.sh << _EOF_
 #!/bin/bash
-pushd $md_inst
-sudo aoss ./reicast.elf -config config:image="\$1"
+pushd "$md_inst"
+echo Reading the entire Reicast emulator into memory to execute from there...
+sudo mkdir tmpfs
+#TODO: Find optimal smaller tmpfs size, I do not believe anywhere near this much is required.  I have only ever seen 54MB utilized during a game of Rush 2049.
+sudo mount -o size=450M -t tmpfs none tmpfs/
+sudo cp -v * tmpfs/
+cd tmpfs
+sudo aoss ./reicast.elf -config config:homedir="$home" -config config:image="$1"
+cd ..
+echo Ensuring any freshly-created VMUs are owned by pi and not root...
+sudo chown -R pi:pi "$home/.reicast"
+echo Freeing up memory...
+sudo umount "$md_inst/tmpfs"
+sudo rm -rf tmpfs
 popd
 _EOF_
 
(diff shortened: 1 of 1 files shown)


joolswills:
@gizmo98 Please can I get your feedback on this - as the repository has changed from yours etc.

scriptmodules/emulators/reicast.sh
```diff
@@ -40,13 +40,25 @@ function install_reicast() {
...
+sudo mount -o size=450M -t tmpfs none tmpfs/
+sudo cp -v * tmpfs/
+cd tmpfs
+sudo aoss ./reicast.elf -config config:homedir=/home/pi -config config:image="$1"
+cd ..
+echo Ensuring any freshly-created VMUs are owned by pi and not root...
+sudo chown -R pi:pi .reicast
+echo Freeing up memory...
+sudo umount /opt/retropie/emulators/reicast/tmpfs
+sudo rm -rf tmpfs
```
joolswills: Does this need to be absolutely pathed ? if so, you can use `$rootdir` but the mount didn't so the umount probably doesn't too if in the right folder.

joolswills:
Apologies for messing up my comments (some on your branch, some on the diff). I didn't notice :/

free5ty1e:
Ah, thanks for the info!  I wasn't sure what to use for the home dir instead of hardcoding, that's exactly what I needed... will update PR soon!

The tmpfs thing works great, and is cleaned up after emulator launch.  Unfortunately, Reicast falls back on file allocation for rpi2 since shm apparently does not work (at least, the way it's currently coded).. so the tmpfs thing allows this file allocation to take place in memory and actually makes the emulator playable.

It's a goal to figure out how to do this all within the main.cpp code of reicast, but for now this enables Dreamcast fun :D

joolswills:
Did you try `sudo mount -o remount,exec /dev/shm` 

see:
https://github.com/reicast/reicast-emulator/issues/601#issuecomment-74723710

joolswills:
Also there may be a possibility to modify the mprotect call in the code if that's what it is failing on.

free5ty1e:
Oh hey, that's great -- I'll try that out and update the PR again :D 

Thanks!!

free5ty1e:
Also, did you want these commits all squashed?

joolswills:
re commits - squashed would be better thanks.

joolswills:
reading the last of the thread linked above. looks like there can be a code fix to avoid the mount too - by calling http://man7.org/linux/man-pages/man2/personality.2.html checking what would be needed and if anyone else has already done it for this code.

free5ty1e:
Even better!

joolswills:
Something like

```
#include <sys/personality.h>
...
int pers, old_pers;
pers = personality(0xffffffff);
pers &= ~READ_IMPLIES_EXEC;
old_pers = personality(pers);
```

free5ty1e:
I can confirm that remounting SHM with the command `sudo mount -o remount,exec /dev/shm` does NOT fix the issues that the tmpfs mount fixes.  The frequent emulator hangs are back as the SD card is accessed during gameplay.  :(

Will try the personality fix after my appointment.  Thanks for the assist!!  :D

joolswills:
I guess we need to debug further - might be worth trying the code above anyway.

joolswills:
Aah they fixed it upstream on a branch

https://github.com/reicast/reicast-emulator/commit/14875df52b547351d37d050fbd4e6f0154b343cb

joolswills:
(assuming this is the same issue - which it may not be) - maybe worth diffing against that branch though.

free5ty1e:
That commit / those changes have been merged in to this branch, that doesn't fix the issue either.  I still have to try messing with the personality but at the moment, the PR as is works great :D

(I took the original branch we were building for RetroPie and rebased that on the latest upstream master, and then based my changes on top of that.)

free5ty1e:
I also have a pull request in with reicast upstream:
https://github.com/reicast/reicast-emulator/pull/666

free5ty1e:
Ah, the personality change you referred to has also already been merged in to this branch.  The following is how it is handled in the source:

```
    printf("Personality: %08X\n", personality(0xFFFFFFFF));
    personality(~READ_IMPLIES_EXEC & personality(0xFFFFFFFF));
    printf("Updated personality: %08X\n", personality(0xFFFFFFFF));
```

free5ty1e:
Squashed the commits.

joolswills:
Thanks.

free5ty1e:
Awesome!  Thanks again; I'm updating the Dreamcast retropie wiki nao :D

PR #872, opened 2015-06-16
shouldn't be any reason to run inputconfiguration as root - all confi…

joolswills:
…gurations should be in $user readable places now (and if not, we need to fix that). But certainly emulationstation and retroarch are

@petrockblog - wanted to run this one by you in case there was some reasoning for this that I missed ?


petrockblog:
Good that you found that - I forgot to remove that. Used it for debugging purposes to rule out access rights problems in the first place. But, I agree, in the end the input configurations must be un $user readable places.

PR #873, opened 2015-06-16
Reorg

joolswills:
Removed unused scripts and the esconfig module as we don't use them any longer.

Only the splashscreen are left in the root supplementary after these changes - perhaps they too can be moved to scriptmodules/supplementary/splashscreen/splashscreens also ? please let me know @petrockblog


joolswills:
I'm happy to leave the splashscreens where they are though. there is platforms.cfg in there too, I forgot about that.

petrockblog:
Good points!

PR #880, opened 2015-06-23
I think it is more logical that right shoulder does page down and lef…

joolswills:
…t does page up in es - in the way that left is back and right is often forwards in guis etc (at least for those using LTR languages)


McPace:
I agree. This change caught me off guard in beta 4. Feels backwards but it's not a terrible feeling

petrockblog:
Sure!

PR #883, opened 2015-06-26
moved capex.cfg and rominfo.fba to be located in FBAcapex_src directory

tednilsen:
moved capex.cfg and rominfo.fba to be located in FBAcapex_src directory
now pifba from source installs without failing.


tednilsen:
lr-nestopia: removed gtk3 stuff from Makefile

PR #884, opened 2015-06-28
rework and speed up processing of the input configuration

joolswills:
… xmlstarlet to parse the input xml in one go and load it into an associative array.

Previously the code run xmlstartlet 5 times for every button, for every inputconfig module.

We also skip checking whether the xml contains a button, as rather than hardcoding them, we get
them directly from the xml.

This brings processing time on a rpi1 from 27 seconds to under 5.

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/emulationstation.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/emulationstation.sh
@@ -8,153 +8,83 @@
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
 #
 
-#### input type: Joystick ###
+function onstart_emulationstation_joystick() {
+    local device_type=$1
+    local device_name=$2
 
-function onstart_inputconfig_emulationstation_joystick() {
-    local deviceType=$1
-    local deviceName=$2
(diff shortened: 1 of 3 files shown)


joolswills:
I have some more ideas planned - I need to test first, but I think the code could be simpler if the logic is handled in each module for each button, rather than inputconfiguration calling a function for each input.  Are you happy for me to re-work this further?

joolswills:
ok. the modules are now rewritten - I updated the documentation too.

petrockblog:
From looking at it and going through the code I like the changes. However, I have used the current beta 4, updated the script, pulled this pull request and tested it and it seems that the RetroArch config is not saved correctly. EmulationStation works - RetroArch not. I am looking at it right now and try to find the reason for that ...

petrockblog:
Ok, got one step further: I have found that `<inputAction type="on finish">...` is missing in `es_input.cfg`.

petrockblog:
`<inputAction type="on finish">` is erroneously removed after the first run of a controller configuration.

petrockblog:
Found it: It must say `if [[ ! -f "$es_conf" ]]; then` and not `if [[ ! -f "$conf" ]]; then`. I will fix that and merge.

joolswills:
aah me renaming variables and breaking stuff during the work. thanks for sorting it :)

PR #904, opened 2015-07-10
The Kodi installer adds a new Tab for Kodi to EmulationStation

cschlonsok:
Most users want to use Retropie and Kodi. If they want to install Kodi out of RetroPie, it would be nice to give Kodi a separated Tab.

### DIFF
--- a/scriptmodules/ports/kodi.sh
+++ b/scriptmodules/ports/kodi.sh
@@ -26,14 +26,34 @@ function install_kodi() {
 function configure_kodi() {
     echo 'SUBSYSTEM=="input", GROUP="input", MODE="0660"' > /etc/udev/rules.d/99-input.rules
 
-    mkRomDir "ports"
+    mkRomDir "kodi"
 
-    cat > "$romdir/ports/Kodi.sh" << _EOF_
+
+    # Install kodi simple theme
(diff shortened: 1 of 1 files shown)


robertybob:
+1 for the idea

HerbFargus:
I also like this idea.

taalas:
I would like that as well!

I do have an open Pull Request with the RetroPie fork of Emulationstation which adds the option to have direct launch items in the Emulationstation menu, meaning Kodi would start without the need for a sub-menu. Might be relevant here, I thought :)

polzy:
Hi,
i need OSMC :(  It's possible to chose kodi or OSMC ?

HerbFargus:
Also just for reference- I've added the imagery from above into the simple theme repo which simplifies the script. An updated script can be found here:

https://github.com/RetroPie/RetroPie-Setup/wiki/KODI#kodi-15

mcobit:
| i need OSMC :( It's possible to chose kodi or OSMC ? 

OSMC is a distribution using kodi. There is no choice ;-)
But there is a way to have both now:
https://discourse.osmc.tv/t/howto-retrosmc-retrogaming-on-osmc/6671

HerbFargus:
This may be more viable now that we have the direct launch function.

joolswills:
This is now implemented.

PR #915, opened 2015-07-24
Reload Setup Script after update automatically.

HerbFargus:
These changes will make it slightly more convenient when updating the setup script so that people don't have to exit out and then go back in again each time they update the setup script.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -175,7 +175,8 @@ function updatescript_setup()
     fi
     popd
     rp_callModule runcommand install
-    printMsgs "dialog" "Fetched the latest version of the RetroPie Setup script. You need to restart the script."
+    printMsgs "dialog" "Fetched the latest version of the RetroPie Setup script."
+    exec "$scriptdir/retropie_setup.sh"
 }
 
 function source_setup()


HerbFargus:
I realised after doing this that as I was compiling other things from the setup script near the end it seems to wig out with dialog and it doesn't clear the text when the end of compiling message comes on. The menu still works but its obscured by lots of blue and left over text. Perhaps my logic in the coding was incorrect.

scriptmodules/admin/setup.sh
```diff
@@ -175,7 +175,9 @@ function updatescript_setup()
     fi
     popd
     rp_callModule runcommand install
-    printMsgs "dialog" "Fetched the latest version of the RetroPie Setup script. You need to restart the script."
+    printMsgs "dialog" "Fetched the latest version of the RetroPie Setup script. Reloading the updated script now."
+    "$scriptdir/retropie_setup.sh"
+    exit 0 
 }
```
joolswills: This will launch the script again, whilst the old script is still loaded (the original script wont exit until the new one exits) - it would probably be better to use "exec" or something to avoid that

http://wiki.bash-hackers.org/commands/builtin/exec

HerbFargus:
thanks for the tip, I'll see what I can come up with.

HerbFargus:
That should do it- I even squashed the commits for you :P

joolswills:
Thanks.

scriptmodules/admin/setup.sh
```diff
@@ -175,7 +175,8 @@ function updatescript_setup()
     fi
     popd
     rp_callModule runcommand install
-    printMsgs "dialog" "Fetched the latest version of the RetroPie Setup script. You need to restart the script."
+    printMsgs "dialog" "Fetched the latest version of the RetroPie Setup script.
```
joolswills: Hvae you tested this - you are missing a quote.

HerbFargus:
It's always the little mistakes... It should be fixed now. Thanks for catching that.

joolswills:
thanks.

PR #916, opened 2015-07-25
Add Kodi 15 Isengard

HerbFargus:
Installs Kodi 15- before running the script people need to have already installed kodi 14.2 as that enables kodi 15 to meet its dependencies. Perhaps in the future HDMI CEC and Joystick support can be added. Also as an added benefit- this version lacks the black screen bug. Note that there may be some add-on issues as the newest version may not yet have support for some add ons.

### DIFF
--- a/scriptmodules/ports/kodi15.sh
+++ b/scriptmodules/ports/kodi15.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+# This file is part of RetroPie.
+# 
+# (c) Copyright 2012-2015  Florian Müller (contact@petrockblock.com)
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+# http://www.gtkdb.de/index_36_2176.html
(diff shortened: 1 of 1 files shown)


joolswills:
I would prefer to add kodi 15 as a debian package with dependencies once that is available (or if we are to do it ourselves, manage our ownh binary/source builds with dependencies included)

HerbFargus:
That's probably a good idea- that way if at all possible we would be able to enable hdmi-cec and joystick support since the build I'm using in this request left joystick support disabled while compiling. A package will probably be provided by kodi soonish considering that Kodi 15 has basically been released?

joolswills:
I'm pretty sure there will be a package available soon - but if not we could put together a source install perhaps with the options we need.

HerbFargus:
Sounds good to me. I suppose we can close this until then.

joolswills:
Also the debian repository we currently use has 15.0 (not final) in the unstable section - so I assume they will do a newer build soon (It's a build from beg July)

HerbFargus:
Fair enough, I'll wait on that repo then, and then if it lacks proper cec and joypad support then I guess we'll look into compiling our own. I'll close this pull request for now.

PR #918, opened 2015-07-25
Add lr-blueMSX primarily for MSX2

HerbFargus:
It also has an option in the core settings to run colecovision- the colecovision
games run fine but they are currently not properly coded for input so
its basically useless for now.


joolswills:
I just ran this and it's throwing errors in the configure - also the paths are not quite right. Looking into it.

PR #919, opened 2015-07-26
Add Miroof's Virtual Gamepad

HerbFargus:
script that installs a virtual gamepad from Miroof (https://github.com/miroof/node-virtual-gamepads) with node.js that enables people to use their smartphone as a gamepad. The code is probably not the most standardised but it was the only way I was able to figure out how to get it to work. if you have any suggestions for improvement I'm all ears.

### DIFF
--- a/scriptmodules/supplementary/virtualgamepad.sh
+++ b/scriptmodules/supplementary/virtualgamepad.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# This file is part of RetroPie.
+# 
+# (c) Copyright 2012-2015  Florian Müller (contact@petrockblock.com)
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="virtualgamepad"
+rp_module_desc="Virtual Gamepad for Smartphone"
+rp_module_menus="4+"
+
+function install_virtualgamepad() {
+    wget http://node-arm.herokuapp.com/node_latest_armhf.deb
+    dpkg -i node_latest_armhf.deb
+    rm node_latest_armhf.deb
+    gitPullOrClone $md_inst https://github.com/miroof/node-virtual-gamepads.git
+    cd $md_inst
+    npm install --unsafe-perm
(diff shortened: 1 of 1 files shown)


HerbFargus:
Ok, thanks for helping me figure it out. I've ammended it and tested it on a fresh build and it is working great!

joolswills:
please can you have a go at squashing these commits and pushing back to this pull request? :)

HerbFargus:
Ha ha so now you want me to squash them... Ha ha sure. I'll give it a go.

joolswills:
well, in this case it's more logical to do so - as it's a restructure in the second commit that isn't needed in the history etc.

joolswills:
basically, on your branch

```
git rebase -i 5d8dd2a8da7c230486396b3e311f20b6f6719122 (the commit before your first commit)
```
edit the second line and change it from pick to squash - save the file.

and you then have a single commit. and you can push -f back 

BTW you could do one more commit first and then squash, as you have some trailing whitespace in the file

joolswills:
You also have incorrect indentation in the install function.

scriptmodules/supplementary/virtualgamepad.sh
```diff
@@ -0,0 +1,29 @@
...
+
+rp_module_id="virtualgamepad"
+rp_module_desc="Virtual Gamepad for Smartphone"
+rp_module_menus="4+"
+
+function sources_virtualgamepad() {
+    wget http://node-arm.herokuapp.com/node_latest_armhf.deb
+    dpkg -i node_latest_armhf.deb
+    rm node_latest_armhf.deb
+}
```
joolswills: I would probably put this in the install section too - as it isn't source but a binary package.

HerbFargus:
When you are referring to the trailing whitespaces- were you referring to EOL? I'm editing things in notepad++ on a windows PC- should I change the EOL format to unix instead of windows?

joolswills:
I believe git is handling EOL conversion automatically, as the files end up as unix line endings at my end. I was referring to the extra spaces you had after `cd "$md_inst"` before the end of the line.

HerbFargus:
ok. that should do it I think. Thanks for being patient with me

joolswills:
Thanks!

PR #926, opened 2015-08-01
as we have additional package functions it would be useful to list th…

joolswills:
…em all. The changes here allow that by using compgen builtin. It does mean that functions will be listed alphabetically, but it means we can now see all the custom functions via retropie_packages.sh

Doing this one as a pull request to make sure you are happy with it @petrockblog 

it only affects the commandline output, but now for example shows:

```
sudo ./retropie_packages.sh

...

901/setup               : GUI based setup for RetroPie               : binaries configure depends experimental individual reboot source supplementary updatescript
```

previously it would only show if the module had any of a limited set of functions.

### DIFF
--- a/scriptmodules/packages.sh
+++ b/scriptmodules/packages.sh
@@ -42,10 +42,10 @@ function rp_listFunctions() {
     for idx in ${__mod_idx[@]}; do
         mod_id=${__mod_id[$idx]};
         printf "%d/%-20s: %-42s : " "$idx" "$mod_id" "${__mod_desc[$idx]}"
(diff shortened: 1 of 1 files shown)


petrockblog:
That is a nice enhancement. And I am getting to know some new bash tools :smile:

joolswills:
Thanks - It's slightly slower than before (fractionally) - but I have an idea to resolve functions earlier, and then parse later. Ill benchmark it to see what is best. I've uploaded a 3.0 btw final. But I will just try and fix a bug with one emulator lr-handy and then we can go 3.0 :)

PR #927, opened 2015-08-02
Create wifi.sh

HerbFargus:
Simple Wifi Script for those that don't wish to use Wicd-Curses. It could be more automated but it covers the basics for now until we can come up with something a little better.

Has options for WPA/WPA2, WEP, and Open Networks.

### DIFF
--- a/scriptmodules/supplementary/wifi.sh
+++ b/scriptmodules/supplementary/wifi.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+
+# This file is part of RetroPie.
+# 
+# (c) Copyright 2012-2015  Florian Müller (contact@petrockblock.com)
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="wifi"
+rp_module_desc="Configure Wifi"
+rp_module_menus="3+"
+rp_module_flags="nobin"
+
+function wifi_list() {
+    ifconfig wlan0 up
+    wifiNetworkList=$(iwlist wlan0 scan | grep ESSID | awk -F \" '{print $2}')
+    printMsgs "dialog" "$wifiNetworkList"
+}
+
+function wifi_wpa() {
+    cmd=(dialog --backtitle "$__backtitle" --inputbox "Please Enter the SSID of the Network You Would Like to Connect to:" 10 60 "$ssid")
+    choice=$("${cmd[@]}" 2>&1 >/dev/tty)
+    if [[ -n "$choice" ]]; then
+        ssid="$choice"
+    fi
+
+    cmd=(dialog --backtitle "$__backtitle" --insecure --passwordbox "Please enter the network password" 10 60 $psk)
+    choice=$("${cmd[@]}" 2>&1 >/dev/tty)
+    if [[ -n "$choice" ]]; then
+        psk="$choice"
+    fi
+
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. It does need a efw changes before being included though I think - Few points:

it would be clearer to use Here Documents rather than echo -e with it all in one line (http://tldp.org/LDP/abs/html/here-docs.html)

It should be possible to have it so it only modifies the wlan portions of interfaces, replacing any existing wlan block and backing up the previous file rather than overwriting the whole file. Perhaps you can clarify on the functionality as I would have though the essid etc can be included just in the interfaces file?

We could probably just show all networks, and then prompt user depending on which they try to connect to - rather than having multiple entries (Looks like there is some code duplication too).

I was planning on coming up with something in regards to the wifi functionality,bu I'm happy to use your script as a starting point, unless you are happy to make the above changes ?

HerbFargus:
I definitely agree it has room for improvement. I was attempting the in here docs but I couldn't quite get the variables to work out properly.

There are really two different ways we can go about modifying configurations, we can either modify interfaces directly or we can have interfaces reference wpa_supplicant.conf (like it is by default) 

Modifying wpa_supplicant.conf has the benefit of allowing multiple networks as well as automatically reconnecting if the network resets for some reason (as long as its setup with wpa-roam rather than wpa-conf)

I created a list of all the main potential configurations for WiFi setups with both just editing interfaces and for wpa_supplicant here: 

https://www.raspberrypi.org/forums/viewtopic.php?t=116023&p=797363

I agree it should be something simpler where one just selects from a list and types in a password and it would automatically determine the encryption and set the configs accordingly. Along with providing an option to back up the current configs. (The PiAssist script has something pretty similar https://github.com/Death259/PiAssist but it has a little room for improvement in determining encryption, etc.)

I currently lack the coding know how to accomplish all of that without the duplicate code, etc. Im still learning but its definitely above my head, So if you're able to do that it would be appreciated.

scriptmodules/supplementary/wifi.sh
```diff
@@ -0,0 +1,114 @@
...
+    echo -e 'ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev\nupdate_config=1\n\nnetwork={\n\tssid="'$ssid'"\n\tkey_mgmt=NONE\n}' > "/etc/wpa_supplicant/wpa_supplicant.conf"
+    ifdown wlan0
+    ifup wlan0
+    printMsgs "dialog" "You are now connected to: $ssid\n\nConfigurations have been saved to /etc/network/interfaces and /etc/wpa_supplicant/wpa_supplicant.conf"
+}
+
+function configure_wifi() {
+
+    local ip_int=$(ifconfig | grep -Eo 'inet (addr:)?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*' | grep -v '127.0.0.1')
+
```
joolswills: This will likely fail on a non English install as eg german ifconfig has "inet Adresse"

I spent some time looking into this before, and the best solution I came up with was

```
ip route get 8.8.8.8 2>/dev/null | head -1 | cut -d' ' -f8
```

which is what is used in the bashwelcometweak.

joolswills:
I'm happy to have a go - some good point regarding using the wpa_supplicant.conf - I guess we have to weight up keeping it simple vs functionality. Appreciate you are also learning as you go - it's a good start :)

Regarding Here Documents, variables will be substituted, unless the EOF is quoted, so if you want `$something` without it being replaced you need to quote the `$` `\$something` for example.

if you don't need any variables in the Here Document you can just do

```
cat > "somefile <<\_EOF_
not $variables here will be replaced
_EOF_
```

joolswills:
I've got some code together- starting from your script. Needs a bit more polishing - I'll submit a pull request for testing before it's committed though - as there are likely bugs.

HerbFargus:
Thanks for the tips on the inhere docs, that's useful. And for sure- I'll be happy to test it!

joolswills:
ok so, ill close this now and open another pull request in a mo. Cheers. thanks for the help.

PR #934, opened 2015-08-06
new module for editing of configs, and setting of common retroarch options

joolswills:
@biscuits99 @HerbFargus 

Please check out this configuration module and let me know what you think.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -158,6 +158,13 @@ function iniGet() {
     ini_value=$(sed -rn "s/^[[:space:]]*$key[[:space:]]*$delim_strip[[:space:]]*$quote(.+)$quote.*/\1/p" $file)
 }
 
+function editFile() {
+    local file="$1"
+    local cmd=(dialog --backtitle "$__backtitle" --editbox "$file" 22 76)
+    local choice=$("${cmd[@]}" 2>&1 >/dev/tty)
+    [[ -n "$choice" ]] && echo "$choice" >"$file"
+}
+
 function hasPackage() {
     PKG_OK=$(dpkg-query -W --showformat='${Status}\n' $1 2>/dev/null|grep "install ok installed")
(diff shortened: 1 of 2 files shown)


joolswills:
I'm interested in bugs, and configuration options that should be added / removed from the retroarch common settings. I added a bunch of video ones basically for now, not sure if others would be useful. Don't want to over complicate it either

HerbFargus:
Wow this is brilliant! I'm testing it right now- it seems to work quite well. I'm trying to think of the settings people want to change- but some simple video configs and terminal gui options for editing some manual controller configs is just about everything most people would need.

joolswills:
Thanks. if you are happy I can squash / merge it. Regarding configuring core settings, it can be done  via rgui, but it would be nice if rgui had a feature to just "save" the core settings. perhaps an upstream feature request.

HerbFargus:
it all looks good to me, I'm sure we can tweak it as needed as we see fit. 

The Rgui has much to be desired for saving settings as that seems to be the biggest issue for a lot of people- it's not very intuitive. It's definitely worth looking into.

joolswills:
Agree regarding RGUI. 

Thanks for the feedback.

biscuits99:
This is a fantastic feature, I love it. The tips at the bottom of the screen are very useful, definitely need to keep those.

Would it be a step too far to allow this tool to choose the shader?

Do the rp-dist files need to be listed though, I thought they should be kept clean?

This could be a great way to make it a more friendly project, as many complain at the first sight of a command line interface.

joolswills:
I could add shader stuff later perhaps, but I'm keen to get 3.0 out first :) Also after the shader cleanups have been done etc.

I left the rp-dist displayed for reference (if wanting to copy config data from them for example).

joolswills:
regarding the tips, they dont all fit, but you can read the full info via the help button. this seems to be a limitation of "dialog" - in regards to displaying only a single lines worth.

Issue #933, opened 2015-08-04
Changing some retroarch defaults


joolswills:
@HerbFargus @petrockblog @gizmo98 

feedback welcome.

I would like to change some default retroarch settings. I would like to turn off video_threaded by default, as it is detrimental to perfect video sync.

https://github.com/libretro/RetroArch/wiki/Getting-optimal-vsync-performance

>Threaded video fallback

>If your video driver has very bad performance, it is possible to run it on a thread to avoid almost all video driver overhead. Set video_threaded = true in config. Butter smooth VSync behavior in this case is impossible however, and latency might increase slightly. Use only if you cannot obtain full speed otherwise.

I think in the majority of cases it is better off. Most emulators are full speed on the pi1 with it off also from my testing. Or we could turn it off just for the pi2 if you think there are pi1 emulators that benefit from the setting.

Also I was thinking about pixel shaders. By default we have video_smooth off, and "some" emulators have some pixel shader configs in the system specific retroarch.cfg. Some users just want video smoothing on, and it's a pain for them as they need to edit lots of files. What would be easier would be to put the shader configs in, but left disabled, and those that want shaders can configure/enable - that way the core retroarch video_smooth option can be easily changed.

What we have currently for example is /opt/retropie/configs/gb/retroarch.cfg

```
input_remapping_directory = /opt/retropie/configs/gb/
video_smooth = false
video_shader = /opt/retropie/emulators/retroarch/shader/hq4x.glslp
video_shader_enable = false
```

What I propose 

```
input_remapping_directory = /opt/retropie/configs/gb/
# video_smooth = false
video_shader = /opt/retropie/emulators/retroarch/shader/hq4x.glslp
# video_shader_enable = true
```

so the user can easily enable the lines above if they want to use a shader. we could easily have a simple gui to then configure video_smooth on or off.

I prefer video_smooth on by default, if I'm lazy to configure shaders, as on modern LCD's without everything looks very blocky. but whatever the default is, it should be easy to switch.

joolswills:
My personal preference out of the box would be video smooth on and shaders off. then have a retroarch configuration tool to allow editing of core retroarch.cfg and have some presets also. but im happy to discuss the defaults. I think the scaling artifacts are more visible with video smooth off also.

HerbFargus:
[...]

PR #935, opened 2015-08-06
changing retropie defaults to allow easier switching of video smoothi…

joolswills:
…ng by only setting it in the core - and turning it on my default. This should look better for most users without having to configure shaders. with the new configuration tool this can easily be changed. also default threaded video to off for better sync. #933


joolswills:
@biscuits99 @HerbFargus @petrockblog 

Here's the default config changes.

Video smoothing on by default, but not overridden in per core configs. 

per core configs by default now look like

```
# Settings made here will only override settings in the global retroarch.cfg if placed above the #include line

input_remapping_directory = /opt/retropie/configs/gb/
# video_smooth = false
video_shader = /opt/retropie/emulators/retroarch/shader/hq4x.glslp
# video_shader_enable = true

#include "/opt/retropie/configs/all/retroarch.cfg"
```

and with the new config edit tool it is easy to switch video smoothing off for those that prefer it (and video threading on etc).

joolswills:
Whether smoothing is on or off by default could be debated, but for most users on modern LCD, i think it looks better on - as scaling artifacts are less obvious (imho), but at least now it can be switched easily for those that dont want to configure shaders for each platform.

HerbFargus:
looks good to me

biscuits99:
sounds good

robertybob:
Does this affect performance on RPi 1?

joolswills:
video threading off could perhaps reduce performance slightly, but from my own testing, it only made a slight different on emulators that were pushing the limits of the pi1 capability. So long as we document the changes it's not an issue to me - we can document how to use the new config editor etc.

Thankfully we have people like @HerbFargus and @biscuits99 to help with all that :)

joolswills:
Im ready to merge this and push out a new image. @robertybob please email me at buzz[at]exotica.org.uk and I will give you an image to test before release.

3.0 is too long coming.. you can never be bugfree. If testing from you guys turns up stuff, mayhbe a rc2, but its been sooo long since 2.6 - I really want people to use the newer versions. so much has changed. cheers

joolswills:
I'm going to merge and offer up a build to you guys.

joolswills:
I would also add, I read a post from a user that tried retropie and thought it looked all "blocky" and then switched to recallbox os. I'm still with the thinking that we should offer basic defaults but adhere to the needs of power users. the video smooth value is on by default on retroarch - and I think it's a better default. Now it can easily be switched, even from a controller.

HerbFargus:
As soon as I've tested the new image I'll overhaul the wiki again and update it to 3.0

joolswills:
thanks. let me know anyway - ill post it via g+ i guess, but you know the url, and should it pass your tests I'll send it on to @petrockblog for release. I will need to make a changelog also i guess, with the most important stuff between releases.

joolswills:
2.6 to 3.0 was too long.. it should have been 3.0 last year and 4.0 now..

HerbFargus:
Maybe this should be a separate ticket but I think maybe it would be a good idea to have a prioritised list of improvements that are reasonable to pursue for future releases.

joolswills:
I think a wishlist is ok, but I may be retiring - well not really, but I need to put some more love into http://www.xbmc4xbox.org.uk ...

joolswills:
maybe a spreadsheet would be useful though

HerbFargus:
Sure, I'll set one up and open up a ticket. (even if you take a leave of absence for a bit, 3.0 is a pretty good place to be. thanks for all the work :) )

robertybob:
@joolswills Thanks for the offer - I don't currently have the time to do any testing (it's a struggle to get a lot of time to play tbh!) however I still follow progress regardless :)   I trust your judgement that it's a worthwhile addition, with minimal slowdown, for the RPi1

PR #937, opened 2015-08-07
Add video splashscreen (omxplayer)

HerbFargus:
added video splashscreen support. I had to rename it splashscreen from asplashscreen as the dependencies for omxplayer aren't met being first alphabetically in /etc/init.d, it gives the error `waiting for dbus address to appear`. So we can keep it named asplashcreen if we can get the right required-start perametre correct in the script. The difference in time loading for the splashcreen image is that of only a couple seconds.


joolswills:
thanks - actually we should probably fix up the dependency based boot stuff. but the rename is fine anyway for now. Unfortunately it looks as though you may have a character encoding issue in the file though - at least via the github gui (check the copyright).

HerbFargus:
Might also be worth integrating this randomiser in some form or fasion: https://github.com/sur0x/retropiesplashscreen

joolswills:
Did you do a git mv before editing on asplashscreen ? Looks like git's rename detection hasn't worked (or did you rm / git add a new file) ?

I think I'm going to leave this for now I'm afraid - for similar reasons as https://github.com/RetroPie/RetroPie-Setup/pull/668 as I want to improve the main module before I mess around with the script.

joolswills:
I also want to look into the init.d script in more detail and do further testing.

joolswills:
It might be enough btw not to rename it but just to add "dbus" to Required-Start - which is one of the things I wanted to test. if renaming we are going to have to clean up any old copies of splashscreen etc.

HerbFargus:
I would also prefer to not have to rename it as it would be much cleaner. Adding just dbus to the required start perametre isn't enough I'm afraid as it still gives the same errors, I'll keep testing today and see if I can figure out a solution, if I figure it out I'll submit a new pull request.

joolswills:
K. Cheers :)

HerbFargus:
Ok I've figured it out- I found it easier to throw a sleep perametre on omxplayer than to sort out the boot depencies. One thing that I'm curious about: in splashscreen.sh there is the line `update-rc.d asplashscreen defaults` which corresponds to runlevel Start (2 3 4 5) Stop (0 1 6) yet the `Default-Start:` perameter is S so its inconsistent (if we changed the default start perameter to the defaults it wouldn't load until the very end of the boot sequence), it would make more sense to use `update-rc.d asplashscreen start 01 S`

PR #938, opened 2015-08-08
Add Nickname to netplay module

HerbFargus:
Coded it so that people can have names with spaces.  The nickname is
really an optional aesthetic thing but I think it's a nice touch. 

One more setting to note is "netplay_client_swap_input" it's a simple true false perametre (default
would be true) that allows users to both use player one controls on
their respective devices even if they are player two in the game (but only for each
user which means you can't both be playing the same player one- just
using player one joypad controls) 

If it isn't already on by default, it should be- it probably isn't necessary to add to the module


joolswills:
thanks.

PR #939, opened 2015-08-08
Add omxplayer video splashscreen

HerbFargus:
and fixed the update-rc.d to match the runlevel in asplashcreen

tested it on RPi 2 and RPi 1- I had to up the sleep perameter to 5 before omxplayer for the rpi 1 because of the slower boot time, you can get away with making it a 3 for the RPi2 if you want the video to boot faster. if you want the video to play all the way through without emulationstation cutting it off you can change `do-start &` to `do-start` in asplashscreen

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -14,7 +14,7 @@ rp_module_menus="3+"
 rp_module_flags="nobin"
 
 function depends_splashscreen() {
-    getDepends fbi
+    getDepends fbi omxplayer
 }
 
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks.

joolswills:
Seems to work with udev as dependency - which should be better than a sleep as that could technically fail based on speed of sdcard etc.

joolswills:
at least it worked once, but not again. I'm going to dig further with this I think

HerbFargus:
Yeah when I used dependencies they seemed to work once... and then never again. I got the same results- quite irksome really.

PR #940, opened 2015-08-09
remove delay from omxplayer start and fix up depdencies

joolswills:
@HerbFargus please try this - please run the script again, so it updates it, and then do enable.

This is the last thing to sort before the release...

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -27,8 +27,7 @@ function default_splashscreen() {
 }
 
 function enable_splashscreen() {
-    # This command installs the init.d script so it automatically starts on boot
-    update-rc.d asplashscreen start 00 S &>/dev/null
+    insserv asplashscreen
 
     # not-so-elegant hack for later re-enabling the splashscreen
     update-rc.d asplashscreen enable >/dev/null
--- a/scriptmodules/supplementary/splashscreen/asplashscreen
+++ b/scriptmodules/supplementary/splashscreen/asplashscreen
@@ -1,7 +1,7 @@
 #! /bin/sh
 ### BEGIN INIT INFO
 # Provides:          asplashscreen
-# Required-Start:    dbus
+# Required-Start:    $local_fs
 # Required-Stop:
 # Default-Start:     S
 # Default-Stop:
@@ -10,16 +10,17 @@
 ### END INIT INFO
 
 do_start () {
-    while read splashline; do
-        echo "Playing splash video or image $splashline"
-        if $(echo "$splashline" | grep -q ".avi\|.mov\|.mp4\|.mkv\|.3gp\|.mpg"); then
-            sleep 5
-            omxplayer "$splashline"
-        else
-            /usr/bin/fbi -T 2 -once -t 10 -noverbose -a "$splashline"
-            sleep 12
-        fi
-    done </etc/splashscreen.list
+    local config="/etc/splashscreen.list"
+    local line="$(head -1 "$config")"
+    if $(echo "$line" | grep -q ".avi\|.mov\|.mp4\|.mkv\|.3gp\|.mpg"); then
+        omxplayer -b "$line"
+    else
+        local count=$(wc -l <"$config")
+        [ $count -eq 0 ] && count=1
+        [ $count -gt 20 ] && count=20
+        local delay=$((20/count))
+        fbi -T 2 -once -t $delay -noverbose -a -l "$config" >/dev/null 2>&1
+    fi
     exit 0
 }
 


joolswills:
The other thing is currently there is a 12 second delay before another image is shown - and now that only a single image is passed to the script at a time, the slideshow functionality with timeout doesn't work correctly, so the previous patches have sort of broken multiple image functionality..

joolswills:
im inclined to change it so it can have a single movie or multiple images.. My bad for not thoroughly checking the patchset.

HerbFargus:
yeah I can do that, I'm just finishing up squashing some commits for my most recent pull request (to add art to videopac and sg-1000) and then I'll give this a go

HerbFargus:
seems to be working fine on my rpi2

joolswills:
I have made some further changes - Would you mind trying it again with the new code. Works here. I have changed Required-Start - I think it starts slightly earlier now. It should work with multiple images (displayed for 5 seconds each), and for a single image it will be shown for 25 seconds (5 seconds + 20 for last image).

HerbFargus:
it froze on me- might have a corrupted sd card. multiple splashscreens worked, I'm just testing out the video splash now

HerbFargus:
I can't seem to get the video splashscreen to boot.

joolswills:
Looks like the Required-Start change isn't working after all - nothing is output because I was redirecting errors. Reverted - it wasn't working here either so no idea how it was before - I am pretty tired though.

joolswills:
Would you mind giving it another go - sorry.

HerbFargus:
yep testing it right now

joolswills:
There are still problems I'm afraid - I am trying to end fbi processes if there are more to display, and then wait for the last one for longer. However, it seems to sometimes leave the system in a bad state and if emulationstation is not running, you can see the shell prompt. Thinking about another workaround.

HerbFargus:
yeah... with the revert the video splash is working now at least.

HerbFargus:
I'm testing multiple images right now.

HerbFargus:
multiple images work too.

HerbFargus:
when I enable custom splashscreens and then go to choose a video it sometimes freezes on me... I wonder if its my pi or if it has something to do with the processes of splashscreen.sh

joolswills:
when you say crashed - is anything displayed ? can you ssh in ? it might just be the framebuffer left in a certain state - also please clarify on what you mean by "choose a video" ?

HerbFargus:
the dialog menu just freezes- so I can't do anything. I choose enable custom splashscreen and then when am going through the list to choose my video splashscreen it just freezes randomly.

joolswills:
it may be related to the parsing/creation of the list - put set -x in the script and you will see the output if its in an infinite loop.

HerbFargus:
actually scratch that- its happening in any of my menus. would this be a sign of a corrupted sd card? or would it have something to do with the splashscreen stuff?

joolswills:
no idea - check the output of dmesg etc.

joolswills:
you could disable the splashscreen stuff and see if it makes a difference.

HerbFargus:
yeah once disabled it works find so it has something to do with the asplashscreen code

joolswills:
I have some new code almost ready - i believe fbi was not exiting cleanly in the way I was trying to manage it before.

joolswills:
ok. I give up trying to make it clever for now. Back to old behaviour, with the video support, so you can have a single video or you can have a few pics. the delay between the pics is set to 20, but i should probably make it 20/number of pics, so more pics, the quicker they are displayed etc.

joolswills:
couple more fixes coming.

joolswills:
ok. really hope it's ok now. functionality is much the same as before, but the timing is better - the splash is shown for 20 second for an image, but if two images, 10 seconds each - but using the slideshow mode as before.

if the first line is a video, only the video is shown. we do not process each line by line and send to FBI as it is problematic.

HerbFargus:
ok, testing with a video now

joolswills:
Thanks.

HerbFargus:
video works. now testing with multiple images

HerbFargus:
my rpi2 boots too fast to see any more than one image if I only have two images on the list- but it isn't a big deal, as I'll only ever use one image or one video. I'll throw on a bunch of images just for kicks though and see what it does

joolswills:
yeh emulationstation shows up in front - if you disable that you get to see more. on a pi1 it can take up to 20 seconds. ideally we would have it so its shown, then killed when the machine is finished, but fbi doesn't like being "killed". Anyway, I don't consider the splash important and would rather work on other things, so I will not revisit this anytime soon I hope. Time to build images again.. about the 6th time today :)

joolswills:
thanks for the help.

HerbFargus:
with 6 on the list it shows two before emulationstation boots so the rpi 2 would just have to have a huge list, but its all good- its more for people with an Rpi 1 anyhow. All that's left is to add a randomiser :P but we can look into that another time. Thanks for giving up your sleep to help get this to work. I appreciate it. 

it's about time 3.0 is finally released :)

joolswills:
Those that want it random can put

```
sudo shuf /etc/splashscreen.list -o /etc/splashscreen.list
``` 

into /etc/rc.local - and be done with it :)

HerbFargus:
ha ha fair enough. thanks for the tip

PR #941, opened 2015-08-10
start splash earlier, and wait for fs to be writable for video (requi…

joolswills:
…red by omxplayer), but not for images

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -28,13 +28,10 @@ function default_splashscreen() {
 
 function enable_splashscreen() {
     insserv asplashscreen
-
-    # not-so-elegant hack for later re-enabling the splashscreen
-    update-rc.d asplashscreen enable >/dev/null
 }
 
 function disable_splashscreen() {
-    update-rc.d asplashscreen disable &>/dev/null
+    insserv -r asplashscreen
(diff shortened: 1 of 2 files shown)


joolswills:
@HerbFargus really sorry, but please could you test this - I start earlier now, wait for fs to be writable for omxplayer. seems to work here. you might want to do inserv -r asplashscreen / insserv asplashscreen

HerbFargus:
Ha ha no worries happy to do it. I'll test it right now.

joolswills:
video shoudl work but start around the same time as before. images should appear a little earlier - before the root disk check as it works when the disk is read-only

HerbFargus:
That's Brilliant! it works great :) now I see 3 out my list of 6 and the video boots a little faster too I think

joolswills:
yay! :D awesome.

HerbFargus:
I'll probably add another option in the menu to automate the randomiser at some point in the future- just for those that want it.

joolswills:
we should move the splashscreen list to config area, and we can have another setting for the randomiser that the startup script looks for.

HerbFargus:
That sounds reasonable

joolswills:
but after 3.0 of course :)

HerbFargus:
of course :P It will be a part of the new menu options

PR #948, opened 2015-08-13
new module for theme installation from RetroPie github repositories

joolswills:


HerbFargus:
looks good mate, I'll give it a go, and if I find any other issues with the themes like I did with zoid, I'll patch them up and submit the fixes

HerbFargus:
I think you may have the installed/ not installed thing backwards. Once I've installed the theme it says (not installed) and if it isn't installed it says (installed)

joolswills:
Tis already fixed :) I'm still testing etc btw so there might be more changes.

HerbFargus:
yeah it looks like there may be some issues with nbba too, so I'll look into that and see what it might be missing. it just shows up as everything being blank right now

HerbFargus:
Looks like NBBA has a bit more than is necessary at the moment, and may take a bit of tweaking before its ready for downloading. I can contact the creator of it and see if we can get the repo into a working state.

joolswills:
I'll disable for now then in the script.

joolswills:
Everything else look ok (now that the simple-dark is renamed) ?

HerbFargus:
yeah it all seems to be working on my end- feel free to confirm
(emulationstation will of course need a refresh if the theme is changed from the start menu)

joolswills:
thanks for testing.

HerbFargus:
Thanks for fixing my horrendous code :P

PR #959, opened 2015-08-15
Create bluetooth.sh

Death259:
Adding capability to add/remove/list Bluetooth devices

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -1,0 +1,142 @@
+#!/usr/bin/env bash
+
+# This file is part of RetroPie.
+# 
+# (c) Copyright 2012-2015  Florian Müller (contact@petrockblock.com)
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - I will test, and possibly make a few tweaks, and them commit.

joolswills:
(But generally looks good - thanks) :)

joolswills:
Don't worry about doing the above changes apart from the logic as I would like clarification  - just put here for information - have made most changes already but still testing. I think the connect logic / error handling could be improved and I would like it to generate a pin for keyboards etc.

joolswills:
merged with changes as part of https://github.com/RetroPie/RetroPie-Setup/commit/72d7d8a5af303cfd04de3ff0b1ec01e41e145773

Would like to handle the pin code via a dialog - giving the user the choice to choose a random pincode (for pairing a keyboard), or entering a pincode (for other devices that might need a specific code). that can then be piped to bluez-simple-agent. The code also currently can say something is paired when it isn't in the case the pincode fails etc.

joolswills:
Thanks for the contribution.

PR #968, opened 2015-08-18
Fixed paths of Super Mario War for RetroPie

HerbFargus:
I created a new repo with a more complete commit history as well as
reprogrammed the splashscreens to be raspberry pi specific and also
fixed the paths so all game files are installed into
/opt/retropie/ports/smw

### DIFF
--- a/scriptmodules/ports/smw.sh
+++ b/scriptmodules/ports/smw.sh
@@ -13,11 +13,11 @@ rp_module_desc="Super Mario War"
 rp_module_menus="4+"
 
 function sources_smw() {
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

HerbFargus:
I just tested it from a fresh image just to be sure, and it works great. so since the paths have changed does that now mean that binaries would be able to be built?

joolswills:
yep. I will test it, build binaries, and then move it out of experimental - but not right now as I'm off to bed ;-)

HerbFargus:
goodness mate-- yeah thats a great idea :P thanks :)

PR #970, opened 2015-08-19
add splashscreen directory to samba shares

HerbFargus:


joolswills:
I was planning to move the splashscreens folder, but I guess it's not a bit deal and it can stay where it is. Will merge. thanks.

HerbFargus:
perhaps we could consolidate them sometime for a future release but for the time being this  is a simple fix

HerbFargus:
out of curiosity were you planning on moving the splashscreens and splashscreen.list to the home folder?

joolswills:
I was thinking about ~/Retropie/splashscreens but it doesn't really matter - it's fine to leave them where they are.

joolswills:
I have changed the location of the user splashscreens now. Rather than share the /opt/retropie/supplementary/splashscreens folder (which is not writable by pi by default), we make a ~/RetroPie/splashscreens folder for users to use. Avoids users changing the system splashes too, which could break later git updates due to local modifications etc.

HerbFargus:
That is very wise. I approve.

McPace:
YEAH!

PR #975, opened 2015-08-20
Add .prg to Commodore 64

HerbFargus:
verified here:
http://blog.petrockblock.com/forums/topic/vice-issues-with-prg-files-and-drive-sound-emulation/


joolswills:
thanks.

Issue #945, opened 2015-08-11
Add Sselph's Scraper as a module


HerbFargus:
Mostly completed- just needs some work on a checklist to choose scrapers by system (function scraper_system)

```
#!/usr/bin/env bash

# This file is part of RetroPie.
# 
# (c) Copyright 2012-2015  Florian Müller (contact@petrockblock.com)
# 
# See the LICENSE.md file at the top-level directory of this distribution and 
# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
#

rp_module_id="scraper"
rp_module_desc="Steven Selph's Scraper"
rp_module_menus="3+"

function scraper_install() {
echo "Downloading and Installing Scraper created by SSELPH..."
scraperVersion=$(wget -qO- https://api.github.com/repos/sselph/scraper/releases/latest | grep tag_name | sed -e 's/.*"tag_name": "\(.*\)".*/\1/')
    if isPlatform "rpi2"; then
        wget https://github.com/sselph/scraper/releases/download/$scraperVersion/scraper_rpi2.zip -q
        unzip scraper_rpi2.zip scraper -d /usr/local/bin/
        rm scraper_rpi2.zip
    else
        wget https://github.com/sselph/scraper/releases/download/$scraperVersion/scraper_rpi.zip -q
        unzip scraper_rpi.zip scraper -d /usr/local/bin/
        rm scraper_rpi.zip
    fi
    printMsgs "dialog" "$scraperVersion Scraper Installed"
}

function scraper_all() {
su pi -c "/usr/local/bin/scraper -scrape_all -thumb_only"
printMsgs "dialog" "All games scraped"
}

function scraper_system() {
#show scraper menu - ask user which console they would like to scrape					
romFolderList=$(grep "<path>" /etc/emulationstation/es_systems.cfg | sed "s/<path>//g" | sed "s/<\/path>//g" | sed "s/~/\/home\/pi/g" | awk '{print $0; print NR; print "OFF";}')
esSystemsList+=$romFolderList
    romFolders=$(whiptail --backtitle "PiAssist" --checklist "Select ROM Folders" 0 0 0 $romFolderList 3>&1 1>&2 2>&3)
    exit_status=$?
	romFolderAcutallySelected=true
	case $exit_status in
	$DIALOG_CANCEL)
	romFolderAcutallySelected=false
	;;
	$DIALOG_ESC)
	romFolderAcutallySelected=false
	;;
	esac
[...]

PR #985, opened 2015-08-29
initial version of module to use Steven Selph's scraper #945

joolswills:
@HerbFargus please give this a go - no config options yet, but it can go in as a basic scraper and be improved later. i've only tested a little bit.

### DIFF
--- a/scriptmodules/supplementary/scraper.sh
+++ b/scriptmodules/supplementary/scraper.sh
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+
+# This file is part of RetroPie.
+# 
+# (c) Copyright 2012-2015  Florian Müller (contact@petrockblock.com)
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="scraper"
+rp_module_desc="Scraper for EmulationStation by Steven Selph" 
+rp_module_menus="4+configure"
+rp_module_flags="nobin"
+
+function install_scraper() {
+    local ver=$(latest_ver_scraper)  
+    mkdir -p "$md_build"
+    local name="scraper_rpi.zip"
+    isPlatform "rpi2" && name="scraper_rpi2.zip"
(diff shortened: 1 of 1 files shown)


HerbFargus:
will do

HerbFargus:
So I take it you opted for .emulationstation storage by default. on inital tests it all seems to work fine. 

I'm working on a theme right now with rookervik that will be meant for use with screenshots instead of thumbnails so at least those two options would be good- I don't know how many other people use different settings. I know thumbnails is preferable for the pi in general just to keep the memory usage down.

HerbFargus:
except images don't show...

joolswills:
I probably made an error with a path. I'll check / debug

HerbFargus:
the other metadata shows up fine

joolswills:
I just scraped mame-mame4all and the images are showing up here. I may need to see copies of your gamelists and have you verify if images were downloaded etc ~/.emulationstation/downloaded_images

HerbFargus:
Yeah they were downloaded, its possible I have conflicts from the manual install i did of sselphs scraper before this. I'm reimaging with a fresh 3.0 image just to make sure.

HerbFargus:
Ok must have been something with my build. works perfectly on a fresh image

joolswills:
I'm going to merge it now, and then do improvements later then (configs etc and so on - and tweaks of parameters). Will be good to get feedback from users. Thanks for the help.

HerbFargus:
No thank YOU! :)

robertybob:
I think now that new themes are being made, that the thumbs option shouldn't be the default. With more users running the RPi2 and some themes being image-only (no release date, details etc), I think scraping the largest size photos possible is the best idea. ES will always resize them down after all. 

Tbh I scrape using the Windows version of sselph's scraper so never use any -flags, then I use an image resizer software to make all images the same size (I'm OCD like that). What I've found is that not all images are the same size on TGDB and it can take a couple of months for replacement images to me moderated. 

Perhaps in the future there could be an option of resizing images through RetroPie - we could come to some kind to agreement here on idea sizes for each console and have one option per system.

joolswills:
The ticket would be a better place as this request is now closed :)

PR #990, opened 2015-09-01
MAME4All: Add support for platform specific compile

MattWatt:
Platform specific compilation.  Use 4 cores for compiling on RPi2.  Pass platform parameter to pending modification to MAME4All Makefile.

### DIFF
--- a/scriptmodules/emulators/mame4all.sh
+++ b/scriptmodules/emulators/mame4all.sh
@@ -17,7 +17,7 @@ function depends_mame4all() {
 }
 
 function sources_mame4all() {
-    gitPullOrClone "$md_build" https://github.com/RetroPie/mame4all-pi.git
+    gitPullOrClone "$md_build" https://github.com/MattWatt/mame4all-pi.git "make-optimizations"
 }
 
 function build_mame4all() {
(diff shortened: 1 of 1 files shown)


joolswills:
parralel compile is handled by retropie build system. we dont use 4 cores due to memory issues, but it can be overridden with MAKEFLAGS if you want.

scriptmodules/emulators/mame4all.sh
```diff
@@ -22,7 +22,11 @@ function sources_mame4all() {
 
 function build_mame4all() {
     make clean
-    make
+    if isPlatform "rpi2"; then
+        make PLATFORM=rpi2 -j 4
+    else
+        make PLATFORM=rpi1
+    fi
     md_ret_require="$md_build/mame"
```
joolswills: see https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/system.sh - you can build on all 4 cores by using MAKEFLAGS as needed. this also handles the compiler flags for platform.

joolswills:
to note - for building binaries for distribution, i use -j10 in an emulated chroot, + with arm targetted distcc on a couple of machines. Just do sudo MAKEFLAGS="-jX" ./retropie_packages.sh etc - if you want to rebuild mame4all specifically with 4 threads do

```
sudo MAKEFLAGS="-j4" ./retropie_packages.sh mame4all
```

you can use __platform=rpi1 / rpi2 to force target platform (useful in emulated environment)

PR #1002, opened 2015-09-11
Command-line/SSH RetroPie system config

Fordi:
I wanted to be able to configure systems and offer advice to do the same without launching EmulationStation, so I broke runcommand.sh into a set of library functions, used what I needed for console-config.sh, and caused the installer to link it to /usr/bin/console-config.  

Additionally, I modified main_menu to check if it has a rom, and not show rom-related options if not.

Pull if you would like the same flexibility.


joolswills:
I think splitting it to that degree will make it a lot more work to maintain - it would be better to integrate a parameter into runcommand itself perhaps rather than split it up, so you can just call it with a config parameter for a system / rom.

Note: You have introduced some tab indentation also - we have a styleguide here which might be of use https://github.com/RetroPie/RetroPie-Setup/wiki/Shell-Style-Guide

PR #1006, opened 2015-09-16
add .z5 to frotz

HerbFargus:
does zmachine need to be added to platforms.cfg?


joolswills:
It could be yeh. Not massively important though.

joolswills:
cheers.

PR #1012, opened 2015-09-21
add 3 more complete themes to the theme installer

HerbFargus:
There are few more tweaks I'll make on logos on and whatnot but it is
fully functional and runs over 50 systems on my rpi2 with default memory
splits.


joolswills:
thanks!

if/when carbon is complete we should have the discussion about making it default with @petrockblog - as it would help with all the out of memory problems users have with too many systems etc.

HerbFargus:
That sounds good to me. We're working out a few issues with offsets on the gamelist but we're hoping it should be completed within the week. I'll open a ticket on github when its finished.

PR #1025, opened 2015-09-29
revert permissions back to user pi to enable savestates in GPSP

HerbFargus:
I couldnt figure out how to get gpsp to not overwrite the romdir.txt
created so for the time being this PR just enables savestates but they
are all currently saved in /opt/retropie/emulators/gpsp. If people wish
for their configs to be saved they need to exit from the F10 menu rather
than pressing escape.

### DIFF
--- a/scriptmodules/emulators/gpsp.sh
+++ b/scriptmodules/emulators/gpsp.sh
@@ -42,6 +42,7 @@ function install_gpsp() {
 
 function configure_gpsp() {
     mkRomDir "gba"
+    chown $user:$user -R "$md_inst"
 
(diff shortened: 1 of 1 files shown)


joolswills:
I would like to look into this further before switching ownership back - will get back to you.

joolswills:
There is something wrong with its loading of romdir.txt - even if I lock the file, it ignores the contents. Will merge this thanks.

HerbFargus:
Yeah I had the same issues unfortunately. Should suffice for now. Really gpsp should be the last emulator used anyhow. Thanks.

PR #1035, opened 2015-10-06
mupen64plus-testing: add GLideN64 video plugin

gizmo98:
Runs best with 640x480 and RPI2 OC. C++11 code needs gcc-4.8 or higher.


joolswills:
Thanks.

PR #1036, opened 2015-10-07
Add Omnija's Turtle Pi theme

HerbFargus:
This theme is also fantastic as it also doesnt have the white screen of
death issue- plus who doesn't like turtles?


joolswills:
thanks.

PR #1041, opened 2015-10-10
mupen64plus: raspbian jessie compatibility

gizmo98:
Add default gcc version check. If default version is 4.7 (whezzy)
install and use gcc-4.8. Else use default gcc version (jessie=4.9).


joolswills:
Looks good - thanks! :)

PR #1045, opened 2015-10-12
add sudo to darkplaces script

HerbFargus:
from here:
http://blog.petrockblock.com/forums/topic/darkplaces-quake-not-functioning-correctly-when-launched-from-ports/#post-107736

dont know why this makes it work but it does. thanks to ronoh for the idea


joolswills:
Thanks - although I would prefer not to run it as root, but rather diagnose why it doesn't work without (eg folder permissions).

joolswills:
Could try changing the config, or the folder to be owned by pi and see if that works, or just open a ticket and I'll look into it. odd that it works fine when not launching from ES though.

Issue #654, opened 2015-02-15
After pairing one PS3 controller, the second cannot be paired (Solutions Included)


kasgore:
We were able to identify several bugs and find fixes for them in the following forum conversation.
http://blog.petrockblock.com/forums/topic/multiple-ps3-controller-via-bluetooth-frustration/#post-87517

List of bugs:
1. Cannot pair a second PS3 controller after pairing the first one
2. Bluetooth adapter is in state down after fresh image copy to SD card from download of 2.5
3. Not enough time to pair bluetooth controller before emulationstation boots and you get the "no controllers detected" which prevents keyboard free operation and convenience. 

Fixes

Issue 1
Before pairing the first controller if you run "hciconfig" the mode is "Up Running pscan:. After pairing the first controller the bluetooth adapter exits discovery mode and hciconfig then only reads "up running"

Fix 1
If you change the mode to PISCAN by running "sudo hciconfig hci0 piscan" after pairing the first controller the bluetooth stays in the discovery mode and it works as expected for the pairing of the second controller.
I fixed this by inserting "hciconfig hci0 piscan" into the rc.local file just before the exit 0 line.

Issue 2
After a fresh install to the SD card and the first boot the bluetooth adapter is in the state down if you run "hciconfig" this could be confusing to new users and prevents a trouble and hassle free install if the intent is to use a bluetooth controller

Fix 2
If you run the command "sudo hciconfig hci0 up" this issue is simply fixed so I am guessing before the image was imaged the bluetooth adapter simply needs to be told to be up so when users install the image it will continue to be that way.

Issue 3
Especially with the RPI2 there is simply not enough time to pair the controllers before the "no controllers connected" dialogue pops up. I propose there is a simple option to enable or disable the popup or another workaround is done, this prevents the user from being able to sit back on the couch with wireless controllers and no keyboard on the PI to hit F4 and pair and then run the emulationstation comand

Fix 3
A simple workaround we found was to setup the keyboard as a controller and then even after the keyboard is removed the popup never comes up again. I am not sure this is the best fix but it does work.

free5ty1e:
[...]

PR #1054, opened 2015-10-18
ps3controller: fix two controller issue

gizmo98:
@joolswills please take a look if everything is fine. I tested these changes on a fresh RP3.1 installation and with two controllers. Rebooted five times. Disabled and enabled controllers at runtime. Controllers always paired. Issue https://github.com/RetroPie/RetroPie-Setup/issues/654 should be solved now.

ps3helper.sh and ps3pair.sh could be merged somehow because they share 90% code.

1. Workaround second controller pairing issue. Bluetooth dongle is no
longer discoverable after first controller is paired. Reenable pscan
with a udev rule (if virtual input device bt controller shows up
reenable pscan).

2. Cleanup module from old stuff which is no longer necessary because
udev rules manage everything.

3. Add some checks to ps3pair.sh. Check if sixad is running and enable
bluetooth dongle or pscan if necessary.

### DIFF
--- a/scriptmodules/supplementary/ps3controller.sh
+++ b/scriptmodules/supplementary/ps3controller.sh
@@ -69,15 +69,50 @@ _EOF_
     chmod +x "$md_inst/bluetooth.sh"
 
     # If a PS3 controller is connected over usb check if bluetooth dongle exits and start sixpair
-    cat > "$md_inst/ps3pair.sh" << _EOF_  
+    cat > "$md_inst/ps3pair.sh" << _EOF_
 #!/bin/bash
 if hcitool dev | grep -q "hci0"; then
-    hciconfig hci0 pscan
-    $md_inst/sixpair
+    # Check if sixad is running
+    if service sixad status | grep -q "sixad is running"; then
+        # activate bt dongle if necessary
+        if !(hciconfig | grep -q "RUNNING"); then
+            hciconfig hci0 up
+        fi
+        # Make bt dongle discoverable
+        if !(hciconfig | grep -q "PSCAN"); then
+            hciconfig hci0 pscan
+        fi
+        # Write bt dongle's mac address into controller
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/ps3controller.sh
```diff
@@ -89,6 +124,7 @@ _EOF_
     cat > "/etc/udev/rules.d/99-sixpair.rules" << _EOF_
 # Pair if PS3 controller is connected
 DRIVER=="usb", SUBSYSTEM=="usb", ATTR{idVendor}=="054c", ATTR{idProduct}=="0268", RUN+="$md_inst/ps3pair.sh"
+SUBSYSTEM=="input", ATTR{name}=="PLAYSTATION(R)3 Controller", RUN+="/opt/retropie/supplementary/ps3controller/ps3helper.sh"
```
joolswills: The path /opt/retropie should not be hardcoded.

joolswills:
As you say those scripts are virtually identical, so they should be merged into one - you should be able to pass an argument in from the UDEV rules for example - please could that be done first and the other fix I mentioned, and then squash the commits and I will merge.

gizmo98:
Thank you very much! I will merge both scripts and replace the path with md_inst.

gizmo98:
Tested all changes with a fresh 3.1 installation again. Works fine.

joolswills:
Thanks.

eddywebs:
I believe this is still an issue with retropie 3.7 using gasia or clone ps3 controller ? 
It actually crashed the entire system:
Steps to reproduce:

 - Pair 1 ps3 controller to the retropie (gasia/clone)
 - Pair 2nd controller by connecting it to the raspberrypi using usb cable, disconnect usb cable press the PS3/home button the lights on 2nd controller keeps on flashing and eventually the system crashess.

Issue #1050, opened 2015-10-17
Add OpenTyrian to Ports


HerbFargus:
I am taking my info from here: http://dosonthepi.blogspot.com/2015/02/install-opentyrian.html

### Install files
```
git clone https://github.com/vanfanel/SDL12-kms-dispmanx.git
sudo mv ~/SDL12-kms-dispmanx/ /opt/sdl-dispmanx
cd /opt/sdl-dispmanx
CFLAGS="-Ofast -mcpu=arm1176jzf-s -mfpu=vfp -mfloat-abi=hard" ./configure --prefix=/opt/sdl-dispmanx --disable-video-opengl --enable-video-dispmanx --disable-video-fbcon --disable-video-kms --disable-video-directfb --disable-oss --disable-alsatest --disable-pulseaudio --disable-pulseaudio-shared --disable-arts --disable-nas --disable-esd --disable-nas-shared --disable-diskaudio --disable-dummyaudio --disable-mintaudio --disable-video-x11 --disable-input-tslib
make && sudo make install
cd
wget -c https://sites.google.com/site/dosonthepi/opentyrian.zip
unzip opentyrian.zip
cd ~/opentyrian
make release
wget -c http://www.camanis.net/tyrian/tyrian21.zip
unzip -j tyrian21.zip -d data
```

### Configure for Retropie
```
cd ~/RetroPie/roms/ports
sudo nano opentyrian.sh
#!/bin/bash
cd /home/pi/opentyrian
/opt/retropie/supplementary/runcommand/runcommand.sh 0 "./opentyrian"
sudo chmod +x opentyrian.sh
```

I'm just posting this as a basis for a module- placing it in a module should simplify the code a bit.

HerbFargus:
This is my preliminary attempt at a module: it needs a bit of work:

```
#!/usr/bin/env bash

# This file is part of The RetroPie Project
# 
# The RetroPie Project is the legal property of its developers, whose names are
# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
# 
# See the LICENSE.md file at the top-level directory of this distribution and 
# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
#

rp_module_id="opentyrian"
rp_module_desc="Tyrian"
rp_module_menus="4+"

function depends_opentyrian() {
}

function sources_opentyrian() {
[...]

PR #1056, opened 2015-10-20
added opentyrian - in experimental for now. Works well on my system -…

joolswills:
… thanks to Herb / dudleydes #1050

### DIFF
--- a/scriptmodules/ports/opentyrian.sh
+++ b/scriptmodules/ports/opentyrian.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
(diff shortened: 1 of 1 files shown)


HerbFargus:
@joolswills works brilliantly mate. Thanks!

dudleydes:
I cannot build OpenTyrian on the Raspberry Pi 1 but it builds fine on the Pi 2.

The issues I raised in https://github.com/RetroPie/RetroPie-Setup/issues/1050 regards hqNx only appears to affect the Pi 1. I can build successfully by deleting video_scale_hqNx.c and commenting out lines 53, 56 and 58 in video_scale.c.

Sorry, should have mentioned this earlier.

joolswills:
out of memory issue - well you can check by looking at dmesg.

256 or 512 pi ? you can reduce gpu mem perhaps - I can add a minimum so more swap gets added, but not major as in experimental etc.

dudleydes:
It's a 512 pi. I tried to build by setting the gpu mem to 16 without success.

What should I be looking for in dmesg?

joolswills:
the system running out of memory and killing gcc. post it on pastebin.com or similar if unsure and ill check.

dudleydes:
Here is the output from dmesg:

http://pastebin.com/pJY8XNmU

joolswills:
Yeh - out of memory

 1664.813469] Out of memory: Kill process 2608 (cc1) score 836 or sacrifice child

PR #1057, opened 2015-10-23
Use rbp-userland-src-osmc package instead of 'hello_pi' tarball to satisfy xroar dependency

samnazarko:
Further to our discussion in https://github.com/osmc/osmc/issues/219. A package is now in the official OSMC APT repository.


joolswills:
Thanks :)

PR #1058, opened 2015-10-23
Add SuperTux to ports

HerbFargus:
So I was thinking, there are quite a few games that can be installed directly with apt-get from the raspbian repositories- would there be any merit to a dynamic menu/script of sorts that builds the games based solely on the repo name as most games scripts will look virtually identical to this one?

That is of course unless you want them to be built from source/ want to be more picky  about what we include with retropie (as some games only work with an x server, some just aren't worth playing, some only require a keyboard etc.)

Just some thoughts anyhow, if you want supertux compiled from source instead of a repo/ don't think its up to snuff then feel free to reject it- just posted it as an option just in case people were interested.

### DIFF
--- a/scriptmodules/ports/supertux.sh
+++ b/scriptmodules/ports/supertux.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+ 
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
For now this is mostly fine. needs  nobin flag though.

joolswills:
thanks :)

HerbFargus:
should be solved now

joolswills:
lovely. thanmks

joolswills:
wrong button :)

Issue #1060, opened 2015-10-24
Retroarch: Use analog stick as a dpad (if system supports no analog sticks) 


gizmo98:
The retroarch option `input_player1_analog_dpad_mode = "1"`tells retroarch to use the left analog stick  as a dpad. (0=no remapping, 1=left analog stick, 2=right analog stick). The actual dpad works at the same time. This option does not interfere with a joypad without analog stick. For up to eight joypads it is only necessary to add the following lines:

configs/all/retroarch.cfg
```
input_player1_analog_dpad_mode = "1"
input_player2_analog_dpad_mode = "1"
input_player3_analog_dpad_mode = "1"
input_player4_analog_dpad_mode = "1"
input_player5_analog_dpad_mode = "1"
input_player6_analog_dpad_mode = "1"
input_player7_analog_dpad_mode = "1"
input_player8_analog_dpad_mode = "1"  
```

There are only two systems which use analog sticks. The option should be disabled in there retroarch.cfgs:
configs/psx/retroarch.cfg and configs/n64/retroarch.cfg
```
input_player1_analog_dpad_mode = "0"
input_player2_analog_dpad_mode = "0"
input_player3_analog_dpad_mode = "0"
input_player4_analog_dpad_mode = "0"
input_player5_analog_dpad_mode = "0"
input_player6_analog_dpad_mode = "0"
input_player7_analog_dpad_mode = "0"
input_player8_analog_dpad_mode = "0"  
```

HerbFargus:
+1 from me

gizmo98:
Verified lr-mupen64plus and lr-psx-rearmed ignore this option. So only configs/all/retroatch.cfg must be modified. If nothing speaks against i would like to add something like this:
```
for i in {1..8}
do
    iniSet "input_player$i_analog_dpad_mode" "1"
done 
```

joolswills:
This has broken D-pads on snes style controllers. What I am confused about is that you seem to have known about the issue beforehand ? https://github.com/libretro/RetroArch/issues/609#issuecomment-37754495

joolswills:
Reverted for now in https://github.com/RetroPie/RetroPie-Setup/commit/897934b0dbf913e01b659567e1928d0f4a82b6c5 - building a 3.2.1 image - will make some manual fix instructions for anyone that downloaded 3.2

gizmo98:
[...]

PR #1062, opened 2015-10-25
retroarch: use analog sticks as dpads

gizmo98:
https://github.com/RetroPie/RetroPie-Setup/issues/1060

And remove code replacement. Search box can be closed with controller
now. Fix is no longer necessary.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -26,9 +26,6 @@ function sources_retroarch() {
     gitPullOrClone "$md_build" https://github.com/libretro/RetroArch.git
     gitPullOrClone "$md_build/overlays" https://github.com/libretro/common-overlays.git
     gitPullOrClone "$md_build/shader" https://github.com/gizmo98/common-shaders.git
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - needs "local i" so i is local scope. Also are you going to include the disabling of this for the emulators you said use the analogue stick ?

The retroarch menu change is unrelated and should be in another commit imho.

joolswills:
just cosmetic, but we use "for blah; do" in the rest of the code - so worth keeping to the same styling. thanks.

gizmo98:
I have tested lr-mupen64plus and lr-psx-rearmed. Both ignore this option and use analog sticks like before.

joolswills:
ok thanks. BTW although the search box can be closed, it is still triggered when entering RGUI or was last time I checked, which is not ideal, so we might still want to include the menu change, but something for another patch.

gizmo98:
Our fix does not work anymore. I was surprised it opens if the wrong key is pressed. So we need a new fix if be don't want search box pop up.

joolswills:
ok.

gizmo98:
I will do a PR with two separate commits.

joolswills:
you could just force commit to this :)

PR #1063, opened 2015-10-25
retroarch fixes

gizmo98:
-remove old search box fix because it does not work and is not necessary anymore
-default use analog sticks as dpads


gizmo98:
Once again ;-)

joolswills:
thanks.

PR #1064, opened 2015-10-25
added ppsspp experimental emulator

joolswills:
Please give it a go. Works here - testing with the excellent "Suicide Barbie" demo currently.

Performance probably could be a bit better as we are not building for arm7 due to current problems with ppsspp and arm7 target on the rpi.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -97,6 +97,9 @@ pc_fullname="PC (x86)"
 pcengine_exts=".pce .cue .zip"
 pcengine_fullname="TurboGrafx 16 (PC Engine)"
 
+psp_exts=".iso .pbp"
+psp_fullname="PlayStation Portable"
+
 psx_exts=".bin .cue .cbn .img .iso .m3u .mdf .pbp .toc .z .znx"
 psx_fullname="PlayStation"
 
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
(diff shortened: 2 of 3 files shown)


HerbFargus:
I'll give it a go- whats the estimated compiling time on the rpi2?

joolswills:
The demo crashed - but it worked for a while :) There will be lots of bugs and unplayable stuff of course. 

Not too about build time as not timed it building just on the rpi2. 30 mins + ?

HerbFargus:
just under an hour to build on the rpi 2- testing games shortly

joolswills:
note you will need to add some config to platforms.cfg - ive not committed that yet (or run it manually from terminal)

HerbFargus:
That is Brilliant! I am really surprised at how well this works. I tested it out with Harvest Moon Hero of Leaf Valley.  There was a bit of an audio lag at parts but ran quite well. haven't changed any default settings or anything. and yeah I noticed the lack of platforms .cfg simple enough to add .iso to the es_systems.cfg

joolswills:
should the libretro core work ok - I will do some de-duplicating between them, and push it to this branch also.

joolswills:
glad it's working for you. :) and it built/ran. which is the main thing

HerbFargus:
Granted it wasn't a direct launch, selecting the game opened up the ppsspp menu and then I had to navigate to my game from there. I also didnt test with a controller, I only tested with a keyboard.  and hopefully the libretro core will work as well :) lr configs would be ideal

joolswills:
the libretro core works, so ill push that shortly

HerbFargus:
This will make many people happy. Brilliant mate :)

joolswills:
added libretro core to this branch now.

joolswills:
if you installed it earlier and you reinstall, you will need to delete /opt/retropie/configs/psp and /home/pi/.config//ppsspp or else the symlinks won't get recreated correctly - I make a few symlinks so that homebrew installed from the ppsspp gui will end up in ~/roms/psp

HerbFargus:
here's the build log for lr-ppsspp that failed:

http://pastebin.com/pnczMbF8

joolswills:
as discussed on hangouts - that's due to not having latest version of all files here - lr-ppsspp relies on ppsspp for building ffmpeg, so there is no code duplication. should be fine now you have a full checkout of my branch.

joolswills:
squashed.

PR #1067, opened 2015-10-29
Add wolf4sdl to ports

dudleydes:
I have written a script module to install wolf4sdl. The wolf4sdl source code builds a distinct binary, wolf3d, on the Pi depending upon the game data to be used. Binaries are not compatible with different versions of the game. For example, the shareware binary will not work with the 3D Realms game data.

The script module will first present a menu to let a user select which game data they wish to use. The script currently supports the following versions:

* Shareware version v1.4
* 3D Realms retail version (Apogee v1.4)
* Steam retail version (GT/ID/Activision)

Depending on the user selection, the script module (lines 29 to 40) will then clone the source code and delete the appropriate line(s) in [version.h](https://github.com/mozzwald/wolf4sdl/blob/master/version.h). I'm not sure whether this is best practice and commenting the necessary lines may be better. I could not find (at least get my head around) the command to do this so I went ahead with deleting option. Maybe this could be improved?

I have successfully tested the script for all three versions of the game data. There is a minor error, though, during the install process as below:
````
===> INSTALL
install: cannot create regular file `/opt/retropie/ports/wolf4sdl/share/man//man6': No such file or directory
Makefile:117: recipe for target 'install' failed
make: *** [install] Error 1
````
The wolf3d binary is installed to the correct location (/opt/retropie/ports/wolf4sdl/bin) and the game launches fine.

Users will need to put their game data in /home/pi/RetroPie/ports/wolf3d folder. File names need to be lower case. To make things easier, the scriptmodule will create a bash script, lower-script.sh, located in the wolf3d folder that can change the file names from the command line.

It may be also be worthwhile checking the CFlags in [Makefile](https://github.com/mozzwald/wolf4sdl/blob/master/Makefile) and [config.default](https://github.com/mozzwald/wolf4sdl/blob/master/config.default) file.

### DIFF
--- a/scriptmodules/ports/wolf4sdl.sh
+++ b/scriptmodules/ports/wolf4sdl.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="wolf4sdl"
+rp_module_desc="Wolf4SDL, port of Wolfenstein 3D Shareware 1.4 version"
+rp_module_menus="4+"
+rp_module_flags="dispmanx"
+
+function depends_wolf4sdl() {
+    getDepends libsdl1.2-dev
+}
+
+function sources_wolf4sdl() {
+    gitPullOrClone "$md_build" https://github.com/mozzwald/wolf4sdl.git
+    # Define Shareware version
+    sed -i 's|#define GOODTIMES|//#define GOODTIMES|g' version.h
+    
+    # Define Steam version
+    # sed -i 's|#define UPLOAD|//#define UPLOAD|g' version.h
+    
+    # Define 3D Realms version
+    # sed -i 's|#define GOODTIMES|//#define GOODTIMES|g' version.h
+    # sed -i 's|#define UPLOAD|//#define UPLOAD|g' version.h
+}
+
+function build_wolf4sdl() {
+    make clean
+    make DATADIR="$romdir/ports/wolf3d/"
+    md_ret_require="$md_build"
+}
+
+function install_wolf4sdl() {
+     mkdir -p "$md_inst/share/man/man6"
+     make install PREFIX="$md_inst" MANPREFIX="$md_inst/share/man"
+}
+
+function configure_wolf4sdl() {
+    mkRomDir "ports"
+    mkRomDir "ports/wolf3d"
+
+    # Get shareware game data
+    wget -q -O wolf3d14.zip http://maniacsvault.net/ecwolf/files/shareware/wolf3d14.zip
+    unzip -j -o -LL wolf3d14.zip -d "$romdir/ports/wolf3d"
+    rm -f wolf3d14.zip
+
+    setDispmanx "$md_id" 1
+
+    addPort "$md_id" "wolf4sdl" "Wolfenstein 3D" "$md_inst/bin/wolf3d"
+}

### CODE REVIEW

File: scriptmodules/ports/wolf4sdl.sh [hunk 1]
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
+
+# This file is part of RetroPie.
+# 
+# (c) Copyright 2012-2015  Florian Müller (contact@petrockblock.com)
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/petrockblog/RetroPie-Setup/master/LICENSE.md.
+#
```
joolswills: We use a newer copyright now - please check another module for details

scriptmodules/ports/wolf4sdl.sh
```diff
@@ -0,0 +1,85 @@
...
+                    ;;
+                3)
+                    sed -i '/UPLOAD/d' version.h
+                    ;;
+            esac
+        else
+            break
+        fi
+}
+
```
joolswills: There can't be user input in the sources part - wont be compatible with out binaries later etc - The best thing to do would be to build all three versions separately.

scriptmodules/ports/wolf4sdl.sh
```diff
@@ -0,0 +1,85 @@
...
+      if [ ! -f \$x ]; then
+        continue
+      fi
+      lc=\`echo \$x  | tr '[A-Z]' '[a-z]'\`
+      if [ \$lc != \$x ]; then
+        mv -i \$x \$lc
+      fi
+      done
+_EOF_
+
```
joolswills: What is this needed for - the prorgam needs lowercase game data or ? Also can just use "rename 'y/A-Z/a-z/' *"

joolswills:
thanks. made a few comments re some changes to be made - regarding the install error, you can probably make the folder first to avoid it, or fix/patch uop the install script so it uses mkdir -p.

dudleydes:
I have written a script module to build a binary for the shareware version on Wolf4SDL. The binary works fine but the addPort function on line 53 isn't creating the correct script to launch Wolf4SDL. This is the script generated

````
#!/bin/bash
"/opt/retropie/supplementary/runcommand/runcommand.sh" 0 _SYS_ wolf4sdl ""
````

What am I missing?

**Comment by joolswills** [comment]
how do you mean it isn't correct - what happens on launch ?

File: scriptmodules/ports/wolf4sdl-wl.sh [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function configure_wolf4sdl() {
+    mkRomDir "ports"
+    mkRomDir "ports/wolf3d"
+
+    # Get shareware game data
+    wget -q -O wolf3d14.zip http://maniacsvault.net/ecwolf/files/shareware/wolf3d14.zip
+    unzip -j -o wolf3d14.zip -d "$romdir/ports/wolf3d"
+    rm -f wolf3d14.zip
```
joolswills: Aah the lowercasing was for the files in the zip. can avoid having to do all of that by forcing it to lowercase when unpacking using unzip -LL (unzip -L may be fine too depending on the origins of the zipfile)

dudleydes:
Aah - script works fine. Sorry, I saw the script and assumed it was incorrect as it didn't look like the other scripts for ports (eduke32, opentyrian). I'll test the scripts for the other binaries.

joolswills:
If you do a binary update the other scripts will look similar. I changed the way ports are launched to match the way emulators are launched. not needed in most cases but simplified the modules scripts themselves, and now allows for multiple commandlines per port, or say switching between 2 different doom engines.

joolswills:
Rather than multiple modules for each port, we should build all versions in one module, and they can all be added as ports and switched by the user as they need. that way we avoid duplication between modules etc.

joolswills:
if unsure how to do that, just include a single module called wold4sdl.sh that builds the shareware and I'll sort it when it's merged.

dudleydes:
Thanks for clearing up the new scripts for ports. I've done a binary update for eduke32 and can see how it now works.

I agree about a single module. I tried multiple ones without success. I've create a single module wolf4sdl.sh that will build the shareware. 

I'm not sure how to create a single module but I have included the code I used to patch the source code to build the Steam and 3D Realms versions.

Should we include Spear of Destiny or build a separate module? I'm happy to have a go at doing this once the Wolf3D module is built.

joolswills:
Thanks. Please can you squash the commits and I will merge (eg as described here -  http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request)

dudleydes:
I followed the instructions in the link but I'm not sure if I've squashed the commits as requested. Maybe it's because I haven't updated the wolf4sdl branch since I created it. Please let me know if I've done it correctly, else I'll try to fix.

joolswills:
no. in fact you have added a heap more commits to the branch.

that rebase probably wouldn't work if your branch is not updated. Instead you could have done "git rebase -i HEAD~4" where 4 is the number of commits you have done on your branch.

joolswills:
Just remove the branch on your local machine (saving your module file), create a new branch with the same name just with the file and commit it and force push it back. Should sort it.

dudleydes:
Okay, I think I've managed to fix it.

joolswills:
Thanks! :)

joolswills:
BTW the upstream repository you use lacks code history - I found another one https://github.com/ljbade/wolf4sdl which also has some newer development - was there a specific reason behind github repo choice ?

dudleydes:
It's the repository that was linked in the [Wolfenstein 3D thread](https://www.raspberrypi.org/forums/viewtopic.php?p=515392#p515392) at the RPi forum. I guess the maintainer hadn't included the history from the original development which can be found [here](http://homepage.o2mail.de/mkroll/).

Incidentally, there's a link to the mozzwald repo at the wiki. Can this be removed and a page created? I have written a draft for the wiki entry so the page need only be a placeholder.

Can we add a --goodtimes flag to the wolf4sdl-spear binary to disable the copy protection quiz?

joolswills:
Where do you mean on the wiki. Are you unable to change it ?

I might try switching to the other repository to test.

will sort the flag when i have a chance (or you can submit a pull request).

HerbFargus:
By all means, feel free to create a wiki page. I just threw the link up there just so people knew it was part of retropie until I had time to test and create a page, but seeing as you added it, you'd probably have a better idea of the documentation.

dudleydes:
Thanks, Herb. If I create a new page and put a link to it in the sidebar instead of the github repo, will that be okay? I really don't want to mess up your excellent work.

HerbFargus:
Of course. Generally I link the sidebar to pages on the wiki and then on the home wiki page and within each wiki page I provide a link to the github source as well. The sidebar is meant to be a quick reference to the rest of the wiki, so its actually preferable to not have any links on the sidebar that take people out of the wiki.

Issue #1073, opened 2015-11-01
PS3 controller pairing script


tiagojpdias:
Hey,

It seems that the script which handles pairing PS3 controllers with bluetooth dongle its not quite right.

It lacks the 3 steps, like
1 - "Connect the USB dongle"
2 - "Connect controller by USB"
3 - "Disconnect USB cable and press PS button to pair"

I found where the possible errors were made by cherrypicking this commit: 5182170d6f7f90beb07e44ad10887599eaffe1a0

Here the script behaves like it should and I successfully pair the controller.
If any more info is needed let me know.

Cheers,

gizmo98:
It's intented. Pairing works anytime if you have retropie 3.2.1. If you use an older SD card image update setup script, remove and reinstall ps3 controller driver.

gizmo98:
If it is still not working please post your hardware setup.

tiagojpdias:
Hi @gizmo98,

I've managed to setup the controller before reporting the issue.
I'll try to make a fresh install of my system and then trying to replicate it (because I don't want to "undo" all the work done to this point)

I'll post the hardware setup just to be documented:

* Raspberry PI B (Raspbian Jessie with kernel 4.1.7+)
```bash
pi@rpi ~ $ lsb_release -a
No LSB modules are available.
Distributor ID: Raspbian
Description:    Raspbian GNU/Linux 8.0 (jessie)
Release:        8.0
Codename:       jessie
pi@rpi ~ $ uname -a
Linux rpi 4.1.7+ #817 PREEMPT Sat Sep 19 15:25:36 BST 2015 armv6l GNU/Linux
```
* Last pull
```bash
pi@rpi ~/RetroPie-Setup $ git show | grep Date
Date:   Sat Oct 31 15:34:21 2015 +0000
```
* Bluetooth:
```bash
pi@rpi ~ $ lsusb | grep Blue
Bus 001 Device 006: ID 0a12:0001 Cambridge Silicon Radio, Ltd Bluetooth Dongle (HCI mode)
```

* PS3 sixaxis controller
```bash
pi@rpi ~ $ dmesg | grep Sony
[   58.801287] input: Sony Computer Entertainment Wireless Controller as /devices/platform/soc/20980000.usb/usb1/1-1/1-1.3/1-1.3.4/1-1.3.4:1.0/bluetooth/hci0/hci0:1/0005:054C:0268.0001/input/input3
[...]

PR #1074, opened 2015-11-02
ps3controller: debian jessie compatibility

gizmo98:
-add check for debian jessie’s „service sixad status“ return value. Jessie returns "active (running)".
-add hciconfig hci0 up/pscan to rc.local. udev rule 10-local.rules is not ready at startup if raspbian jessie is used.
https://github.com/RetroPie/RetroPie-Setup/issues/1073

### DIFF
--- a/scriptmodules/supplementary/ps3controller.sh
+++ b/scriptmodules/supplementary/ps3controller.sh
@@ -52,7 +52,19 @@ function install_ps3controller() {
     checkinstall -y --fstrans=no
     insserv sixad
 
-    # If a bluetooth dongle is connected set state up and enable pscan
+    # If a bluetooth dongle is present "at startup" set state up and enable pscan
+    sed -i 's/exit 0//g' "/etc/rc.local"
+    cat >> "/etc/rc.local" <<\_EOF_
+# PS3 PROFILE START
+if hciconfig | grep -q "hci0"; then
+    hciconfig hci0 up
+    hciconfig hci0 pscan
+fi
+# PS3 PROFILE END
+exit 0
+_EOF_
+
+    # If a bluetooth dongle is connected "at runtime" set state up and enable pscan
     cat > "$md_inst/bluetooth.sh" << _EOF_
 #!/bin/bash
 if hciconfig | grep -q "hci0"; then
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. I think I would prefer something that doesn't involve rc.local, maybe it's possible to workaround it another way. I'll take a look anyway.

gizmo98:
I would prefer something without rc.local as well. Any ideas are welcome.

gizmo98:
@joolswills 
I just wanted to add gasia gamepad support but ended up with an own sixad repo with all udev rules and scripts. I solved the rc.local whilst moving everything into the sixad daemon. If sixad starts hci0 states will be set up. Even though the module looks much cleaner now. ;-)

joolswills:
Thanks - as this is a core thing, please can we move the repository to the RetroPie organisation.

joolswills:
(And please can you squash the two above commits)

gizmo98:
We can. But i don't have permissions to do this.

joolswills:
I could fork it if that would be ok - to move it you will need to get admin permissions from @petrockblog

gizmo98:
No Jools. I don't have permissions to fork or transfer this repo. I need your or Florians help. If you do so please fork https://github.com/gizmo98/common-shaders as well.

joolswills:
both forked.

gizmo98:
All right. I will wait until i have tested branch gasia with a 3rd pary controller. There is no sixad driver for original and clone controllers without function degradation. If this controller works i would like to add a second install option.
1. Install/Pair PS3 controller
2. Install/Pair PS3 controller (3rd party controller support)
3. Remove PS3 controller configurations

petrockblog:
@gizmo98 Sure. I have set up a temporary team with you as team maintainer so that you can transfer the repository (if I have not missed any important configuration).

gizmo98:
@petrockblog 
Thank you!

@joolswills 
I will add the new setup script directly. A cheapo ps3 controller should arrive next week. If branch gasia works i will modify script again. is there any trick to pass variables into a module function like:
`rp_callModule ps3controller source gasia`

`function sources_ps3controller() {
    branch=$1
    gitPullOrClone "$md_build/sixad" https://github.com/RetroPie/sixad.git $branch`

joolswills:
rp_callModule will pass any additional parameters onto the function.

PR #1075, opened 2015-11-03
Mupen64plus: cleanup module

gizmo98:
-All modifications were added upstream. Remove old fixes.
-Add raspbian jessie fix again.


joolswills:
thanks.

PR #1087, opened 2015-11-17
retroarch config script tweaks

gizmo98:
-remove duplicate onend_retroarch_joystick() function.
-add sanity check for controllers without select button.

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
@@ -238,16 +238,6 @@ function map_retroarch_joystick() {
     done
 }
 
-function onend_retroarch_joystick() {
-    local device_type=$1
-    local device_name=$2
-    local file="${device_name// /}.cfg"
-    if [[ -f "/opt/retropie/configs/all/retroarch-joypads/$file" ]]; then
-        mv "/opt/retropie/configs/all/retroarch-joypads/$file" "/opt/retropie/configs/all/retroarch-joypads/$file.bak"
-    fi
-    mv "/tmp/tempconfig.cfg" "/opt/retropie/configs/all/retroarch-joypads/$file"
-}
-
 function map_retroarch_keyboard() {
     local device_type="$1"
     local device_name="$2"
--- a/scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
@@ -309,6 +309,26 @@ function map_retroarch_keyboard() {
 function onend_retroarch_joystick() {
     local device_type=$1
     local device_name=$2
+    
+    # hotkey sanity check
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Good spot with the duplicate function - must have been some c&p fail :)

scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
```diff
@@ -319,6 +309,26 @@ function map_retroarch_keyboard() {
...
+        iniSet "input_exit_emulator_btn" ""
+        iniSet "input_state_slot_decrease_axis" ""
+        iniSet "input_state_slot_increase_axis" ""
+        iniSet "input_reset_axis" ""
+        iniSet "input_menu_toggle_axis" ""
+        iniSet "input_load_state_axis" ""
+        iniSet "input_save_state_axis" ""
+        iniSet "input_exit_emulator_axis" ""
+    fi
+
```
joolswills: Was just looking at this piece of code - we don't know if they will be _btn or _axis so they should be removed with sed or something (without the _btn/_axis suffix)

It would be easier just not to add them though if there is no hotkey enable button set ?

gizmo98: We could use (or fix) iniDel(). Last time i used it it didn't remove an entry.

joolswills: I can check it, but it wouldn't work in the case of a partial key where some might be _btn and others _axis etc.

scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
```diff
@@ -319,6 +309,26 @@ function map_retroarch_keyboard() {
...
+        iniSet "input_save_state_btn" ""
+        iniSet "input_exit_emulator_btn" ""
+        iniSet "input_state_slot_decrease_axis" ""
+        iniSet "input_state_slot_increase_axis" ""
+        iniSet "input_reset_axis" ""
+        iniSet "input_menu_toggle_axis" ""
+        iniSet "input_load_state_axis" ""
+        iniSet "input_save_state_axis" ""
+        iniSet "input_exit_emulator_axis" ""
+    fi
```
joolswills: Sorry to comment on an ancient PR - but I had a question about the above block (I am making some changes to how hotkey configuration works).

Why do these need to be set to empty if the user has no hotkey enable (because they have no select button) ? Does retroarch enable them by default ? I would have thought if they didn't exist in the controller file, they are not enabled. I did a test and I couldn't see any issues either.

joolswills: Don't worry - I forget we set the hotkeys anyway so need to remove them - I might use iniDel though. With the new configurable hotkey (not yet merged), we can probably do this differently (in the future).

PR #1092, opened 2015-11-20
Add Eudora Themes from AmadhiX

HerbFargus:
They look quite brilliant actually- a lot of customisability options with colour sets.


joolswills:
thanks.

Issue #729, opened 2015-03-12
Modify N64 InputAutoCfg.ini if a controller is registered


gizmo98:
If a controller is registered we can use the retroarch autoconfig file to modify /opt/retropie/configs/n64/InputAutoConfig.ini. Most button numbers are equal. AnalogDeadzone and AnalogPeak is always the same. Retroarchs input_device is the section name of InputAutoConfig.ini.

Example:
https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/udev/PS3Controller.cfg
input_device = "PLAYSTATION(R)3 Controller"
input_driver = "udev"
input_b_btn = "14"
input_y_btn = "15"
input_select_btn = "0"
input_start_btn = "3"
input_up_btn = "4"
input_down_btn = "6"
input_left_btn = "7"
input_right_btn = "5"
input_a_btn = "13"
input_x_btn = "12"
input_l_btn = "10"
input_r_btn = "11"
input_l2_btn = "8"
input_r2_btn = "9"
input_l_x_plus_axis = "+0"
input_l_x_minus_axis = "-0"
input_l_y_plus_axis = "+1"
input_l_y_minus_axis = "-1"
input_r_x_plus_axis = "+2"
input_r_x_minus_axis = "-2"
input_r_y_plus_axis = "+3"
input_r_y_minus_axis = "-3"

https://github.com/mupen64plus/mupen64plus-input-sdl/blob/master/data/InputAutoCfg.ini
[Sony PLAYSTATION(R)3 Controller]
plugged = True
plugin = 2
mouse = False
AnalogDeadzone = 4096,4096
AnalogPeak = 32768,32768
DPad R = button(5)
DPad L = button(7)
DPad D = button(6)
DPad U = button(4)
Start = button(3)
Z Trig = button(10)
B Button = button(15)
A Button = button(14)
C Button R = axis(2+)
C Button L = axis(2-)
C Button D = axis(3+)
C Button U = axis(3-)
R Trig = button(11)
L Trig = button(8)
Mempak switch = 
Rumblepak switch = 
X Axis = axis(0-,0+)
Y Axis = axis(1-,1+)

gizmo98:
I will try to create a mupen64plus config script.

HerbFargus:
I'm happy to test once you come up with something

HerbFargus:
I try to find time this evening to give it a go. will hotkeys also be autogenerated with this or is that a separate thing?

[...]

PR #1093, opened 2015-11-20
inputconfig: add mupen64plus script

gizmo98:
-add mupen64plus.sh. Modify InputAutoConfig.ini
-inputconfiguration.sh: add iniGet and iniDel 

Herbfargus and i tested this script with four controllers. 
https://github.com/RetroPie/RetroPie-Setup/issues/729#issuecomment-15829
2871

This script is not intended to setup hotkeys. Another script is necessary to modify mupen64plus.cfg.
--> https://github.com/gizmo98/mupen64plus-retropie-hotkeys

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
@@ -1,0 +1,267 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+function onstart_mupen64plus_joystick() {
+    local device_type=$1
+    local device_name=$2
+
+    # write temp file header
+    echo "; ${device_name}_START " > /tmp/mp64tempconfig.cfg
+    echo "[${device_name}]" >> /tmp/mp64tempconfig.cfg
+    iniConfig " = " "" "/tmp/mp64tempconfig.cfg"
+    iniSet "plugged" "True"
+    iniSet "plugin" "2"
+    iniSet "mouse" "False"
+    iniSet "AnalogDeadzone" "4096,4096"
+    iniSet "AnalogPeak" "32768,32768"
+    iniSet "Mempak switch" ""
+    iniSet "Rumblepak switch" ""
+}
+
+function map_mupen64plus_joystick() {
+    local device_type="$1"
+    local device_name="$2"
+    local input_name="$3"
+    local input_type="$4"
+    local input_id="$5"
+    local input_value="$6"
+
+    local keys
+    local dir
+    case "$input_name" in
+        up)
+            keys=("DPad U")
+            dir=("Up")
+            ;;
+        down)
+            keys=("DPad D")
+            dir=("Down")
+            ;;
+        left)
+            keys=("DPad L")
+            dir=("Left")
+            ;;
+        right)
+            keys=("DPad R")
+            dir=("Right")
+            ;;
+        b)
+            keys=("A Button")
+            ;;
+        y)
+            keys=("B Button")
+            ;;
+        a)
+            keys=("C Button D")
+            ;;
+        x)
+            keys=("C Button U")
+            ;;
+        leftbottom)
+            keys=("Z Trig")
+            ;;
+        rightbottom)
+            keys=("R Trig")
+            ;;
+        lefttop)
+            keys=("L Trig")
+            ;;
+        start)
+            keys=("Start")
+            ;;
+        leftanalogleft)
+            keys=("X Axis")
+            dir=("Left")
+            ;;
+        leftanalogright)
+            keys=("X Axis")
+            dir=("Right")
+            ;;
+        leftanalogup)
+            keys=("Y Axis")
+            dir=("Up")
+            ;;
+        leftanalogdown)
+            keys=("Y Axis")
+            dir=("Down")
+            ;;
+        rightanalogleft)
+            keys=("C Button L")
+            dir=("Left")
+            ;;
+        rightanalogright)
+            keys=("C Button R")
+            dir=("Right")
+            ;;
+        rightanalogup)
+            keys=("C Button U")
+            dir=("Up")
+            ;;
+        rightanalogdown)
+            keys=("C Button D")
+            dir=("Down")
+            ;;
+        *)
+            return
+            ;;
+    esac
+
+    local key
+    local value
+    #iniConfig " = " "" "/tmp/mp64keys.cfg"
+    for key in "${keys[@]}"; do
+        # read key value. Axis takes two key/axis values.
+        iniGet "$key"
+        case "$input_type" in
+            axis) 
+                # key "X/Y Axis" needs different button naming
+                if [[ "$key" == *Axis* ]]; then
+                    # if there is already a "-" axis add "+" axis value
+                    if   [[ "$ini_value" == *\(* ]]; then
+                        value="${ini_value}${input_id}+)"
+                    # if there is already a "+" axis add "-" axis value
+                    elif [[ "$ini_value" == *\)* ]]; then
+                        value="axis(${input_id}-,${ini_value}"
+                    # if there is no ini_value add "+" axis value
+                    elif [[ "$input_value" == "1" ]]; then
+                        value="${input_id}+)"
+                    else 
+                        value="axis(${input_id}-,"
+                    fi
+                elif [[ "$input_value" == "1" ]]; then
+                    value="axis(${input_id}+) ${ini_value}"
+                else
+                    value="axis(${input_id}-) ${ini_value}"
+                fi
+                ;;
+            hat)
+                if [[ "$key" == *Axis* ]]; then
+                    if   [[ "$ini_value" == *\(* ]]; then
+                        value="${ini_value}${dir})"
+                    elif [[ "$ini_value" == *\)* ]]; then
+                        value="hat(${input_id} ${dir} ${ini_value}"
+                    elif [[ "$dir" == "Up" || "$dir" == "Left" ]]; then
+                        value="hat(${input_id} ${dir} "
+                    elif [[ "$dir" == "Right" || "$dir" == "Down" ]]; then
+                        value="${dir})"
+                    fi
+                else
+                    if ! [[ "$dir" == "" ]]; then 
+                        value="hat(${input_id} ${dir}) ${ini_value}"
+                    fi
+                fi
+                ;;
+            *)
+                if [[ "$key" == *Axis* ]]; then
+                    if   [[ "$ini_value" == *\(* ]]; then
+                        value="${ini_value}${input_id})"
+                    elif [[ "$ini_value" == *\)* ]]; then
+                        value="button(${input_id},${ini_value}"
+                    elif [[ "$dir" == "Up" || "$dir" == "Left" ]]; then
+                        value="button(${input_id},"
+                    elif [[ "$dir" == "Right" || "$dir" == "Down" ]]; then
+                        value="${input_id})"
+                    fi
+                else
+                    value="button(${input_id}) ${ini_value}"
+                fi
+                ;;
+        esac
+
+        iniSet "$key" "$value"
+    done
+}
+
+function onend_mupen64plus_joystick() {
+    local device_type=$1
+    local device_name=$2
+
+    local bind
+    local axis
+    local axis_neg
+    local axis_pos
+    for axis in "X Axis" "Y Axis"; do
+        if [[ "$axis" == *X* ]]; then
+            axis_neg="DPad L"
+            axis_pos="DPad R"
+        else
+            axis_neg="DPad U"
+            axis_pos="DPad D"
+        fi
+
+        # analog stick sanity check
+        # replace Axis values with DPAD values if there is no Axis 
+        # device setup
+        if ! grep -q "$axis" /tmp/mp64tempconfig.cfg ; then
+          iniGet "${axis_neg}"
+          bind=${ini_value//)/,}
+          iniGet "${axis_pos}"
+          ini_value=${ini_value//axis(/}
+          ini_value=${ini_value//hat(/}
+          ini_value=${ini_value//button(/}
+          bind="${bind}${ini_value}"
+          iniSet "$axis" "$bind"
+          iniDel "${axis_neg}"
+          iniDel "${axis_pos}"
+        fi
+    done
+
+    echo "; ${device_name}_END " >> /tmp/mp64tempconfig.cfg
+    echo "" >> /tmp/mp64tempconfig.cfg
+
+    # abort if old device config cannot be deleted.
+    # keep original mupen64plus-input-sdl configs.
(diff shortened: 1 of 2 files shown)


scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,266 @@
...
+
+    local key
+    local value
+    #iniConfig " = " "" "/tmp/mp64keys.cfg"
+    for key in "${keys[@]}"; do
+        # read key value. Axis takes two key/axis values.
+        iniGet "$key"
+        if [[ "$input_type" == "axis" ]]; then
+            # key "X/Y Axis" needs different button naming
+            if echo "$key" | grep -q "Axis"; then
```
joolswills: Please check out using bash regular expressions over running echo / grep  `[[ "$blah" =~` etc

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,266 @@
...
+    # analog stick sanity check
+    # replace Axis value with DPAD values if there is no Axis 
+    # device setup
+    local bind;
+    if ! grep -q -e "X Axis" -e "Y Axis" /tmp/mp64tempconfig.cfg; then
+        iniGet "DPad U"
+        bind=`echo "$ini_value" | sed -e "s/)/,/g"`
+        iniGet "DPad D"
+        ini_value=`echo "$ini_value" | sed -e "s/button(//g"`
+        ini_value=`echo "$ini_value" | sed -e "s/hat(//g"`
```
joolswills: `${string/substring/replacement}`

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,266 @@
...
+        cp "$file" "${file}.bak"
+        sed -i /"${device_name}_START"/,/"${device_name}_END"/d "$file"
+        if grep -q "$device_name" "$file"; then
+            return
+        fi
+    else
+        cat > "$file" << _EOF_
+; InputAutoCfg.ini for Mupen64Plus SDL Input plugin
+
+[Keyboard]
```
joolswills: keyboard input is hardcoded or ? if keyboard is configurable we should try and configure that in the case that the user uses a keyboard to configure ES (although this could come later).

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
joolswills: in some of the cases where you need a simple match it might be cleaner to use case / esac with wildcards etc.

joolswills: Even quicker is probably [[ "$blah" == *something* ]] for simple pattern matching.

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
gizmo98: Keyboard is hardcoded. I would at '; KEYBOARD_START' '; KEYBOARD_END' frame here to replace keyboard with sed if necessary. IniSet does not support sections?!

joolswills: no sections yet - it's one of the things i had planned (needed for fba).

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
gizmo98: So `$(ini_value/button/)` is doing the same?

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
gizmo98: Wildcards are a really good idea!

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
joolswills: See http://tldp.org/LDP/abs/html/string-manipulation.html

joolswills: If you need to removev substring from front / back there are is also specific string manipulation for that.

joolswills:
Regarding bash specific code , if you are stripping/removal parts with sed to extract substrings, you can also do regexp matching with () and use $BASH_REMATCH[1]

eg
http://aplawrence.com/Linux/bash-regex.html

gizmo98:
Ok. I keep the sed substring replacement as a one liner for now. The substring replacement example does not work. Apart from that i worked in most of your suggestions.

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,260 @@
...
+function onend_mupen64plus_joystick() {
+    local device_type=$1
+    local device_name=$2
+    
+    local bind
+    local axis_neg
+    local axis_pos
+    for axis in "X Axis" "Y Axis"; do
+        if [[ $axis == *"X"* ]]; then
+            axis_neg="DPad L"
```
joolswills: Missing "local axis"
"$axis" needs quoting *X* doesn't

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,260 @@
...
+            axis_neg="DPad U"
+            axis_pos="DPad D"
+        fi
+        
+        # analog stick sanity check
+        # replace Axis values with DPAD values if there is no Axis 
+        # device setup
+        if ! grep -q "$axis" /tmp/mp64tempconfig.cfg ; then
+          iniGet "${axis_neg}"
+          bind=`echo "$ini_value" | sed -e "s/)/,/g"`
```
joolswills: Should work without sed

```
echo `echo ")))" | sed -e "s/)/,/g"`
,,,

foo=")))"
echo ${foo//)/,}
,,,
```

joolswills:
made a couple more comments - if you can sort them, then please squash the commits / force push back to the bvranch and I'll merge

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
gizmo98: But you are right. 
Strange. Why does it run anyway? I have not seen a error message and remapping of DPAD to AXIS works?

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,260 @@
...
+    local value
+    #iniConfig " = " "" "/tmp/mp64keys.cfg"
+    for key in "${keys[@]}"; do
+        # read key value. Axis takes two key/axis values.
+        iniGet "$key"
+        case "$input_type" in
+            axis) 
+                # key "X/Y Axis" needs different button naming
+                if [[ $key == *"Axis"* ]]; then
+                    # if there is already a "-" axis add "+" axis value
```
joolswills: Please quote $key, but no need to quote Axis.

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,260 @@
...
+                fi
+                ;;
+            hat)
+                if [[ $key == *"Axis"* ]]; then
+                    if   [[ $ini_value == *"("* ]]; then
+                        value="${ini_value}${input_id} ${dir})"
+                    elif [[ $ini_value == *")"* ]]; then
+                        value="axis(${input_id} ${dir},${ini_value}"
+                    elif [[ "$dir" == "Up" ]] || [[ "$dir" == "Left" ]]; then
+                        value="button(${input_id} ${dir},"
```
joolswills: just fyi (not important), but bash allows this without the additional group. e.

```
elif [[ "$dir" == "Up" || "$dir" == "Left" ]]; then
```

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
joolswills: seems bash is forgiving and allows it but still fits our styleguide to quote it anyway.

joolswills: Just seen "Hint: Inside the conditional expression ([[ ]]) Bash doesn't perform word splitting, and thus you don't need to quote your variable references - they are always seen as "one word".

gizmo98:
Didn't thought it would be so easy to rebase. ;-)

HerbFargus:
I just tested the latest changes and there are a few things that may need to be addressed:

Is this intended to wipe the old auto config file and create a whole new inputautoconfig.ini, or is it just supposed to append it? When it just appends it there must be some conflicts (like there was originally with the retroarch autoconfigs when we had them all in the same folder together) where my controller was added, but it seemed to take a different controllers configs (for my cheap snes gamepad) x was supposed to be a C button but it ended up being my A button. Once I deleted the config and reconfigured input from emulationstation it created the same config but this time it worked properly. Perhaps it may be best to start an autoconfig from scratch like we have done with retroarch instead of keeping all the autoconfigs as to eliminate conflicts like I experienced.

The other thing is the hat directions for the dpad aren't being encoded (in the case of my logitech controller) I don't think:
This is the configs before all the code changes:

http://pastebin.com/yGn1K8rB

This is the config for the latest changes:

http://pastebin.com/QGvVpwkD

gizmo98:
@joolswills 
If you think it's clean enough you an merge it. I have tested with my controllers and had no issues.

HerbFargus:
just for a more convenient perusal:

https://www.diffchecker.com/dvsslk2x

gizmo98:
Ok. Jools please don't merge this for now.

gizmo98:
@HerbFargus i did not see your post ;-)
I added something for hats bound to C Buttons and DPADS. Please test again.

HerbFargus:
ha ha you beat me to it... I'll test the latest now

HerbFargus:
seems good to me :+1: 

Just for reference: http://pastebin.com/ZcXLZqkT

HerbFargus:
actually I take it back- mixups with the analog and hats....

```
; Logitech Gamepad F310_START
[Logitech Gamepad F310]
plugged = True
plugin = 2
mouse = False
Mempak switch =
Rumblepak switch =
C Button D = button(1) axis(4+)
C Button L = axis(3-)
R Trig = button(5)
L Trig = axis(2+)
Start = button(7)
Y Axis = button(0 Up,0 Down)
Z Trig = button(4)
DPad U = axis(1-)
C Button U = button(3) axis(4-)
A Button = button(0)
DPad D = axis(1+)
X Axis = axis(0 Left,0 Right)
DPad R = axis(0+)
B Button = button(2)
DPad L = axis(0-)
C Button R = axis(3+)
; Logitech Gamepad F310_END
```

gizmo98:
@HerbFargus I need a controller with hats! There was a really idiotic bug.

HerbFargus:
Everything is good... except I can only turn right with my analog on the logitech f310 the left analog doesnt work for some reason

```
; Logitech Gamepad F310_START 
[Logitech Gamepad F310]
plugged = True
plugin = 2
mouse = False
Mempak switch = 
Rumblepak switch = 
C Button D = button(1) axis(4+) 
C Button L = axis(3-) 
R Trig = button(5) 
L Trig = axis(2+) 
Start = button(7) 
Y Axis = hat(0 Up,0 Down)
Z Trig = button(4) 
DPad U = axis(1-) 
C Button U = button(3) axis(4-) 
A Button = button(0) 
DPad D = axis(1+) 
X Axis = hat(0 Left,0 Right)
DPad R = axis(0+) 
B Button = button(2) 
DPad L = axis(0-) 
C Button R = axis(3+) 
; Logitech Gamepad F310_END 
```

HerbFargus:
So I think that the X and Y axis don't need to be hats, but the dpad is the hat for the f310 as this was the config that worked 

```
;Logitech Gamepad F310 CONFIG START
[Logitech Gamepad F310]
plugged = True
plugin = 2
mouse = False
AnalogDeadzone = 4096,4096
AnalogPeak = 32768,32768
Mempak switch =
Rumblepak switch =
C Button D = axis(4+)
C Button L = axis(3-)
R Trig = button(5)
L Trig = axis(2+)
Start = button(7)
Y Axis = axis(1-,1+)
Z Trig = button(4)
DPad U = hat(0 Up)
C Button U = axis(4-)
A Button = button(0)
DPad D = hat(0 Down)
X Axis = axis(0-,0+)
DPad R = hat(0 Right)
B Button = button(2)
DPad L = hat(0 Left)
C Button R = axis(3+)
;Logitech Gamepad F310 CONFIG END
```

gizmo98:
I map up/down/left/right to "dpad" and analogleft to "axis" at the moment. I map dpad to axis if there is no "axis" device. I will take a look if hats are not supported as analog devices. If so dpad must be remapped as well.

gizmo98:
@HerbFargus 
Could you check something for me? There is no example for a hat device mapped to axis anywhere. I have taken a look in mupen64plus source. The correct mapping name should be:
X Axis = hat(0 Left Right)
Y Axis = hat(0 Up Down)

If this works i have to modify the code for hats.

HerbFargus:
testing now

HerbFargus:
yes it works

gizmo98:
I updated hat axis handling, so mapping should work now. Strange enough no one else uses this. Our configs will always differ from the old autoconfig file ;-)

HerbFargus:
Brilliant mate :) 

http://pastebin.com/eHWct1Gn

so when I do configs for emulationstation the dpad is what is listed as the hat, the right analog is just an axis: see the following diagrams

http://i.stack.imgur.com/wjVTm.png
http://i.imgur.com/nmycqTC.jpg

gizmo98:
Sounds good! If you think the mapping is correct i will keep the current code.

HerbFargus:
Seems to work fine as it is. I may create some new diagrams for the wiki at some point based on your new mappings (since they differ from the retroarch build) There may be tweaks in the future depending on how it works for other peoples controllers, but for now it seems good

gizmo98:
@joolswills @HerbFargus 
Should we merge the current state or test more controllers? Herbfargus and i tested "Logitech F310", "Playstation 3 controller", "USB GamePad" and "HuiJia GamePad". I have no more controllers :-(.

joolswills:
I have a couple I could test, but I'm happy to merge as it is now - looks good - thanks :)

scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
```diff
@@ -0,0 +1,267 @@
...
+Y Axis = key(273,274)
+; Keyboard_END
+
+_EOF_
+    fi
+
+    # read temp device configuration and append to InputAutoCfg.ini
+    local deviceconfig=`cat /tmp/mp64tempconfig.cfg`
+    echo "$deviceconfig" >> "$file"
+    rm /tmp/mp64tempconfig.cfg
```
joolswills: I am about to change this block of code, but wanted to check just in case there is something I have overlooked - Why is there a variable used here over just doing `cat /tmp/mp64tempconfig.cfg >>"$file"` - going via a variable and using echo will strip the LF I believe - was that needed ?

joolswills: aah no, that only happens if quotes are not used on the echo. I'll just make this a cat.

Issue #1061, opened 2015-10-25
mupen64plus: add hotkey combos for exit, save state and load state


gizmo98:
Even if mupen64plus-input-sdl knows your controller there are no hotkeys for exit, save and load at the moment. The mupen64plus-core is responsible for exit, load and save hotkeys and it is working a little bit different. The setup of a exit hotkey for two ps3 controllers looks like "J0B16,J1B16". If you have controllers with different layouts it is important which one was connected first because the joypad number changes. If you connect in the wrong order buttons are wrong --> "J0B16,J1B1" or "J0B1,J1B16". Do overcome the "connect in the right order" problem i have written a udev rule which runs if a controller is added or removed. It uses controller config of emulationstation and writes known "select+start=exit", "select+L=load state" and "select+R=save state" combos to mupen64plus.cfg.     

I have opened up a repo for testing:
```
git clone --depth=1 https://github.com/gizmo98/mupen64plus-retropie-hotkeys.git
cd mupen64plus-retropie-hotkeys
sudo ./install.sh
```

If it does not work or you dislike this behavior type:
```
cd mupen64plus-retropie-hotkeys
sudo ./remove.sh 
```
Any suggestions?

HerbFargus:
I guess I'm not familiar enough with the configs in order to get this working. Do I need to reinstall mupen64 from source or do I have to set manual configs in mupen64plus.cfg?

My controller is logitech f310 so it is in the inputautocfg,ini so do I still need to set controls manually to incorporate hotkey support? Is what you discussed above related to this commit? https://github.com/mupen64plus/mupen64plus-core/commit/14bf40b236af1d5a5e7b86cd15a6cb628b4153fd

gizmo98:
You have found the right mupen64plus commit! Which video plugin are you using? I had no crash with GLideN64. 

A reboot is not really necessary. Just disconnect and reconnect your controller. udev rule will be triggered if a joypad/joystick device is added or removed. 

Your config file looks good!

HerbFargus:
I was on gles2n64 playing mario kart

gizmo98:
I rewrote the script. It's a normal bash script with hat and axis supports now.

HerbFargus:
Brilliant :+1:  Even works with the second player too. still crashes on load state but I think thats a mupen64plus bug unrelated to the hotkeys. (happened with glesn64 as well as gliden64)

gizmo98:
[...]

PR #1098, opened 2015-11-25
mupen64plus: add hotkey remapping

gizmo98:
https://github.com/RetroPie/RetroPie-Setup/issues/1061

### DIFF
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -1,0 +1,231 @@
+#!/usr/bin/env bash
+VIDEO_PLUGIN="$1"
+ROM="$2"
+configdir="/opt/retropie/configs"
+
+# arg 1: hotkey name, arg 2: device number, arg 3: retroarch auto config file
+function getBind() {
+    local key="$1"
+    local m64p_hotkey="J$2"
+    local file="$3"
+    local input_type
+    
+    iniConfig " = " "" "$file"
+
+    # search hotkey enable button
+    for input_type in "_btn" "_axis"; do 
+        iniGet "input_enable_hotkey${input_type}"
+        ini_value="${ini_value// /}"
+        if [[ -n "$ini_value" ]]; then
+            case "$input_type" in
+                _axis)
+                    ini_value="${ini_value//\"/}"
+                    m64p_hotkey+="A${ini_value:1}${ini_value:0:1}/"
+                    break
+                ;;
+                _btn)
+                    # if ini_value contains "h" it should be a hat device
+                    if [[ "$ini_value" == *h* ]]; then
+                        ini_value="${ini_value//\"/}"
+                        local dir="${ini_value:2}"
+                        ini_value="${ini_value:1}"
+                        case $dir in
+                            up)
+                                dir="1"
+                                ;;
+                            right)
+                                dir="2"
+                                ;;
+                            down)
+                                dir="4"
+                                ;;
+                            left)
+                                dir="8"
+                                ;;
+                        esac
+                        m64p_hotkey+="H${ini_value:0:1}V${dir}/"
+                        break
+                    else
+                        m64p_hotkey+="B${ini_value//\"/}/"
+                        break
+                    fi
+                ;;
+            esac
+        fi
+    done
+
+    # search hotkey
+    for input_type in "_btn" "_axis"; do 
+        # add hotkey and append enable hotkey button
+        # return if hotkey exists
+        iniGet "${key}${input_type}"
+        ini_value="${ini_value// /}"
+        if [[ -n "$ini_value" ]]; then
+            case "$input_type" in
+                _axis)
+                    ini_value="${ini_value//\"/}"
+                    m64p_hotkey+="A${ini_value:1}${ini_value:0:1}"
+                    echo "$m64p_hotkey"
+                    return
+                    ;;
+                _btn)
+                    # if ini_value contains "h" it should be a hat device
+                    if [[ "$ini_value" == *h* ]]; then
+                        ini_value="${ini_value//\"/}"
+                        local dir="${ini_value:2}"
+                        ini_value="${ini_value:1}"
+                        case $dir in
+                            up)
+                                dir="1"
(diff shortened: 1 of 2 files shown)


scriptmodules/emulators/mupen64plus/mupen64plus.sh
```diff
@@ -0,0 +1,265 @@
...
+function getDeviceName() {
+    local device="$1"
+    local device_name=""
+
+    # read device name
+    device_name=`cat /sys/class/input/${device}/device/name`
+
+    echo "\"$device_name\""
+    return
+}
```
joolswills: device_name isn't needed since the value is echoed, but I would just get rid of the whole function - only used in one place, and /sys/class/input/${device}/device/name can be used directly (note in bash you can do value=$(<file) instead of cat

scriptmodules/emulators/mupen64plus/mupen64plus.sh
```diff
@@ -0,0 +1,265 @@
...
+    local input_type
+    
+    iniConfig " = " "" "$file"
+
+    # search hotkey enable button
+    for input_type in "_btn" "_axis"; do 
+        iniGet "input_enable_hotkey${input_type}"
+        ini_value="${ini_value// /}"
+        if [[ "$ini_value" != "" ]]; then
+            case "$input_type" in
```
joolswills: styleguide - please use "[[ -n "$ini_value" ]]

scriptmodules/emulators/mupen64plus/mupen64plus.sh
```diff
@@ -0,0 +1,265 @@
...
+    local devices
+    local device_numbers
+    local device_counter=0
+    local device_number
+    
+    # get lists of all present js device numbers and device names
+    # get device count
+    for device_number in {0..10}; do
+        if [[ -h "/sys/class/input/js$device_number" ]]; then
+            devices[$device_counter]=`getDeviceName "js${device_number}"`
```
joolswills: please use $() instead of ``

scriptmodules/emulators/mupen64plus/mupen64plus.sh
```diff
@@ -0,0 +1,265 @@
...
+
+function remap() {
+    local devices
+    local device_numbers
+    local device_counter=0
+    local device_number
+    
+    # get lists of all present js device numbers and device names
+    # get device count
+    for device_number in {0..10}; do
```
joolswills: instead of hardcoding and looping through 0-10 you could loop through the actual existing devices (using ls or so) - there would be no need to check if the device existed then (if you need the id # you can easily extract that if needed)

scriptmodules/emulators/mupen64plus/mupen64plus.sh
```diff
@@ -0,0 +1,265 @@
...
+    local j
+    
+    for i in {0..2}; do
+        bind=""
+        for j in `seq 0 $device_counter`; do
+            # get name of retroarch auto config file
+            file=`getFile "${devices[$j]}"`
+            # abort if file does not exist
+            if [[ -f "$file" ]]; then
+                if [[ "$bind" != "" ]]; then
```
joolswills: `if [[ -f "$file" && -n "$bind ]]; then` or `[[ -f "$file" && -n "$bind ]] && bind+=","

joolswills:
ignore my last comment (I removed it now) I misread some code.

joolswills:
This is what I mean with the for loop (needs testing). associative array simplifies things, and we loop through actual existing devices.

```bash
function remap() {
    declare -A devices

    local device
    local device_num
    # get all joystick device and name
    for device in /dev/input/js*; do
        device_num=${device/\/dev\/input\/js/}
        devices[$device_num]=$(</sys/class/input/js${device_num}/device/name)
    done

    # read retroarch auto config file and use config for mupen64plus.cfg
    local file
    local bind
    local hotkeys_rp=( "input_exit_emulator" "input_load_state" "input_save_state" )
    local hotkeys_m64p=( "Joy Mapping Stop" "Joy Mapping Load State" "Joy Mapping Save State" )
    local i

    for i in {0..2}; do
        bind=""
        for device_num in "${!devices[@]}"; do
            # get name of retroarch auto config file
            file=$(grep --exclude=*.bak -rnw '/opt/retropie/configs/all/retroarch-joypads' -e "${devices[$device_num]}")
            file=${file%:*}
            file=${file%:*}
            # abort if file does not exist
            if [[ -f "$file" ]]; then
                [[ -n "$bind" ]] && bind+=","
                bind+=$(getBind "${hotkeys_rp[$i]}" "${device_num}" "$file")
            fi
        done
        
        # write hotkey to mupen64plus.cfg
        iniConfig " = " "\"" "/opt/retropie/configs/n64/mupen64plus.cfg"
        iniSet "${hotkeys_m64p[$i]}" "$bind" 
    done
}
```

joolswills:
updated above function.

joolswills:
Updated it again - the grep can have "-l" to list files. no need to process output.

joolswills:
I don't think grep needs `-r or -n` either.

joolswills:
and it can probably lose `-w` but adding `-F` might be useful

gizmo98:
Thank you. I will sort this out later.

gizmo98:
Ok. Tested with four controllers.

HerbFargus:
I can give it a test a bit later when I have some time

joolswills:
thanks.

HerbFargus:
Initially with the f310 and snes controller plugged in, input would work but the hotkeys wouldnt, I tried with just the snes controller and it worked fine. so I can confirm it's working, and the f310 issue is a driver or firmware bug unrelated to the hotkey functionality as there are similar issues with retroarch.  I eventually got both working after unplugging them and rebooting a few times.

gizmo98:
Thanks herbfargus. If problem pops up again post your mupen64plus.cfg please.

HerbFargus:
just for reference here it is

http://pastebin.com/nZyaRwnD

I'm pretty sure it's correct as far as I know. seems to be working

gizmo98:
Thanks. Looks correct.

PR #1099, opened 2015-11-26
Make omxplayer play over emulationstation (removes the need for the a…

HerbFargus:
…utostart delay)

thanks to Chito
http://blog.petrockblock.com/forums/topic/how-can-you-get-video-splash-screen-to-play-all-the-way-to-when-es-is-loaded/


joolswills:
Ta

PR #1102, opened 2015-11-27
add warning to kill emulationstation before using sselph's scraper

HerbFargus:
from http://blog.petrockblock.com/forums/topic/scraper-little-details-to-solve/#post-110776

Unless you feel there is a better programmatic way to automate killing of processes when people start the setup script from within emulationstation


joolswills:
We should probably give an option of exiting es or something but this is a good step. cheers.

PR #1105, opened 2015-11-29
n64: use mupen64plus as default emulator

gizmo98:
-switch default emulator from lr-mupen64plus/rice to mupen64plus/rice.
Don’t use glesn64 or GLideN64 because glesn64 does crash with certain
roms and GLideN64 is to slow at HD resolutions.
-remove zip file extension because mupen64plus does not support zip
files.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -76,7 +76,7 @@ megadrive_fullname="Sega Mega Drive / Genesis"
 msx_exts=".rom .mx1 .mx2 .col .dsk .zip"
 msx_fullname="MSX / MSX2"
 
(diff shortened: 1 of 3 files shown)


joolswills:
any details regarding the switch ?

gizmo98:
lr-mupen64plus is broken as hell. The only working video plugin is rice at the moment. lr-mupen64plus's rice plugin has visual issues. Lighting is broken (compare Super Mario 64 intro). There are visual corruptions in Zelda games an Mario Kart. It is slow even if you use low resolutions.

With the new input script(s) mupen64plus can be setup as easy as a libretro core.

joolswills:
thanks.

PR #1109, opened 2015-12-02
Add reicast upstream repo

gizmo98:
-use reicast upstream repo
-remove or replace cpu related flags
-use system libs instead of precompiled libs
-link emu.cfg and mappings dir into /opt/retropie/configs/dreamcast/
-update input script. create controller config files under
/opt/retropie/configs/dreamcast/mappings
-reicast does not have a dynamic setup. The current event device and
mapping file muste be stored in emu.cfg.
-add startup script. Startup script sets the current /dev/input/event*
device and related mapping file in
/opt/retropie/configs/dreamcast/emu.cfg.

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -15,65 +15,69 @@ rp_module_menus="4+"
 rp_module_flags="!rpi1"
 
 function depends_reicast() {
-    getDepends libsdl1.2-dev alsa-oss
+    getDepends libsdl1.2-dev python-dev python-pip alsa-oss
 }
 
 function sources_reicast() {
-    gitPullOrClone "$md_build" https://github.com/free5ty1e/reicast-emulator.git free5ty1e/rpi2/retropie-reicast-updated
+    gitPullOrClone "$md_build" https://github.com/reicast/reicast-emulator.git
+    # tune flags
+    sed -i 's/-mtune=cortex-a9//g' $md_build/shell/linux/Makefile
+    sed -i 's/-mfpu=neon/-mfpu=neon-vfpv4/g' $md_build/shell/linux/Makefile
+    sed -i 's/-mfloat-abi=softfp/-mfloat-abi=hard/g' $md_build/shell/linux/Makefile
+    # don't use precompiled stuff. use system libs
+    sed -i 's|-L../linux-deps/lib||g' $md_build/shell/linux/Makefile
+    sed -i 's|-I../linux-deps/include||g' $md_build/shell/linux/Makefile
+    # fix reicast-joyconfig script
+    sed -i 's|"dreamcast", "btn_escape"|"emulator", "btn_escape"|g' $md_build/shell/linux/tools/reicast-joyconfig.py
 }
 
 function build_reicast() {
-    cd "$md_build/shell/rapi2"
+    pushd "$md_build/shell/linux"
     make clean
-    make 
-    md_ret_require="$md_build/shell/rapi2/reicast.elf"
+    make platform=rpi2
+    popd
+    md_ret_require="$md_build/shell/linux/reicast.elf"
 }
 
 function install_reicast() {
+    pushd "$md_build/shell/linux"
+    make PREFIX="$md_inst" install
+    popd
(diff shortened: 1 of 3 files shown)


joolswills:
Thanks!

joolswills:
rather than symlinking certain folders/files in $home/.reicast it would probably be best to just symlink the entire folder to  $configdir/dreamcast - which is what we do with other modules.

gizmo98:
True, but if we do this BIOS files (stored under .reicast/data/) will be in the same directory as config files (.reicast/ and .reicast/mappings/). Moreover reicast writes some other stuff into ./reicast which is neither config nor bios file.

joolswills:
Well, I think it simplifies things but ok. note this is no different from mame where we have savegames/nvram and so on - we symlinks the entire $home folders for everything else, so it is more consistant imho.

joolswills:
It doesn't build here correctly - it builds, and then builds again during install and fails. I'll make some tweaks as it should inherit our cflags etc. I think we might as well let it use the shipped with dependencies like libpng - it's a lot newer than the wheezy one etc.

joolswills:
the make clean doesn't clean the right target either. I have it mostly working here now. just a few more tweaks.

gizmo98:
Initial commit build some hours earlier. I removed all old reicast directories and config files and reicast started.

joolswills:
well, unless i broke it with some of my tweaks ;-)

gizmo98:
No problem here:
http://pastebin.com/gJBmSD03

Crosscompiler issue?

joolswills:
it's because the platform flag gets set correctly when building on the real pi, as it checks /proc/cpuinfo and sets platform=rpi2 then. So it does need correcting to include platform=rpi2 for the make clean / make install to work on say a chroot etc.

joolswills:
The symlinks need to be the other way around too - we want to link the /home/pi/.reicast to the new location  in /opt/retropie/configs so the actual configs live in /opt/retropie/configs etc

joolswills:
I will correct this - I think I'm just going to go for a single symlink though - data files can be ignored.

joolswills:
made a bunch of changes. you might need to delete your /opt/retropie/configs/dreamcast folder due to the reversal of the symlinks. I forked the repository to simplify the customisations rather than having lots of seds. I also did some fixes to the configuration section (files ownership etc). needs some testing in case anything got broken along the way. the files created by the emulator contain the bios settings etc right so it does actually make sense to have them in the retropie config area so they can be backed up etc. I made some changes in the launch script too as it had hardcoded /home/pi paths (retropie-setup can be installed by any username)

HerbFargus:
not sure if I'm configuring controls incorrectly... but my controls aren't working. 

emu.cfg creates them properly:
```
[Logitech Gamepad F310]
axis.0 = Axis_X
axis.1 = Axis_Y
button.0 = Btn_A
button.1 = Btn_B
button.2 = Btn_X
button.3 = Btn_Y
button.4 = Axis_LT
button.5 = Axis_RT
button.6 = Quit
button.7 = Btn_Start

[USB gamepad]
axis.0 = Axis_X
axis.1 = Axis_Y
button.0 = Btn_Y
button.1 = Btn_B
button.2 = Btn_A
button.3 = Btn_X
button.4 = Axis_LT
button.5 = Axis_RT
button.8 = Quit
button.9 = Btn_Start
```

I don't see any equivalent in the mappings folder though, perhaps I misunderstand how they work. 

I deleted all reicast folders, updated the setup script from the latest reicast related commit, and reconfigured my controllers from emulationstation and transferred over my bios files. It all loads up but there are no controls working.

joolswills:
it is of course quite possible I have broken something with a bunch of recent changes :) Possibly due to the changes I've made to the configuration symlinking etc. I will check on my controller when I get a mo.

joolswills:
did you re-install emulationstation btw ? you will need to do that to get the latest reiccast controller setup script (and then reconfigure controller)

HerbFargus:
May be the step I missed, giving it a go.

HerbFargus:
Yep my bad, sorry for the noise. Works great.

PR #1112, opened 2015-12-03
Remove Wiki Folder (Outdated)

HerbFargus:
I don't know if it's worth the trouble to keep wiki images with the retropie setup script, I change the wiki as frequently as retropie is developed so I don't see the value in keeping the images here, and if I did keep them somewhere, it would probably be on a separate repo. It seems that github keeps their image upload links indefinitely (as far as I've seen)- so its likely I don't need a repo at all.


joolswills:
Agreed.

Issue #1077, opened 2015-11-04
Enabling option to configure Retroarch controller


biscuits99:
I know that recently the option to configure a retroarch controller has been removed to avoid confusion that it may be required now that ES handles the config file generation:
https://github.com/RetroPie/RetroPie-Setup/commit/6d7d74cd781af3537b190ecac022bd29938f1d03

However, I've found some instances where the ES generation doesnt always match the old style of generation. An example is with the 8bitdo controllers, there are mismatches on some of the buttons although the axis are fine. This is referred to here in step 5 and 6:
https://github.com/RetroPie/RetroPie-Setup/wiki/Setting-up-a-Bluetooth-controller

I think this user *may* also be having the same issues with the non-descript controller:
http://blog.petrockblock.com/forums/topic/neogeo-exit-button-not-working/#post-109124

Would it be possible to re-instate this menu option, albeit with a warning that it shouldnt be required unless you know what you are doing type message?
I'm happy to provide the retroarch cfg file with ES and the old file for comparison if that helps?

biscuits99:
Different button mapping shown here:

http://pastebin.com/raw.php?i=zpJaMY7P

joolswills:
It was only removed from the ES menu - it's still available from retropie-setup. The problem was people kept using it, and usually it isn't needed.

I would need the es_temporaryinput.cfg file too which contains the full config generated by ES sent to our input configuration tool - and if you know which bits are generated wrong, it will save me time from comparing each config line etc :)

joolswills:
updated above post.

joolswills:
based on the configs above - I can see that the ID for button B is different - I cannot see how that can happen unless a different button is pressed for B in ES vs what is pressed for B in retroarch-joyconfig.

joolswills:
(I see a lot of the button ids are higher - so there may be something odd at play regarding how ES handles the mapping vs ES, but very strange). The controllers were connected in exactly the same way / in the same mode in both cases ? what does button B output from jstest ?

joolswills:
Also, does it make a difference if you switch the input driver in retroarch from udev to SDL2 (or linuxraw) ?

joolswills:
Sorry for ticket spam - I have updated the above message so please re-read thread.

biscuits99:
Ah, I see - my mistake, I thought it was being removed from the retropie-setup. Ok, no real problem then, at least there is a fallback option.

I have put the es_temporaryinput.cfg here:
[...]

PR #1115, opened 2015-12-05
configedit.sh: add input joypad driver option

gizmo98:
https://github.com/RetroPie/RetroPie-Setup/issues/1077


joolswills:
thanks.

PR #1129, opened 2015-12-15
Add logs and tmp folders in the root to gitignore

mamoit:
Both the tmp and logs folders located in the root are generated, and should be ignored.


joolswills:
thanks.

PR #1149, opened 2015-12-25
retroarch input script: don't set keyboard hotkeys

gizmo98:
Keyboard hotkeys should not be changed here. There are other modules
doing this.
https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/supplementary/retroarchinput.sh


gizmo98:
Ok. Xin-Mo controllers with keyboard emulation need this. But it still interferes with hotkey module. And there is no fallback to disable all hotkeys if no menu toggle button is mapped for keyboards.

joolswills:
also my picade etc.

joolswills:
hotkey module is destined to be remade at some point - it's currently a bit of a hack and predates the input configuration stuff

Issue #1148, opened 2015-12-25
Add platform x86


gizmo98:
I play with the idea to use a cheap x86 barebone for RetroPie as a speed upgrade.
I have modified the script a little bit so platform x86 will be detected:
https://github.com/RetroPie/RetroPie-Setup/compare/master...gizmo98:x86?expand=1

Tested with Debian Jessie, Cinammon, VirtualBox, snes9x-next, nestopie, genesis-plus-gx and emulationstation. ;-)

The user has to be in group admins to run setup script. RetroPie app launcher can be found in start menu (type retropie in search box).

joolswills:
Looks good! :)

PR #1150, opened 2015-12-25
Initial "platform x86" PR

gizmo98:
https://github.com/RetroPie/RetroPie-Setup/issues/1148

@joolswills @HerbFargus 
I don't see any problem using this alongside with rpi platforms. Please take a look if i have missed something. Retroarch cores (vba-next, snes9x-next, gambatte, genesis-plus-gx, etc.) and emulationstation run under an emulated x86 Debian Jessie, X11 and OpenGL support.

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -12,6 +12,7 @@
 rp_module_id="advmame"
 rp_module_desc="AdvanceMAME"
 rp_module_menus="2+"
+rp_module_flags="!x86"
 
 function depends_advmame() {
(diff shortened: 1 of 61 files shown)


joolswills:
Thanks. Looks good.

I think I will change around some of the platform stuff soon - possibly a separate variable, and have it so it exclude by default, which may make it simpler than having module exclusions etc.

gizmo98:
Oh. I had to squash all stuff again. Don't know why github desktop separated all commits again. :(

joolswills:
Don't worry. Most of the commits are descriptive additions rather than corrections, so it seems fine to have them split.

PR #1161, opened 2015-12-31
emulationstation: swap a/b buttons

gizmo98:
Swap a/b buttons for modern controllers:
a = ok = X (Playstation) = A (XBOX)
b = abort = O (Playstation) = B (XBOX)

Old mapping is:
a = ok = O (Playstation) = B (XBOX)
b = abort = X (Playstation) = A (XBOX)

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/emulationstation.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/emulationstation.sh
@@ -54,8 +54,14 @@ function map_emulationstation_joystick() {
         rightbottom)
             key="pagedown"
             ;;
-        up|right|down|left|a|b|start|select)
+        up|right|down|left|start|select)
             key="$input_name"
+            ;;
+        a)
+            key="b"
+            ;;
+        b)
+            key="a"
             ;;
         *)
             return


HerbFargus:
Sounds fair to me

gizmo98:
I was always wondering why i should press red O to accept and blue X to abort. A playstation behaves the other way around ;-)

jaimerod:
Thanks for this update. It was driving my wife crazy.

zerojay:
On PlayStation controllers, the main facebuttons have specific uses. X = cancel, O = accept, Square = menu, Triangle = viewpoint. (The colors of the buttons are meaningless.) Sony in North America decided to swap X and O and has caused endless annoyance to us developers for decades now.

tynor88:
:thumbsup: for this one

joolswills:
I'm not sure about this as it swaps it for all controller including snes style pads etc correct?. I'm certain we will have other users complaining if we change this now. I think we should get some more feedback on it at least first.

joolswills:
Right now I would say that I am against the change though.

InsecureSpike:
correct me if I'm wrong, but can't the user just edit their .cfg files? and once edited they just need to be backed up?

joolswills:
Yes.

gizmo98:
What do you think of a detection like "if analog sticks are present it is a modern controller" ;-)

joolswills:
That makes sense, but I would like to hear some user feedback on the forum. I don't want to push out the next release and have lots of complaints about this change :)

joolswills:
If we have a poll or ask on the forum, we can at least argue that we did get feedback first to those that don't like the change.

gizmo98:
Sounds good.

Gaander:
Where can I find said .cfg to swap the buttons in ES?

Gaander:
oh... found it. I didn't know it was attached here.... my bad

gizmo98:
@joolswills 
What do you think of a optIn option like `getAutoConf emulationstationSwapButtons && A=B`? So it is only necessary to enable button swap in our autoconfig file.

joolswills:
im fine with that, but i think "swap_a_b" or "swap_buttons" something as a key ? (not camelcase)

gizmo98:
Looks good. Much shorter and better to read.

gizmo98:
Close this one because emulationstation module of this branch is too old.

joolswills:
you could rebase it on top of upstream/master and force push etc btw.

joolswills:
(upstream being whatever your repository name for the main repo is)

gizmo98:
It was faster this way :)

MetalManTN:
Ok guys, I'm sorry for the stupid question but how do I enable this new option. I hate using circle to enter/select and X to go back on my PS3 DualShock 3. I want to enable the swap, but cannot figure out how. I have updated binaries, and have verified that I have the updated code. Thanks.

gizmo98:
Update RetroPie-Setup and do a source based installation of emulationstation. Open /opt/retropie/configs/all/autoconf.cfg and set `es_swap_a_b = 1`. Configure input again and restart es. X/O button mapping should be correct now.

MetalManTN:
@ gizmo98 Thank you so much! I had switched the 0 to 1, but I must not have done the other steps properly. I really appreciate the help. The button swap is wonderful!

PR #1163, opened 2015-12-31
Set FBA, NeoGeo and Arcade to use MAME Scraper

zerojay:
Set up scraper to use the -mame parameter (scrape by filename) when scraping system names beginning with "fba", "neogeo" and "arcade" as mentioned on https://github.com/sselph/scraper.


scriptmodules/supplementary/scraper.sh
```diff
@@ -47,7 +47,7 @@ function scrape_scraper() {
     params+=(-output_file "$gamelist")
     params+=(-rom_dir "$romdir/$system")
     params+=(-workers "4")
-    [[ "$system" =~ ^mame- ]] && params+=(-mame -mame_img t,m,s)
+    [[ "$system" =~ ^mame- ]] || [[ "$system" =~ ^fba ]] || [[ "$system" =~ ^neogeo ]] || "$system" =~ ^arcade ]] && params+=(-mame -mame_img t,m,s)
```
joolswills: We only use ^mame- to match all the rom folders prefixed with mame - if you want to use others just either match individual strings or use a regular expression with |

`[[ "$system" =~ ^mame- || "$system" =~ fba|neogeo|arcade ]]`

joolswills:
I don't think we have an "arcade" system either currently either.

scriptmodules/supplementary/scraper.sh
joolswills: can probably be simplified to

`[[ "$system" =~ ^mame-|fba|neogeo ]]`

zerojay:
Indeed, there is not an "arcade" system, however one of the popular multiboot SD card distribuions has custom made an arcade system which allows users to consolidate all their MAME/FBA/NeoGeo games into one set. As it's such a simple change in RP, I figure it wouldn't be a problem to add it to the list of possible supported systems for the scraper, but I'll let you make a decision on that.

scriptmodules/supplementary/scraper.sh
zerojay: Apologies, forgot that we now just have a single fba rom directory.

joolswills:
I will leave out arcade for now - if we ever add it ourselves we can put it in. Please can you squash these commits and I will merge. thanks.

zerojay:
Closing as requested.

joolswills:
I'm not asking for it to be closed. I wanted the commits squashed and force pushed back to the branch. I'll manually implement the changes though. thanks.

joolswills:
I squashed your commits and merged. thanks.

PR #1164, opened 2016-01-02
Wonderswan and Neogeo Seperations

Omnija:
Would like Neogeo Pocket/Color and Wonderswan/Color as seperate folders.
1 for organization purposes and as well to maintain seperate folders for seperate systems.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -85,8 +85,11 @@ nds_fullname="Nintendo DS"
 neogeo_exts=".fba .zip"
 neogeo_fullname="Neo Geo"
 
-ngp_exts=".ngp .ngc .zip"
-ngp_fullname="Neo Geo Pocket (Color)"
+ngp_exts=".ngp .zip"
+ngp_fullname="Neo Geo Pocket"
+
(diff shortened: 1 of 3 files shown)


joolswills:
Thanks but I don't want to split these up currently as it means lots of theme work, and for most people I think it's fine with a single rom folder for them.

Omnija:
I already submitted a small update to the theme so you wouldn't have to worry about it. I also thought it would be a good idea for the same reasons gb and gbc are seperate.

joolswills:
OK. I will get some feedback from others and decide. @HerbFargus / @gizmo98 - what do you think ?

HerbFargus:
I have no problem with it, as he does make a valid point that they are technically closer to the gb and gbc than they are the Genesis/megadrive. And he did add them to the theme, so that's not an issue (and I don't mind updating themes for some others)

gizmo98:
Makes sense. We have separate folders for handhelds gb and gbc as well.

joolswills:
ok. Will merge. thanks for feedback.

PR #1165, opened 2016-01-04
Odroid-c1

joolswills:
initial support for odroid-c1 (other mali boards should be possible based on this with a few changes).

probably will add some platform flags, so we can do logic like `hasPlatformFlag "mali"` etc

also want to change the module code so we have platform flags in a separate variable perhaps, but include them rather than exclude, as I think it will be simpler that way.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -42,7 +42,8 @@ function sources_retroarch() {
 
 function build_retroarch() {
     local params=(--disable-x11 --disable-ffmpeg --disable-sdl --disable-oss --disable-pulse --disable-al --disable-jack --enable-floathard)
-    isPlatform "rpi2" && params+=(--enable-dispmanx --enable-sdl2 --enable-neon)
+    isPlatform "rpi" && params+=(--enable-dispmanx --enable-sdl2)
+    isPlatform "rpi2" && params+=(--enable-neon)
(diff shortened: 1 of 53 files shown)


scriptmodules/emulators/retroarch.sh
```diff
@@ -30,11 +35,15 @@ function sources_retroarch() {
...
+    if isPlatform "odroid"; then
+        sed -i 's|struct mali_native_window native_window|fbdev_window native_window|' gfx/drivers_context/mali_fbdev_ctx.c
+    fi
 }
 
 function build_retroarch() {
-    local params=(--disable-x11 --enable-dispmanx --disable-oss --disable-pulse --disable-al --disable-jack --enable-sdl2 --enable-floathard)
-    isPlatform "rpi2" && params+=(--enable-neon)
+    local params=(--disable-x11 --disable-ffmpeg --disable-sdl --disable-oss --disable-pulse --disable-al --disable-jack --enable-floathard)
+    isPlatform "rpi2" && params+=(--enable-dispmanx --enable-sdl2 --enable-neon)
```
gizmo98: Shouldn't there be a extra line for rpi1 and rpi2 like `isPlatform "rpi" && params+=(--enable-dispmanx --enable-sdl2)`?

joolswills: yes. Good spot. Will sort.

joolswills:
I think I'm, happy to merge this now. Can improve it later. Let me know if you spot anything else though - cheers.

gizmo98:
Looks good! :+1:

PR #1171, opened 2016-01-05
Use safer range expression in get_save_vars

### LINKED ISSUE #1166: $dispmanx_conf is ignored on some systems
I've installed RetroPie via binary installation on [OSMC](https://osmc.tv) with help of [mcobit/retrosmc](https://github.com/mcobit/retrosmc) script. Everything works well so far except for emulators/ports that need dispmanx despite having it properly configured in `$configdir/all/dispmanx.cfg`.

My example is OpenTyrian
Running it from EmulationStation always ends with such entry in `/tmp/runcommand.log`:
> `failed to initialize any supported video`

1. I've read in [discussion](https://github.com/RetroPie/RetroPie-Setup/issues/1050#issuecomment-148933668) under #1050 that this is caused by unset dispmanx flag.
2. I double-checked file permissions of `$configdir/all/dispmanx.cfg` and its contents. File permissions were fine, and `dispmanx.cfg` had line:
> opentyrian="1"

3. I've also added to `runcommand.sh` tiny debug-purpose line `echo $save_emu > /tmp/save_emu`
right after calling `get_save_vars`. Resulting file after starting the game from ES was blank.
4. Then I tried also to comment out [line 91](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/supplementary/runcommand/runcommand.sh#L91) and after that OpenTyrian worked flawlessly and `/tmp/save_emu` had one line with 'opentyrian` in it.

It seems then that the problem lies in this particular substitution:

```
save_emu=${save_emu//[^a-Z0-9_\-]/}
```

Right now I don't have any idea why this substitution doesn't work. Maybe it's related with some bash-4.3 (this is the bash version on OSMC) changes? I believe on Raspbian Wheezy there is still bash-4.2. Could be anything else of course - I'm not a Bash-related expert :)

#### Sidenote:
Before installation of OpenTyrian I had to:
- install **mercurial** via `apt-get install mercurial`. `retropie_setup.sh` didn't succeed with that for some reason.
- install **unzip** via `apt-get install unzip` - default version of unzip on OSMC doesn't have option `-j` used by `retropie_setup.sh` during extraction of *tyrian21.zip*.

mruwek:
a-Z depends on LC_COLLATE or LC_ALL and can produce weird bugs if user
have set there something other than what we're using by default.

a-zA-Z should work just everywhere.

Fixes #1166 and mcobit/retrosmc#14

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -88,7 +88,7 @@ function get_params() {
 function get_save_vars() {
     # convert emulator name / binary to a names usable as variables in our config files
     save_emu=${emulator//\//_}
-    save_emu=${save_emu//[^a-Z0-9_\-]/}
+    save_emu=${save_emu//[^a-zA-Z0-9_\-]/}
     save_emu_render="${save_emu}_render"
     fb_save_emu="${save_emu}_fb"
     save_rom=r$(echo "$command" | md5sum | cut -d" " -f1)


joolswills:
Thanks for debugging this.

PR #1172, opened 2016-01-05
Add dependency for unzip in OpenTyrian port

mruwek:
This will force the installation script to use proper version of unzip on systems using other implementations (e.g. busybox unzip).


joolswills:
I added unzip dependencies for everything as it was used by other modules so this isn't needed now. But cheers.

PR #1176, opened 2016-01-07
emulationstation: fix shutdown/restart

gizmo98:
If sudo requests a password (raspbian user !pi or debian/ubuntu) emulationstation cannot shutdown or restart a system. Sudo is hardcoded. Emulationstation expects sudo without
password request. https://github.com/Aloshi/EmulationStation/blob/646bede3d9ec0acf0ae37841
5edac136774a66c5/es-core/src/platform.cpp#L47

### DIFF
--- a/scriptmodules/supplementary/emulationstation.sh
+++ b/scriptmodules/supplementary/emulationstation.sh
@@ -124,4 +124,8 @@ _EOF_
     mkdir -p "/etc/emulationstation"
 
     configure_inputconfig_emulationstation
+    
+    # run sudo without password so emulationstation can shutdown and restart system
+    local file="/etc/sudoers"
+    grep -q "$user" $file || echo "$user ALL=(ALL) NOPASSWD:ALL" >> $file
 }


joolswills:
I don't think we should change stuff like this from retropie-setup - it is up to the user if installing themselves to set this up and we can document it, or add it to raspbian-tools or something.

joolswills:
but messing with security related things like sudo config from the ES script is a big no no from me. sorry.

joolswills:
BTW i am going to change the shutdown and restart from ES so its handled by the launch script, in which case a password could be entered if needed.

gizmo98:
It's how the user pi is setup in raspbian. If you do mind to add this fix we should remove emulationstation restart/shutdown menu items because emulationstation (or the whole system) hangs if this fix is not applied.

joolswills:
See my previous comment

gizmo98:
Too late ;)

joolswills:
as with recalbox, I will just create a file when doing shutdown/restart system/quit (and add a restart emulationstation). our launch script will then handle each case. It's the simplest solution and a bit of a compromise, but I couldn't get ES code to do things like rescan roms and update gui safely (sort of worked, but there were problems).

joolswills:
I don't mind if this is added as a feature to raspian-tools (this needs splitting out / renaming as some stuff isn't raspbian specific - eg updating packages), but it should say what it is doing. also you can use /etc/sudoers.d/ rather than editing sudoers directly. Easier to maintain then.

joolswills:
I see you updated your post. Those users who want to shutdown without password should use our image which is ready set up, or be aware they need to allow their user to restart without a password.

joolswills:
other methods to shutdown without root, or using sudo (btw sudo could be setup to allow specific commands of course), would be to make shutdown SUID, but there are security concerns with this.

joolswills:
also via systemd for jessie possibly  - systemctl poweroff etc.

gizmo98:
I will test systemctl poweroff. It's no raspbian stuff. RetroPie-Setup runs quite nice on Ubuntu 15.10 x86_64 (and Debian Jessie) now. It's just like:
-do a sourced based installation
-install an es theme
-enable autostart
The only thing bothering me is shutdown/restart hang ;-)

joolswills:
it's not a big issue to add that in the notes though.

gizmo98:
I will add it as a note to a x86 installation guide. For now i will dig around and get some more modules working.

PR #1178, opened 2016-01-08
[xboxdrv.sh] Added user-configurable analog deadzones and number of controllers to enable.

zerojay:
Added new menu items to the xboxdrv.sh scriptmodule:

* Number of controllers to attempt to enable (default: 2)
* Size of analog deadzone to use for all controllers (default: 4000)

### DIFF
--- a/scriptmodules/supplementary/xboxdrv.sh
+++ b/scriptmodules/supplementary/xboxdrv.sh
@@ -31,7 +31,21 @@ function install_xboxdrv() {
 }
 
 function enable_xboxdrv() {
-    local config="\"$md_inst/bin/xboxdrv\" --daemon --detach --dbus disabled --detach-kernel-driver --id 0 --led 2 --deadzone 4000 --silent --trigger-as-button --next-controller --id 1 --led 3 --deadzone 4000 --silent --trigger-as-button"
+    if [[ -n "$1" ]]; then
+
+        # Because function return codes are limited to 0-255 range, we could not leave this calculation in the deadzone_xboxdrv routine or we'd get weird results.
+        local deadzone="$((($2-1) * 500))"
+
+	local config="\"$md_inst/bin/xboxdrv\" --daemon --detach --dbus disabled --detach-kernel-driver --id 0 --led 2 --deadzone $deadzone --silent --trigger-as-button"
+        local loop="1"
+
+	while [ "$loop" -lt "$1" ]
+        do
+            config+=" --next-controller --id $loop --led $(($loop+2)) --deadzone $deadzone --silent --trigger-as-button"
+            loop=$(($loop+1))
+        done
+    fi
+    
     if ! grep -q "xboxdrv" /etc/rc.local; then
         sed -i "s|^exit 0$|${config}\\nexit 0|" /etc/rc.local
         printMsgs "dialog" "xboxdrv enabled in /etc/rc.local with the following config\n\n$config\n\nIt will be started on next boot."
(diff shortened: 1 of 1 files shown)


joolswills:
thanks - made a few comments - if you can make those changes and squash the commits / force push back to the branch I will merge. Cheers.

zerojay:
I believe that I've addressed just about everything you've mentioned with a few more commits tonight. Regarding the $controllers_wanted being a global, to my knowledge (which is admittedly very limited) local variables created within a code block are accessable by anything called within that code block, so I wouldn't call it a global exactly. If it is still an issue as it is now, can you point me to an example of how we're using function parameters elsewhere in the project? I'm also seeing similar examples that don't appear to be using function parameters (such as scriptmodules/supplementary/retronetplay.sh). I'm willing to make the change, I just would like to get pointed in the right direction. I'm still new to learning some of this stuff, so my apologies.

Once I'm able to address this, I'll then squash/force push everything.

joolswills:
retronetplay is older and needs some work. Not that many require parameters bu esthemes.sh / t ps3controller / retroarchautoconf.sh do. You can then call them via retropie_packages with parameters also which can be useful for automation.

sudo ./retropie_packages module function parameter.

scriptmodules/supplementary/xboxdrv.sh
```diff
@@ -31,7 +31,18 @@ function install_xboxdrv() {
 }
 
 function enable_xboxdrv() {
-    local config="\"$md_inst/bin/xboxdrv\" --daemon --detach --dbus disabled --detach-kernel-driver --id 0 --led 2 --deadzone 4000 --silent --trigger-as-button --next-controller --id 1 --led 3 --deadzone 4000 --silent --trigger-as-button"
+    if [[ -n "$controllers_wanted" ]]; then
+
+	local config="\"$md_inst/bin/xboxdrv\" --daemon --detach --dbus disabled --detach-kernel-driver --id 0 --led 2 --deadzone $deadzone_wanted --silent --trigger-as-button"
+        local loop="1"
+
+	while [ "$loop" -lt "$controllers_wanted" ]
```
joolswills: could use a for loop:

```
local controllers="$1"
local loop
for (( loop=1; loop<=$controllers; loop++ )); do
```

joolswills: doesn't actually metter - just mentioning it. please use double brackets though as per our styleguide https://github.com/RetroPie/RetroPie-Setup/wiki/Shell-Style-Guide

zerojay:
I've used function parameters as you requested (I believe) and everything's been squashed and force pushed. This should be ready to merge. Please let me know if there's anything else that might need to be changed.

joolswills:
Thanks.

HerbFargus:
Great job on the script!

Just out of curiosity for usage of this script:

As far as I understand the process, 
- One first needs to select the number of controllers to enable, 
- then enable the driver (which writes the config to rc.local) (if no controllers are selected first, what is the default number of controllers it enables?)
- then if one wishes to change the controls then he must disable the controls (which removes the config from rc.local)
- then choose a new amount of controllers 
- then choose enable. 

For brevity's sake and usability would it make more sense to include the amount of controllers as a submenu of the enable controller option as users may not understand the order of setting the number of controllers first? Or should I just  update the wiki to explicitly state the order?

Also this is just me being picky but Is there a reason why we don't just use sed/grep to automatically replace the line for a new enable rather than making the user disable it first before they can change it?

joolswills:
by default it would set up for 2 controllers

it will only write to /etc/rc.local using "enable" if no config is already there - to replace a config you have to disable / re-enable

in regards to your points, I think that would be useful, and I had some other ideas too, but thought I would sort that later (I have already made some changes since the pull request)

HerbFargus:
Ok, I'll hold off on updating the wiki until its all finished then. I only have two xbox controllers, but I'm happy to test any new adjustments.

zerojay:
You've got it right and I agree. It was just how the script worked originally so I didn't mess with how it did that yet. Ideally, I'd like to be able to allow players to configure their controllers independently in a menu somewhere along these lines (excuse the ascii menus, use your imagination!):

-= Xboxdrv =-
1 - Xboxdrv Options (Enabled)
2 - Controller 1 (Enabled)
3 - Controller 2 (Enabled)
4 - Controller 3 (Disabled)
5 - Controller 4 (Disabled)

-=Xboxdrv Options=-
1 - Xboxdrv (Enabled)
2 - (whatever that option is to write something to /boot/config.txt)
3 - (whatever that option is to remove that something we wrote to /boot/config.txt)

-= Controller Menu =-
1 - Controller 1 (Enabled)
2 - Treat Controller 1 D-pad as buttons (Enabled)
3 - Treat Controller 1 triggers as buttons (Enabled)
4 - Set Controller 1 analog stick deadzone (4000)
5 - Set Controller 1 trigger deadzone (5%)

So... baby steps, but that's what I have in my head as an eventual goal. Feel free to let me know if you feel this wouldn't work or that there's a better way to go about this.

joolswills:
im done for now. I just made a small improvement so if there is an existing config it will ask if you want to update.

PR #1180, opened 2016-01-10
[xboxdrv] Switch to xboxdrv repo with pending pull requests/fixes

zerojay:
Updating to point to a new xboxdrv repo that integrates the following pull requests from the original xboxdrv repo:

137 - Fix zero-length rumble packets causing infinite rumble on some controllers.
169 - Add support for MadCatz Xbox 360 Marvel vs Capcom Tournament Edition Fightstick.
173 - Fix minor typo.
174 - Fix USB read error causing controller to be unusable until xboxdrv is restarted manually. (Happens somewhat frequently to me, personally.)
175 - Add support for Xbox One controllers (rumble and LED lights not functional yet).

### DIFF
--- a/scriptmodules/supplementary/xboxdrv.sh
+++ b/scriptmodules/supplementary/xboxdrv.sh
@@ -27,7 +27,7 @@ function depends_xboxdrv() {
 }
 
 function sources_xboxdrv() {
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

PR #1183, opened 2016-01-10
x86 stuff

gizmo98:
-Add mupen64plus
-Add lr-desmume
-Add lr-ppsspp
-Add desktop file with app icon (type RetroPie in startmenu)
-Refine autostart
-Add case amd64
-Use -march=native because platform x86 is always a source based installation


scriptmodules/emulators/mupen64plus.sh
```diff
@@ -41,28 +50,22 @@ function sources_mupen64plus() {
...
 function build_mupen64plus() {
     rpSwap on 750
 
     local dir
-    local params
+    local params=()
     for dir in *; do
         if [[ -f "$dir/projects/unix/Makefile" ]]; then
             make -C "$dir/projects/unix" clean
-            params=()
```
joolswills: It's not necessary to reset params for each plugin etc anymore ?

scriptmodules/supplementary/emulationstation.sh
```diff
@@ -17,7 +17,8 @@ function depends_emulationstation() {
     getDepends \
         libboost-locale-dev libboost-system-dev libboost-filesystem-dev libboost-date-time-dev \
         libfreeimage-dev libfreetype6-dev libeigen3-dev libcurl4-openssl-dev \
-        libasound2-dev cmake libsdl2-dev libsm-dev
+        libasound2-dev cmake libsdl2-dev
```
joolswills: libsm-dev is required as a dependency as if it exists on the system cmake decides to link with it.

scriptmodules/supplementary/emulationstation.sh
```diff
@@ -124,4 +130,8 @@ _EOF_
     mkdir -p "/etc/emulationstation"
 
     configure_inputconfig_emulationstation
-}
+
+    # run sudo without password so emulationstation can shutdown and restart system
+    local file="/etc/sudoers"
+    grep -q "$user" $file || echo "$user ALL=(ALL) NOPASSWD:ALL" >> $file
```
joolswills: We already discussed this - accident or ?

You should use sudoers.d btw (but not as part of the ES module)

scriptmodules/emulators/mupen64plus.sh
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
 # This file is part of The RetroPie Project
-# 
+#
 # The RetroPie Project is the legal property of its developers, whose names are
 # too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
-# 
+#
```
joolswills: All the copyright headers contain this whitespace. We should clean all if changing it so search/replace is easier in future. I would leave it as is in this file for now.

scriptmodules/supplementary/emulationstation.sh
```diff
@@ -106,15 +107,20 @@ _EOF_
         iniSet "gpu_mem_1024" 256
         iniSet "overscan_scale" 1
     else
+        mkdir -p /usr/share/icons/retropie
+        cp "$scriptdir/scriptmodules/$md_type/$md_id/retropie.svg" "/usr/share/icons/retropie/"
         cat > /usr/share/applications/retropie.desktop << _EOF_
```
joolswills: I believe we should use /usr/local/share as this is not packaged stuff

scriptmodules/supplementary/emulationstation.sh
gizmo98: Oops. I will remove this.

scriptmodules/supplementary/emulationstation.sh
gizmo98: Yes. /usr/local/share is listed in $XDG_DATA_DIRS.
I will move desktop file and icon there.

scriptmodules/emulators/mupen64plus.sh
gizmo98: Ok. We could also move the isPlatform->setParam stuff out of the loop.

joolswills: yeh if they are not dependent on which plugin we can do that - thanks.

scriptmodules/supplementary/emulationstation.sh
joolswills: (and libsm-dev is installed with sdl1.2 etc, so it exists on the system I use to build binaries. A user installing just emulationstation on raspbian jessie lite noticed this). Actually, the cmakefiles for ES could be fixed for rpi so X is ignored - it's pulled in by one of the configs)

gizmo98:
Ok. Moved desktop file and icon to /usr/local/share. Tested modules emulationstation, mupen64plus and autostart again.

joolswills:
Thanks/

PR #1189, opened 2016-01-17
Platform flags

joolswills:
This allows more flexibility in modules - so instead of checking specifically for a platform you can check for something in common with a platform. in the same way isPlatform "rpi" matches rpi1 and rpi2, now you can do isPlatform "mali" to check if the current platform has a mali gpu.

the addtional flags are set in the system.sh

thi makes it easier to add new platforms (I have included a generic X11 and armv7+mali platform), without additional work in modules. Or for example you can do isPlatform "armv7" to decide whether to enable armv7 optimisations.

not for merging yet as I need to do some more testing.

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -139,7 +139,7 @@ function configure_advmame() {
         iniSet "device_keyboard" "raw"
         iniSet "device_sound" "alsa"
         iniSet "display_vsync" "no"
-        if isPlatform "rpi1"; then
+        if isPlatform "armv6"; then
             iniSet "sound_samplerate" "22050"
         else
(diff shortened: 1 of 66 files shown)


joolswills:
@gizmo98 please do have a look over this 

From now on we can also use isPlatform "x11" if building something for an X11 target so it could be possible to build on arm with X11 etc.

joolswills:
eg in scriptmodules/supplementary/emulationstation.sh we can do `isPlatform "x11" && getDepends gnome-terminal` instead of `isPlatform "x86" && getDepends gnome-terminal`

gizmo98:
Looks good. As i can see we can add other flags like gles easily as well if needed.

joolswills:
yeh. if there was an additional specific flag needed it can be added when needed.

PR #1190, opened 2016-01-17
handle quitting/restarting es from the launch script - relies on http…

joolswills:
…s://github.com/RetroPie/EmulationStation/issues/16


joolswills:
https://github.com/RetroPie/RetroPie-Setup/pull/1190

Issue #1181, opened 2016-01-10
Bluetooth daemon fails to start


PeterPablo:
I am running a raspbian jessie with a recently cloned retropie setup script. During boot my bluetooth daemon fails to start:

```
root@rpi2:~# journalctl -xn
-- Logs begin at Sa 2016-01-09 15:47:13 CET, end at So 2016-01-10 10:49:31 CET. --
Jan 10 10:49:24 rpi2 su[11048]: + /dev/pts/0 root:root
Jan 10 10:49:24 rpi2 su[11048]: pam_unix(su:session): session opened for user root by pi(uid=0)
Jan 10 10:49:31 rpi2 sudo[11074]: root : TTY=pts/0 ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/service bluetooth start
Jan 10 10:49:31 rpi2 sudo[11074]: pam_unix(sudo:session): session opened for user root by pi(uid=0)
Jan 10 10:49:31 rpi2 systemd[1]: Starting Bluetooth service...
-- Subject: Unit bluetooth.service has begun with start-up
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit bluetooth.service has begun starting up.
Jan 10 10:49:31 rpi2 systemd[11098]: Failed at step EXEC spawning /usr/lib/bluetooth/bluetoothd: Permission denied
-- Subject: Process /usr/lib/bluetooth/bluetoothd could not be executed
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- The process /usr/lib/bluetooth/bluetoothd could not be executed and failed.
-- 
-- The error number returned while executing this process is 13.
Jan 10 10:49:31 rpi2 systemd[1]: bluetooth.service: main process exited, code=exited, status=203/EXEC
Jan 10 10:49:31 rpi2 systemd[1]: Failed to start Bluetooth service.
-- Subject: Unit bluetooth.service has failed
-- Defined-By: systemd
-- Support: http://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- Unit bluetooth.service has failed.
-- 
-- The result is failed.
```

I think this is due to the execute flag being removed [here](https://github.com/RetroPie/sixad/blob/master/sixad#L67): `$SUDO chmod 644 "$BLUETOOTHD"`.

[...]

PR #1191, opened 2016-01-17
added some bluez5 / jessie workarounds for the bluetooth module

joolswills:
* include some test scripts from the bluez5 repository as they are not readily available on jessie - meaning we don't have to rewrite everything
 * use bluetoothctl for discovery of devices as hcitool scan lists devices, but doesn't make them available via bluez dbus api

Would be interested in anyone who can test this please on Raspbian Jessie. I was able to pair my keyboard with it at least.


PeterPablo:
Does this apply to pairing a PS3 controller over bluetooth as well? (#1181) Should I retest by applying this PR?

joolswills:
Not used for ps3 sorry.

PR #1203, opened 2016-01-23
lr-gambatte: add default green yellow gb classic palette

gizmo98:
http://blog.petrockblock.com/forums/topic/gameboy-overlay-settings-psd/


gizmo98:
Ok. Should be sorted out now.

joolswills:
thanks.

PR #1221, opened 2016-01-31
Add Arcade System

HerbFargus:
Still not completely sure of some of the configs (i.e. advmame, mame4all)

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -120,6 +120,7 @@ function install_advmame() {
 }
 
 function configure_advmame() {
+    mkRomDir "arcade"
     mkRomDir "mame-advmame"
 
     # delete old install files
@@ -152,6 +153,7 @@ function configure_advmame() {
 
         default=0
         [[ "$version" == "0.94.0" ]] && default=1
+        addSystem 0 "$md_id-$version" "arcade" "$md_inst/$version/bin/advmame %BASENAME%"
         addSystem $default "$md_id-$version" "mame-advmame arcade mame" "$md_inst/$version/bin/advmame %BASENAME%"
     done
 }
--- a/scriptmodules/emulators/pifba.sh
+++ b/scriptmodules/emulators/pifba.sh
@@ -43,6 +43,7 @@ function install_pifba() {
 }
 
 function configure_pifba() {
+    mkRomDir "arcade"
     mkRomDir "fba"
     mkRomDir "neogeo"
 
@@ -62,6 +63,7 @@ function configure_pifba() {
 
     local def=0
     isPlatform "rpi1" && def=1
+    addSystem 0 "$md_id" "arcade" "$md_inst/fba2x %ROM%"
     addSystem $def "$md_id" "neogeo" "$md_inst/fba2x %ROM%"
     addSystem $def "$md_id" "fba arcade" "$md_inst/fba2x %ROM%"
 }
--- a/scriptmodules/libretrocores/lr-fba-next.sh
+++ b/scriptmodules/libretrocores/lr-fba-next.sh
@@ -45,13 +45,16 @@ function install_lr-fba-next() {
 }
 
 function configure_lr-fba-next() {
+    mkRomDir "arcade"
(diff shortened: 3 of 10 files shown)


scriptmodules/emulators/pifba.sh
```diff
@@ -62,6 +63,7 @@ function configure_pifba() {
 
     local def=0
     isPlatform "rpi1" && def=1
+    addSystem $def "$md_id" "arcade" "$md_inst/fba2x %ROM%"
```
joolswills: instead of $def - 0 would be better as we don't want either of the fba emulators as default for arcade. I think no default should be set - the user will be asked then on first run by runcommand.cfg

scriptmodules/libretrocores/lr-fba-next.sh
```diff
@@ -45,13 +45,16 @@ function install_lr-fba-next() {
...
+    mkRomDir "arcade"
     mkRomDir "fba"
     mkRomDir "neogeo"
+    ensureSystemretroconfig "arcade"
     ensureSystemretroconfig "fba"
     ensureSystemretroconfig "neogeo"
 
     local def=1
     isPlatform "armv6" && def=0
+    addSystem $def "$md_id" "arcade" "$md_inst/fba_libretro.so"
```
joolswills: as above

joolswills:
apart from my comment above it looks good - have you tried it ?

If so please fix the above / squash and I will merge. Cheers.

HerbFargus:
Would I also want to create a new platforms.cfg for arcade?

joolswills:
it would need that too I think yeh - this needs testing before we merge (you will notice any issues then).

HerbFargus:
Of course. Just making sure I have all my basis covered first. I'll make the changes and give it a test.

joolswills:
k. sounds good. thanks.

scriptmodules/emulators/advmame.sh
```diff
@@ -152,6 +153,7 @@ function configure_advmame() {
 
         default=0
         [[ "$version" == "0.94.0" ]] && default=1
+        addSystem $default "$md_id-$version" "arcade" "$md_inst/$version/bin/advmame %BASENAME%"
```
HerbFargus: How come it is "mame-advmame arcade mame" and not just "mame-advmame" ?

scriptmodules/emulators/advmame.sh
```diff
@@ -152,6 +153,7 @@ function configure_advmame() {
...
         addSystem $default "$md_id-$version" "mame-advmame arcade mame" "$md_inst/$version/bin/advmame %BASENAME%"
```
HerbFargus: How come it is "mame-advmame arcade mame" and not just "mame-advmame" ?

joolswills: The three keywords are assigned like this:

mame-advmame - system (used for romdir and <name> in es_systems.cfg)
arcade - platform (used in es_systems.cfg as <platform>)
mame - theme (used in es_systems.cfg as <theme)

In your arcade case, assuming you are going to make a new arcade theme part, the single parameter is fine.

HerbFargus: Do I also want to set the $default to 0 as well here like I did for fba?

joolswills: yep.

HerbFargus:
Tested them and it seems to work, still need to squash the commits. the last question I have pertains to  samples, artwork etc.  mame-advmame and mame-mame4all have those folders. do we need to symlink them somehow or should we just expect users to put them in their respective folders?

joolswills:
yeh those will have to stay where they are. not a major issue - just needs to be mentioned on the wiki or something.

HerbFargus:
Ok, sounds good. I'll make the last changes and squash the commits.

HerbFargus:
Should be good to merge.

joolswills:
Thanks herb.

HerbFargus:
Yep, Rookervik is working on the accompanying theme right now.

joolswills:
nice!

PR #1222, opened 2016-01-31
Add CoolCV

HerbFargus:
Of course we'd first need to add the binary to the archive url

Binary is here (6.2):
http://atariage.com/forums/topic/240800-coolcv-emulator-for-mac-os-x-linux-windows-and-raspberry/page-1

I'm also not super sure on the bios for it and if I'd need to symlink
that to the bios folder.

Don't know if we want to wait until nanochess releases 6.3 for the
Pi with a potential configurable keyboard. Having the source code would make it a bit easier.

### DIFF
--- a/scriptmodules/emulators/coolcv.sh
+++ b/scriptmodules/emulators/coolcv.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="coolcv"
+rp_module_desc="CoolCV Colecovision Emulator"
+rp_module_menus="4+"
+
+function install_coolcv() {    
+    wget -O- -q $__archive_url/coolcv.tar.gz | tar -xvz -C "$md_inst"
+}
+
+function configure_coolcv() {
+    mkRomDir "coleco"
+
+    addSystem 1 "$md_id" "coleco colecovision colecovision" "$md_inst/coolcv/coolcv_pi %ROM%"
+}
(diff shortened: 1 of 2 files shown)


scriptmodules/emulators/coolcv.sh
```diff
@@ -0,0 +1,29 @@
...
+function install_coolcv() {    
+    wget $__archive_url/coolcv.zip -O "$md_inst/coolcv.zip"
+    unzip -n "$md_inst/coolcv.zip" -d "$md_inst"
+    rm "$md_inst/coolcv.zip"
+}
+
+function configure_coolcv() {
+    mkRomDir "colecovision"
+
+    chown -R $user:$user "$md_inst/coolcv_pi"
```
joolswills: -R is for recursion so not needed here - I shouldn't think the executable needs to be owned by $user though anyway

scriptmodules/emulators/coolcv.sh
```diff
@@ -0,0 +1,29 @@
...
+    chmod +x "$md_inst/coolcv_pi"
```
joolswills: is the file not executable in the zip ? if not perhaps we should repack as a tar.gz

scriptmodules/emulators/coolcv.sh
```diff
@@ -0,0 +1,29 @@
...
+
+    addSystem 0 "$md_id" "colecovision" "$md_inst/coolcv_pi %ROM%"
```
joolswills: 1 to make it the default in this case, as there are no other colecovision emulators.

joolswills:
Thanks

HerbFargus:
I have no problem with a tar.gz (the original zip is packaged with binariesfor all the systems it supports)  So all we would need is coolcv_pi and the readme. If you are able to put the binary up with proper permissions I can fix the script and test it.

joolswills:
Will do.

joolswills:
coolcv.tar.gz is uploaded.

should be just this on-liner to install it now.
`
wget -O- -q $__archive_url/coolcv.tar.gz | tar -xvz -C "$md_inst"
`

HerbFargus:
Should be ready to merge, tested and seems to work fine with a keyboard ( Feel free to test and verify) . Joypad configs currently will have to be done manually and may be easier with future releases.

joolswills:
Looks good. Thanks.

PR #1236, opened 2016-02-04
Clean launch

joolswills:
This has been requested on a few occasions - making the launching less verbose and slightly prettier.
changes are:

Dialog when launching by default.

stdout/stderr logged to /tmp/runcommand.log unless a console app is used like frotz

can switch on use_art for runcommand to show ES artwork before launch - configurable from the runcommand config in retropie-setup

### DIFF
--- a/scriptmodules/emulators/frotz.sh
+++ b/scriptmodules/emulators/frotz.sh
@@ -33,5 +33,6 @@ function configure_frotz() {
 
     chown -R $user:$user "$romdir/zmachine/"*
 
-    addSystem 1 "$md_id" "zmachine" "pushd $romdir/zmachine; frotz %ROM%; popd" "Z-machine" ".dat .z5"
+    # CON: to stop runcommand from redirecting stdout to log
+    addSystem 1 "$md_id" "zmachine" "CON:pushd $romdir/zmachine; frotz %ROM%; popd" "Z-machine" ".dat .z5"
 }
--- a/scriptmodules/supplementary/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand.sh
@@ -23,6 +23,14 @@ function install_runcommand() {
     cp "$scriptdir/scriptmodules/$md_type/$md_id/joy2key.py" "$md_inst/"
     chmod a+x "$md_inst/runcommand.sh"
     chmod a+x "$md_inst/joy2key.py"
+    if [[ ! -f "$configdir/all/runcommand.cfg" ]]; then
+        mkUserDir "$configdir/all"
+        iniConfig "=" '"' "$configdir/all/runcommand.cfg"
+        iniSet "use_art" "0"
+        iniSet "disable_joystick" "0"
+        iniSet "governor" ""
+        chown $user:$user "$configdir/all/runcommand.cfg"
+    fi
 }
 
 function governor_runcommand() {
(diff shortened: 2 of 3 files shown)


joolswills:
Please give this a try and let me know what you think.

joolswills:
Can also disable joystick control now - as some people were asking how to disable it as they were triggering the configuration menu by accident.

HerbFargus:
Brilliant :D I am testing it right now and it is quite lovely. the blue flash of a screen is very fast (almost epileptic...)

HerbFargus:
I also think it killed my keyboard input (perhaps when I pressed a button when the box art was up)

HerbFargus:
I learned a new thing: Dialog colours can be customised, go figure

```
sudo dialog --create-rc etc/dialogrc
```

Set:

```
# Turn color support ON or OFF
use_colors = ON
```

and 

```
# Screen color
screen_color = (CYAN,BLACK,ON)
```

joolswills:
would you prefer a black background ?

Also should we increase the "wait for keypress" to 2 seconds perhaps ?

HerbFargus:
I would definitely prefer a black background. (could probably change some other colours around as well if we were feeling ambitious) and yeah 2 seconds I think might be good. the dialog takes a bit longer to load than text so you can barely read it before the game loads.

HerbFargus:
Yeah and if you exit out of the emulator before fbi has its time to go through, when you go back into the terminal with f4 the keyboard becomes unusable.

joolswills:
i'll leave background for now, as making a custom config for one dialog seems overkill - ill have a think (no parameter for it as far as I can see)

joolswills:
which emulator ? and exiting emulationtation right away or ? BTW I added more commits since the first, did you check with the latest (delay rather than killing fbi etc)

joolswills:
can't reproduce that here currently.

joolswills:
aah it just needs a terminal reset ? try typing "reset" into the terminal (blindly)

HerbFargus:
I only tested with lr-mame2003, and I loaded it while you were making changes so I may have missed some, I'll try updating it and see if it has the same problems.

HerbFargus:
Yes typing reset fixed it.

joolswills:
we can always make sure we wait for fbi when exiting the ES launch script, and do a reset perhaps. I've made a couple of changes that may help, but the reset may need to be in ES script to run after fbi.

HerbFargus:
doesnt seem to be as bad with the keyboard now, and the delay is much better.

joolswills:
added the ability to disable the launcher menu completely (and the launch dialog). images will still show if enabled.

joolswills:
I think im happy with it for now as far as this pull request goes.

HerbFargus:
I was just going to suggest that. always 3 steps ahead of me. All looks great. We can look at aesthetics later with dialog (easy enough to change manually if needs be - or we can add as another menu option later)

HerbFargus:
Minor nuance - if images are still on and runcommand menu is disabled, when you press a button it kills the image and shows the filepath to the picture instead.

joolswills:
thats because fbi captures the button press I guess - I guess if disabled the joystick should be disabled too. a keyboard will still interfere though probably.

joolswills:
I changed it so it shows the rom name to be launched with the emulator in brackets.

HerbFargus:
yeah its only any issue with the keyboard, no issue with joystick really (though yeah either should be documented in the wiki or just automatically disabled when the launch menu is disabled)

HerbFargus:
and the rom name and emulator is great.  works brilliantly.

joolswills:
I shall merge then :) thanks for testing.

HerbFargus:
dont know if it matters, but the latest change i don't think changes the menu- so it may set joypad to off in the settings but the menu still shows as on

HerbFargus:
nvm I think its working now.

HerbFargus:
So just to clarify when it says to turn off pre launch menu and it says Currently ON that means that the menu is off yes?

joolswills:
no that means it is on, and you choose the option to turn it off.

joolswills:
I will reword it.

HerbFargus:
unless it was just put in backwards. when the turn off pre launch menu is currently on the joystick  control cannot be turned on. so as long as thats how its intended then it works.

joolswills:
Disable pre launch menu (currently enabled) ? well same problem depending how you read it, but it's disable not disabled.

joolswills:
not sure how you mean?

HerbFargus:
to say off is on is whats confusing me. just ignore my ignorance. maybe if its set to true of false. I don't know the best way to word it.

joolswills:
the options do what they say in the text when selected - the bit in brackets is how the option is currently.

HerbFargus:
I'm sure its fine, I'll just make sure to document it proper. It looks great :) Good work

joolswills:
We can rework it later - I see your point - I actually tried a few different wordings and they all had ways of misreading it depending on whether you read it as a status, or an option to change.

joolswills:
the fbi printing text should be sorted - I accidentally enabled edit mode (misread docs)

joolswills:
(fbi will quit though if a key is pressed)

HerbFargus:
minor details ;)

HerbFargus:
For when we revisit: 

Menu Suggestion (In Order):

- Pre-Launch Menu (ENABLED/DISABLED)
- Pre-Launch Menu Boxart (ENABLED/DISABLED)
- Pre-Launch Menu Joystick Control (ENABLED/DISABLED)
- Pre-Launch Menu CPU Configurations

PR #1239, opened 2016-02-05
Add Direct Launch to Helper Script (set essystem function)

HerbFargus:
Preliminary support for direct launch through emulationstation as a result of https://github.com/RetroPie/EmulationStation/pull/1

For example in implementation with kodi: The direct launch path and system name are really only needed and then you indicate a 'true' at the end to enable directlaunch :

```
setESSystem 'Kodi' 'kodi' '' '' '/opt/retropie/supplementary/runcommand/runcommand.sh 0 "kodi-standalone" "kodi"' '' '' 'true'
```

Kodi seems to crash on exit when using directlaunch which may be a separate bug or some conflict with runcommand which may need to be further investigated. I have not tested with other systems and am unsure if it is specific to kodi or if it will happen with all systems.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -301,6 +301,7 @@ function setESSystem() {
     local command=$5
     local platform=$6
     local theme=$7
+    local directlaunch=$8
 
     local conf="/etc/emulationstation/es_systems.cfg"
     mkdir -p "/etc/emulationstation"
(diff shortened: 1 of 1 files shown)


HerbFargus:
Ah shoot missed a minor detail. rebasing hold on.

HerbFargus:
Ok I think it should all be good now

joolswills:
I think you are missing some code there - lost in rebase ?

HerbFargus:
Ah right you are. (this is why I do pull requests) fixing.

HerbFargus:
I think thats all of it?

joolswills:
thanks. merged

PR #1245, opened 2016-02-07
Add support for scraping Arcade system as MAME for sselph scraper.

zerojay:
Now that Retropie supports having a merged arcade folder in Retropie 3.5, add support for scraping the arcade folder as MAME.


joolswills:
thanks.

PR #1247, opened 2016-02-07
Add Beetle PSX emulator

HerbFargus:
Rubbish performance even on the pi2, mostly for x86 really.  Has the
PlayStation Splashscreen, requires scph5500.bin scph5501.bin
schp5502.bin. see here for more details
http://wiki.libretro.com/index.php?title=Beetle/Mednafen_PSX same issues
with bios and save directory as other mednafen systems.

### DIFF
--- a/scriptmodules/libretrocores/lr-beetle-psx.sh
+++ b/scriptmodules/libretrocores/lr-beetle-psx.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
probably best to exclude it from the rpi then with !rpi in module flags or else I foresee complaints regarding performance :)

joolswills:
actually !arm would be better which would exclude all arm platforms for now - I doubt any are quick enough currently (maybe some upcoming boards will be). depending how slow it is.

HerbFargus:
Will do. I'll change the 3do and sharp while I'm at it.

HerbFargus:
should be sorted now

PR #1268, opened 2016-02-14
mame2003 core option defaults

objectchris:
skip_disclaimer skips the nag screen
dcs-speedhack speeds up mk2/3


joolswills:
Thanks.

btw you could have force pushed back to your other branch - which would have saved opening a new pull request etc.

Issue #684, opened 2015-02-23
Duke Nukem 3D Mouse problems


PowerWiesel:
I can’t get my mouse running for aiming and shooting etc. For example it’s working wiith Quake out of the box.

In the menu (megaton edition) you can move the curser, but you can’t click anything.

Ingame there is also a moving courser, but you can’t shoot or aim.

I tried in_mouse 1 and MouseAiming = 1 and several little things that dind’t work

joolswills:
Thanks - Yeh I can confirm this.

agenttank:
same here, but i don't really care ;)
fresh install of retropie 2.6 full image

luxiouronimo:
i can confirm and add:

raspberry pi 2, installed retropie-v3.0beta2-rpi2.img, and install from yesterday's (5/19/15) binaries..

in the in-game menu the mouse works and i can click menu items, but in-game the mouse does not affect the aim/view or movement, even when toggling mouse move with the 'u' key.

clicking mouse1 does however shoot.

joolswills:
It's a known issue as above.

luxiouronimo:
to be clear, relating to v3.beta2-rpi2, you can click the in-game menu and that works, and in-game you can shoot.

joolswills:
yep - this is the issue. the mouse works in the menus, but not for aiming.

davidj1987:
Is Duke 3D eduke or a retroarch port? Is this issue related to this issue: https://github.com/libretro/Genesis-Plus-GX/issues/31

In this emulator mouse buttons work but the cursor doesn't move. :/

joolswills:
unrelated issue.

PowerWiesel:
Thank you very much:)

davidj1987:
I'm having mouse issues with eduke...is there something I need to enable?

HerbFargus:
This issue is closed. Please see the wiki (https://github.com/retropie/retropie-setup/wiki/Duke-Nukem-3D)  and use the forum for support

PR #1274, opened 2016-02-17
Add mouse aim support to DukeNukem3D

HerbFargus:
By Switching to SDL2

You still need to press "U" on the keyboard to toggle it on first.

Will fix this issue: https://github.com/RetroPie/RetroPie-Setup/issues/684

### DIFF
--- a/scriptmodules/ports/eduke32.sh
+++ b/scriptmodules/ports/eduke32.sh
@@ -15,18 +15,19 @@ rp_module_menus="2+"
 rp_module_flags="dispmanx !mali"
 
 function depends_eduke32() {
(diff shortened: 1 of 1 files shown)


joolswills:
Wheezy doesn't have sdl2-mixer so it will need to use sdl1.2 for wheezy and 2 for jessie.

joolswills:
nasm is only needed for x86 target also.

joolswills:
I committed this with some changes to dependencies so sdl1 is still used on wheezy, and so that the parameters are changed depending on target etc. 335ed5fcf26cbd3c9fa851dcd3887f4019b7436c

joolswills:
The other problem which I remember seeing before when I tested with SDL some time ago, is that the mouse doesn't move very quickly - I think this is some SDL2 issue on the rpi as I've seen it with other software.

joolswills:
It's usable after sensitively tweaks anyway.

PR #1278, opened 2016-02-20
wait for 5 frames when hotkey enable is pressed before blocking - wor…

joolswills:
…ks around some issues with sharing a hotkey with a button used in a game. based on https://github.com/recalbox/recalbox-buildroot/pull/72


HerbFargus:
Looks really good. I'll give it a test

HerbFargus:
Works great on my end. tested with cadillacs and dinosaurs on lr-fba

HerbFargus:
Also tested on super mario world just to make sure that the hotkey blocking still carried over for the second keypress and it does, so very well done.

joolswills:
thanks -  well kudos to the original author - it's a lot simpler than my attempt at a workaround, and seems to work around our problem cases.

Issue #1271, opened 2016-02-16
Directlaunch Runcommand Crash


HerbFargus:
Before we actively pursue setting certain systems (e.g. Kodi see https://github.com/RetroPie/RetroPie-Setup/pull/904)  as directlaunch from emulationstation we need to

- create a function for directlaunch
- address crashing on exit when implemented with runcommand

From testing kodi works fine when launched directly with kodi-standalone but if runcommand is used it crashes.

HerbFargus:
Just for reference this is my kodi crash log:

http://pastebin.com/KHT2eTXZ

joolswills:
I just tested and it seems fine here

launching with

```
  <system>
    <name>kodi</name>
    <fullname>Kodi</fullname>
    <command>/opt/retropie/supplementary/runcommand/runcommand.sh 0 kodi-standalone</command>
    <theme>kodi</theme>
    <directlaunch>true</directlaunch>
  </system>
```

joolswills:
```
  <system>
    <name>kodi</name>
    <fullname>Kodi</fullname>
    <command>/opt/retropie/supplementary/runcommand/runcommand.sh 0 _SYS_ kodi</command>
    <theme>kodi</theme>
    <directlaunch>true</directlaunch>
  </system>
```
works too (and gives the ability to the user to control the commandline etc)

HerbFargus:
No crash on exit? May have been addressed with your last modification to fb_restore. If you're happy I'm happy.

joolswills:
must have been that. or something else. I will switch kodi to direct launch in a PR and we can test.

HerbFargus:
Brilliant :)

HerbFargus:
Seems resolved now.

HerbFargus:
Upon further testing it seems there is an issue with directlaunch in that once implemented it will crash emulationstation when switching themes through the start menu in emulationstation.

To reproduce:

- Install updated kodi from the retropie setup script
- install a different theme than the default
- switch to the new theme - Will crash.

joolswills:
This should be opened against ES then.

HerbFargus:
opened here: https://github.com/RetroPie/EmulationStation/issues/37

PR #1279, opened 2016-02-21
runcommand.sh: fix wrong fb_res restore

gizmo98:
If overscan is enabled `fb_res_h = display_res_h - overscan_r
-overscan_l` and `fb_res_v = display_res_v - overscan_t -overscan_b`.
It is false to use display resolution. Just store original fb
resolution and use this value for restore_fb().

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -210,6 +210,12 @@ function load_mode_defaults() {
         fi
     fi
 
+    # get default fb_res
+    fb_orig=""
+    fb_orig=$(fbset)
+    fb_orig=${fb_orig##*mode \"}
+    fb_orig=${fb_orig%%\"*}
+
     mode_def_emu=""
     mode_def_rom=""
     fb_def_emu=""
@@ -557,7 +563,7 @@ function restore_mode() {
 
 function restore_fb() {
     sleep 1
-    switch_fb_res "${mode_orig[2]}x${mode_orig[3]}"
+    switch_fb_res "$fb_orig"
 }
 
 function config_dispmanx() {


joolswills:
Thanks.

gizmo98:
@joolswills 
I would expect fb_res must be always a fraction of the current display resolution if overscan is enabled. So if for example `fb_res = 3/4 * disp_res` we should not set `fb_res=disp_res` if we change resolution. Things get rendered outside of the viewable area (I remember we had such issues in the past). I think we should calculate the ratio between fb_res and disp_res and recalculate fb_res in line 549.
https://github.com/RetroPie/RetroPie-Setup/blob/eea1e7087ccd37bc064d094329293acf726a5d55/scriptmodules/supplementary/runcommand/runcommand.sh#L549

Not tested yet.

joolswills:
You may be right - overscan is not taken into consideration, and may need to be different from the config.txt overscan if switching modes anyway - but most of the time framebuffer res doesn't need changing (is useful for X stuff afair, and perhaps frotz on high res screens).

joolswills:
Other stuff should not be affected by the framebuffer change if it renders via sdl or dispmanx (just that it needs changing for terminal text to be readable etc)

joolswills:
https://github.com/RetroPie/RetroPie-Setup/commit/0810f984acfc830325da24a437b248f24ad1891f

gizmo98:
It could be graphics_get_display_size() returns fb resolution instead of display resolution. I think i have seen retroarch showing deviant resolution if overscan is enabled. Retroarch, mupen64plus and SDL2 use this function. So if fb_res defines the viewable area we should set it right. But not tested yet. Just an expectation.

joolswills:
the framebuffer res is not allowed to be configured for retroarch so I don't think anyone would have done that manually. SDL1 and SDL2 (using eduke32 as a test) are unaffected too so I don't think this is the case.

joolswills:
(sorry correction, the framebuffer will be switched if video mode changing - so there is a possibility there is some odd case here with resolution if using overscan, so that needs checking perhaps)

joolswills:
(but then our current behaviour should be correct - as X/Y would match the res)

joolswills:
/* If the console framebuffer has active overscan settings, 
    * the user must have overscan_scale=1 in config.txt to have 
    * the same size for both fb console and dispmanx. */

Although I believe rpi firmware author mention overscan_scale=1 should be avoided. We use it though. Possible that a mode switch would lose overscan settings then, but mode switching is not required for retroarch in most cases.

gizmo98:
Yeah. We don't need mode switching in the most cases anymore. But there were problems in the past with overscan and mode switching. I will take a look if these problems are still there or if there is some small room for improvement.

joolswills:
thanks.

joolswills:
there is one more improvement regarding restoring framebuffer btw - in the case the user had a different depth before - we currently restore to 16 bpp. This was a problem with the framebuffer changing on odroid (because at that time the framebuffer was always reset). we don't always reset it now - that was also causing a problem with the dispmanx driver in retroarch. we could change the function to perhaps not only take the X/Y but also depth or so.

joolswills:
We have the possible problem since I made this change that if we don't reset it and something leaves it in a bad state, it will stay like that - not sure what to do without some additional logic, but I believe most of the apps do leave the framebuffer in a usable state these days. If we had a case, we would need to change the logic again to - reset if mode change, or fb change or not retroarch, and not odroid or something, but unless we get some case where it is needed I'm happy to leave it as is.

HerbFargus:
Is it possible this is related to this discussion?
https://github.com/RetroPie/RetroPie-Setup/issues/1271

objectchris:
this change also might have an effect on https://github.com/libretro/RetroArch/issues/2562 - i noticed when removing the restore_fb call it stopped dispmanx causing a display crash when RGUI was invoked. never figured out why.

joolswills:
@HerbFargus no, I don't think so since the framebuffer is not reset by default on exit (unless the user has switched screenmode). Unless not resetting it is an issue, but I don't think so.

@objectchris as above

from above:
> we don't always reset it now - that was also causing a problem with the dispmanx driver in retroarch.

I was referring to that ticket.

joolswills:
(this behaviour was changed in db09ea97500f8ca123a66933d934e49e4db3b65c)

objectchris:
@joolswills that fixed my issue, awesome :)

PR #1285, opened 2016-02-22
Uninstall

joolswills:
some uninstall functionality

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -294,6 +294,26 @@ function individual_setup()
     done
 }
 
+function uninstall_setup()
+{
+    dialog --yesno "Are you sure you want to uninstall RetroPie?" 22 76 2>&1 >/dev/tty || return 0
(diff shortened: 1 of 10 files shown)


HerbFargus:
Works great. As the error text in the script mentions symlinks still have to be removed manually but it would seem everything else was uninstalled properly. 

list of symlinks left over:

.advance
.atari800.cfg
.dosbox
.eduke32
.fuserc
.gngeo
.hatari
.reicast
.scummvm
.scummvmrc
.vice

and I think kodi wasn't uninstalled either

joolswills:
Thanks . The symlinks need to be dealt with in each module and I will do that later on. Kodi will need a remove function (as will some other experimental modules)

HerbFargus:
not sure if we want to remove raspi-config though as its part of raspbian

joolswills:
I don't ?

joolswills:
ok it was removing alsa-utils which is a raspi-config dependency. I have changed this now.

PR #1292, opened 2016-02-26
Set sselph scraper to use the OpenVGDB

zerojay:
Set the sselph scraper to scrape from OpenVGDB until thegamesdb is back up and running again.


joolswills:
it is up and running currently.

joolswills:
Better would be to add it as an option in the menu etc so users can switch.

zerojay:
Agreed. I'll throw something in for it soonish.

PR #1294, opened 2016-02-27
Allow users to select the scraper to use with sselph's scraper.

zerojay:
Add the option to allow users to select which scraper the sselph scraper should scrape from.
Updated menu items to make more sense.


joolswills:
Thanks :)

PR #1298, opened 2016-02-27
Added libretro core for latest MESS.

zerojay:
Added libretro core for latest MESS pulled from git, currently 0.171.
Some additional symlinks or folders may need to be created for systems to work properly.
Documentation on how to set up MESS using this core will be coming in the future to the RetroPie wiki and has yet to be written.
Systems that have been set up for use with this MESS core are only those found to work approximately 100% speed. These include the following: Nintendo Gameboy (uses "gameboy" directory instead of "gb"), Nintendo Entertainment System and Colecovision with more added in the future as they get tested.

Systems that have been tested and that are not up to speed and therefore disabled at the moment include the following (as tested by a Raspberry Pi 2 overclocked to 1067mhz):

Amstrad GX4000 (60% speed)
Atari 2600 (70-85% speed)
Atari 5200 (50-60% speed)
Nintendo Gameboy Advance (50% speed)
Sega Game Gear (80% speed)
Sega Genesis (50% speed)
Sharp X68000 (14% speed)
SNK Neo Geo AES (60-65% speed)

Compilation of the core takes approximately 3 hours as I did not include the number of jobs to compile at once. Let me know if you require information for testing.

### DIFF
--- a/scriptmodules/libretrocores/lr-mess.sh
+++ b/scriptmodules/libretrocores/lr-mess.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-mess"
+rp_module_desc="MESS emulator - MESS Port for libretro"
+rp_module_menus="4+"
+
+function sources_lr-mess() {
+    gitPullOrClone "$md_build" https://github.com/libretro/MAME.git
+}
+
+function build_lr-mess() {
+    make -f Makefile.libretro clean
+    make -f Makefile.libretro SUBTARGET=mess
+    md_ret_require="$md_build/mess_libretro.so"
+}
+
+function install_lr-mess() {
+    md_ret_files=(
+        'mess_libretro.so'
+    )
+}
+
+function configure_lr-mess() {
+    mkRomDir "nes"
+    mkRomDir "gameboy"
+    mkRomDir "coleco"
+    ensureSystemretroconfig "nes"
+    ensureSystemretroconfig "gameboy"
+    ensureSystemretroconfig "coleco"
+
+    setRetroArchCoreOption "mame_softlists_enable" "enabled"
+    setRetroArchCoreOption "mame_softlists_auto_media" "enabled"
+    setRetroArchCoreOption "mame_boot_from_cli" "enabled"
+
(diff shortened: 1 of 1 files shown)


joolswills:
Looks good (certainly enough for experimental). thanks.

BenWlson:
Sorry... I'm still figuring out how to use github. I found your full description in the closed pull request section.

Darknior:
Really an excellent news 😃 
@BenWlson is real and MESS can help a lot to add many old cool systems to Libretro !
To have all the shaders, savestate ... and remove to Retropie the other emulators 👍 

The problem was MESS is not a fast emulator ?
Not a problem @HCF found a solution to really speed up this emulator on xBox 1 with a little trick and make CDI working at full speed !!! 😆 

http://www.emuxtras.net/forum/viewtopic.php?f=189&t=5730&p=73415&sid=a0feb61fc00ba8fb8e1701d97be4a47b#p73415

> This is not the kind of optimization that the MESS team is going to include in its code, so I will not commit it to their repository. MESS targets for accuracy, and this trick targets for speed. Although our trick doesn't "break" anything, it is obviously making the emulation less real, because we are removing some work that the real machine actually DOES. The magic of this thing, is that it is a work that the real CDi does... but we cannot see it in the gameplay under certain circunstancies.
So, as an Xbox user, I thought that all this not-noticeable work that the CDi does, should be removed to boost the emulation ;) 
Anyways, this trick could certainly be added in other ports of MESS (not in the official branch)... If the Raspberry PI has a port of MESS, of course this trick could be added. 

That's cool for us 👍 
Write him; it is a really cool cool cool guy, ask for his help !

> First of all, to avoid misunderstandtings: the optimization that we have done is valid in MESS, so it can be applied to many of the systems that MESS emulates. It is not a trick that could be applied in other emulators (I'm sorry if someone understood this from my words). As Madmab said, I discovered that MESS is doing some redundant work that, in some systems, is wasting a lot of CPU time. Removing this work has a great improvement in performance! The bad point is that the trick must be applied once per each system that we want to optimize, so it is a bit tedious and we are only going to do it in those systems that really need a boost (in systems that already work well, it is not needed).
Although the optimization is quite simple to do, the performance boost is really huge: CDi ran at 30% of its real speed in my stock Xbox, and now it runs at full speed :D

xBox has a 700Mhz proc and 64 Mo of RAM only !

>On stock xbox. CDI runs at 100% with frameskip of 10 out of 10. According to HCF the frameskipping in MESS is not like your normal emulators.
> Yes, the concept of frameskip is different in MESS than in other emulators. Frameskip 6 in MESS is like frameskip 1 in a normal emulator (it means "skipping 6 frames per each 12 frames"), so we can say that the Trusty runs at full speed with almost no frameskip :D
Frameskip 9 in MESS is equivalent to frameskip 3 in a normal emulator, so we can say that the stock Xbox plays at full speed with a tolerable frameskip. As you can see in the video, it is absolutely playable and enjoyable... if you can enjoy CDi games!  :lol:

I wish i have help 👍

PR #1302, opened 2016-02-28
Fix MESS BIOS hashes incorrect placement.

zerojay:
The hashes were being placed into $biosdir/mame instead of $biosdir/mame/hash.

### DIFF
--- a/scriptmodules/libretrocores/lr-mess.sh
+++ b/scriptmodules/libretrocores/lr-mess.sh
@@ -42,7 +42,7 @@ function configure_lr-mess() {
     setRetroArchCoreOption "mame_boot_from_cli" "enabled"
 
     mkdir "$biosdir/mame/hash"
-    cp -rv "$md_build/hash" "$biosdir/mame"
+    cp -rv "$md_build/hash/*" "$biosdir/mame/hash/"
     chown -R $user:$user "$biosdir/mame"
     addSystem 0 "$md_id" "nes" "$md_inst/mess_libretro.so"
     addSystem 0 "$md_id" "gameboy" "$md_inst/mess_libretro.so"


joolswills:
This doesn't look right (the code before looks correct). The wildcard is inside the quotes - so i don't think that will work.

joolswills:
the previous code without the mkdir would seem to be correct. But I will check.

joolswills:
also the "$biosdir/mame" folder needs creating first - or else you need to use mkdir -p

joolswills:
```
diff --git a/scriptmodules/libretrocores/lr-mess.sh b/scriptmodules/libretrocores/lr-mess.sh
index a50d946..877fe02 100644
--- a/scriptmodules/libretrocores/lr-mess.sh
+++ b/scriptmodules/libretrocores/lr-mess.sh
@@ -41,8 +41,8 @@ function configure_lr-mess() {
     setRetroArchCoreOption "mame_softlists_auto_media" "enabled"
     setRetroArchCoreOption "mame_boot_from_cli" "enabled"
 
-    mkdir "$biosdir/mame/hash"
-    cp -rv "$md_build/hash/*" "$biosdir/mame/hash/"
+    mkdir "$biosdir/mame"
+    cp -rv "$md_build/hash" "$biosdir/mame"
     chown -R $user:$user "$biosdir/mame"
     addSystem 0 "$md_id" "nes" "$md_inst/mess_libretro.so"
     addSystem 0 "$md_id" "gameboy" "$md_inst/mess_libretro.so"
```

joolswills:
(added missing /)

zerojay:
Got a way to quickly debug these scripts so I don't have to wait 3 hours? :)

joolswills:
use a pc and run steps independently.
```
sudo ./retropie_packages.sh lr-mess sources
sudo ./retropie_packages.sh lr-mess build
sudo ./retropie_packages.sh lr-mess install
sudo ./retropie_packages.sh lr-mess configure
```

etc

joolswills:
I have committed me above changes.

PR #1307, opened 2016-02-29
Rpi3

joolswills:
Added a neon flag so we don't have to do isPlatform "armv7" for neon optimisations

RPI3 will id as rpi2 currently, which is fine - I expect it to use the rpi2 binaries. rpi3 runs in 32bit mode on Raspbian. if retropie-setup is run on aarch64 people can try __platform=rpi3 to build from source

this is all untested currently.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -17,33 +17,7 @@ function setup_env() {
     __memory_phys=$(free -m | awk '/^Mem:/{print $2}')
     __memory_total=$(free -m -t | awk '/^Total:/{print $2}')
 
-    if [[ -z "$__platform" ]]; then
-        case $(sed -n '/^Hardware/s/^.*: \(.*\)/\1/p' < /proc/cpuinfo) in
-            BCM2708)
-                __platform="rpi1"
-                ;;
-            BCM2709)
-                __platform="rpi2"
-                ;;
-            ODROIDC)
-                __platform="odroid-c1"
-                ;;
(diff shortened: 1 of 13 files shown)


scriptmodules/system.sh
```diff
@@ -181,18 +194,30 @@ function platform_rpi2() {
...
     __qemu_cpu=cortex-a15
     __has_binaries=1
 }
 
+function platform_rpi3() {
+    platform_rpi2
+}
+
+function platform_rpi3-64() {
+    __default_cflags="-O2 -mcpu=cortex-a53 -mfpu=neon-vfpv4 -mfloat-abi=hard"
```
gizmo98: Ok. I have not find any info if -mfpu=neon-fp-armv8  is 64Bit exclusive. Will do some testing if the new pi arrives.

joolswills: I will include "-mcpu=cortex-a53 -mfpu=neon-fp-armv8" by default - not sure but I thought it made a tiny difference in high res on quake - but could be my imagination. not enough for me to want to maintain another set of binaries. I think -mcpu is fine over -mtune etc as no makefiles have -march/mtune in them afaik

BTW. We might want to experiment on rpi2 by using "-ffast-math" by default - see https://github.com/libretro/Genesis-Plus-GX/commit/410119d5a0faf8bd57da332b5c06f2ab0ad2b548#commitcomment-16361790

Issue #1288, opened 2016-02-25
Change video_smooth default from true to false


dankcushions:
After a bit of feedback before I do the pull request (it's a 1 line change). Currently the default for the Retropie script is for video_smooth to be true, which means the image when running retroarch cores is subjected to a bilinear filter, making it slightly blurred. I guess we all know what it looks like by now :)

Personally, I think the default to be 'false'. My reasoning is:
- If it's on it makes shaders display incorrectly, yet by default shaders are also enabled (at least, enabled to be used, if not turned on)
- 'Off' seems more like a default to me. it's the raw image and doesn't make presumptions about the user's TV setup or preference.
- Such filtering settings have typically been off in other emulator systems I use (eg, retroarch standalone). I think they are turned off in the non-libretro cores in retroarch also(?)
- There's a number of threads on the forum for people asking how to turn it off.

I realise it's easy to change in Retropie-Setup, but I guess that argument goes both ways :)

joolswills:
in reply to your points:

1) shaders are not enabled by default. Many people never bother with them.

2) it is subjective - I prefer it on (depending on which monitor I use). Others prefer it off - if you can provide a decent poll that most want it off, I'll switch the default.

3) retroarch has it on by default - check their config.

4) there were a number of threads before when it was on by default with people saying it looks blocky.

joolswills:
https://github.com/libretro/RetroArch/blob/8f73779acb0595b12008a0c4f220ead5e00a479c/config.def.h#L466

HerbFargus:
I was actually thinking about this the other day, and I think I agree with you on this point (particularly in that users have been asking about it quite a bit recently)

It really is a matter of personal preference. Not that it matters all that much but some users have been using recalbox because they think it looks better but it's really just a matter of this setting.  

But @objectchris if you want to take a screenshot with and a screenshot without and then put up a poll on the forum or something so we can get feedback from the community and we'll go from there.

joolswills:
it's worth noting that it can also depend on the monitor resolution it seems how it ends up looking or at least it looks quite different on a couple of my screens (it looks much better on, on the lower resolution monitor - the effect is less).

I'm not against switching, but I'm not going to be switching this back and forth - hence a poll is the best way which we can them point to when users complain about the default.

joolswills:
https://github.com/RetroPie/RetroPie-Setup/issues/933

objectchris:
poll created :)
[...]

PR #1313, opened 2016-03-01
Changed default video_smooth from true to false.

dankcushions:
As per https://github.com/RetroPie/RetroPie-Setup/issues/1288
Looks like debate has settled in the forum thread.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -86,7 +86,7 @@ function configure_retroarch() {
     iniSet "system_directory" "$biosdir"
     iniSet "config_save_on_exit" "false"
     iniSet "video_aspect_ratio_auto" "true"
-    iniSet "video_smooth" "true"
+    iniSet "video_smooth" "false"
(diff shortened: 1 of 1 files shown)


joolswills:
it would have been polite to wait for agerement on the ticket from me - Don't know why you think I would need a one line pull request on this ?

dankcushions:
No offence intended. I thought we were all in agreement in the ticket that the users had spoken.

joolswills:
We were in agreement as to how the results looked - but bear in mind how few people responded considering the size of the userbase. I am not as convinced as you, hence I want to make other changes alongside it which you will see I have mentioned on the ticket and would have been brought up had you asked before jumping ahead.

joolswills:
anyway - no offence, just don't want this rushed - you will get your change. We also need to consider whether it's worth changing the shader default from off to on (but the shader unset), as this was all connected to the change to video smooth being set on originally.

joolswills:
We should probably move discussion to ticket but I agree there are things to be considered. ;-)

PR #1321, opened 2016-03-03
Added Solarus engine.

zerojay:
Added the Solarus open source Zelda LttP engine with 3 games made with it.


HerbFargus:
I tested it and it works great. the resolution in runcommand has to be set to 640x480 in order to be fullscreen though

joolswills:
thanks. Note this won't work on wheezy due to lacking sdl2-image

PR #1322, opened 2016-03-03
Use stable URLs to download Solarus games

christopho:
Download games compatible with Solarus 1.4 rather than the latest ones


joolswills:
thanks :)

PR #1324, opened 2016-03-04
added tyrquake standlone to experimental

joolswills:


joolswills:
Added this for access to console / addition settings etc (useful for benchmarking too)

PR #1325, opened 2016-03-04
Added Cannonball engine script.

zerojay:
Added Cannonball engine script.

Jools, there's a compilation warning that seems harmless when compiling on the Pi 2 that I'm worried about when it comes to running on Pi 3. The repository that this is being pulled from hasn't been updated for anything Pi 3, so it's possible that his compile flags here may cause some issues on it. For each file compiled, the following warning is spit out: 

warning: switch -mcpu=cortex-a7 conflicts with -march=armv6 switch

So if you could give it a quick test whenever you get your Pi 3, I'd appreciate it.

### DIFF
--- a/scriptmodules/ports/cannonball.sh
+++ b/scriptmodules/ports/cannonball.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="cannonball"
+rp_module_desc="cannonball - An Enhanced OutRun Engine"
+rp_module_menus="4+"
+rp_module_flags="!x11 !mali"
+
+function depends_cannonball() {
+    getDepends libsdl2-dev libsdl2-image-dev libsdl2-ttf-dev libboost-dev
+}
+
+function sources_cannonball() {
+    gitPullOrClone "$md_build" https://github.com/djyt/cannonball.git
+    sed -i "s/-march=armv6 -mfpu=vfp -mfloat-abi=hard//" $md_build/cmake/sdl2_rpi.cmake $md_build/cmake/sdl2gles_rpi.cmake
+}
+
+function build_cannonball() {
+    
+    mkdir build
+    cd build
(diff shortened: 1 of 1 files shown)


joolswills:
The armv6 flag needs removing / overriding for rpi2 also.

scriptmodules/ports/cannonball.sh
```diff
@@ -0,0 +1,54 @@
...
+    touch "$md_inst/hiscores.xml"
+    chown pi:pi "$md_inst/"
+    chown pi:pi "$md_inst/cannonball"
+    chown pi:pi "$md_inst/config.xml"
+    chown pi:pi "$md_inst/hiscores.xml"
+    chown pi:pi "$md_inst/res"
+    chown pi:pi "$md_inst/roms"
+    chown pi:pi "$md_inst/res/tilemap.bin"
+    chown pi:pi "$md_inst/res/tilepatch.bin"
+}
```
joolswills: do all the files need to be chowned? retropie-setup can be installed under any user - so you need to use $user also.

scriptmodules/ports/cannonball.sh
```diff
@@ -0,0 +1,54 @@
...
+
+function sources_cannonball() {
+    gitPullOrClone "$md_build" https://github.com/VanFanel/cannonball.git
+}
+
+function build_cannonball() {
+    mkdir build
+    cd build
+    cmake -DTARGET=sdl2_rpi -DCMAKE_INSTALL_PREFIX:PATH="$md_inst" ../cmake
```
joolswills: also - if the module is rpi only other platforms need to be excluded in the flags like x11/mali but it should be possible to x86/x11 with this.

scriptmodules/ports/cannonball.sh
```diff
@@ -0,0 +1,54 @@
...
+    chown pi:pi "$md_inst/hiscores.xml"
+    chown pi:pi "$md_inst/res"
+    chown pi:pi "$md_inst/roms"
+    chown pi:pi "$md_inst/res/tilemap.bin"
+    chown pi:pi "$md_inst/res/tilepatch.bin"
+}
+
+function configure_cannonball() {
+    addPort "$md_id" "cannonball" "Cannonball - OutRun Engine" "$md_inst/cannonball"
+    __INFMSGS+=("You need to unzip your OutRun set B from latest MAME (outrun.zip) to $md_inst/cannonball/roms. They should match the file names listed in the roms.txt file found in the roms folder. You will also need to rename the epr-10381a.132 file to epr-10381b.132 before it will work.")
```
joolswills: generally $md_inst is considered an area not to be used by the user - if it requires the roms installed there you can symlink it to a $romdir folder

joolswills:
Also the changes from vanfanel are merged upstream - so you can use the upstream repository.

PR #1331, opened 2016-03-05
Added core option defaults to lr-fba-next.

dankcushions:
Now, if Retropad Start button is held, the game's diagnostic menu appears, allowing various dipswitch settings to add free play and offer effects to FBA and Neogeo games. These settings are saved in the /roms/<system>/ folder as .fs files, and are automatically loaded on game load.


joolswills:
thanks

PR #1338, opened 2016-03-05
Add Stratagus

timfel:
This adds the Stratagus engine that can be used to play Warcraft I + II and Starcraft (the latter in a more limited fashion). Extracting the game data into the Stratagus format is not part of the engine. The data has to be extracted from the original media (can be done by using the installers for https://github.com/Wargus/wargus, https://github.com/Wargus/war1gus,  https://github.com/Wargus/stargus on Windows, their debian packages, or their OS X bundles). The extracted folders must then be placed in the roms directory with the folder name ending in `.data`.

I don't know how to get OpenGL mode to work (it doesn't launch for me in that case), I suspect there needs to be work in the engine to support the Raspberry Pi. However, this is a simple 2d strategy engine and software mode works just fine for the three Blizzard games.

### DIFF
--- a/scriptmodules/emulators/stratagus.sh
+++ b/scriptmodules/emulators/stratagus.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

PR #1340, opened 2016-03-05
Config editor

joolswills:
more configuration editor improvements including addition of a simple render resolution selection in the basic configuration screen.


joolswills:
There are some known bugs with the aspect ratio currently. will sort later.

joolswills:
have added the analogue sticks as dpad config - in the basic config i just included players 1 to 4 which seems reasonable (the rest are in the advanced settings). Want to keep the basic settings list reasonable short.

joolswills:
@HerbFargus ready for testing :-) Should be mostly bug free now.

HerbFargus:
I think we should include player index in the config editor so that people can easily specify ports if they use different controls by system. doesnt necessarily need to be in the basic editor but I think it should be somehwere

joolswills:
agreed, but I will do that in the next round of changes (the framework is improved now for adding more stuff more easily).

HerbFargus:
of course. Works great! 

If we are looking at a simple menu it for the next round it may be worth splitting the basic settings into two menus:

- Basic Libretro Video Settings
- Basic Libretro Controller Settings
- Advanced Configurations

The latest changes work great, no issues in my short testing,

joolswills:
I agree, already had that in mind, although there are not enough controller settings yet to justify it imho. as things get added it will be reorganised.

HerbFargus:
I trust your judgment on the matter

PR #1347, opened 2016-03-06
updated scummvm to 1.8 - use github directly for source, and enable sdl2

joolswills:
The only thing I'm thinking of changing is a few of the parameters which are set when building for the RPI. The savegame timestamps are switched off - this was because the RPI doesn't have a battery backed up clock, but internet connected RPI do have the correct time, so we might want this.

Also the hq scalers are switched off. Sure for performance we don't want them, but some people may want to run in a higher res with a scaler to get a "better image". so It might be worth including them. The event recorder is also switched off, but i dont personally think it will cause any sd card wear to worry about (no more than the system logging etc).

### DIFF
--- a/scriptmodules/emulators/scummvm.sh
+++ b/scriptmodules/emulators/scummvm.sh
@@ -12,10 +12,10 @@
 rp_module_id="scummvm"
 rp_module_desc="ScummVM"
 rp_module_menus="2+"
-rp_module_flags="dispmanx !mali"
+rp_module_flags="!mali"
 
 function depends_scummvm() {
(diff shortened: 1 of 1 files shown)


HerbFargus:
Looks good to me. I can give it a test

joolswills:
you can switch on opengl rendering and it will switch the interface to a higher res and do scaling with bilinear filtering.

HerbFargus:
## Default: 

## OpenGL

## OpenGL (no filtering)

## Myst


## Pajama Sam

HerbFargus:
Works great :)

joolswills:
thanks - looks good - im just building binaries and then I will merge.

HerbFargus:
Perfect. I'll add a mention on the wiki for the rendering options

PR #1348, opened 2016-03-07
Add Commander Genius (Commander Keen Engine)

HerbFargus:


scriptmodules/ports/cgenius.sh
```diff
@@ -0,0 +1,55 @@
...
+function depends_cgenius() {
+    getDepends build-essential libvorbis-dev libogg-dev libsdl2-dev libsdl2-image-dev libgl1-mesa-dev libboost1.55-dev libboost-dev 
+}
+
+function sources_cgenius() {
+    wget -O- -q "https://github.com/gerstrong/Commander-Genius/archive/v180release.tar.gz" | tar -xvz --strip-components=1 -C "$md_build"
+}
+
+function build_cgenius() {
+    cd $md_build
```
joolswills: you dont need to cd to $md_build as thats the default folder for the build functions

scriptmodules/ports/cgenius.sh
```diff
@@ -0,0 +1,55 @@
...
+    cmake -DUSE_SDL2=yes -DCMAKE_INSTALL_PREFIX="$md_inst"
```
joolswills: if you are not using "make install" later you probably dont need the install prefix parameter

joolswills:
The above issues are not that important, but I thought to mention them. Will merge.

HerbFargus:
appreciate the feeback

PR #1352, opened 2016-03-08
Added AlephOne engine (Marathon, Marathon 2, Marathon Infinity).

zerojay:


scriptmodules/ports/alephone.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function install_alephone() {
+    make install
+    mkRomDir "ports"
+    mkRomDir "ports/$md_id/"
+    mv "$md_build/Marathon" "$romdir/ports/$md_id/"
+    mv "$md_build/Marathon 2" "$romdir/ports/$md_id/"
+    mv "$md_build/Marathon Infinity" "$romdir/ports/$md_id"
+}
```
joolswills: Game data should be downloaded and installed at the configure stage (see quake examples).

mkRomDir etc also - the install stage is never run when things move to a binary distribution - just the $ms_inst location is unpacked and configure is then run, so apart from files in $md_inst everything else has to be done under configure.

joolswills:
thanks. please see my comments above.

scriptmodules/ports/alephone.sh
```diff
@@ -0,0 +1,52 @@
...
+
+function configure_alephone() {
+    wget https://github.com/Aleph-One-Marathon/alephone/releases/download/release-20150620/Marathon-20150620-Data.zip
+    wget https://github.com/Aleph-One-Marathon/alephone/releases/download/release-20150620/Marathon2-20150620-Data.zip
+    wget https://github.com/Aleph-One-Marathon/alephone/releases/download/release-20150620/MarathonInfinity-20150620-Data.zip
+    unzip Marathon-20150620-Data.zip
+    unzip Marathon2-20150620-Data.zip
+    unzip MarathonInfinity-20150620-Data.zip
+    mkRomDir "ports"
+    mkRomDir "ports/$md_id/"
```
joolswills: Please put the mkrom folders first and only install the game data if it isn't already installed. You also should clean up any zip files after. configure could be run more than once, so if the existing wget downloads are there it would save to zip.1 files currently.  see lr-tyrquake (it doesn't need to be in a separate function that is just done in lr-tyrquake so it can be used shared by the standalone tyrquake port)

joolswills: I don't think md_build will be the location at this stage - you can unpack to $__tmpdir/archivename.zip which would be safest

zerojay:
Updated pull request.

joolswills:
thanks.

PR #1363, opened 2016-03-09
ZDOOM: Added logic to find and add other iWAD files to EmulationStati...

zerojay:
…on if the files exist already.


joolswills:
I have plans for a selector so don't want to add them  like this, sorry. (which also adds loads of extra config folders).

joolswills:
in the meantime users can edit the default emulators.cfg for "doom" and add additional lines for launching other wads.

zerojay:
Okay, makes sense to me.

PR #1364, opened 2016-03-10
Create rc.local

pancakeykakes:
Suggest that xboxdrv and this code be added to your rc.local file being included in future RetroPie image releases for two major reasons:
(1) More accurate input reads: The default driver being used erroneously reads inputthat is not actually being sent (most noticeable when trying to assign trigger buttons during initial controller setup on first boot), where as using this configuration fixes it.
(2) retroarch. Retroarch does not initially recognize controller input after a fresh image install, and therefore requires a keyboard for initial setup.

The rc.local proposal here will allow the 360 controller be recognized in such a way that anyone running a new RetroPie image will not ever need a keyboard for initial use unless they want to do low-level modifications (which I would assume would just be done via SSH anyways). It is a very lightweight solution which will benefit many xbox360 controller users with virtually no impact to startup time, and without any risk of hangupsfor users not electing to use an xbox360 controller.

### DIFF
--- a/rc.local
+++ b/rc.local
@@ -1,0 +1,4 @@
+#!/bin/sh -e
+sudo chmod +x /dev/uinput
+
+xboxdrv --trigger-as-button --wid 0 --led 2 --detach-kernel-driver --quiet --silent & sleep 1


joolswills:
Would be better just as an issue as there is no benefit from merging a file called rc.local into the root of our project.

If I was sure running the xboxdrv wouldn't mess with any other setup (incuding other 3rd party drivers) it could be included. Why the sleep 1? Don't really want to put delays in rc.local - not sure about the chmod either - if uinput permissions are wrong they can be corrected with a udev rule

pancakeykakes:
I have a habit of adding extra sleeps, and therefore you can just take that
out.

The chmod prevents libusb errors on boot. When I was working on my project
that this came from, it was the only solution that worked. My not be
necessary on your images. The important part is having xboxdrv launch with
those arguments via rc.local.

And my apologies, I didn't think you would take to well to this being
posted as an issue. I know it would be moot to actually merge the file when
you can just have a command make the addition or have it be in the rc.local
that comes with your image.

joolswills:
have you tested running the driver as above when using other drivers ? gamecon / ps3 / xarcade2jstick etc ?

joolswills:
my xbox controller works out of the box on the image (Just have to redo the shoulder config sometimes), but it works well enough to use to go to the menu to install the xboxdrv.

JustinAiken:
As someone who only uses SNES-type controllers, I wouldn't want `xboxdrv` running on my pi :)

joolswills:
@HerbFargus agreed.

joolswills:
I'm closing this as I don't want these changes as illustrated at least.

joolswills:
but thanks for the contribution :)

PR #1365, opened 2016-03-10
Added Crispy Doom script.

zerojay:


scriptmodules/ports/crispy-doom.sh
```diff
@@ -0,0 +1,52 @@
...
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="crispy-doom"
+rp_module_desc="Crispy Doom - Enhanced port of the official DOOM source"
+rp_module_menus="4+"
+rp_module_flags="!mali !x86"
+
+function depends_crispy-doom() {
+    getDepends libsdl1.2-dev libsdl-net1.2-dev libsdl-mixer1.2-dev python-imaging
```
joolswills: does it need any of the autotools dependencies ? (as it is using autogen.sh)

joolswills:
what does this doom port offer over existing ports ? Ijust wondering as there are even more doom engines, and we don't really want to add them all unless they are needed (every module increases the time to load retropie_setup, so I would like to know we are adding things through need rather than just because they exist etc) :)

zerojay:
It was a request from the forums. I'm okay with it being left out and just living in my github, honestly. To my understanding, it's meant to lift some of the limits of the engine for more detailed levels, that sort of thing. Still plays the original levels the same way. Put it this way, I'm a huge DOOM fan and I hadn't heard of this particular source port before.

joolswills:
thanks. ok might be easiest to do that for for now. afaik zdoom does that also - although without knowing more I don't know the technically differences.

I might try and think of a mechanism when making it easier to install/remove independent modules to speed up loading, perhaps with a cache so only modules that are needed are loaded or something. It's not needed right away, but if we continue to grow at this pace it will be needed at some point.

scriptmodules/ports/crispy-doom.sh
zerojay: I added automake and autoconf to it now.

joolswills:
np though :)

zerojay:
I created my own repo (https://github.com/zerojay/RetroPie-Extra) that I'll use for all the scripts I'll be working on for stuff like this and you can feel free to dip in and grab what you feel will be right for RetroPie, suggest fixes, etc... Things in the master branch will be stuff I consider mature enough and tested and things that are still unstable/unready/has issues will be in their own branches.

I'll still be sending PRs for stuff that makes sense for mainline too. :)

joolswills:
sounds great thanks :)

PR #1381, opened 2016-03-15
Fix typo in ISSUE_TEMPLATE

adrianmoisey:


joolswills:
I've reworded it so this change is not needed. thanks.

PR #1384, opened 2016-03-16
Rpi1 default res

joolswills:


joolswills:
Assuming everyone is happy then. :)

PR #1393, opened 2016-03-19
Config editor

joolswills:
Switching to use quoted values for retroarch.cfg - when retroarch saves the configs itself it uses quotes, since we were not, it broke the config editor - so we might as well just use quotes (even though their skeleton config file does not.

### DIFF
--- a/scriptmodules/libretrocores/lr-pcsx-rearmed.sh
+++ b/scriptmodules/libretrocores/lr-pcsx-rearmed.sh
@@ -47,9 +47,5 @@ function configure_lr-pcsx-rearmed() {
     mkRomDir "psx"
     ensureSystemretroconfig "psx"
 
-    # system-specific, PSX
-    iniConfig " = " "" "$configdir/psx/retroarch.cfg"
-    iniSet "rewind_enable" "false"
-
     addSystem 1 "$md_id" "psx" "$md_inst/libretro.so"
 }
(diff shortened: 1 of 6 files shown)


scriptmodules/libretrocores/lr-pcsx-rearmed.sh
```diff
@@ -48,7 +48,7 @@ function configure_lr-pcsx-rearmed() {
     ensureSystemretroconfig "psx"
 
     # system-specific, PSX
-    iniConfig " = " "" "$configdir/psx/retroarch.cfg"
+    iniConfig " = " '"' "$configdir/psx/retroarch.cfg"
     iniSet "rewind_enable" "false"
```
HerbFargus: I don't believe this config will work as it shows up after the #include line in the system specific config. It is likely that we'll need to tweak or add a function to included system based configs above the #include line. Would also be helpful for mednafen type cores until upstream issues are fixed.

joolswills: good point. - been broken for ages then. we should remove it maybe - or do a "insert before" ini thing

HerbFargus: rewind_enable is already set to false system-wide by default. so the line can be removed.

PR #1401, opened 2016-03-22
Added missing quotes around file variable

Lauszus:
This affected [scriptmodules/emulators/mupen64plus/mupen64plus.sh](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/mupen64plus/mupen64plus.sh) in case the retroarch configuration file has spaces in the file name.


Lauszus:
This was a problem for me as the retroarch auto config file for my PS3 controller was called "PLAYSTATION(R)3 Controller.cfg".

joolswills:
Thanks! well spotted :)

joolswills:
I thought we dealt with spaces (and other chars) in the retroarch configs - but perhaps not - I will review it, although less important now as we used to try and match the filename of the now removed config tool they shipped with - in any case, string parameters should be quoted :) cheers

PR #1404, opened 2016-03-25
Ports

joolswills:
It was getting confusing with so many "ports" config folders in /opt/retropie/configs/

this changeset moves all ports configs to $configdir/ports/NAME and that location is now used.

`$md_conf_root` points to $configdir usually, but is adjusted to "$configdir/ports" in addPort (addPort needs to be first now so this variable is set correctly).

This needs some testing.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -580,6 +580,14 @@ function setRetroArchCoreOption() {
     chown $user:$user "$configdir/all/retroarch-core-options.cfg"
 }
 
+# sets module config root to subfolder from $configdir - used for ports that are not actually in the ports etc
+function setConfigRoot() {
+    local dir="$1"
+    md_conf_root="$configdir"
+    [[ -n "$dir" ]] && md_conf_root+="/$dir"
+    mkUserDir "$md_conf_root"
+}
+
 # add a framebuffer mode to /etc/fb.modes - useful for adding specific resolutions used by emulators so SDL
 # can use them and utilise the rpi hardware scaling
 # without a 320x240 mode in fb.modes many of the emulators that output to framebuffer (stella / snes9x / gngeo)
@@ -654,7 +662,7 @@ function addSystem() {
 
         # automatically add parameters for libretro modules
         if [[ "$id" =~ ^lr- ]]; then
-            cmd="$emudir/retroarch/bin/retroarch -L $cmd --config $configdir/$system/retroarch.cfg %ROM%"
+            cmd="$emudir/retroarch/bin/retroarch -L $cmd --config $md_conf_root/$system/retroarch.cfg %ROM%"
         fi
     fi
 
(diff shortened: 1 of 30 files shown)


zerojay:
Let me know and I'll tackle converting my scripts over ASAP. I agree with this change, config was getting very crowded.

joolswills:
basically - to update a module

* Put addPorts first
* Use $md_conf_root instead of "$configdir" (you till need to include the name after eg $md_id or doom (configs for doom go in doom, but there are multiple modules that implement a doom port)
* no need for mkRomDir "$roms" (but subfolders have to be made if needed)

HerbFargus:
All seems fine to me, tested all the defaults and a few experimental like cgenius and xrick, no issues. 

One thing that I wasn't sure about was with my keyboard before I was pretty sure I could use hotkeys to exit but now it uses the escape key so I'm wonder if its missing a retroarch.cfg somewhere when it loads the libretro ports. which its looking like you may have addressed above

joolswills:
aah no. There is a bug with the paths hmm and I may need to introduce a new function to set the configdir perhaps rather than doing it in addPorts (so addPorts can be at the end like addSystem, but another function would need to be called first).

joolswills:
I need to have a think about the cleanest way of doing this, so that I don't have to keep reworking the modules :-)

joolswills:
I want to keep it so the same variables are used between modules, to keep it simple, but the problem is we have a inconsistent structure currently (eg love is in ports, but is actually a system in regards to config / rom folder structure, some ports are actually libretro cores and so on)

InsecureSpike:
I hope you don't mind me throwing in an idea, if the ports at the moment are including actual ports and systems could these not be split into two? ports stay as ports but experimental systems could go under another menu as experimental, then when they have been tested then they get moved out as normal and have their own system menus?

joolswills:
I'm going with a solution that means another function call is needed to set the config location for things that don't use the obvious default for the scriptmodule type (eg ports in libretrocores)

joolswills:
@InsecureSpike the way stuff is organised in the retropie-setup menu will be changed, with "core, main, and optional, and experimental modules" - that is separate from the work I am doing now, which is basically reorganising the config structure to match the roms directory structure more closely (so ports are not mixed amongst all the system folders)

InsecureSpike:
ahh I see, sorry for the confusion

joolswills:
@HerbFargus updated branch - I think the libretro ports should load their configs correctly now

HerbFargus:
Can confirm the libretro configs are fixed

joolswills:
Thanks for testing. I will merge.

joolswills:
One thing to note is for existing libretro ports (lr-prboom / lr-tyrquake / lr-nxengine)  - when it moves their configs, it won't overwrite the retroarch.cfg after with a new config so the retroarch.cfg will contain an old path for "input_remapping_directory" - it will create a retroarch.cfg.rp-dist with the new config though which can be copied over.

PR #1412, opened 2016-03-31
fix compiling scummvm from source

MarcLandis:
Running ./configure results in the following error:
ERROR: The keymapper and the event recorder cannot be enabled simultaneously currently, please disable one of the two

As keymapper is explicitly enabled already, disable eventrecorder.


joolswills:
Thanks. Somehow this error wasn't thrown before when building the binaries.

PR #1416, opened 2016-04-04
fix updating RetroPie-Setup from git

MarcLandis:
Current error:
You are not currently on a branch. Please specify which
branch you want to merge with. See git-pull(1) for details.

    git pull <remote> <branch>

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -169,7 +169,7 @@ function updatescript_setup()
         return
     fi
     local error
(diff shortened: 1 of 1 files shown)


joolswills:
it's not broken - you have just switched your local branch. I don't want to hardcode it in, since this is useful for testing when working on another branch that is tracking another upstream

joolswills:
(do `git checkout master` and a plain git pull should be fine again)

MarcLandis:
oops sorry - you are right.

PR #1418, opened 2016-04-06
Add LXDE install option to Raspbian Tools

HerbFargus:


HerbFargus:
should be sorted

joolswills:
Thanks :)

PR #1419, opened 2016-04-07
rework autostart code to use raspi-config on jessie, and to show opti…

### LINKED ISSUE #1344: autostart improvements
Fix up autostart to use raspi-config if possible on jessie (takes parameters now).

joolswills:
…ons to boot to console or desktop, apart from on X11 where it has just an enable/disable autostart - fixes #1344


joolswills:
this may need a little bit more testing.

joolswills:
confirmed ok.

PR #1422, opened 2016-04-09
added new gamelist.xml with descriptions and icons for retropiemenu.

joolswills:
removed dispmanx configuration as most users don't need to use it
big thanks to Rookervik and Herb

@HerbFargus - I have tested this on x86 and it seems ok, but let me know.


HerbFargus:
Will test now on my pi3

HerbFargus:
Looks great :) just fixing up pixel so it looks nicer with the new icons, but the code on this end it just fine.

joolswills:
Thanks!

PR #1433, opened 2016-04-15
Exclude downloaded_images from configedit

wetriner:
All scraped images are loaded into the configuration editor list, so depending on how many games are scraped, it can get impractical to navigate.


joolswills:
thanks.

Issue #1212, opened 2016-01-28
Replacement version of Linapple


dabonetn:
I've made a lot of changes to linapple to better support frontends.

I'd like to get it included in retropie.

It now supports direct launching of floppy images, and joystick remapping and quitting from joystick.
It also looks in the users home directory .linapple directory for the linapple.conf and Master.dsk.

Project is located here.

https://github.com/dabonetn/linapple-pie

For integration into retropie that I see the following changes needed.

Change to needed files to the following

        'CHANGELOG'
        'INSTALL'
        'LICENSE'
        'linapple'
        'Master.dsk'
        'README'
	'README-linapple-pie'
	'linapple.conf'

Make a symbolic link from /home/pi/.linapple to /opt/retropie/configs/apple2

Place the linapple.conf and Master.dsk into /opt/retropie/configs/apple2

Change the startup scripts to reflect the command line changes.

(linapple -1 %ROM%)

Change the media types to refect apple disk images (.po .PO .dsk .DSK .nib .NIB)

Get controller mappings from the main config into the linapple.conf file.

Apple joysticks only have max 2 buttons on joystick one, and one button on joystick two.

The following is from the linapple.conf file, I tried to make it as simple as possible to reconfigure the joysticks.

```
#######################################################################
#
#	Joysticks, those sticks of joy! There may be 2 joysticks at the same time
# Possible values are:
#	0	- joystick disabled
#	1	- use PC joystick #1 or #2, for corresponding joystick

#	2	- Keyboard standard
#	3	- Keyboard centered
# When joysticks used as a keyboard, they are stuck to Numpad keys (1-9 - axis movement, 0 - button1, . - button2)
#  when centered used, axis of joystick will be centered after releasing any cursor (Numpad 1..9) key.
# otherwise it assumed to be pressed always

#	4	- Use mouse as a joystick. Rather interesting thing, try it. Useful in Fantavision(tm)by Broderbund Software

# Default values are 2 for Joystick 0, and 0 for Joystick 1

	Joystick 0	=	1
	Joystick 1	=	0

# For Joysticks you can define which Joystick index number, axis number, and buttons.
# 
# Default for Joystick 1 is index 0, axis 0 and 1, buttons 0 and 1.
# Default for Joystick 2 is index 1, axis 0 and 1, button 0.


	Joy0Index	=	0
	Joy1Index	=	1
	Joy0Button1	=	0
	Joy0Button2	=	1
	Joy1Button1	=	0
[...]

Issue #1055, opened 2015-10-20
LinApple improvements


joolswills:
http://blog.petrockblock.com/forums/topic/how-to-get-linapple-disk-images-running-from-emulation-station-in-retropie/

Should be easy enough to patch the source on the fly, and improve the launch script based on the above.

HerbFargus:
added as of https://github.com/RetroPie/RetroPie-Setup/pull/1437

PR #1437, opened 2016-04-17
Update linapple with dabone's changes

HerbFargus:
supports autolaunching disks (still needs a reset to start a game though) but this will remove the need for a start script for the apple 2.

should address #1055 and #1212

### DIFF
--- a/scriptmodules/emulators/linapple.sh
+++ b/scriptmodules/emulators/linapple.sh
@@ -19,8 +19,7 @@ function depends_linapple() {
 }
 
 function sources_linapple() {
-    wget -O- -q $__archive_url/linapple-src_2a.tar.bz2 | tar -xvj --strip-components=1
-    addLineToFile "#include <unistd.h>" "src/Timer.h"
+    gitPullOrClone "$md_build" https://github.com/dabonetn/linapple-pie.git
 }
 
 function build_linapple() {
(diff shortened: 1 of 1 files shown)


joolswills:
This doesn't merge currently - your branch must have been out of date. I assume we will need to copy the default config file etc also still ?

HerbFargus:
Ah yeh, my bad I'll sort that,  by copying the default config do you mean linapple.conf?

joolswills:
Yeh - assuming it still uses that (so people can tweak it if needed)

HerbFargus:
Ok it should be squashed and ready unless there are any other glaring issues with the module.

scriptmodules/emulators/linapple.sh
```diff
@@ -37,31 +36,18 @@ function install_linapple() {
...
-        'splash.bmp'
         'Master.dsk'
         'README'
+        'README-linapple-pie'
+        'linapple.conf'
     )
     # install linapple.conf under another name as we will copy it
     cp -v "$md_build/linapple.conf" "$md_inst/linapple.conf.sample"
+    cp -vf "$md_build/Master.dsk" "$md_conf_root/apple2/Master.dsk"
 }
```
joolswills: This line shouldn't be needed, but also the install stage should only affect files in the $md_inst area.

joolswills: If the Master.dsk needs to exist in $md_conf_root/apple2 it needs to be copied at the configure stage as with the linapple.conf

HerbFargus: yeah it is needed for dabones changes but I can move it to the configure function

HerbFargus:
when copying from the configure function instead of $md_build do I need to copy from $md_inst?

joolswills:
Yeah - $md_inst needs to contain everything that is needed (apart from things that can be downloaded from configure such as wads for doom etc), so you install to $md_inst from install, then copy from there anything to the configuration folder as needed.

joolswills:
that way when we archive up a binary package, it includes everything that is needed ($md_build would not exist in this case)

HerbFargus:
make sense. I think it should be good now - I dont have my pi to test right now though

joolswills:
I will give a a try in a bit also - Cheers!

HerbFargus:
just as a reference its supposed to make it more friendly for frontend launching but as it was originally a computer there isnt really a clean way of running it all with a controller as it still requires a reset to start the disk and also other keyboard buttons to initialise a game, but at least this will remove the need for the start script

HerbFargus:
I tested it and it seems to be working as intended.

joolswills:
Apologies for being slow on this - still need to test etc.

joolswills:
I'm sure it is fine actually.. I'll merge - don't want to be seen as holding back development ;-) thanks

PR #1439, opened 2016-04-18
Don't fail installation if /dev/tty doesn't exist

adrianmoisey:
I use ansible to provision my RetroPie. This fails for me now, because it depends on `/dev/tty` existing.
This tries to get around that.

### DIFF
--- a/retropie_packages.sh
+++ b/retropie_packages.sh
@@ -73,5 +73,5 @@ fi
 printMsgs "console" "${__INFMSGS[@]}"
 
 if [[ -n $__joy2key_pid ]]; then
-    kill -INT $__joy2key_pid 2>/dev/null
+    kill -INT $__joy2key_pid 2>/dev/null || echo "Failed to kill joy2key"
 fi
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -40,7 +40,11 @@ for arg in sys.argv[2:]:
 event_format = 'IhBB'
 event_size = struct.calcsize(event_format)
 
-tty_fd = open("/dev/tty", "w")
+try:
+    tty_fd = open('/dev/tty', 'w')
+except:
+    print 'Unable to access /dev/tty, exiting'
+    sys.exit(0)
 js_fd = open(sys.argv[1], "rb")
 
 buttons_state = 0
@@ -65,7 +69,7 @@ while True:
     if js_type == JS_EVENT_BUTTON:
         if js_number < len(button_codes) and js_value == 1:
             hex_chars = button_codes[js_number]
-    
+
     if js_type == JS_EVENT_AXIS:
         if js_number % 2 == 0:
             if js_value <= JS_MIN * JS_THRESH:


retropie_packages.sh
```diff
@@ -73,5 +73,5 @@ fi
 printMsgs "console" "${__INFMSGS[@]}"
 
 if [[ -n $__joy2key_pid ]]; then
-    kill -INT $__joy2key_pid 2>/dev/null
+    kill -INT $__joy2key_pid 2>/dev/null || echo "Failed to kill joy2key"
```
joolswills: I would prefer to make sure this code is not called if joy2key is not actually running rather than showing an error - but I am happy to add the checking to joy2key.py - thanks.

joolswills: I think actually, we can just not worry - if the process has already died for whatever reason, we silently fail here as before.

adrianmoisey: The problem I'm fixing with this change, is when joy2key isn't around, and kill tried to kill it. Kill fails with a non-zero exit code, and the fails.

joolswills: the script should keep going unless your automation software is stopping due to an error being thrown ? The above change won't stop kill from exiting with a non zero code.

joolswills: Are you running the bash with `set -e` on ?

adrianmoisey: I could be mistaken, but without the || the entire script's exit code will be non-zero if the kill is non-zero, which causes my automation to fail.

adrianmoisey: It's executing it with the settings passes in via the shebang.

joolswills: if this line causes it to stop due to be running with "exit on error" set, there are many other places which will fail - so I think this will need to be disabled via your automation software.

joolswills: I have never used the software before but - http://docs.ansible.com/ansible/playbooks_error_handling.html seems to point to a setting.

adrianmoisey: > if this line causes it to stop due to being run with "exit on error" set, there are many other places which could fail - so I think this will need to be disabled via your automation software.

That sort of defeats the point of having an exit code then. In this particular case it's ok for joy2dev to not be running, because of the change I made above. Ignoring exit codes could mean that something breaks (which is actually broken), but everything will continue, and it will be almost impossible to debug.
I've tested this branch, and this is currently the only script that exists non-zero.

Would you be fine with this change instead:
```
if [[ -n $__joy2key_pid ]] && ps -p $__joy2key_pid; then
    kill -INT $__joy2key_pid 2>/dev/null
fi
View
```

joolswills: then the correct fix would be to add an exit 0 at the end of the code if the kill if this is the cause, but there are lots of places where non 0 return codes do not mean an error, eg with grep etc. I'm happy to add an exit 0 at the end if it will help.

joolswills: I will consider adding a || true but the code still needs further changes as it won't give fail return codes for failures currently even though return codes are used internally.

scriptmodules/supplementary/runcommand/joy2key.py
```diff
@@ -40,7 +40,11 @@ def signal_handler(signum, frame):
 event_format = 'IhBB'
 event_size = struct.calcsize(event_format)
 
-tty_fd = open("/dev/tty", "w")
+try:
+    tty_fd = open('/dev/tty', 'w')
+except:
+    print 'Unable to access /dev/tty, exiting'
+    sys.exit(0)
 js_fd = open(sys.argv[1], "rb")
```
joolswills: Also if following your argument regarding exit codes being valid, then joy2key should not return a 0 if it fails to run, and should return a non 0 error.

joolswills:
I have addressed this in a different way in https://github.com/RetroPie/RetroPie-Setup/commit/e1a2960c30b2526bbc2b7687fabb9255b58f279b which should now return a failure / success correctly from the rp_callModule function.

joolswills:
thanks.

PR #1440, opened 2016-04-18
mupen64plus: add videocore video plugin

gizmo98:
tachiweasel's mupen64plus video plugin is a little bit faster but also less compatible as any other video plugin. 
https://github.com/tachiweasel/mupen64plus-video-videocore


joolswills:
Thanks. Will rebuild the binaries.

joolswills:
Done.

PR #1447, opened 2016-04-23
Added a theme gallery using fbi

wetriner:
A menu option for esthemes that handles cloning and displaying a sample image of all themes.

### DIFF
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -66,6 +66,15 @@ function configure_esthemes() {
         local options=()
         local status=()
         local i=1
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - looks good. We will probably want to copy/fork the gallery into the RetroPie org at some point if you are ok with that (Keeps everything together then).

wetriner:
Sounds good to me. You can fork or copy at any point. I don't really care about having my username involved with anything.

joolswills:
Cheers. The only other thing is it needs a bit of extra logic to use another viewer for X11 target as FBI wouldn't work from the desktop env on a pc etc. That can be added at some point.

Issue #1118, opened 2015-12-05
Documentation, FAQs, Guides, etc [Feature Request]


ProxyCell:
Howdy,

Would it be a good idea to add some kind of documentation/guides/FAQs/etc into RetroPie itself? Possibly accessible from Emulation Station?

I know this is kind of crazy-sounding but if we can agree on a way to load/display them on screen with just a gamepad I don't think its so bad.

robertybob:
One idea I had was to have add a help section, perhaps next to RetroPie within Emulationstation. Within that are shortcuts to PDFs for different things (adding roms, installing new themes etc) - there are a few pdf readers for Raspbian and I figure this would be quicker and easier than linking to a website or GUI screen with help info.

joolswills:
In terms of the additional work to add and maintain this I prefer it as it is currently with the documentation online.

HerbFargus:
I would prefer to only maintain one set of documentation. Perhaps if RetroPie were more static and wasn't in constant development, it would be more viable.

joolswills:
I did a bit of reading earlier, and I may consider this as a retropie module if we have it so we checkout the documentation repository,  and then provide a simple console render of the markdown documentation - that way we don't need to maintain two sets of documentation. However, I have other things I want to do first.

joolswills:
eg. using the "markdown" command, and viewing in lynx on terminal.

HerbFargus:
Sounds doable, images might be a bit of a trick.

joolswills:
yeh it wouldn't include images with this method, but it would be simple and easy to update from the rpi.

ProxyCell:
oh wow, you guys have been really running with this idea I see
i just felt it would be nice to have something offline and accessible for users of the device itself

i was even thinking of creating some myself so you guys didn't have to maintain anything new. i was really thinking along the lines of just a few articles/FAQs for features like the SAMBA share, USB roms, etc

HerbFargus:
this is more or less integrated.

PR #1455, opened 2016-05-01
Create wikiview.sh

wetriner:
Github Wiki viewer using pandoc and lynx. Used joy2key to have reasonable joystick navigation in lynx. This only will display one page and then go back to the menu. None of the links work for local viewing without editing the links, so they are disabled. The end result isn't remarkable. The markdown conversion to html has its limits. Inspired by issue #1118

### DIFF
--- a/scriptmodules/supplementary/wikiview.sh
+++ b/scriptmodules/supplementary/wikiview.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/wikiview.sh
```diff
@@ -0,0 +1,104 @@
...
+# See the LICENSE.md file at the top-level directory of this distribution and 
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="wikiview"
+rp_module_desc="RetroPie-Setup Wiki Viewer"
+rp_module_menus="3+"
+rp_module_flags="nobin"
+
+rootdir="/opt/retropie"
```
joolswills: This path is already defined (and should not be hardcoded).

joolswills:
Thanks. please remove the rootdir define and squash and I will merge. I may make a few tweaks after to de-duplicate the joy2key code with existing joy2key functionality

wetriner:
First time doing a squash, so hope it's correct. I've removed the line defining rootdir.

joolswills:
Looks good - thanks for the contribution :-)

PR #1459, opened 2016-05-03
Filename change. Avoids script error when rebuilding from source.

dankcushions:
Binaries will subsequently need to be rebuilt to avoid a similar error from installing from old binaries, I presume.

Samples now a core option, rather than hard-coded, so added that, even though they default to enabled. Cheats and sample rates are a core option, but I think these should remain user-preferences.

This is all following https://github.com/libretro/mame2003-libretro/pull/29
Should this PR be extended so the script moves the old paths to the new ones? I believe that should be possible, but may end up making this all a bit messy. I could just update the [wiki](https://github.com/RetroPie/RetroPie-Setup/wiki/lr-mame2003) with a warning?

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -37,7 +37,7 @@ function build_lr-mame2003() {
 function install_lr-mame2003() {
     md_ret_files=(
         'mame2003_libretro.so'
-        'README'
+        'README.md'
         'changed.txt'
         'whatsnew.txt'
         'whatsold.txt'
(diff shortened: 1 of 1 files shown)


joolswills:
thanks :)

joolswills:
I have now updated the RetroPie binaries for mame2003

joolswills:
Regarding paths - we can just mention it in the wiki perhaps ? Keeps it simple.

dankcushions:
Thanks :) I've updated the wiki https://github.com/RetroPie/RetroPie-Setup/wiki/lr-mame2003

joolswills:
Cheers!

PR #1463, opened 2016-05-06
update linapple depencies and configs

HerbFargus:


joolswills:
Merging - it would be useful in future to have a commit message that details why a change it made so It's easier for people to follow. eg. - this isn't requires due to new emulator binary etc.

PR #1465, opened 2016-05-07
Add reicast input setup module

gizmo98:
If retropie controller configuration fails this module can be used to setup controller.

### DIFF
--- a/scriptmodules/supplementary/reicastinput.sh
+++ b/scriptmodules/supplementary/reicastinput.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - I don't think this warrants a module of its own however. You could add it as a gui_ function to the existing reicast.sh and add it to the setup menu, or perhaps add it to the ES menu, and handle it from the existing launch script.

gizmo98:
I thought retropieinput.sh would be a good place to put this. There are functions for keyboard setup and there was a similar joypad config function some time ago. But it is a "retroarch" input module. So a new file was necessary. ;) 
retropieinput.sh
reicastinput.sh

Do you mean i should modify the launch menu and add this option or add a menu item in ES?!

joolswills:
I was thinking a menu item like the existing "systemManager.cdi". But perhaps that will confuse users as they may think they have to use it to configure.

I am moving the retroarchinput.sh into the configurationeditor module btw along with some manual controller config stuff, just got to think of the best way to handle it. Perhaps it should be included there, but I need to think of a way to do it, so module code is still together - perhaps by having some special menu functions that each module can use to "add configuration" to the configuration editor.

Maybe I will accept this as it's own module for now, but move it later then.

joolswills:
Going to merge and move around later. thanks.

PR #1469, opened 2016-05-08
Add default gcc version for raspbian jessie

gizmo98:


joolswills:
This shouldn't be needed I think - we only had this set for Wheezy as the default gcc was 4.6 on that which had some issues.

gizmo98:
Reicast for example has gcc 5.x issues. I thought this value could be used to check the system gcc version. Did not know it was for wheezy.

joolswills:
you could force gcc 4.9 in that module (add it as a dependency etc), although I assume this is more likely to be a code issue than a gcc bug (but possible).

joolswills:
I'm going to merge this as I'm getting reports of people with old gcc as default on Jessie - not sure how they are doing that. I will allow it to be overridden though later.

joolswills:
(i will re-implement it manually as branch deleted)

PR #1471, opened 2016-05-08
Add lr-reicast for platfom x86

gizmo98:
The current source code cannot be used for rpi. But x86 build works quite well.


joolswills:
thanks.

PR #1480, opened 2016-05-13
Updated splashscreen.sh and asplashscreen

wetriner:
* Changed from a directory search to a file search to allow for future restructuring in the splashscreen github to not require files in directories.
* Added an append option to make up for the lost ability to select a directory and append all contained files to the splashscreen.list.
* Added Preview options for images and videos.
* Restructured menus to have toggle options.
* Added a randomizer option that won't show up until an updated asplashscreen has been copied over to /etc/init.d/ during the splashscreen update.

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -24,7 +24,7 @@ function install_splashscreen() {
     gitPullOrClone "$md_inst" https://github.com/RetroPie/retropie-splashscreens.git
 
     mkUserDir "$datadir/splashscreens"
-    echo "Place your own splashscreen in here, each one with its own folder" >"$datadir/splashscreens/README.txt"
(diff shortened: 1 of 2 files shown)


joolswills:
thanks!

joolswills:
I was short on time to review this earlier and it mostly looks good - thanks for the work.

There are a few things I think could be changed though or improved on:

* not sure combining both submenus into a single function offers any advantage over having multiple functions - and would probably be more readable as wouldn't need the additional logic.
* we generally split if/then/fi onto multiple lines.
* videoextens should be a local var

HerbFargus:
also just as a reference I've been going through the splashscreen repo on my reorganisation branch [here:](https://github.com/HerbFargus/retropie-splashscreens/tree/reorganisation) to standardise the naming conventions and files and probably to slim it down a bit by removing some and maybe optimising others. still a work in progress.

PR #1482, opened 2016-05-14
splashscreen:

joolswills:
* addresses some of the points from https://github.com/RetroPie/RetroPie-Setup/pull/1480#issuecomment-219145076
 * make functions more standalone and simplify logic - eg choose_splashscreen just returns a splashscreen - setting / preview etc is handled in their own functions
 * put extensions into functions
 * paths for splashscreen locations is now handled in own function

this needs some testing

### DIFF
--- a/scriptmodules/inifuncs.sh
+++ b/scriptmodules/inifuncs.sh
@@ -54,6 +54,10 @@ function iniProcess() {
 
     [[ "$cmd" == "unset" ]] && key="# $key"
 
+    # escape backslashes and pipes for sed
+    value="${value//\\/\\\\}"
+    value="${value//|/\\|}"
+
     local replace="$key$delim$quote$value$quote"
     if [[ -z "$match" ]]; then
         # add key-value pair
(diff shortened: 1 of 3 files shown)


joolswills:
I need to do a couple more things also - I haven't looked at asplashscreen yet but I want to put the extensions in there based on the values from the module so we don't duplicate that

HerbFargus:
I will test now, you can also see my notes on https://github.com/RetroPie/retropie-splashscreens/issues/11 but I'm happy with whatever you deem to be most produnt

HerbFargus:
all seems good to me

joolswills:
I have added support so the extensions used in the module are added to the init.d script. I haven't had a chance to test it yet though.

joolswills:
I found another issue just now - the /home/pi folder is hardcoded in asplashscreen so this will need sorting as RetroPie-Setup can be installed under any user.

joolswills:
ok one more set of changes committed to change the hardcoding of the folders. still untested - your help is appreciated thanks :)

HerbFargus:
from my testing everything seems to be in order

joolswills:
I shall merge it, and we can fix up any problems as they arise. thanks again.

PR #1483, opened 2016-05-15
reicast.sh: use command line parameters

gizmo98:
Use command line parameters instead of ini tweaks. This method is safer
and more reliable.


joolswills:
Thanks!

PR #1488, opened 2016-05-23
bluetooth: fix connection problems

gizmo98:
After installation of newest pi-bluetooth package bluetooth is broken. Add necessary package rapberrypi-sys-mods.

@joolswills I don't know if this should be already merged. If raspberrypi-sys-mods will be installed there is a risk users will be asked if an udev rule should be changed.


joolswills:
I have fixed it on my image - but yeh it may prompt a user to overwrite using an older image - we could possibly force it - http://askubuntu.com/questions/56761/force-apt-get-to-overwrite-file-installed-by-another-package

PR #1489, opened 2016-05-24
Modular

joolswills:
Initial rewrite of the setup menu, including per package/module management  
 * rp_module_menus is no more - instead use rp_module_section with one of 
   - core (core packages)
   - main (main packages - which will be installed by default in the image)
   - opt (optional packages)
   - exp (experimental packages)
   - driver (driver packages)
   - config (configuration packages / tools)

 * The setup menu organises the data based on the above sections. more could be added in the future if needed. Packages (internally modules), can be added / configured / removed individually, or as entire sections.

 * The setup menu will automatically detect if a binary is available to be installed. the nobin flag is no longer needed. modules that install directly from binary via aptInstall or via a prebuilt binary need to use an install_bin_ function instead of install_

 * rp_module_section of type "config" will have the "gui_" function called first if it's available, otherwise a standard depends/sources/build/install/configure will be called

 * configure is no longer used for "gui" configuration function - use gui_ instead. Had already started to move to this before, but now it is required.

This is still a WIP and there may be bugs, but I welcome anyone to test it and give feedback. The current selection of what is in "main" and what is "optional" is my own choice and we can tweak the sections. The idea is to have most common stuff in main, and let users install what they need as they need it from optional/experimental packages


joolswills:
And I've just realised I missed out most of the ports, so need to reorganise them and update this branch.

joolswills:
Fixed a bunch of bugs - should be in better shape for testing now.

HerbFargus:
Its looking pretty good

in terms of tweaks:

- Configs: there are two main places for configs- hidden folders/files in the home folder, and the 'opt/retropie/configs' folder.  I presume the configs are something a user has to remove manually?  For example with vice it has the symlinked hidden `.vice` folder in the home folder that remains after removal - `/opt/configs/c64` also remains, etc. 
- ROM folders: I'm not sure when removing how it will handle launch scripts like with UAE4ALL, I presume all the stuff in the roms folders will just require a manual removal as well?

I've uninstalled and reinstalled a bunch of stuff, there were a few flukes when I messed with the core settings when I installed all of them together from binaries from not finding runcommand.sh when launching a game (which was fixed by installing runcommand binaries individually rather than with all of them for some reason) to emulationstation not finding any systems even though I had roms installed and had reinstalled the emulator, so I'm going to have to play with it a bit more to get those errors to be fully reproducible, but for the most part it seems to work well enough.

joolswills:
thanks. Currently neither symlinks nor configs are removed. I may be able to clean up symlinks in the future with some additional changes, but I think configs can be left for the user to remove. Perhaps an option can be added for that but it would require that each module reports what it has installed (or what will be created by the emulator). so it requires more work. Post 4.0 I think for that.

Similar with rom folders, although I was planning to implement some of this now. Each module references what system it is, and based on that we can clean up the ES config, rom folders (if empty), and so on (There are cases where multiple emulators share the same rom folder which it will need to take into account). Initially I will make it clean up the es config (if there are no other emulators using it).

joolswills:
In regards to updating, since we will ship with the core and main stuff installed, the update installed emulators should be enough. If in a new retropie-setup we decide we want to add more emulators to main, it would be up to the user to install them from the list. That way we don't install new stuff a user may not want when they have "customised" their setup.

HerbFargus:
Sounds good, looks like it will greatly simplify things including updates. 

When I get off work I'll set up a Jessie Lite image and build off of that and see how it fares.

HerbFargus:
so I installed it fresh from a raspbian Jessie Lite image and it seems to be working well.  

- As mentioned above, when installing all core packages at once from binaries, the runcommand menu seems to miss the `/opt/retropie/supplementary/runcommand/runcommand.sh` - updating/installing runcommand individually resolves it.

Other than that everything seems to be working really well. it is possible I missed some things but nothing major popped up.

joolswills:
fixed the runcommand issue, and added code for better removal of modules. removes entries from emulators.cfg and removes entry es_system.cfg if it is the last emulator for that system installed. Ports works similar, except it removes from es_systems.cfg is the roms/ports has no more launch scripts left

HerbFargus:
I can confirm that the runcommand issue is resolved

HerbFargus:
es_systems.cfg system removal also seems to function as intended (albeit with some minor empty lines but thats just formatting)

HerbFargus:
re-adding the RetroPie menu doesnt seem to be an option in setup unless I overlooked it...

joolswills:
I'm thinking of merging this soon, after a bit further testing from myself, and once I have added some more functions for building the binaries for uploading  - Was wondering if @gizmo98 or @petrockblog had a chance to try the branch out - would be interested in getting some more feedback. Thanks for your help @HerbFargus

petrockblog:
Thanks for the heads up - I am testing it right now.

petrockblog:
I also like the new interface.
So far, I only found some minor issues from which I am even not sure that they are directly related to the changes in this pull request. Here are my findings:

 - px68k did not build successfully, but it seems to be an issue within the sources itself.
 - Updating EmulationStation from binaries in the core menu yields this in one step:
```bash
Running action 'install_theme' for 'esthemes' : Install themes for Emulation Station

error: Your local changes to the following files would be overwritten by merge:
        oric/art/controller.svg
        oric/art/system.svg
Please, commit your changes or stash them before you can merge.
Aborting
/home/pi/RetroPi
```

Besides that I installed all packages, removed some, also tested the command line behavior. I did not find any other bugs so far.

joolswills:
Thanks.

I will look into px68k but it's unrelated to these changes. The install issue is odd - that should only happen if some files were changed on your local system for the themes repository. I've not seen that here but I can try and reproduce.

HerbFargus:
The theme thing may have been my bad. Possibly a bad merge on a change I'd made.... I'll look into it.

joolswills:
It's to do with a mismatch of git settings from the machine I build the images on (autocrlf = input) to the default on the retropie image (autocrlf = false), and the fact some files in that repo have windows line endings in.

I will change to "autocrlf = false" on my local machine for future images, but we should avoid adding CRLF files to the repository to avoid things like this.

I will change the theme updating script to do a clean checkout for now, to workaround the issue.

joolswills:
This branch is basically done now. I will build any missing binaries for the main/opt sections though before I merge.

If anyone has a reason why this should not be merged, please let me know :-) thanks.

PR #1492, opened 2016-05-25
mobile gamepad configuration file

sbidolach:
Configuration for new mobile gamepad implementation

### DIFF
--- a/scriptmodules/supplementary/mobilegamepad.sh
+++ b/scriptmodules/supplementary/mobilegamepad.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. There are some major changes coming, so this will need to wait till after (see https://github.com/RetroPie/RetroPie-Setup/pull/1489) - it will require some minor tweaks to work, but I can sort that as needed. How does this differ to https://github.com/miroof/node-virtual-gamepads btw ?

sbidolach:
Many thanks for reply. The reason why I decided to write own mobile gamepad was that in node-virtual-game i couldn't jump diagonally (ex. right-up,  left-up). Without these moves, gamepad is useless in many games. In my opinion also use existing pad layout on mobile is no sense because layout should be fitted to mobile possibilities and this is why I changed buttons positions and added two more. In future I want also use motion sensors which will be using to for view position and see how this will be dealing with games like quake III. I'm using this mobile gamepad for myself and in most games works perfectly so the I thought that I can share with others :)

joolswills:
sounds good thanks!

PR #1494, opened 2016-05-28
Add option to set -append and -use_nointro_name=false flags

sselph:
I've seen this pop up in issues and emails so thought I'd
add the option to tweak these flags.


joolswills:
Thanks!

PR #1513, opened 2016-06-08
scraper: add option to use filenames as ROM name.

### LINKED ISSUE #1506: Option for scraper to use filenames (minus extensions...)
This is a feature of sselph's scraper, can we please have an option to use the filename instead of the current two options of NoIntro and GamesDB?

sselph:
Fixes #1506


joolswills:
thanks.

Issue #1435, opened 2016-04-15
Texas Instruments TI-99 Emulator Support


stevetb:
I would like to put in a request a Texas Instruments TI-99 Emulator to be added into RetroPie.  I found a link to several emulators which may work or already be supported.  ##

http://www.99er.net/emul.shtml

Please let me know if I can do anything to assist in moving this into an experimental build.

Thank you,
Stevetb

HerbFargus:
Already been done on zerojays repo:

https://github.com/zerojay/RetroPie-Extra/blob/master/scriptmodules/emulators/ti99sim.sh

stevetb:
Why is this not added to the main RetroPie Experimental Install Section?  This was a huge system back in the 1980's, seems worthy to me.

HerbFargus:
We are working on some fundamental changes to the menu before we add anymore new systems, but you can easily install the extra modules from zerojay's repository in the meantime: https://github.com/zerojay/RetroPie-Extra

stevetb:
Thank you HerbFargus.  I went through the installer and had some errors tonight.  I will go through the logs and see if it is related to 3.7 and submit to ZeroJay.  I appreciate the help.  Please give the developers a nudge to add if you can :-)

HerbFargus:
I am one of the developers ;)

stevetb:
Be gentle on yourself then, hahaha

HerbFargus:
added as of https://github.com/RetroPie/RetroPie-Setup/pull/1514

PR #1514, opened 2016-06-08
add ti-99/4A

HerbFargus:
tested on both the pc and pi and it seems to be working. I didnt symlink the bios cause it only works in the rom folder but I'm sure we can tweak it as needed after more testing. Thanks to @zerojay for the module

Should address https://github.com/RetroPie/RetroPie-Setup/issues/1435


joolswills:
thanks.

Issue #1511, opened 2016-06-07
Better mupen64plus defaults


loganmc10:
As you are probably aware, many users have issues getting N64 emulation to works properly, here are my suggestions:

Make mupen64plus-GLideN64 the default, not gles2rice. Honestly if it was me, I would get rid of gles2rice, gles2n64, and the videocore plugin. GLideN64 works betters than all those in every game I have tried. Just leave mupen64plus-GLideN64 and lr-mupen64plus.

Make CEA-1 the default resolution for N64 games. If you don't use CEA-1 then mupen64plus is going to try to upscale the image to 1080p (assuming they are using an HDTV for instance). That creates big slowdowns for people.

Set gpu_freq=500 in config.txt (at least for the rpi3). This doesn't void the warranty and seems quite stable, it is also needed for a lot of N64 games to work well.

I would also increase the gpu_mem amount. It looks like it's setting ```gpu_mem_1024``` to 256. I would probably increase that to around 500MB-600MB. GLideN64 incorporates a texture cache, and the default texture cache max size is 500MB, so if someone plays for a long time there is the possibility for that cache to fill up.

joolswills:
@gizmo98 thoughts ?

Regarding resolution - we could try and force that res, but you assume all screens will support it . runcommand does have some code for this to try and pick a working res, but it's not ideal - eg if you are on a small screen etc it may not work at all.

we are definitely not overclocking by default. This should be up to end users.

Increasing GPU mem may upset other users too, I think our current defaults are ok.

joolswills:
I don't want to remove choice/configurability either - this is what RetroPie is all about. You may not think those plugins are needed, but I can assure you there would be great upset by many if they were removed.

loganmc10:
@joolswills all good points.

If you aren't going to increase the gpu_mem then you might want to consider changing the default texture cache size in GLideN64.

in mupen64plus.cfg under ```[Video-GLideN64]``` the default is:
```
# Size of texture cache in megabytes. Good value is VRAM*3/4
CacheSize = 500
```

loganmc10:
Also, in regards to the resolution. There is another way!

I submitted a patch to the SDL developers here:
https://bugzilla.libsdl.org/show_bug.cgi?id=3353

[...]

PR #1518, opened 2016-06-09
Added some defaults for GlideN64 mupen64plus plugin. Do not delete cf…

dankcushions:
…g every install/update.

see https://github.com/RetroPie/RetroPie-Setup/issues/1511
probably needs some changes as per my comment there  :)

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -125,19 +125,53 @@ function configure_mupen64plus() {
     cp "$scriptdir/scriptmodules/$md_type/$md_id/mupen64plus.sh" "$md_inst/bin/"
     chmod +x "$md_inst/bin/mupen64plus.sh"
 
-    # to solve startup problems delete old config file
-    rm -f "$md_conf_root/n64/mupen64plus.cfg"
-    # remove default InputAutoConfig.ini. inputconfigscript writes a clean file
-    rm -f "$md_inst/share/mupen64plus/InputAutoCfg.ini"
     mkUserDir "$md_conf_root/n64/"
-    # Copy config files
-    cp -v "$md_inst/share/mupen64plus/"{*.ini,font.ttf,*.conf} "$md_conf_root/n64/"
-    su "$user" -c "$md_inst/bin/mupen64plus --configdir $md_conf_root/n64 --datadir $md_conf_root/n64"
 
-    iniConfig " = " '"' "$md_conf_root/n64/mupen64plus.cfg"
+    # Copy config files
+    cp -v "$md_inst/share/mupen64plus/"{*.ini,font.ttf} "$md_conf_root/n64/"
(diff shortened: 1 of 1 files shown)


scriptmodules/emulators/mupen64plus.sh
```diff
@@ -125,19 +125,37 @@ function configure_mupen64plus() {
     cp "$scriptdir/scriptmodules/$md_type/$md_id/mupen64plus.sh" "$md_inst/bin/"
     chmod +x "$md_inst/bin/mupen64plus.sh"
 
-    # to solve startup problems delete old config file
-    rm -f "$md_conf_root/n64/mupen64plus.cfg"
+    # if the user has an existing mupen64plus config we will not overwrite it, but instead copy the
+    # default configuration to mupen64plus.cfg.rp-dist so any new options can be manually
+    # copied across as needed without destroying users changes
```
joolswills: please add `local config`

joolswills:
Looks good apart from the missing local config - please add that and I will merge. thanks.

joolswills:
@gizmo98 just want to make sure this is ok as you wrote this part - running mupen64plus creates the initial config, or do we copy it from the intall folder or both ? - just wondering if that will interfere with the rp-dist changes.

dankcushions:
> Looks good apart from the missing local config - please add that and I will merge. thanks.

done :)

joolswills:
@dankcushions thanks. I think there could be another problem, but I've not tested it - just from looking at the code. I believe running the mupen64plus creates an initial config - in which case, the rp-dist wont actually be a default config, but just contain some parts in the case that a default config already exists - please can you check that ?

gizmo98:
Sorry. I cannot test any changes at the moment. I'm on vacation, i have no pi and the internet connection is not the best. Actually i dislike rpi related defaults here. Just ask gonetz to add rpi related defaults here:
https://github.com/gonetz/GLideN64/blob/master/src/Config.cpp
Use `#ifdef VC` (VC = videcore) for raspberry pi stuff. There is an example in the current source code.

joolswills:
that's a good point. and actually yeh if we are adding defaults here we need to do them only for the rpi

dankcushions:
Squashed some commits following some changes suggested by @joolswills. I agree - I think these defaults would be better moved to the GLideN64 repo, but it's moving so fast with RPI-related changes that maybe we should do that at a later stage? I think the ones here so far aren't really that significant - was hoping others would suggest more :)

loganmc10:
I disagree with having the rpi defaults baked into GlideN64, the bilinear mode thing is just a preference. The cache size at 192 is also just a preference, the rpi can handle the default of 500, I have gpu_mem at 600 on my rpi and I use the default 500MB cache

joolswills:
there will always be people who disagree with the defaults - but you can still change them. For now I'm happy with our default config though (which you are free to change). If people want to push some changes upstream, that can be discussed with the author.

PR #1520, opened 2016-06-10
help improvements

joolswills:
* get rid of _INFMSGS and instead have a rp_module_help field with information in.
   this can then be displayed in the setup menus, and not just after configuring.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -117,6 +117,11 @@ function package_setup() {
             options+=(X "Remove")
         fi
 
+        local help="${__mod_desc[$idx]}\n\n${__mod_help[$idx]}"
+        if [[ -n "$help" ]]; then
+            options+=(H "Package Help")
+        fi
+
         cmd=(dialog --backtitle "$__backtitle" --cancel-label "Back" --menu "Choose an option for ${__mod_id[$idx]}" 22 76 16)
         choice=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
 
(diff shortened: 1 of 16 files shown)


joolswills:
@HerbFargus if you want a not very fun documentation job - you could now improve the internal documentation for each module via rp_module_help. Not intended to replace the wiki, but just to give the basics (eg rom locations).

joolswills:
No pressure though - completely optional ;-) Cheers.

HerbFargus:
Sounds fun :)

HerbFargus:
Just so I'm consistent with the docs would a good template be:

```
Copy your <emulator> roms to $romdir/<folder>.  

Copy the <BIOS name> to $biosdir
```

joolswills:
sounds good to me - and then any extra info the users may need. note that only global variables like $romdir are usable in the help variable - not things like $md_id (due to the point at which they are processed).

HerbFargus:
Yeah I figure I'll just hardcode in the rom folder names anyways as they don't match the module Id anyhow. In terms of line breaks do I need to keep them on the same line or can I add \n breaks?

joolswills:
I think \n should work - please check. if you think it would be best split, that's ok too. bear in mind this all has to be maintained, so I think a summary is best, with the rest on the wiki (ie not too verbose).

joolswills:
and thanks! sorry for suggesting the worst jobs for you ;-)

HerbFargus:
Yep I'll just stick with rom paths and bios if applicable. Anything more than that can be referred to the wiki. Thanks

PR #1521, opened 2016-06-10
Help docs update

HerbFargus:
I left off  most of the ports for the time being as they mostly don't require any special instructions and the ones that do are a little more extensive than a few lines. this should cover all the libretro cores and standalone emulators.

I haven't yet tested this so I will do so promptly once I get a little time with my pi, but feel free to verify as well.

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -11,6 +11,7 @@
 
 rp_module_id="advmame"
 rp_module_desc="AdvanceMAME"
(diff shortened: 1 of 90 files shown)


HerbFargus:
seems to be working fine

HerbFargus:
hold off on a merge, I'm probably going to put the accepted extensions while I'm at it as I realised there is more space for stuff than I had anticipated.

HerbFargus:
Alright I think its complete now. All Libretro Cores and Emulators should be accounted for. Each help doc covers:
- ROM extensions
- ROM folder
- BIOS file and folder if applicable

joolswills:
awesome thanks. Ill have a quick read through and merge a bit later.

PR #1522, opened 2016-06-11
Game data

joolswills:
Changes to split off game data install - and so it only installs in "install" mode - as we call the configure in remove mode to remove packages. Also corrects a few issues I found along the way and a few cleanups.

### DIFF
--- a/scriptmodules/emulators/frotz.sh
+++ b/scriptmodules/emulators/frotz.sh
@@ -23,21 +23,26 @@ function remove_frotz() {
     aptRemove frotz
 }
 
+function game_data_frotz() {
+    if [[ ! -f "$romdir/zmachine/zork1.dat" ]]; then
+        cd "$__tmpdir"
+        local file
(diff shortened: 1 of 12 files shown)


HerbFargus:
I presume this primarily for ports that have external game data applied. Would it also apply to things like oric that have the demos in with the repo or is that a different case?

joolswills:
Yep - ill do that in a mo thanks for the heads up.

Anything that has external data installed at the configure stage needs doing - or else when you uninstall it will technically re-install the data again. plus it makes the configure functions cleaner and enables you to just manually install game data from retropie_packages.sh (we can then use it for stuff later to easily re-install game data)

joolswills:
oricutron is slightly less important than the others - since it just copies from md_inst - rather than downloading large data from the internet

HerbFargus:
Yeah I i figured it was primarily for external datasets. I'll give this a test

joolswills:
thanks.

PR #1529, opened 2016-06-14
pifba - added initial input configuration script (player 1 only)

joolswills:


HerbFargus:
seems to be working, using my xbox controller with xboxdrv, for cadillacs and dinosaurs A=punch X=jump which I'm guessing is the correct mapping. I'm used to A and B but I think thats something I usually manually change rather than a configuration error.

joolswills:
thanks.

PR #1530, opened 2016-06-14
Create directories for lr-fba-next and lr-mame2003. lr-mame2003 does …

dankcushions:
…this on-demand, but I think it's worth creating them on our own terms, and also it lets the user know the structure ahead of a game looking for it.

Tested on a new and existing install.. seems to behave!

### DIFF
--- a/scriptmodules/libretrocores/lr-fba-next.sh
+++ b/scriptmodules/libretrocores/lr-fba-next.sh
@@ -53,6 +53,7 @@ function configure_lr-fba-next() {
     ensureSystemretroconfig "neogeo"
 
     # Create samples directory
+    mkUserDir "$biosdir/fba/"
     mkUserDir "$biosdir/fba/samples/"
 
     # Set core options
--- a/scriptmodules/libretrocores/lr-fba-next.sh
+++ b/scriptmodules/libretrocores/lr-fba-next.sh
@@ -52,6 +52,9 @@ function configure_lr-fba-next() {
     ensureSystemretroconfig "fba"
     ensureSystemretroconfig "neogeo"
 
+    # Create samples directory
+    mkUserDir "$biosdir/fba/samples/"
+
     # Set core options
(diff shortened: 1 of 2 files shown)


scriptmodules/libretrocores/lr-fba-next.sh
```diff
@@ -52,6 +52,9 @@ function configure_lr-fba-next() {
     ensureSystemretroconfig "fba"
     ensureSystemretroconfig "neogeo"
 
+    # Create samples directory
+    mkUserDir "$biosdir/fba/samples/"
```
joolswills: mkUser dir will only chown the lowest directly, so as fba may not exist you need to create that first, or chown "$biosdir/fba" manually.

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -49,10 +49,24 @@ function configure_lr-mame2003() {
...
+        ensureSystemretroconfig "$mame_dir"
+
+        mkUserDir "$romdir/$mame_dir/mame2003/cfg/"
+        mkUserDir "$romdir/$mame_dir/mame2003/ctrlr/"
+        mkUserDir "$romdir/$mame_dir/mame2003/diff/"
+        mkUserDir "$romdir/$mame_dir/mame2003/hi/"
+        mkUserDir "$romdir/$mame_dir/mame2003/inp/"
+        mkUserDir "$romdir/$mame_dir/mame2003/memcard/"
+        mkUserDir "$romdir/$mame_dir/mame2003/nvram/"
+        mkUserDir "$romdir/$mame_dir/mame2003/snap/"
```
joolswills: another for loop would be cleaner . eg 

```
local dir
for dir in cfg ctrlr diff hi inp memcard nvram snap; do
```

joolswills: can use mkRomDir `$mame_dir/mame2003/$dir/"`also instead of mkUserDir

PR #1531, opened 2016-06-15
add fds support for lr-fceumm

HerbFargus:
from the forum: https://retropie.org.uk/forum/topic/1981/famicom-disk-system-fds-and-lr-fceumm

I'm also not sure if I got the numbers right on the addSystem function. couldnt remember which was default.

### DIFF
--- a/scriptmodules/libretrocores/lr-fceumm.sh
+++ b/scriptmodules/libretrocores/lr-fceumm.sh
@@ -11,7 +11,7 @@
 
 rp_module_id="lr-fceumm"
 rp_module_desc="NES emu - FCEUmm port for libretro"
-rp_module_help="ROM Extensions: .nes .zip\n\nCopy your NES roms to $romdir/nes"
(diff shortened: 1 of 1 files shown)


scriptmodules/libretrocores/lr-fceumm.sh
```diff
@@ -41,8 +41,11 @@ function configure_lr-fceumm() {
...
     mkRomDir "nes"
+    mkRomDir "fds"
     ensureSystemretroconfig "nes"
+    ensureSystemretroconfig "fds"
 
     delSystem "lr-fceu-next" "nes"
-    addSystem 1 "$md_id" "nes" "$md_inst/fceumm_libretro.so"
+    addSystem 0 "$md_id" "nes" "$md_inst/fceumm_libretro.so"
+    addSystem 1 "$md_id" "fds" "$md_inst/fceumm_libretro.so"
 }
```
joolswills: addSystem 0 means it's not set as default - addSystem 1 means it is (Although existing defaults are never changed)

So in the case above you have made it so fceumm is no longer the default nes emulator which is wrong. It could be the default fds emulator if it is tested to be better than the other.

HerbFargus:
I see. well I've changed it back, and left fds as default on nestopia for now until we can see if its any better.

PR #1534, opened 2016-06-16
initial integration of image building into retropie-setup

joolswills:
This is a WIP. Just for reference. Still untested.


joolswills:
This is working now, although the created images need some further testing. @HerbFargus I will merge it and build some images for testing.

HerbFargus:
Sounds good to me

PR #1546, opened 2016-06-24
adding joystick selection for RetroArch players 1-4

meleu:
This module shows the available joysticks and let the user choose what controller to use for RetroArch Player 1-4.

Short description of how it does its job:
- puts at the beginning of $configdir/all/retroarch.cfg an "#include" pointing to the file $configdir/all/input-selection.cfg
- let the user manage the input-selection.cfg through dialogs (this file contains the configs for input_playerN_joypad_index for players 1-4).

OBS. 1: the joystick selection doesn't work if the config_save_on_exit is set to true, because the retroarch.cfg is overwritten frequently.
OBS. 2: This program relies on the output of the jslist program to get the available joysticks and their respective indexes. This program will be put at /opt/retropie/supplementary/jslist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+# Use new trusty images, should yield newer compilers and packages
+sudo: required
+dist: precise
+language: cpp
+
+matrix:
+  include:
+    - compiler: gcc
+      addons:
+        apt:
+          sources:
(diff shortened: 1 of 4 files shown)


meleu:
ooops! I did a Pull Request in the wrong branch...

PR #1554, opened 2016-06-29
Moved all advmame dirs to the right place. Turned smp on for multicor…

dankcushions:
…s.

the smp option probably doesn't do anything massively useful. It just threads the GPU stuff, which according to their docs is only useful with lots of post-processing/scaling. Still, might aswell :)

Some caveats:
- if a user has an existing install, this won't do anything because iniSet doesn't change existing settings (at least, that's my understanding?), so all their existing files should be reachable as usual if they have an existing install.
- things get a bit confusing with the dual directories in the .rc - from my testing it will read from both, but will only save into the first one (in this case /roms/mame-advmame/*/ as it is before /roms/arcade/advmame*/ in the .rc). Maybe it would be better to setup symlinks so that arcade sub folders are actually symlinked to the mame-advmame master?

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -164,7 +164,18 @@ function install_advmame() {
 
 function configure_advmame() {
     mkRomDir "arcade"
-    mkRomDir "mame-advmame"
+
+    local mame_dir
+    for mame_dir in arcade/advmame mame-advmame
+    do
+        mkRomDir "$mame_dir"
+
+        local mame_sub_dir
+        for mame_sub_dir in artwork diff hi inp memcard nvram sample snap sta
+        do
+            mkRomDir "$mame_dir/$mame_sub_dir"
+        done
+    done
 
     # delete old install files
     rm -rf "$md_inst/"{bin,man,share}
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - 

iniSet will change existing settings - it adds new or updates old.

Have you tested the smp option ? the docs suggest it may need to do more work in some cases ?

symlinks would probably be better - means users don't have to do anything if they change their mind (arcade should symlink to mame-advmame). it also would simplify the config.

dankcushions:
> Have you tested the smp option ? the docs suggest it may need to do more work in some cases ?

it didn't seem to make a measurable difference in a game that was slowing down before, and i don't have a way of testing ALL mame games to see if there's a regression, so i'll just take it out. more fun to code than probably useful :)

> symlinks would probably be better

done! plz check if the order is the right way around as i wasn't sure what way you mean't - i've put the 'master' in mame-advmame, and the symlink in arcade. also, i am unable to test this as my rom drive is fat32. i'm pretty sure it should work, though.

If all good, I'll squash.

joolswills:
Looks good thanks, ill try it out.

dankcushions:
yikes - something up with this: https://retropie.org.uk/forum/topic/2559/retropie-not-booting-after-total-update

i can't test the symbolic link stuff right now but i don't see what's wrong:
```
    local mame_sub_dir
    for mame_sub_dir in artwork diff hi inp memcard nvram sample snap sta
    do
         mkRomDir "mame-advmame/$mame_sub_dir"
         ln -sf "$romdir/mame-advmame/$mame_sub_dir" "$romdir/arcade/advmame/$mame_sub_dir"
    done
```

joolswills:
I dont think i tested it. will check.

joolswills:
I think I see the problem.

joolswills:
This should be fixed now. Anyone who has the broken structure will need to manually remove it and re-install.

joolswills:
I added a workaround to fix up broken links, and fixed up some indent issues.

PR #1561, opened 2016-07-04
Support multiple device IDs when auto-mapping controllers.

MisterTea:
The script that auto-configures retroarch assumes that there's only one inputconfig.  In my case, I have two different usb joysticks that I'm using, so there are two entries.  With the current script, it concatenates the scripts and is not desired behavior.  With this pull request, it loops through the configs and handles each inputconfig correctly.

### DIFF
--- a/scriptmodules/supplementary/emulationstation/inputconfiguration.sh
+++ b/scriptmodules/supplementary/emulationstation/inputconfiguration.sh
@@ -65,56 +65,58 @@ function inputconfiguration() {
     # check if we have the temporary input file
     [[ ! -f "$es_conf" ]] && return
 
-    local line
-    while read line; do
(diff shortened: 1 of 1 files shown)


joolswills:
I thought that the temporaryinput.cfg which it works off only has a single entry for the last configured controller though ? Unless I am misunderstanding your change ?

joolswills:
Please make sure you read the "guidelines for contributing" above before opening an issue. Support questions should go to the forum. Please post to the forum first, as many "issues" can be solved with configuration changes. Please remove this text before posting.

MisterTea:
Hey, I haven't had a chance to look at this yet but I'll take a look later
this week.

joolswills:
Closing due to no activity. Can be re-opened if needed.

PR #1563, opened 2016-07-07
Optimize mupen64plus

gizmo98:
-Default render resolution is 320x240 for best performance. Use dispmanx to upscale video.
-Use GlideN64 as default video plugin.
-Enable GlideN64 optimizations (NEON assembler).
-Update GlideN64 configVersion.
-Disable GlideN64 FBEmulation. I don't know why but games like MK64 and Castlevania have visual glitches since this morning. Mupen64plus-Core repo was updated today. There could be a correlation. @loganmc10 do you have visual issues with the newest Mupen64plus-Core as well?

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -73,7 +73,7 @@ function build_mupen64plus() {
     # build GLideN64
     $md_build/GLideN64/src/getRevision.sh
     pushd $md_build/GLideN64/projects/cmake
-    cmake -DMUPENPLUSAPI=On ../../src/
+    cmake -DMUPENPLUSAPI=On -DOPT ../../src/
     make
     popd
 
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -116,12 +116,13 @@ function install_mupen64plus() {
 }
 
 function configure_mupen64plus() {
-    addSystem 0 "${md_id}-GLideN64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM%"
     if isPlatform "rpi"; then
-        addSystem 1 "${md_id}-gles2rice" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-rice %ROM%"
+        addSystem 1 "${md_id}-GLideN64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM%"
+        addSystem 0 "${md_id}-gles2rice" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-rice %ROM%"
         addSystem 0 "${md_id}-gles2n64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-n64 %ROM%"
         addSystem 0 "${md_id}-videocore" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-videocore %ROM%"
     else
+        addSystem 0 "${md_id}-GLideN64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM%"
         addSystem 1 "${md_id}-glide64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-glide64mk2 %ROM%"
     fi
 
@@ -156,7 +157,7 @@ function configure_mupen64plus() {
         su "$user" -c "$cmd"
     fi
 
-    iniConfig " = " '"' "$config"
+    iniConfig " = " "" "$config"
     iniSet "ScreenshotPath" "$romdir/n64"
     iniSet "SaveStatePath" "$romdir/n64"
     iniSet "SaveSRAMPath" "$romdir/n64"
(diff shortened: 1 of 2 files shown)


loganmc10:
@gizmo98 Can you be more specific? What do you mean visual issues?

To get Framebuffer Emulation working on the Raspberry Pi I had to do this "hack":
https://github.com/gonetz/GLideN64/blob/master/src/FrameBuffer.cpp#L710-L713

If that is not there you get visual glitches like this:
https://cloud.githubusercontent.com/assets/848146/15656484/fe0715ea-2664-11e6-9f9f-f08231583081.png

However, I still get issues sometimes. I'll call it "ghosts"? Basically sometimes when I play GoldenEye all of a sudden that problem will reappear, and I will see the sky everywhere. In Pokemon Stadium I'll sometimes see the outline of a pokemon where it shouldn't be, etc...

You can see some of the discussion around getting Framebuffer Emulation working here:
https://github.com/gonetz/GLideN64/issues/994
And the discussion about these strange graphic issues here:
https://github.com/gonetz/GLideN64/issues/1011
And the commit that enabled FBE on the Raspberry Pi here:
https://github.com/gonetz/GLideN64/commit/1cbbea81830978e73fedd3ab7f6468a4ecd88b0d


So basically Framebuffer Emulation is still  not perfect, but it's not a black screen either, so that's something...

gizmo98:
MK64 is broken if FBEmulation is enabled. This does not happen yesterday. After i rebuild everything today this happens:
dsc01626
dsc01627
Even if i user older GlideN64 source code this happens now.

loganmc10:
Interesting, I'll look into that in a bit

loganmc10:
I can confirm I get the same graphical glitches, but building the mupen64plus-core from the commit in April doesn't help, neither does an earlier version of GLideN64. I can guarantee you this is the same issue I was describing. I am able to fix it by adding glFinish() to the saveBuffer method in GLideN64 as described here:

https://github.com/gonetz/GLideN64/issues/1011#issuecomment-227951422

Which means it is the same issue. I don't think I ever tested MK64 with Framebuffer Emulation since I submitted that patch. I normally leave it disabled unless a game needs it (like Pokemon Stadium or Mario Golf).

loganmc10:
I'm kind of glad you found this actually. I haven't been working on fixing it since it only happened sporadically for me before, this happens 100% of the time for me so it will be much easier to debug.

gizmo98:
It's strange it did not happen yesterday. Zelda OoT menu screen rendered fine without any delay, so FBEmulation was enabled. I think it is a good idea to keep it disabled until there is a fix.

dankcushions:
this is good! however i just downloaded your fork and attempted a source rebuild of mupen64plus:
```
    CXX _obj/VCConfig.o
    CXX _obj/VCDebugger.o
../../src/VCConfig.cpp:35:12: warning: ‘int VCConfig_GetInt(const toml::Value&, const char*, int)’ defined but not used [-Wunused-function]
 static int VCConfig_GetInt(const toml::Value &topValue, const char *name, int defaultValue) {
            ^
../../src/VCConfig.cpp:59:14: warning: ‘char* VCConfig_GetString(const toml::Value&, const char*, const char*)’ defined but not used [-Wunused-function]
 static char *VCConfig_GetString(const toml::Value &topValue,
              ^
    CXX _obj/VCRenderer.o
    CXX _obj/VCUtils.o
    CXX _obj/VI.o
    LD  mupen64plus-video-videocore.so
make: Leaving directory '/home/pi/work/RetroPie-Setup/tmp/build/mupen64plus/mupen64plus-video-videocore/projects/unix'
current revision "b71705f"
/home/pi/work/RetroPie-Setup/tmp/build/mupen64plus/GLideN64/projects/cmake /home/pi/work/RetroPie-Setup/tmp/build/mupen64plus /home/pi/work/RetroPie-Setup
Parse error in command line argument: -DOPT
Should be: VAR:type=value
CMake Error: No cmake script provided.
CMake Error: Problem processing arguments. Aborting.

make: *** No targets specified and no makefile found.  Stop.
/home/pi/work/RetroPie-Setup/tmp/build/mupen64plus /home/pi/work/RetroPie-Setup
Removing additional swap
/home/pi/work/RetroPie-Setup
Could not successfully build N64 emulator MUPEN64Plus (GLideN64/projects/cmake/plugin/release/mupen64plus-video-GLideN64.so not found).

Log ended at: Thu  7 Jul 21:22:02 BST 2016
Total running time: 0 hours, 9 mins, 25 secs
```

loganmc10:
Ya I think it should be -OPT=On, not just -OPT

loganmc10:
The FBE issues should be fixed with a pull request I sent to GLideN64:

https://github.com/gonetz/GLideN64/pull/1055

dankcushions:
> Ya I think it should be -OPT=On, not just -OPT

that fixes it! i now have a successful build, and i've done a quick test of this PR and it looks good to me.

loganmc10:
One of semi-related thing, mupen64plus shouldn't' depend on mesa on the rpi:

https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/mupen64plus.sh#L19

I'm not really sure how RetroPie handles dependencies and whatnot, but libgl1-mesa-dev shouldn't be a requirement.

gizmo98:
@dankcushions 
Oops. ;-)

@loganmc10 
Thank you for fixing this bug so fast. If gonetz adds your commit FB emulation can be enabled. I will do a test if libgl1-mesa-dev is still necessary. As far as i know you have fixed GlideN64's cmake file which required gl headers.

joolswills:
Thanks - is this ready to be merged or are we waiting on anything ?

loganmc10:
The fix for framebuffer emulation was merged so @gizmo98 as long as your test of Mario Kart 64 is noe successful you should be able to enable framebuffer emulation before this is merged

gizmo98:
I think this can be merged now. I do not use DOPT anymore . Zelda OoTs Navi dislikes this option.
dsc01628

Some games (for example Castlevania) dislike FBEmulation as you can see here:
dsc01631

And GlideN64 seems to have a new problem with Smash Brothers menu screen (FBEMU on and off):
dsc01634

In spite of everything it is the best video plugin at the moment.

joolswills:
Thanks very much for this. Do I need to rebuilt the binaries ?

gizmo98:
Yes. You need to rebuild binaries. GlideN64 changed a lot. If an old version of this video plugin is used the screen stays black if FBEmulation is enabled.

joolswills:
binaries have been rebuilt. Cheers!

PR #1564, opened 2016-07-09
Update mupen64plus Pt.4

gizmo98:
-Use bilinear filtering standard mode to fix super smash brothers and
mario kart texture corruption. https://github.com/RetroPie/RetroPie-Setup/pull/1563#issuecomment-231420842
-Disable audio OMX audio output auto configuration. We use SDL audio
now.
-Disable FBEmulation for now. Conker BFD, Bomberman, Castlevania, Banjo
Kazooie, Super Smash Brothers, Goldeneye and Rayman 2 have extreme visual corruptions if this feature is
enabled. Need to find a way to enable it for Zelda games and Star Wars - SOTE. 
-Use native resolution if FBEmulation is enabled.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -171,20 +171,18 @@ function configure_mupen64plus() {
         # Settings version. Don't touch it.
         iniSet "configVersion" "11"
         # Bilinear filtering mode (0=N64 3point, 1=standard)
-        iniSet "bilinearMode" "0"
+        iniSet "bilinearMode" "1"
         # Size of texture cache in megabytes. Good value is VRAM*3/4
         iniSet "CacheSize" "192"
         # Disable FB emulation until visual issues are sorted out
-        iniSet "EnableFBEmulation" "True"
+        iniSet "EnableFBEmulation" "False"
+        # Use native res
+        iniSet "nativeResFactor" "1"
     fi
 
     chown -R $user:$user "$md_conf_root/n64"
 
-    if isPlatform "rpi"; then
-        addAutoConf mupen64plus_audio 1
-    else
-        addAutoConf mupen64plus_audio 0
-    fi
+    addAutoConf mupen64plus_audio 0
     addAutoConf mupen64plus_hotkeys 1
     addAutoConf mupen64plus_compatibility_check 1
 }


loganmc10:
@gizmo98 It does look like using standard bi-linear filtering gets rid of those "green" graphical glitches, good find.

I'm curious, why don't you use audio-omx anymore? I use that, the only issue I've had with it is that savestates don't always work.

Also, what issues are you having with FBEmulation? I just tested GoldenEye and Super Smash with FBE turned on and everything looks perfect to me.

gizmo98:
@loganmc10 
I do not use audio omx anymore because it does not respect system volume (always 100%) and it does not respect alsa audio output configuration. Our audio output autoconf works more or less. There were always some users complaining about no sound. Audio sdl always uses the right audio output path.

loganmc10:
Those graphical glitches look like the issues that occurred before I submitted the latest patch to GLideN64 that fixed the issues with Mario Kart 64, I'll have to look into that more, you're sure it's using the latest version of GLideN64?

Also i don't think mupen64plus uses GLideN64.custom.ini at all, so something else is going on there

loganmc10:
So I can duplicate some of those issues. Bomberman, Smash Bros, and GoldenEye work fine for me. The menus in Castlevania and Rayman 2 are messed up. It probably is best to disable FBE by default until these issues get sorted.

gizmo98:
Mario Kart works fine but Goldeneye does not. Keep in mind i do not use -DOPT at the moment. I will do a test which optimization is not zelda fairy compatible. https://github.com/gizmo98/GLideN64/commit/b3bc8ce689cbf93ca399c6be0781a7d4651e0173

gizmo98:
@loganmc10 
The new CRC optimization is not zelda fairy compatible. ;-)

loganmc10:
Ok I've created a pull request to fix some of those issues. It seems to be pretty good now, but you guys might want to come up with some kind of system to determine when to use Framebuffer Emulation, since it does cause a performance hit.

gizmo98:
I have written already something. I just wait if my GLideN64 PRs are merged so i can add NEON and VEC4 OPT and the game related FBEmulation enable mechanism.

Jools will kill me if i came up every day with a new PR ;-) 

Is Majora's Mask running with your new patch? With the old source it is not working correct w/wo FB emulation. I have already written a rice fallback.

loganmc10:
What didn't work correctly? It looks good to me, this is with FBE enabled;
snapshot
snapshot2

gizmo98:
If FBEmu is disabled I cannot see the transformation into a deku boy. 
If FBEmu is enabled everything flickers.

loganmc10:
Hmm well I didn't play very far into the game but with FBE on I didn't see any flickering

gizmo98:
Ok. I deleted my old config file and flickering is gone. There are still some games with depth problems even with your new patch. Start Conker BFD and start a game. In the first level you can see the waterfall through the ground. Play Majora's Mask a little bit longer. If link falls into the abysm the whole horror kid/deku boy animation seems to be hidden behind a black rect.

loganmc10:
Ok I look into it over the next few days. It's probably going to be hard to fix it 100% since it is some kind of GPU driver bug.

PR #1572, opened 2016-07-14
minor cosmetic tweak to bash welcome

HerbFargus:
### Original
bash_original
### New Changes
bash_1

### DIFF
--- a/scriptmodules/supplementary/bashwelcometweak.sh
+++ b/scriptmodules/supplementary/bashwelcometweak.sh
@@ -71,7 +71,6 @@ function retropie_welcome() {
     local bfgwht="${bld}$(tput setaf 7)"
 
     local logo=(
-        "           "
         "${fgred}   .***.   "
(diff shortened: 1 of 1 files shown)


joolswills:
I'm not sure - I did actually try increasing the size of the bottom, and I thought it looks the wrong shape then. I'll ask rook.

joolswills:
Unless it was supposed to have that, and the bottom was cut off in the ascii I file I had

joolswills:
(or maybe with `"` chars so it's not as tall)

joolswills:
eg

bash

ill get some feedback from the ascii artist though.

HerbFargus:
yeah either way, it just seemed a little squatty without that last line (whichever characters we end up using) 

I apologise for my recent absence, in the process of moving/ new jobs etc.

joolswills:
No worries - it's just a hobby - hope all goes well with your move/job :)

PR #1576, opened 2016-07-16
8bitdo mapping improvements

joolswills:
* add autoconf.cfg parameter 8bitdo_hack (default 1) to decide whether to add the offset to the inputs for retroarch - which is required on the older firmware
 * make sure autoconf.cfg is owned by $user - it would be created as root previously
 * add setAutoConf function
 * cleanup retroarch config generation code

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -156,6 +156,9 @@ function configure_retroarch() {
 
     copyDefaultConfig "$config" "$configdir/all/retroarch.cfg"
     rm "$config"
+
+    # remapping hack for old 8bitdo firmware
+    addAutoConf "8bitdo_hack" 1
 }
 
 function gui_retroarch() {
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -252,11 +252,20 @@ function gitPullOrClone() {
 
 function ensureRootdirExists() {
     mkdir -p "$rootdir"
+
     # make sure we have inifuncs.sh in place and that it is up to date
     mkdir -p "$rootdir/lib"
     if [[ ! -f "$rootdir/lib/inifuncs.sh" || "$rootdir/lib/inifuncs.sh" -ot "$scriptdir/scriptmodules/inifuncs.sh" ]]; then
         cp --preserve=timestamps "$scriptdir/scriptmodules/inifuncs.sh" "$rootdir/lib/inifuncs.sh"
     fi
+
+    # create template for autoconf.cfg and make sure it is owned by $user
+    local config="$configdir/all/autoconf.cfg"
+    if [[ ! -f "$config" ]]; then
+        echo "# this file can be used to enable/disable retropie autoconfiguration features" >"$config"
+    fi
+    chown $user:$user "$config"
+
     mkUserDir "$datadir"
     mkUserDir "$configdir"
     mkUserDir "$configdir/all"
--- a/scriptmodules/inifuncs.sh
+++ b/scriptmodules/inifuncs.sh
@@ -108,16 +108,22 @@ function addAutoConf() {
        default="1"
     fi
 
-    if [ ! -f "$file" ]; then
-        echo "# this file can be used to enable/disable retropie autoconfiguration features" >> "$file"
-    fi
-
     iniConfig " = " "" "$file"
     iniGet "$key"
     ini_value="${ini_value// /}"
     if [[ -z "$ini_value" ]]; then
         iniSet "$key" "$default"
     fi
+}
+
+# arg 1: key, arg 2: value
+function setAutoConf() {
+    local key="$1"
+    local value="$2"
+    local file="$configdir/all/autoconf.cfg"
+
+    iniConfig " = " "" "$file"
+    iniSet "$key" "$value"
 }
 
 # arg 1: key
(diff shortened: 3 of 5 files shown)


joolswills:
@biscuits99 - if you have the time I would be very interested in this code being tested

to test it out on an existing install

```
git checkout -b joolswills-8bitdo_hack master
git pull https://github.com/joolswills/RetroPie-Setup.git 8bitdo_hack
```

to go back to main retropie-setup again do `git checkout master`

to test the functionality, you need to re-install emulationstation from binary to get the latest input configuration scripts, and it's worth reinstalling retroarch also from binary, so the hack is set to on by default (although you can just set it from the bluetooth settings also).

you can switch the "8bitdo hack" on / off via the bluetooth configuration. You then need to re-input your device configuration into emulationstation so the config gets regenerated.

with the new code, it will not do any workarounds if the hack is "off" which is needed to work on my limited edition FC30 Pro which I assume have a new firmware. I need to do some more testing myself, and will update my SFC30 which currently needs the hack on.

biscuits99:
Sure, I tried it out with my FC30 Pro. I can try with a SFC30 if you need it

`RPi3`
`USB - Keyboard`
`USB - iBuffallo`

```
Wrote 3.8.1 image
Added iBuffalo via ES setup
```

```
Quit ES to CLI
SSH with PuTTY
cd RetroPie-Setup
git checkout -b joolswills-8bitdo_hack master
git pull https://github.com/joolswills/RetroPie-Setup.git 8bitdo_hack
Updated ES from binary
Updated RetrArch from binary
```

```
Registered bluetooth FC30 Pro (hack on) (1.68)
Added udev rule
Powered FC30 Pro on (always using just the "power" button)
Rebooted - didnt auto connect
Changed connect mode to "boot"
Rebooted - connected fine
Added FC30Pro in ES (no issues but misses the top left/top right shoulder as usual)
[fc30pro.zip](https://github.com/RetroPie/RetroPie-Setup/files/367416/fc30pro.zip)
Controller works fine in RetroArch games including hotkeys etc..
```

```
Turned hack off
Regenerated controller config in ES
Button mapping incorrect - not usuable in game
```

```
Upgraded FC30Pro to 1.69
Rebooted
Wouldnt auto connect
Removed old bluetooth connection, re-added it again
Rebooted
Connected fine
Controller works fine in RetroArch games including hotkeys etc..
```

All config files attached

joolswills:
excellent - thanks - so apart from the shoulder button issue, all works as expected? (Just to make sure I have correctly interpreted results).

(I haven't changed the background connect in this branch, although I made some changes the other day, to stop it trying to connect when a pad was already connected - were you using the background connect or?)

biscuits99:
Yes, it worked very well. 1.68 required hack, 1.69 didnt.
Once I was on 1.69, the configs were perfect with hack turned off.

I had the connect set to "boot", I didnt try the background connect.
It just seemed to get confused (wouldnt connect on boot) when I updated the firmware, and wouldn't reconnect unless I removed the registration and re-registered the controller - but then given the firmware change maybe thats fair.

I think this fix will help a lot of people as they will probably ship the 8bitdo controllers with the newer (non-beta) firmware now that requires hack turned off. So I presume that would be the best default?

biscuits99:
Looks like SFC30 still needs hack on though.

```
Added SFC30 (Firmware 2.65) in Bluetooth (hack off)
Added udev rule
Rebooted and it connected
Created ES input configs
Didnt work in game (as expected as I had hack off)
```

```
Updated to firmware 2.67
```

```
Restarted - wouldnt reconnect
Removed device - Readded it - Rebooted
Connected fine on boot
Back in game, mapping still incorrect.
```

```
Rebooted
Recreated ES controller input with SFC30 (produced same config)
Back in game, mapping incorrect.
```

```
Changed setting to hack on
Recreated ES input configs
Everything perfect.
```

Files attached
[SFC30.zip](https://github.com/RetroPie/RetroPie-Setup/files/367577/SFC30.zip)

joolswills:
I thought about having it off by default, but I suspect there are more old controllers out there currently, that are not updated - and if we set it on by default, it could break existing setups. I think we could change  the default in the future perhaps and ask people to upgrade their firmware.

joolswills:
thanks very much for all the testing.

biscuits99:
Yep - I thought the SFC30 would also be fine with hack off, but as it isnt sounds good to keep hack on.

So by default with hack on the SFC30 should work with latest (and earlier) official firmware.
If someone has latest firmware (1.69+) for FC30Pro/NES30Pro they flip the hack off.

joolswills:
Yeah.

I'll mention it in the release text - might be useful if someone with awesome youtube skills covered it in a video ;-)

jravetch:
@joolswills @biscuits99 So when I used my SNES30 controller with 2.68Beta5 firmware, ES was fine, but Retroarch and all the games only the D-Pad worked.  Only by downgrading to 2.67 stable did the controllers work again in Retroarch after unpairing, pairing, then setting them up in ES.  Should I update Retropie to fix this issue addressed here?  If so, what method should I use?

joolswills:
Please use the forum for support.

jravetch:
@joolswills Sorry about that, added the question to https://retropie.org.uk/forum/topic/1835/8bitdo-snes30-bluethooth-input-issues/20

PR #1581, opened 2016-07-18
Scraper improvements

joolswills:


HerbFargus:
minor detail, just only got to testing this but with the launch menu art for runcommand, it doesnt seem to display the images when they are scraped into the rom folders.

joolswills:
i have changed it in the scraper - should be images as sselph's scraper uses that by default

HerbFargus:
Okey dokey!

PR #1591, opened 2016-07-23
[dosbox] hide commands and change to c:

superjamie:
When launching DOSBox, the mount of the C drive is visible which looks messy, and it doesn't change to C: drive automatically which is inconvenient.

Use @ to hide the commands from the user, and automatically change to the mounted C: drive.

The MOUNT command message displays the mounted path so there is no loss of information about where DOSBox has gotten the C drive from.


joolswills:
Thanks!

PR #1592, opened 2016-07-23
Suport for i.MX6 platforms

mbroncano:
Added detection and compiler flags for Freescale i.MX6 platforms such as Udoo

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -166,6 +166,9 @@ function get_platform() {
             ODROIDC)
                 __platform="odroid-c1"
                 ;;
+            "Freescale i.MX6 Quad/DualLite (Device Tree)")
+                __platform="imx6" 
+                ;;
             *)
                 case $architecture in
                     i686|x86_64|amd64)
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks!

joolswills:
So how well does it all work so far ? Do you have accelerated framebuffer with GLES or does it all run under X? (In which case the x11 flag should be added).

mbroncano:
So far I have tested:
- emulation station
- xboxdrv

It works well under X (I'll be adding the flag right away) and I have still to try in with the FB. 

Later today I'll be starting testing emulators, but it seems to be working quite well already.

joolswills:
the flags are used by modules - so some stuff won't show up if it doesn't have the x11 flag (eg sdl1.2 emulators require rpi or x11)

joolswills:
actually, not quite true, as many of them just have !mali (no framebuffer sdl1.x on odroid-c1), but some will check platforms flags to set the configure options etc.

mbroncano:
I see it now. By the way, maybe the worst problem I'm finding so far if the poor state of the udon repo, missing packages and all (e.g. libcaca-dev, required for libsdl1.2-dev). But it is certainly progressing.

PR #1594, opened 2016-07-25
Always use all CPUs for build

Pitel:
At least fix `-j 2` typo, causing infinte jobs


joolswills:
It is set to -j2 intentionally on the RPI due to memory constraints you can override it with `sudo MAKEFLAGS="-jX" ./retropie_setup.sh` etc

joolswills:
Will fix the space (it's not a supported platform yet btw)

joolswills:
The space is technically correct according to make man file.

Pitel:
Oh, I remembered vaguely it once wasn't, and caused unlimited jobs. But it might be many years ago. :)

PR #1595, opened 2016-07-26
Client can define __joy2key_dev

meleu:
A client of the scriptmodules functions can define another device for __joy2key_dev.

I'm playing with customized scripts and sourcing the useful scriptmodules available with RetroPie-Setup. I would like to define another device to use with joy2key, and this simple change will let me do this by just setting another device in __joy2key_dev variable before call joy2keyStart:
```
    __joy2key_dev=${__joy2key_dev:-$(ls -1 /dev/input/js* 2>/dev/null | head -n1)}
```


joolswills:
Thanks. I will implement this functionality, but I think it will be clearer to do it with a simple `[[ -z __joy2key_dev ]] && ` as it is more readable imho.

PR #1602, opened 2016-07-27
define __joy2key_dev in runcommand.cfg

meleu:
I was wrong, user can't define `__joy2key_dev` in `runcommand-onstart.sh` because it's executed in a subshell.
As far as I know there is no way to make a child shell change a parent shell variable.
A possible solution is invoke `runcommand-onstart.sh` using `source`, but I think users can mess up some variables or even functions.
What do you think about a `joy2key_device` variable in `runcommand.cfg`?

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -68,6 +68,8 @@ function get_config() {
         iniGet "disable_menu"
         disable_menu="$ini_value"
         [[ "$disable_menu" -eq 1 ]] && disable_joystick=1
+        iniGet "joy2key_device"
+        __joy2key_dev="$ini_value"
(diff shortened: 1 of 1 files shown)


joolswills:
Just export `__joy2key_dev` in ~/.bashrc or something - I envisaged the start/end scripts for other uses (others have asked for the functionality too in the past).

joolswills:
actually bashrc wont be any good depending on how you launch ES, but you could use `/opt/retropie/configs/all/autostart.sh` and export it before ES

meleu:
I was just thinking in something more dynamic. To let the user changing (almost) "on the fly".

joolswills:
at some point I will do some more work on this, eg generated config, so for now just use the env variable export as mentioned - I don't want a "ini" config for this.

PR #1603, opened 2016-07-29
Mupen64plus

joolswills:
de-duplicated getBinds

added workaround for 8bitdo controllers that need +11 hack (or -11 in this case)


joolswills:
@gizmo98 This has had very limited testing - and it's possible when de-duplicating your getBinds function I missed something. If you have any time, it would be good if you could test it - cheers.

gizmo98:
This looks good. I will try to test this this weekend.

gizmo98:
I just tested one controller. Mupen64plus.cfg looks good. In my opinion this can be merged.

joolswills:
Thanks! :)

Issue #1358, opened 2016-03-08
make runcommand save a file with some info about what it is launching


joolswills:
could be used by the screenshot module for example

joolswills:
/dev/shm/runcommand.info now contains

```
system
emulator
rom
full commandline
```

PR #1610, opened 2016-08-04
giving to on[start|end] access to runcommand arguments

meleu:
There's some useful stuff to do if runcommand-on[start|end] have access to the ROM name (usually the 4th runcommand argument).

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -760,7 +760,7 @@ function check_menu() {
     return $dont_launch
 }
 
-[[ -f "$configdir/all/runcommand-onstart.sh" ]] && bash "$configdir/all/runcommand-onstart.sh"
+[[ -f "$configdir/all/runcommand-onstart.sh" ]] && bash "$configdir/all/runcommand-onstart.sh $@"
 
 # turn off cursor and clear screen
 tput civis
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - but prefer to pass in the actual info of what it is running including emulator choice (and link it in to https://github.com/RetroPie/RetroPie-Setup/issues/1358).

meleu:
But it won't hurt anyone...
Is there a way to make on[start|end] get the ROM name?

meleu:
caver01 gave some good reason to add this feature:
https://retropie.org.uk/forum/topic/2211/a-question-about-runcommand-sh/13

And I can think some others...

joolswills:
You are welcome to maintain your own version, but this adds functionality and I don't want to merge it in and then have to change it later - as users would then need to adjust their scripts. You don't get the emulator name in the parameters by doing the above.

To avoid working on stuff that won't be accepted you could discuss it with me first on the forum.

joolswills:
I didn't say I was against the idea - just the implementation.

meleu:
I tagged you on the forum thread.
Thanks for the answer.

joolswills:
You did. 44 mins ago and I didn't have a chance to reply - I'm at work.

meleu:
That "thanks" was about the answer here, at github! It is sincere. :-) I said I won't be kidding (neither ironic) anymore!
I'm at work too. Let's get back!

PR #1613, opened 2016-08-05
Add GLupen64 N64 libretro core

gizmo98:
Optional module for now. Better alternative for lr-mupen64plus.

### DIFF
--- a/scriptmodules/libretrocores/lr-glupen64.sh
+++ b/scriptmodules/libretrocores/lr-glupen64.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
+# See the LICENSE.md file at the top-level directory of this distribution and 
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks! :)

gizmo98:
@joolswills 
Thanks! 👍 

@loganmc10
There is a module now!

loganmc10:
@gizmo98 have you tested it? How does it work for you? I would love some more people to test it

gizmo98:
I have tested it. F-Zero, Mario games, Goldeneye, Perfect Dark and other games run fine. Speed is much better compared to mupen64plus-libretro. The only game i had a serious problem was Yoshis Story. My PI seemed to hang up.

loganmc10:
When the module is built do you guys use "platform=rpi" or "platform=rpi2/3"?

joolswills:
looking at the module it uses platform=rpi1 or platform=rpi2  / rpi3

loganmc10:
OK good then it will pull in the correct cflags

loganmc10:
Just FYI I've got Framebuffer Emulation working with this core, so it should be pretty much at feature parity with mupen64plus+GLideN64

joolswills:
thanks! :)

Issue #1614, opened 2016-08-05
check/update libretro cores after renames


joolswills:
There has been a bunch of repository renaming and target name changes in libretro upstream. Need to go through our libretro cores and fix up any that are broken. Not considering renaming our modules at this stage though.

gizmo98:
I have seen most snes cores changed:
pocketsnes --> snes9x2002
SNES9x NEXT --> snes9x2010
CATSFC --> snes9x2005
fba --> fbalpha

PR #1615, opened 2016-08-05
Update libretro cores after renames + fix broken scripts

RetroPieNerd:
Fixing stuff from here: https://github.com/RetroPie/RetroPie-Setup/issues/1614


joolswills:
Ok thanks. I had actually done most of it already locally, but I'll merge this - might be worth mentioning if taking on a ticket that is currently assigned to someone else to avoid duplicated work. Cheers.

Issue #1609, opened 2016-08-03
Add snes9x/pisnes config script


superjamie:
Currently native snes9x and PiSNES do not have their controls configured in `/opt/retropie/configs/snes/snes9x.cfg` by the ES configscripts.

I intend to fix this, just logging an Issue to track it.

joolswills:
Thanks. You can probably use the pifba one as a basis (similar config). The pifba sdl1_map array can be moved to inputconfiguration.sh and used by both probably also.

superjamie:
I thought about moving the SDL keyboard map into `inputconfiguration.sh` but not every emulator uses SDL key mappings for the keyboard.

It seems more appropriate to put the SDL keymap into maybe `$rootdir/lib/`, then input config modules can `source` that keymap file during their `onstart_$module_keyboard` as required.

Thoughts?

joolswills:
the mapping isn't useful outside of the input configuration, so it's best to include it in a function in the main code or something.

joolswills:
just checked the module - looks like the mapping code is identical to pifba - so it could just use the pifba functions as mentioned on the PR. even simpler then.

PR #1616, opened 2016-08-06
pisnes - add input configuration script

superjamie:
Tested with Logitech F710 (xpad) and iBuffalo SNES (usbhid)

Resolves https://github.com/RetroPie/RetroPie-Setup/issues/1609

Signed-off-by: Jamie Bainbridge <jamie.bainbridge@gmail.com>

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/pisnes.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/pisnes.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+# 
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# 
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks. It would be best to de-duplicate the mapping as mentioned in the issue if it is the same as pifba mapping.

joolswills:
If in fact, if the mapping part is the same, it can just call the pifba function

scriptmodules/supplementary/emulationstation/configscripts/pisnes.sh
```diff
@@ -0,0 +1,58 @@
...
+function onstart_pisnes_keyboard() {
+    _split_config_pisnes
+
+    iniConfig "=" "" /tmp/pisnes-kb.cfg
+
+    _piemu_sdlkeymap
+}
+
+function map_pisnes_keyboard() {
+    map_pifba_keyboard "$1" "$2" "$3" "$4" "$5" "$6"
```
joolswills: Just for reference you could have done "$@"

joolswills:
Thanks.

PR #1617, opened 2016-08-06
M64p opts

gizmo98:
-Add blacklist for games which crash the pi or aren't playable with any video plugin. Exit script if game is found.
-Disable glesn64 autores feature. Use resolution set in gles2n64.conf and dispmanx upscaling.
-Enable FBEmulation for quake.
-Cleanup startup script video plugin check.


joolswills:
Thanks

PR #1627, opened 2016-08-10
mupen64plus: highres fix

gizmo98:
Don’t add highres options for video plugins which do not use
„—resolution“ startup parameter.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -119,6 +119,8 @@ function install_mupen64plus() {
 
 function configure_mupen64plus() {
     if isPlatform "rpi"; then
+        addSystem 0 "${md_id}-gles2n64$name" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-n64 %ROM% 320x240"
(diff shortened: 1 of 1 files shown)


joolswills:
Didn't realise some didn't use it - what happens ? Do we want to exclude the parameter completely then instead of passing 320x240 to it ?

joolswills:
I would move them to the end - remove the $name variable from them and the 320x240 parameter (and exclude the --resolution parameter completely in the launch script if blank.

gizmo98:
Two plugins do not use mupen64plus API as intended. video-videocore uses SDL1. video-glesn64 does not use global video settings declared in mupen64plus.cfg.

joolswills:
Thanks. Please make the changes I mentioned above and push/squash and I will merge.

gizmo98:
In my opinion it is not necessary to modify startup script. If we pass no resolution into the startup script "--resolution 320x240" will be set and ignored. So it should be enough to remove $name and the 320x240 parameter.

joolswills:
It's tidier - as it can be seen right away the parameter is not used then.  If you do the above, I will make that change then after.

gizmo98:
Thanks Jools! It was a little bit to late yesterday.

PR #1638, opened 2016-08-16
mupen64plus: Update gliden64 configVersion

gizmo98:
https://github.com/gonetz/GLideN64/pull/1087/files

@joolswills 
Mupen64plus needs to be updated. Old gliden64 plugins will not work properly with configVersion 13.


joolswills:
binaries rebuilt.

gizmo98:
Thanks!

Issue #1633, opened 2016-08-12
[Feature Request] Dolphin for x86 systems


j-r0dd:
I would love to see an option for dolphin in maybe the experimental package section for x86 systems. I know the pi definitely could not handle this.

joolswills:
This is included now.

j-r0dd:
Awesome! BTW the hotkeys get stored in ~/.config/dolphin-emu/Hotkeys.ini. Here's what mine look like after only changing select + start to stop the emulator:

`[Hotkeys1]                                                                                                                                                                                      
Device = evdev/0/PLAYSTATION(R)3 Controller
Keys/Open = (!`Alt_L` & !(`Shift_L` | `Shift_R`) & (`Control_L` | `Control_R` )) & O
Keys/Toggle Pause = `F10`
Keys/Stop = `Button 0` & `Button 3`
Keys/Toggle Fullscreen = (`Alt_L` & !(`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & Return
Keys/Take Screenshot = (!`Alt_L` & !(`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `F9`
Keys/Connect Wiimote 1 = (`Alt_L` & !(`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `F5`
Keys/Connect Wiimote 2 = (`Alt_L` & !(`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `F6`
Keys/Connect Wiimote 3 = (`Alt_L` & !(`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `F7`
Keys/Connect Wiimote 4 = (`Alt_L` & !(`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `F8`
Keys/Connect Balance Board = (`Alt_L` & !(`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `F9`
Keys/Disable Emulation Speed Limit = Tab
Keys/Freelook Decrease Speed = (!`Alt_L` & (`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `1`
Keys/Freelook Increase Speed = (!`Alt_L` & (`Shift_L` | `Shift_R`) & !(`Control_L` | `Control_R` )) & `2`
[...]

PR #1641, opened 2016-08-18
Add experimental gc/wii emulator dolphin

gizmo98:
-x86 only.
-no hotkeys. (dolphin seems to have no hotkey support?!)
-runs zelda ww with 15-30fps on a Pentium N3050 (Atom based) which is
real good for such a low end platform.
-only system wide installation atm.
-build time > 1h on a Pentium N3050.
-you need a mouse to get emulator window in front.
https://github.com/RetroPie/RetroPie-Setup/issues/1633

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -103,6 +103,12 @@ ngpc_fullname="Neo Geo Pocket (Color)"
 nes_exts=".nes .zip"
 nes_fullname="Nintendo Entertainment System"
 
+gc_exts=".iso"
+gc_fullname="Nintendo Gamecube"
+
+wii_exts=".iso"
+wii_fullname="Nintendo Wii"
+
 pc_exts=".bat .com .exe .sh"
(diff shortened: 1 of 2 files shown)


joolswills:
thanks.

joolswills:
I missed that fact it installs to /usr/local - I'm looking into that now.

joolswills:
Doesn't seem to like the -b option either ?

joolswills:
I also get no video output at all from dolphin-emu-nogui unless it is built without -DENABLE_HEADLESS - so I'll remove that.

joolswills:
See https://github.com/RetroPie/RetroPie-Setup/commit/1c3ea4262ff2cae4939d061600565614bff42aff

gizmo98:
I wonder why it was  running at all. Thanks for removing -DENABLE_HEADLESS. A working nogui exec is much more handy.

PR #1645, opened 2016-08-20
Ignore accelerometer in runcommand

PermanentWaves:
Ignore axis higher than 8 to prevent accelerometer from causing
unintentional movements in curses GUI. Such as happens with DS4
controllers.


joolswills:
Thanks.

PR #1650, opened 2016-08-24
Update lr-glupen64.sh

loganmc10:
I've done away with the submodules, since it forces pulling in the history of each submodule, which is going to be a lot. So you don't need this anymore when pulling the repo.

You guys can use --depth=1 to speed it up as well but I don't know what the syntax for that would be with the macros you are using.


dankcushions:
this annoyed me also, but it looks like git heard us! see http://stackoverflow.com/a/38895397

however it looks like it needs the latest version of git, which almost no-one would have i guess. something for the future, perhaps.

joolswills:
@loganmc10 Thanks. gitPullOrClone will use --depth 1 by default.

PR #1663, opened 2016-08-28
m64p: some modifications

gizmo98:
common:
-use „-flto“. All mupen64plus plugins use OPTFLAGS = -flto. We
overwrite OPTFLAGS with our CFLAGS until now.

x86:
-add CXD4 and Z64 RSP LLE plugins
-add GLideN64 LLE mode. Allows for example SW Rogue Squadron to start
up.
-add GLideN64 Intel OGL3.3 graphics workaround. This workaround should
be obsolete when Ubuntu 16.04.2 arrives (new mesa 12.1 with OGL4.3
support). There are glitches because OGL3.3 does not support all necessary features.
-enable SSSE3 optimizations. Even ten year old CPUs support SSSE3.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -39,6 +39,8 @@ function sources_mupen64plus() {
     else
         repos+=(
             'mupen64plus video-glide64mk2'
+            'mupen64plus rsp-cxd4'
+            'mupen64plus rsp-z64'
         )
     fi
     local repo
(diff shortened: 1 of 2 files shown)


loganmc10:
I don't think mesa 12.1 is going to solve the MESA_GL_VERSION_OVERRIDE=3.3COMPAT problem.

There is an upstream issue tracking this https://github.com/gonetz/GLideN64/issues/454

gizmo98:
I thought it is a intel driver problem but now it seems it is a common problem. Do you think your VBO branch will be merged upstream master anytime in the future? As i can see GLupeN64 can be build with -DCORE to overcome this problem.

joolswills:
Sorry missed this PR before - is this ready to be merged ? Thanks.

loganmc10:
@gizmo98 gonetz is preparing for an official release so he doesn't want to merge large changes like that right now, so it might be a fews week or however long. https://github.com/gonetz/GLideN64/pull/1096

GLupeN64 is now compiled with the VBO changes and -DCORE set by default, so this isn't an issue with GLupeN64

gizmo98:
@loganmc10 do you think it is ok to add an extra emu option with MESA_GL_VERSION_OVERRIDE=3.3COMPAT until there is a general fix?

loganmc10:
Ya it's no problem, it only applies to desktop OpenGL, not GLES, and it should only apply to mupen64plus+GLideN64 if that is possible (sorry I haven't look closely through the commit to see if that is how it is done)

gizmo98:
All right.

@joolswills this can be merged.

joolswills:
awesome. thanks.

PR #1671, opened 2016-09-04
Fix lr-gambatte

gizmo98:
Makefile location changed:
https://github.com/libretro/gambatte-libretro/commit/4e4fb2b4a6a4e86242a
9b63c6fc235ef25303d06
https://retropie.org.uk/forum/topic/3789/gambatte-fails-to-install-from-
source-4-0-2-x86


joolswills:
Thanks!

PR #1672, opened 2016-09-04
Update lr-gambatte.sh

jtothef:
I am new to programming and github so forgive me if this change request is not submitted correctly. 
The location of gambatte_libretro.so in the current lr-gambatte.sh file is actually one directory too deep.

The following changes I made to this file seem to work for me.

Thanks!


joolswills:
The pr is fine, but there was one 2 hours earlier tlwith the fix. Cheers.

PR #1673, opened 2016-09-05
Update ZEsarUX to 4.1

dudleydes:
Version 4.1 of ZEsarUX has improved sound using the SDL audio driver and now has an on screen keyboard available. I've added the ability to remap the joystick button to bring up screen keyboard in the config file as the default button 5 is right on the D-pad on my PS3 controller.

ZEsarUX version 4.1 is able to emulate Sam Coupe and Amstrad CPC 464, two systems currently available in RetroPie.


joolswills:
Thanks - will implement this manually as would like to self host the archive (due to unreliability of sf in the last). I will make the emulator available for the sam coupe and cpc also. PS. You have merge commits in your PR  - looks like you need to rebase your changes.

PR #1674, opened 2016-09-05
explicit use of rgui as the menu_driver

meleu:
As reported by @RetroResolution [in the forum](https://retropie.org.uk/forum/topic/3728/playstation-ffmpeg-recording-causes-seg-fault-in-retropie-4-0-2-recording-megadrive-vcs-etc-still-fine/), the default `menu_driver` for RetroArch on raspi3 is xmb (raspi3 owners can reproduce it after reinstall RetroArch from source and don't explicit set anything for `menu_driver` in retroarch.cfg).

@dankcushions said [here](https://retropie.org.uk/forum/post/23421) that he checked with libretro guys and "apparently it's defaulted differently on different systems". I checked here that raspi3 defaults to `xmb` and raspi2 and 1 defaults to `rgui`. For the sake of uniformity, it's better to explicitly set `menu_driver = "rgui"`, otherwise raspi3 users will get confused by the `xmb`.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -104,6 +104,7 @@ function configure_retroarch() {
     iniSet "core_options_path" "$configdir/all/retroarch-core-options.cfg"
     iniSet "assets_directory" "$md_inst/assets"
     iniSet "overlay_directory" "$md_inst/overlays"
+    iniSet "menu_driver" "rgui"
     isPlatform "x11" && iniSet "video_fullscreen" "true"
 
     # set default render resolution to 640x480 for rpi1


dankcushions:
can you show where RA defaults the pi2 to RGUI and pi3 to XMB? i think you're missing a link above.

i wonder if we shouldn't just use XMB for both. it is pretty slick and it may well work ok on a pi2.

joolswills:
I'll merge this for now, and I will do some testing with the other gui and we can see if it's worth switching.

meleu:
@dankcushions 
> can you show where RA defaults the pi2 to RGUI and pi3 to XMB?

Do you mean in RetroArch code? I checked it empirically, not in the code.

> i think you're missing a link above.

The forum thread link? I think it's OK.

I agree that xmb is beautiful, but not sure if it will run fine in raspi1 and zero. Also, I needed some minutes browsing through the xmb options to get used to it, so some users can have problems. In my android devices I use it, but in RetroPie I use rgui (maybe it's because a historical reason). 

@joolswills If you guys decide to use xmb by default it needs some tweaking to show those fancy icons instead of these black squares:
xmb

I think @HerbFargus have some info about it, as he mentioned [here](https://retropie.org.uk/forum/post/24080).

joolswills:
Yep. - I am aware of it requiring additional assets etc.

joolswills:
I'm not going to switch to it unless it works properly :-)

dankcushions:
@meleu 
> I checked here that raspi3 defaults to xmb and raspi2 and 1 defaults to rgui

i asked because of what you said above - that read to me like you were missing a link or explanation (where's "here"? :)), but i guess you mean on your local systems. i looked in the code (briefly) and couldn't find it. doesn't really matter to this discussion, i was just curious.

meleu:
Just correcting a wrong info I said in my first post...
I have just updated RetroArch today (to get the feature I talked about [here in the forum](https://retropie.org.uk/forum/topic/3882/now-you-can-control-rgui-using-any-connected-joypad)) and realized that the raspi2 defaults to xmb as well!
The PR keeps useful anyway (even more useful, one can say :smile:)

Since an updating from source don't overwrite the current `all/retroarch.cfg`, but creates an `retroarch.cfg.rp-dist`, I think we'll get some confused users talking about this `xmb` issue in the forum soon.

dankcushions:
the xmb (and GLUI) also currently causes a crash when invoked in GL cores (libretro-mupen64plus, glupen64), when threaded-video is turned on (which it is by default in retropie). annoying! https://github.com/libretro/RetroArch/issues/2655

so yeah, right now i'm in favour of RGUI until that issue is ironed out, although i do really like the way the xmb looks with the pixel theme :)

PR #1678, opened 2016-09-07
m64p-rpi3: use armv8 crc

gizmo98:
https://github.com/gonetz/GLideN64/commit/a471b130e79c71de1791c624ece068
468e89b615

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -78,6 +78,7 @@ function build_mupen64plus() {
     pushd $md_build/GLideN64/projects/cmake
     params=("-DMUPENPLUSAPI=On" "-DVEC4_OPT=On")
     isPlatform "neon" && params+=("-DNEON_OPT=On")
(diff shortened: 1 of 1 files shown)


dankcushions:
We should add the same (and the other flags) for the glupen64 script, although looking at https://github.com/loganmc10/GLupeN64/blob/3caf3b0eaf9fd4d61fb27abcbe3abe4c844a15cd/GLideN64/projects/cmake/Readme.txt maybe `-DCRC_OPT=On` can be safely used for non rpi3?

i'll see if i can test them out later :)

dankcushions:
actually i might be wrong about the first part - glupen is built with make, not cmake, so maybe it sets all those anyway...

gizmo98:
Loganmc10 has already enabled -DNEON_OPT. Take a look into the MakeFile. HAVE_NEON = -DNEON_OPT. -DCRC_OPT uses a simplified hash calculation which can cause crashs and glitches.  Something like HAVE_ARMV8_CRC could be included.

PR #1679, opened 2016-09-09
fs-uae integration

verybadsoldier:
This integrates Amiga emulator fs-uae into Retropie.
Multi-disk games are supported but will only use one drive.
Removed fs-uae-launcher since it is IMO not needed anymore after Retropie integration.

### DIFF
--- a/scriptmodules/emulators/fs-uae.sh
+++ b/scriptmodules/emulators/fs-uae.sh
@@ -11,7 +11,7 @@
 
 rp_module_id="fs-uae"
 rp_module_desc="Amiga emulator - FS-UAE integrates the most accurate Amiga emulation code available from WinUAE"
-rp_module_help="ROM Extension: .adf"
+rp_module_help="ROM Extension: .adf .ipf .dms .adz .zip\n\nCopy your Amiga games to $romdir/amiga\n\nCopy the required kickstart file kick13.bin to $biosdir"
 rp_module_section="exp"
 rp_module_flags="!arm"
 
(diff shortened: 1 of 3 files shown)


joolswills:
as the default config is small it might be best to just generate it from the module (using iniConfig / iniSet). You also want to avoid overwriting existing configs. See vice.sh as an example

The launch script could do with a few styling tweaks - also use of local vars etc (and not using uppercase vars)

joolswills:
I may hold off merging this as is, as mentioned on the forum I did do some code for uae4arm launching, and I may well combine the functionality / logic (The logic for getting set of disks etc could be shared)

verybadsoldier:
Ok, so I better won't rework this according to your review?

joolswills:
it's worth holding on for now - thanks.

vincenzobluesparklabs:
I found a bug on fs-uae.sh script that prevents the correct installation on Ubuntu (and I guess on Debian as well).
The function depends_fs-uae install the dependencies needed to install the uae packages but the first "case" check is based on "__raspbian_ver" variable instead of "__raspbian_name". The _ver variable value is 8 in case of Ubuntu and 7 in case of Debian but the case statement checks "ubuntu" or "debian"

Hope to be helpful ;)

joolswills:
@vincenzobluesparklabs thanks - fixed.

verybadsoldier:
It just turned out that fs-uae-**launcher** is able to handle multi-disk games inside one zip nicely when passing a zipped game from the command line. So you can store the whole game into one ZIP (I would prefer ZIP anyway) and the launcher will take care to insert the disks into the floppies.
Maybe it should be considered to drop the whole multi-disk logic from Retropie (or better: do not write it :)) and just use fs-uae-launcher. It would make fs-uae integration in Retropie really lean and clean.
The downside would be that multi-disk games would be **only** supported insides ZIPs but I think it is ok.

joolswills:
uae4arm will still require multi-disc logic which could still be useful to be shared with fs-uae for those that keep disks separately.

stuaxo:
Seems like upstream in FS-UAE would might be the best place to patch / feature request for the multi-disk support.

verybadsoldier:
You are right, it would only help fs-uae. Also it turned out that fs-uae-launcher is not part of the normal fs-uae package and would pull in lots of other dependencies... too much overhead just for the ZIP functions in my opinion...

@stuaxo 
Yes I did that and they told me that fs-uae-launcher is already capable of multi-disk ZIPs.

verybadsoldier:
Going to rework this using the ZIP functions...

joolswills:
You don't need to close it - can reuse this PR etc. but np.

PR #1685, opened 2016-09-11
adding depends

j-r0dd:
pulseaudio-utils is needed to get pasuspender

### DIFF
--- a/scriptmodules/ports/openpht.sh
+++ b/scriptmodules/ports/openpht.sh
@@ -15,6 +15,7 @@ rp_module_section="opt"
 rp_module_flags="!arm"
 
 function depends_openpht() {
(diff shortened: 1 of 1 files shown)


scriptmodules/ports/openpht.sh
```diff
@@ -42,6 +43,7 @@ function install_bin_openpht() {
 
 function remove_openpht() {
     aptRemove openpht
+    rp_callModule openpht depends remove
```
joolswills: Please remove this line (we don't force removal of package dependencies and this will remove pulseaudio)

joolswills:
I just added the dependency directly to save time. thanks.

j-r0dd:
Thanks!

Issue #1238, opened 2016-02-04
Launch Menu Art crashing ScummVM


BenWlson:
If I enable the new "Launch Menu Art" runcommand feature, ScummVM will crash when trying to load a scraped .svm file. ScummVM will still work fine if I load the main GUI from the .sh file, but it crashes for every game shortly after the box art in displayed when loading the games from .svm files.  Everything works fine again after disabling the "Launch Menu Art" feature.

joolswills:
we might need to disable it for scummvm then, or make scummvm wait for fbi to complete.

joolswills:
Please update retropie-setup from the gui, and try now.

BenWlson:
All better now

PR #1686, opened 2016-09-12
let the user display a custom "launching" image

meleu:
This is my tweak to display a custom "launching" image instead of the traditional runcommand dialog.

The intention is to use an image rather than the dialog infobox only if the original conditions to show the dialog are satisfied **AND** there is a file named "launching.jpg" (or .png) in `$configdir/$system` or `$configdir/all` directory (`$system` takes precedence). If there is no "launching" image, the traditional dialog will be shown.

Some points:
- it by no means changes the functionality of the runcommand configs (launch menu, launch menu art, and launch menu joystick control).
- Keep the capability to access the runcommand menu while launching a game ("press a button to config"). It's possible because fbi runs in background.
- The launching image can be system specific (in `$configdir/$system/`) or a more general one (in `$configdir/all/`).

My inspiration was the @rookervik [forum post here](https://retropie.org.uk/forum/topic/3262/loading-a-custom-image-on-the-run-command) and I tested with some cool launching images @lilbud made and put [here in the forum](https://retropie.org.uk/forum/topic/36/splashscreens/97).

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -751,13 +751,28 @@ function show_launch() {
     if [[ -n "$image" ]]; then
         fbi -1 -t 2 -noverbose -a "$image" </dev/tty &>/dev/null
     elif [[ "$disable_menu" -ne 1 ]]; then
-        local launch_name
-        if [[ -n "$rom_bn" ]]; then
-            launch_name="$rom_bn ($emulator)"
+        # checking for a custom "launching" images
+        for path in "$configdir/$system" "$configdir/all" ; do
+            if [[ -f "$path/launching.jpg" ]]; then
+                image="$path/launching.jpg"
+                break
+            elif [[ -f "$path/launching.png" ]]; then
+                image="$path/launching.png"
+                break
+            fi
+        done
+        # display the custom "launching" image if it was found
+        if [[ -n "$image" ]]; then
+            fbi -1 -t 5 -noverbose -a "$image" </dev/tty &>/dev/null &
         else
-            launch_name="$emulator"
+            local launch_name
+            if [[ -n "$rom_bn" ]]; then
+                launch_name="$rom_bn ($emulator)"
+            else
+                launch_name="$emulator"
+            fi
+            DIALOGRC="$configdir/all/runcommand-launch-dialog.cfg" dialog --infobox "\nLaunching $launch_name ...\n\nPress a button to configure\n\nErrors are logged to $log" 9 60
         fi
-        DIALOGRC="$configdir/all/runcommand-launch-dialog.cfg" dialog --infobox "\nLaunching $launch_name ...\n\nPress a button to configure\n\nErrors are logged to $log" 9 60
     fi
 }
 


RetroPieNerd:
Huh, this looks interesting...

robertybob:
Just read the forum post, what a great idea!

joolswills:
backgrounding fbi causes issues with some sdl programs - https://github.com/RetroPie/RetroPie-Setup/issues/1238

joolswills:
It can be added before the `if [[ -n "$image" ]]; then` using the existing image variable and fbi line though.

meleu:
It's a pitty we have this issue. Because I put fbi running in background to keep the access to the runcommand menu.

I'm thinking in two solutions:

**Solution 1**: keep my changes but check if the user is launching scummvm to avoid the issue #1238 . I agree this isn't the most elegant solution, but I really would like to keep the access to the runcommand menu.

**Solution 2**: give up the runcommand menu access and show the "launching" image when "launch menu art" is enabled **AND** there's no image for the game being launched.

I prefer the solution 1. Are you @joolswills OK with it?

joolswills:
I don't want to go with solution 1 (we don't want app specific hacks in, and it could affect other applications also - hence I didn't do this before when I fixed the issue). It may not affect scummvm any more since it moved to sdl2, but may well affect other apps that use the framebuffer.

Not sure why the logic is going to change in solution 2 from how it is currently - the logic can be kept the same.

joolswills:
BTW you still have access to the runcommand menu without backgrounding it - that's because fbi displays for a couple of seconds then exits and you still have the 2 second delay when you can trigger the menu as before.

It might just be quicker for me to implement this.

joolswills:
I might put some logic in to background fbi if using retroarch - but it will need some testing.

meleu:
yeah, I am aware of it. But there's an inconvenience, as Rookervik reported [here](https://retropie.org.uk/forum/post/19482).

A last attempt before finally go to the solution 2: to avoid the Rookervik's issue would you agree if I set the fbi timeout to 2 and run it in background AND change the `read` timeout to 3?

meleu:
> I might put some logic in to background fbi if using retroarch - but it will need some testing.

yeah, enabling this feature when launching retroarch only would be nice too.
would you agree if I do an ` if [[ "$command" =~ retroarch ]]` before my logic on searching the launching images?

joolswills:
I did that when I did the other changes - check commit log.

robertybob:
@meleu Any change of a how-to guide on the Wiki?

meleu:
@joolswills I've just saw it now. I see an improvement in the "launch menu art" too. Really cool! Thank you very much.

@robertybob There's no actual change on the configs, it's just a new feature.

HerbFargus:
If it's going to be on the wiki it should just a subset of this page: https://github.com/retropie/retropie-setup/wiki/runcommand

(And if onstart/onend are documented they should also be on that page)

meleu:
I'm writing doc for the custom launching image feature. But I'm not sure how to reference the availability of this feature. Because RetroPie is currently 4.0.2 before and after this implementation.
Should I refer to the commit date?
Or is @joolswills considering to use 4.0.3 soon?

joolswills:
I can bump to 4.0.3.

I was planning on a larger bump to a 4.1 soon, but if it helps I can do that. The trouble with rolling updates :-)

meleu:
I'll use 4.0.3+ on the doc then. Thanks!

PR #1688, opened 2016-09-13
Add L'Abbaye Des Morts

RetroPieNerd:
[Here is the website.](https://www.locomalito.com/abbaye_des_morts.php)
[Here is the Google Code archive.](https://code.google.com/archive/p/abbaye-for-linux/)
I had to make the necessary changes to the source code so it will run on RetroPie/Raspberry Pi.
[Here is my repository.](https://github.com/RetroPieNerd/LAbbayeDesMortsRPI)

### DIFF
--- a/scriptmodules/ports/labbayedesmorts.sh
+++ b/scriptmodules/ports/labbayedesmorts.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+ 
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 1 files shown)


joolswills:
Your branch has hardcoded retropie paths which will need changing to use a PREFIX var or something. Your branch lacks the google code commit history which is worth preserving btw. Might just be worth having a RPI branch against the original code with the few changes needed for the RPI (unless I have missed something it looks like just a few makefile tweaks and the removal of HWSURFACE ?)

The module will also need to include dispmanx support (if it works with the dispmanx sdl2).

joolswills:
sdl1.2 even

RetroPieNerd:
Ok I will get that sorted out soon.

joolswills:
Don't open a new PR btw - use this existing one.

RetroPieNerd:
Never mind.

PR #1690, opened 2016-09-15
Add support for x and y buttons

zigurana:
... so they can be used in ES branches that use them. This does not impact the vanilla ES, as there is no action related to these key-presses atm.


joolswills:
They should just be added to the line above `up|right|down|left|start|select)` - reduces code duplication

joolswills:
I've added them as mentioned.

zigurana:
Ok, thanks for that!

PR #1692, opened 2016-09-15
show custom launching image when es art not found

meleu:
show the custom launching image even if `use_art` is on.
Well... it's unfair to not show the custom launching art if the user enabled an option named "launch menu art" :-)
If it will be merged I'll change the wiki accordingly.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -738,8 +738,7 @@ function show_launch() {
             "$HOME/.emulationstation/downloaded_images/$system/${rom_bn}-image"
         )
     fi
-    images=(
-        "${images[@]}"
+    images+=(
         "$configdir/$system/launching"
         "$configdir/all/launching"
     )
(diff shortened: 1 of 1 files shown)


joolswills:
The idea is if a user enables the art they want to see the game art - in this case this would only show the launching if game art was missing. Not sure it is an improvement in logic, but ill merge.

joolswills:
Actually, sorry just checked the code - you should do images+=

joolswills:
Sorry I already went and changed the code myself  - ps please squash commits like this - I prefer that when reviewing over having to remember to squash when merging.

meleu:
I noticed that when the game art is not found, runcommand shows the dialog infobox. I think that this change fits in the "launch menu art" idea.

joolswills:
ok

joolswills:
I agree actually, and I think I will remove the dialog if using art (even if there is none)

joolswills:
I don't agree. Disabling the launch menu means you cannot access it at all. With art people want to be able to access the menu which they can, they just see art instead of the launch dialog.

joolswills:
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

joolswills:
I can guarantee also if we displayed the launch dialog after showing the art there would be heaps of complaints - the feature was added specifically to show art instead of the dialog.

meleu:
I mean show the dialog if there's no game art and no launching art. Not showing it after the art. :-)

P.S.: Thanks for the link.

joolswills:
Well that's what we had before, but I think it's somewhat inconsistent - and graphically prettier for those that want art instead of the dialog to just have a black screen instead. I'm not going to change it now anyway.

PR #1695, opened 2016-09-19
replaced terminal-dependent escape sequences by curses mappings

verybadsoldier:
Removed hardcoded escape sequences (cursor keys etc.) and replaced by terminal-independent curses calls that uses terminfo to map to sequences working for the actual terminal.

The problem was that on x86 Lubuntu the DPad of my X360 Gamepad was not working in runcommand because the hardcoded escape sequences were not working for my terminal:
https://retropie.org.uk/forum/topic/4115/xbox360-dpad-not-working-in-runcommand-menu

Is this ok for you? Please do not merge yet, I did not test it yet for RPi but I would like to ask before if you would support this PR.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -693,7 +693,7 @@ _EOF_
 function joy2keyStart() {
     local params=("$@")
     if [[ "${#params[@]}" -eq 0 ]]; then
-        params=(1b5b44 1b5b43 1b5b41 1b5b42 0a 20)
+        params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20)
     fi
     # check if joy2key is installed
     [[ ! -f "$rootdir/supplementary/runcommand/joy2key.py" ]] && return 1
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -10,6 +10,7 @@
 #
 
 import sys, struct, time, fcntl, termios, signal
+import curses
 
 #    struct js_event {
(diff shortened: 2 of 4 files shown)


joolswills:
if it works on the rpi, then im ok with the changes. retropie-setup will need changing too

joolswills:
thanks.

verybadsoldier:
Okay, tested it on RPi3 and I had to change "kent" back to 0x0A to make it work on RPi. But in my opinion its fine now!

verybadsoldier:
Ohh I just noticed that the hardcorded escape sequences are also contained in helpers.sh (function joy2keyStart) which I have to replace also!

joolswills:
Yeah - that's what I meant when i said retropie-setup will need changing too. You have replaced all the cases now. thanks.

verybadsoldier:
Ok, just tested retropie-config and wikiview on x86 and RPi and seems ok now!

joolswills:
Thanks. Will give it a test myself and merge if all is fine. I can squash this manually when merging, but it is preferable for PR to be squashed by the submitter.

verybadsoldier:
Ok, done, I wasn't sure if you prefer it squashed. Thank you.

joolswills:
Many thanks - sorry for the delay (been v. busy)

meleu:
I tried to update RetroPie-Setup and got a problem that seems to be related to this PR.

retropie-setup-update-issue

joolswills:
yeah there will be an issue because the update hooks puts the new joy2key in place, which gets called after updating so the new parameters were being sent to the old joy2key. I will do some workaround.

PR #1697, opened 2016-09-19
Allow glupen64 to rise from the shadows

loganmc10:
This core has received a fair amount of testing now on Linux, Windows, Android, and the Raspberry Pi. It is stable and matches the features of mupen64plus+GLideN64.

It's performance is probably still slightly worse than standalone mupen64plus, but that is because of the RetroArch rendering pipeline (the extra work it does to support custom shaders and whatnot), so there is not much that can be done at this point.

Framebuffer Emulation is disabled by default still for the Raspberry Pi since there are still hardware specific bugs present with it (same with standalone mupen64plus)

### DIFF
--- a/scriptmodules/libretrocores/lr-glupen64.sh
+++ b/scriptmodules/libretrocores/lr-glupen64.sh
@@ -12,7 +12,7 @@
 rp_module_id="lr-glupen64"
 rp_module_desc="N64 emu - GLupeN64 for libretro"
 rp_module_help="ROM Extensions: .z64 .n64 .v64\n\nCopy your N64 roms to $romdir/n64"
-rp_module_section="opt"
+rp_module_section="main"
 rp_module_flags="!mali"
 
 function sources_lr-glupen64() {


dankcushions:
Maybe there's also a case for removing mupen64plus standalone from the 'main' emulators, if glupen is made 'main'? I need to test more on the performance differences, but if it's subtle/not noticeable then I think i'd lean towards a libretro core being the default.

joolswills:
@gizmo98 thoughts ?

loganmc10:
Can't they both be main? I seem to remember there was a time when mupen64plus-libretro and mupen64plus were both "main".

I wouldn't make glupen64 the default right away, I just want it to have a bit more visibility so more people can try it out, and it if turns out more people gravitate to it then it could be made default later on I would think.

joolswills:
That is not really the purpose of main - main is what is shipped by default out of the box, and it's limited also to keep our image size down. The most popular emulators are included in main. We also keep most emulators than need manual configuration out of main. If it was to be the new default perhaps it would make more sense to move to main but to give it more visibility - optional is fine (the forum / wiki is the place to highlight it)

loganmc10:
I think most systems have a libretro emulator option in "main". Personally I think it would be more beneficial to have "mupen64plus+GLideN64" or "lr-GLupeN64" as the "main" options instead of mupen64plus + 4 different video plugins

joolswills:
there can be only "one" default emulator (defaults are never overwritten, so a user with an old default when updating would have to change it manually).

joolswills:
currently the default on RPI is mupen64plus+GLideN64 - if the libretro core offers an advantage over the standalone we can make that default, otherwise, I'm not sure of the benefit over the existing default ? It's not a massive problem to add it to main if there is a good argument for it imho. I'd like to hear @gizmo98 ' feedback anyway, as he has been looking after a lot of the mupen stuff in retropie.

loganmc10:
Ya I wouldn't change the default, I would add lr-glupen64 to main and I would actually get rid of some of the lesser used mupen64plus video plugins to keep the list simpler for the end user (like get rid of the videocore and rice plugins, are there any games that actually work better there than in GLideN64?)

Anyway I'm also interested to hear @gizmo98's input

joolswills:
I would be reluctant to permanently remove any of the existing options unless we are 100% sure they are not being used (it would only remove them for new installs technically). I'm not sure if the videocore plugin is actually better for anything, and I don't use the n64 enough to make a judgement (but users can be very vocal when we change stuff).

gizmo98:
mupen64plus is a little bit faster and has more game related settings. Framebuffer emulation will be enabled if possible. Games will not be launched if the pi would hang up. Glupen64 module doesn't have such checks at the moment.  
@loganmc10 is it possible to add a ini with game related settings?

loganmc10:
@gizmo98 I believe you have a script that automatically modifies mupen64plus.cfg before it launches the game? The only real solution would be to write a similar script that would modify retroarch-core-options.cfg before launching the game (it's really the same solution).

gizmo98:
@loganmc10 exactly this. Shouldn't be so hard but needs to be added.

loganmc10:
hm ok, I would argue that since it is a lot easier for the user to toggle Framebuffer Emulation on/off in RetroArch a script like that isn't as necessary. Anyway I don't actually use RetroPie so I won't be able to write a script like that, this can be closed if that is a requirement.

gizmo98:
I will take a look and write something in the next weeks. I really appreciate your work and a good libretro core should be always be preferred.

loganmc10:
ok cool thanks!

joolswills:
I'm going to merge this - it's only 2.7mb and if is will replace the other mupen64 as default in the future, we might as well move it now.

Also appreciate all your work on this thanks.

PR #1700, opened 2016-09-25
resetromdirs: Actually fix permissions

superjamie:
People end up with bad permissions for a variety of reasons. This script
module currently just resets ownership, it makes no changes to
permissions, even though it claims to.

Add `chmod -R ug+rwX` to fix permissions.

Signed-off-by: Jamie Bainbridge <jamie.bainbridge@gmail.com>

### DIFF
--- a/scriptmodules/supplementary/resetromdirs.sh
+++ b/scriptmodules/supplementary/resetromdirs.sh
@@ -19,4 +19,6 @@ function gui_resetromdirs() {
     mkUserDir "$biosdir"
     chown -R $user:$user "$romdir"
     chown -R $user:$user "$biosdir"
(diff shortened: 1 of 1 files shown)


joolswills:
This module dates back to when we had a more complicated permissions structure - https://github.com/RetroPie/RetroPie-Setup/commit/7c71277f7cfb83d372009331579c91ecfd1ab350 (to stop users removing rom directories). Also I found that in most cases it is just ownership issues - if the actual file permissions have been broken, there are going to be other issues such as scripts losing execute bits and so on (which this PR won't be able to fix - although less important now as it isn't required for ports etc).

I will merge this for now, but this old module may be removed or changed in the future.

PR #1702, opened 2016-09-27
Update the Minecraft Pi module

RetroPieNerd:
I have updated the Minecraft Pi module to be more convenient and relevant.
:)

### DIFF
--- a/scriptmodules/ports/minecraft.sh
+++ b/scriptmodules/ports/minecraft.sh
@@ -9,34 +9,34 @@
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
 #
 
-rp_module_id="minecraftpi"
+rp_module_id="minecraft"
 rp_module_desc="Minecraft - Pi Edition"
 rp_module_section="exp"
 rp_module_flags="!mali !x86"
 
-function _update_hook_minecraftpi() {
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks - can you be more specific as to improvements with the packaged version.

I would probably avoid the module rename - saves having an extra update hook to run and it will need to clean up the old folder etc. I think it might just be best to leave it called minecraft for now - if we need the namespace in the future that can be changed then.

RetroPieNerd:
The packaged version throws Python 2 out the window as it is legacy stuff now, in favor of Python 3.

RetroPieNerd:
Also, the accompanying Python module will be installed globally along with the game.

joolswills:
Thanks. The update hook won't be needed now since the install folder will now exist - I will need to perhaps remove the contents of it though if not empty. If you can remove the update, I will merge and do the folder cleanup bit.

RetroPieNerd:
Well, thanks! I just learned a little more about the RetroPie module format!

joolswills:
I made a couple of changes - also keeping the addPort name as just "Minecraft" otherwise you get an additional launch script (which means we would need to clean up the old one - but I'm ok with just plain Minecraft.sh)

RetroPieNerd:
Ok.

PR #1705, opened 2016-10-01
added documentation builder module and api documentation for the main helpers.sh

joolswills:
sudo ./retropie_packages.sh docs

to build the documentation - will end up in "docs/" folder off the RetroPie-Setup folder.

### DIFF
--- a/scriptmodules/admin/docs.sh
+++ b/scriptmodules/admin/docs.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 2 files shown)


joolswills:
Basic output

https://retropie.org.uk/api/

Note I will document the ini stuff and the packages etc but not all will be in the doxygen output as some functions are considered "private".

joolswills:
just to clarify - Providing API documentation does not mean this is designed to be a framework for others to build on, even though it's not too hard. The API can change at any point, and the public documentation is provided just as it was requested by a few people - primarily I wanted to get documentation into the code for the most important functions. There may well be mistakes and bits left out for now.

I also won't be doing tutorials or howtos - I want to work on the code :-)

PR #1706, opened 2016-10-03
Modify platforms

gizmo98:
Add armv64 and x64 platforms and platform flags. Some emulators have 32Bit or 64Bit optimizations which can be selected with separate flags.

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -21,6 +21,10 @@ function get_ver_sdl2() {
     echo "$ver"
 }
 
+function get_arch_sdl2() {
+    echo "$(dpkg --print-architecture)"
+}
+
 function depends_sdl2() {
     # Dependencies from the debian package control + additional dependencies for the pi (some are excluded like dpkg-dev as they are
     # already covered by the build-essential package retropie relies on.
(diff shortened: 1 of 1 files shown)


joolswills:
which emulators have different optimisations that need to be selected depending on 32/64bit ?

I am ok with the idea of the changes, but not sure I want the sdl2 changes like that - can just use `dpkg --print-architecture`

gizmo98:
Actually i added these new platforms to create i386 and amd64 SDL2 packages. There are other reasons we should split 32 and 64bit platforms. Dolphin has no official 32Bit support anymore. I have not tested Ubuntu 16.04 32bit but problems are possible (https://de.dolphin-emu.org/blog/2014/05/19/obituary-32bit/). Reicast differs between x86 and x64 and should use different dynarecs (https://github.com/reicast/reicast-emulator/blob/0a1689cb40753d7c4ec572304f9de57ae76897d8/shell/linux/Makefile#L91). Mupen64plus disables new_dynarec if x86_64 is present (https://github.com/mupen64plus/mupen64plus-core/blob/master/projects/unix/Makefile). There was also a emulator without auto detection but i cannot find the right repo at the moment.

joolswills:
Neither reicast nor mupen64plus require us to pass a parameter in though as they both use uname

joolswills:
Neither reicast nor mupen64plus require us to pass a parameter in though as they both use uname (as you are aware). I wonder if we could just use a generic 64bit flag though rather than one for arm and one for 64bit. I am wary of adding stuff as something for the future without an actual need as when we do need it we may want to do it differently.

joolswills:
if you added them for the sdl2 (which you didn't say in the PR you said - Some emulators have 32Bit or 64Bit optimizations which can be selected with separate flags.), then they are not needed for that and sdl2 can just use the dpkg comment I mentioned.

joolswills:
Thanks for the sdl2 changes. Please leave the system.sh changes for now (As I want to do it differently), and I will merge the sdl2 changes.

gizmo98:
dpkg is a really good idea. I have replaced isPlatform queries with dpkg. I have added mupen64plus and reicast just to show there are emulators with 32 and 64 bit optimizations. I know there was an emulator which needed parameters passed. I cannot find it atm. I have not tested dolphin under a 32bit OS. It is not officially supported.

gizmo98:
All right! Commits are cleaned up.

joolswills:
Thanks! :)

PR #1710, opened 2016-10-08
Add script for jacobfk20's EmulationStation mod

RetroPieNerd:
This could be in experimental until it gets implemented into RetroPie...


joolswills:
Thanks - but sorry - not yet.

RetroPieNerd:
Created a gist instead just in case someone wants to install it.
https://gist.github.com/RetroPieNerd/e319fbe347f2da23239e2eefbe93c5a8

PR #1717, opened 2016-10-14
Delete ARMSNES script as repo no longer exists

RetroPieNerd:


joolswills:
thanks. Wonder why they removed it. I guess because not much development over the upstream pocketsnes.

After merging, I then found a fork, so I think I will add it back with a fork in RetroPie for now.

HerbFargus:
Probably has something to do with this fiasco: https://t.co/dq8Ji8HVtC

joolswills:
Aah. Removing the repository doesn't really help with that. I appreciate RetroArch are upset about it, but as we have learnt there is little that can be done about these people ripping off others work for profit. A shame though.

joolswills:
Looks Looks like the project got pulled / removed ? good.

HerbFargus:
See also: https://github.com/ekeeke/Genesis-Plus-GX/issues/109 but yeah. Glad they did the right thing. Unlike kickstarter....

joolswills:
Yeah - Nice !

PR #1722, opened 2016-10-21
reicast: added message box if no joystick is connected

verybadsoldier:
After stupid me searched about 1 hour why my Reicast was just crashing at startup, I discovered that I had just no joystick connected (command line parameter was then like "--player-nb=" which crashes Reicast). I thought it would be cool to pop a message to the user in that case.
Also I tried to simplify the if-conditions in the script to reduce code indentation.

Better look twice: don't trust my bash skills.

### DIFF
--- a/scriptmodules/emulators/reicast/reicast.sh
+++ b/scriptmodules/emulators/reicast/reicast.sh
@@ -69,17 +69,32 @@ function mapInput() {
     params+="-config input:joystick_device_id=-1 "
     params+="-config players:nb=$device_counter "
     echo "$params"
+
+    if [[ -z $device_counter ]]; then
+        exit 1
+    fi
 }
 
-if [[ -f "$HOME/RetroPie/BIOS/dc_boot.bin" ]]; then
-    params="-config config:homedir=$HOME -config x11:fullscreen=1 "
-    getAutoConf reicast_input && params+=$(mapInput)
-    params+=" -config audio:backend=$AUDIO -config audio:disable=0 "
-    if [[ "$AUDIO" == "oss" ]]; then
-        aoss "$rootdir/emulators/reicast/bin/reicast" $params -config config:image="$ROM" >> /dev/null
-    else
-        "$rootdir/emulators/reicast/bin/reicast" $params -config config:image="$ROM" >> /dev/null
-    fi
-else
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. This launch code is from @gizmo98 - so I will get this feedback before merging - cheers.

joolswills:
Also - should reicast work with keyboard only ? If so, we should probably just make sure we don't pass this parameter if no joypads connected.

I can look into this further if required.

gizmo98:
Reicast should run with a keyboard as well. At least there was a working fallback in the past:
https://github.com/verybadsoldier/RetroPie-Setup/blob/1814bf2f1f1353e414a63a6c828ad152c10a3117/scriptmodules/emulators/reicast/reicast.sh#L67

Could be --player-nb=1 must be set for keyboard only setups. If no joystick is found this value is zero now. I will take a look next week.

joolswills:
Logic look like it needs a fix as it does after

```
params+="-config input:joystick_device_id=-1 "
params+="-config players:nb=$device_counter "
```

which gets called even if ``$device_counter == ""`

I

joolswills:
Sorry - didn't read your reply properly :) I see you just said that

joolswills:
I made that 1 for keyboard only, and it seems to work, please test the latest retropie-setup @verybadsoldier

joolswills:
(Will need to update reicast install also)

verybadsoldier:
@joolswills 
Looks fine to me: with no joystick attached Reicast starts now correctly and keyboard is working.

verybadsoldier:
@joolswills 
Hm, when pressing ESC (in the clock screen) then I get this:
`vbs@wz-retro:/opt/retropie/emulators/reicast/bin$ ./reicast.sh oss /home/vbs/RetroPie/roms/dreamcast/Power\ Stone\ 2.cdi 
evdev: open: Permission denied
./reicast.sh: line 86:  6970 Trace/breakpoint trap   (core dumped) aoss "$rootdir/emulators/reicast/bin/reicast" $params -config config:image="$ROM" >> /dev/null
`

Not so sure, but "evdev: open: Permission denied" sounds wrong at first glance. Maybe it still tries to open one joystick?

verybadsoldier:
@joolswills 
Sorry I think that message was related to me messing around with xboxdrv before. After doing a clean reboot that "permission denied" message is gone. The "core dumped" message still stays though.

PR #1730, opened 2016-10-30
add luminous, minilumi, pixel-meta, and pixel-tft ES Themes from RooK

HerbFargus:


RetroPieNerd:
Sounds like inception with es-theme-minilumi...

joolswills:
thanks.

PR #1735, opened 2016-11-05
VICE multi-disk ZIP support

verybadsoldier:
Out of the box VICE can load ZIP files but only uses the first disk from the ZIP. This PR should make it possible to put all the disks belonging to a game into one ZIP. Upon startup the ZIP will be extracted into a temp directory and VICE will be started to autoload the first disk from that directory.
The temp directory (containing all unzipped disks) will be passed to VICE as working directory so when the user uses the disk swap functions in VICE he will see the (temp) directory containing all the relevant disks.

This is pretty much just a proof of concept. If you like it I will polish it a bit and test it more. Also I would like to put the unzip functions into a common shell script because I would also like to use it for emulators that do not support ZIP natively (fs-uae currently).

### DIFF
--- a/scriptmodules/archivefuncs.sh
+++ b/scriptmodules/archivefuncs.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+## @file archivefuncs.sh
+## @brief RetroPie archivefuncs library
+## @copyright GPLv3
+
+readonly arch_dir="/tmp/retropie-archive"
+
+## @fn archiveExtract()
+## @param src_file Full path to archive file to extract
+## @param disk_exts Space seperated list of supported disk extensions (e.g. ".d64 .t64")
+## @brief Extracts an archive to a temporary directory
+## @details After calling this the variable arch_dir will contain the directory that was used
+## for extraction. Also the variable arch_files will contain an array of filenames that are
+## considered as game disks according to parameter disk_exts.
+function archiveExtract() {
+    local src_file="$1"
+    local disk_exts="$2"
+
+    # clean temp directory if needed
+    archiveCleanup
+    mkdir "$arch_dir"
+
+    local ext="${src_file##*.}"
+
+    case "${ext,,}" in
+        zip)
+            unzip "$src_file" -d "$arch_dir"
+            ;;
+        *)
+            echo "Unsupported archive: $src_file"
+            return 1
+            ;;
+    esac
+
+    # build a regex portion from the passed extensions
+    regexPart=$(echo ${disk_exts} | sed 's/ /\\|/g')
+
+    IFS=$'\n' read -d '' -r -a arch_files < <(find "$arch_dir" -iregex ".*.\(${regexPart}\)$")
+
+    if [[ ${#arch_files[@]} -eq 0 ]]; then
+        return 2
+    fi
+}
+
+## @fn archiveCleanup()
+## @brief Purges archive temp directory from previous calls to archiveExtract
+function archiveCleanup() {
+    [[ -d "$arch_dir" ]] && rm -rf "$arch_dir"
+}
--- a/scriptmodules/archivefuncs.sh
+++ b/scriptmodules/archivefuncs.sh
@@ -43,9 +43,9 @@ function archiveExtract() {
     esac
 
     # build a regex portion from the passed extensions
-    regexPart=$(echo ${disk_exts} | sed 's/ /\\|/g')
+    local regex="${disk_exts// /\\|}"
 
-    IFS=$'\n' read -d '' -r -a arch_files < <(find "$arch_dir" -iregex ".*.\(${regexPart}\)$")
+    IFS=$'\n' read -d '' -r -a arch_files < <(find "$arch_dir" -iregex ".*.\(${regex}\)$" | sort)
 
     if [[ ${#arch_files[@]} -eq 0 ]]; then
         return 2
--- a/scriptmodules/emulators/vice.sh
+++ b/scriptmodules/emulators/vice.sh
@@ -45,15 +45,44 @@ function install_vice() {
 }
 
 function configure_vice() {
+    # get a list of supported extensions
+    c64_exts=$(getSystemExtensions c64)
+
+    # install the vice start script
+    mkdir "$md_inst/bin/"
+    cat > "$md_inst/bin/vice.sh" << _EOF_
+#!/bin/bash
+
+BIN="\${0%/*}/\$1"
+ROM="\$2"
+
+romdir="\${ROM%/*}"
+ext="\${ROM##*.}"
+source "$rootdir/lib/archivefuncs.sh"
+
+archiveExtract "\$ROM" "$c64_exts"
+
+# check successful extraction and if we have at least one file
+if [[ \$? == 0 ]]; then
+    ROM="\${arch_files[0]}"
+    romdir="\$arch_dir"
+fi
+
+"\$BIN" -chdir "\$romdir" "\$ROM"
+archiveCleanup
+_EOF_
+
+    chmod +x "$md_inst/bin/vice.sh"
+
     mkRomDir "c64"
 
-    addSystem 1 "$md_id-x64" "c64" "$md_inst/bin/x64 %ROM%"
-    addSystem 0 "$md_id-x64sc" "c64" "$md_inst/bin/x64sc %ROM%"
-    addSystem 0 "$md_id-x128" "c64" "$md_inst/bin/x128 %ROM%"
-    addSystem 0 "$md_id-xpet" "c64" "$md_inst/bin/xpet %ROM%"
-    addSystem 0 "$md_id-xplus4" "c64" "$md_inst/bin/xplus4 %ROM%"
-    addSystem 0 "$md_id-xvic" "c64" "$md_inst/bin/xvic %ROM%"
-    addSystem 0 "$md_id-xvic-cart" "c64" "$md_inst/bin/xvic -cartgeneric %ROM%"
+    addSystem 1 "$md_id-x64" "c64" "$md_inst/bin/vice.sh x64 %ROM%"
+    addSystem 0 "$md_id-x64sc" "c64" "$md_inst/bin/vice.sh x64sc %ROM%"
+    addSystem 0 "$md_id-x128" "c64" "$md_inst/bin/vice.sh x128 %ROM%"
+    addSystem 0 "$md_id-xpet" "c64" "$md_inst/bin/vice.sh xpet %ROM%"
+    addSystem 0 "$md_id-xplus4" "c64" "$md_inst/bin/vice.sh xplus4 %ROM%"
+    addSystem 0 "$md_id-xvic" "c64" "$md_inst/bin/vice.sh xvic %ROM%"
+    addSystem 0 "$md_id-xvic-cart" "c64" "$md_inst/bin/vice.sh 'xvic -cartgeneric' %ROM%"
 
     [[ "$md_mode" == "remove" ]] && return
 
--- a/scriptmodules/emulators/vice.sh
+++ b/scriptmodules/emulators/vice.sh
@@ -46,7 +46,7 @@ function install_vice() {
 
 function configure_vice() {
     # get a list of supported extensions
-    c64_exts=$(getSystemExtensions c64)
+    c64_exts="$(getSystemExtensions c64)"
 
     # install the vice start script
     mkdir "$md_inst/bin/"
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -926,7 +926,7 @@ function getSystemExtensions() {
     iniConfig "=" '"' "$config"
     iniGet "$1_exts"
 
-    echo $ini_value
+    echo "$ini_value"
 }
 
 ## @fn applyPatch()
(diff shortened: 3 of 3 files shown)


joolswills:
I like the idea. in regards to common functions etc, it would depend how much code is being shared etc, but I generally do prefer to avoid duplication of course.

scriptmodules/emulators/vice/vice.sh
```diff
@@ -0,0 +1,32 @@
...
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+BIN=$(dirname "$0")/"$1"
```
joolswills: since we are using bash `BIN="${0%/*}/$1"`

scriptmodules/emulators/vice/vice.sh
```diff
@@ -0,0 +1,32 @@
...
+ROM="$2"
+
+romdir="$datadir/roms/c64"
```
joolswills: can get romdir from the ROM path to avoid hardcoding.

joolswills: also, $datadir is not set anywhere either in this script?

verybadsoldier:
Thanks for pointing those out. I basically rewrote most of it taking your points into account. I created a shell library "archivefuncs.sh" which can handle different compressors (right now only ZIP actually :)). When invoked it will set two variables $arch_dir and $arch_files to be used in the emulator start script afterwards.

I am not very happy that I hardcoded the file extensions supported by VICE into its startscript. The same information (without ZIP) is also maintained in es_systems.cfg but I couldn't find a good way to access it besides parsing the XML.

As always: I am not well trained at bash but still tried to do it correctly. But feel free to post suggestions how I can do better.

joolswills:
Can read the extensions from `platforms.cfg` in RetroPie-Setup/ (note checking /opt/retropie/configs/all/platforms.cfg first if it exists).

scriptmodules/emulators/vice/vice.sh
```diff
@@ -0,0 +1,29 @@
...
+romdir="${ROM%/*}"
+ext="${ROM##*.}"
+source "$rootdir/lib/archivefuncs.sh"
+
+archiveExtract $ROM "crt,d64,g64,prg,t64,tap,x64,vsf"
+
+# check successful extraction and if we have at least one file
+if [[ $? == 0 && ${#arch_files[@]} -gt 0 ]]; then
+    ROM=${arch_files[0]}
+    romdir=$arch_dir
```
joolswills: minor, but you could quote these for consistency.

scriptmodules/helpers.sh
```diff
@@ -345,9 +345,12 @@ function ensureRootdirExists() {
 
     # make sure we have inifuncs.sh in place and that it is up to date
     mkdir -p "$rootdir/lib"
-    if [[ ! -f "$rootdir/lib/inifuncs.sh" || "$rootdir/lib/inifuncs.sh" -ot "$scriptdir/scriptmodules/inifuncs.sh" ]]; then
-        cp --preserve=timestamps "$scriptdir/scriptmodules/inifuncs.sh" "$rootdir/lib/inifuncs.sh"
-    fi
+    declare -a helper_libs=(inifuncs.sh archivefuncs.sh)
```
joolswills: `local helper_libs=(inifuncs.sh archivefuncs.sh)` would suffice. Actually, at some point I intend to do some renaming of vars such as CAPS for global vars and make constants read only, but for now it's fine to leave the declare out.

scriptmodules/emulators/vice/vice.sh
joolswills: This launch script might be small enough to just embed in the module btw - and that way you can avoid hardcoding "rootdir". You have to quote some of the additional vars, but it might be worth it.

you could return 2 for empty archive to save the check here and do it in the archive function ?

joolswills:
made a few comments. but looks generally good and I'm happy to proceed with your idea.

verybadsoldier:
Ok thanks for those hints, very appreciated! I started to feel a bit more comfortable in bash. Here is the next try. I tried to incorporate all your suggestions. There is no (?) hardcoded stuff anymore. rootdir and also the list of supported extensions are printed into the startscript (which is generated now).
I added a function to helpers.sh to get the list of extensions for a given system from platforms.cfg.

In general, do you prefer further commits after your review squashed into the previous commit or better to have them separated so its easier to see the last changes?

verybadsoldier:
I think I missed some quotes again... Coming back to this soon...

scriptmodules/archivefuncs.sh
```diff
@@ -0,0 +1,59 @@
...
+    if [[ ${ext,,} == "zip" ]]; then
+        __zipExtract $src_file $arch_dir
+    else
+        echo "Unsupported archive: $src_file"
+        return 1
+    fi
+
+    # get a list of extracted files matching the desired extensions
+    shopt -s nocaseglob nullglob
+    read -r -a arch_files <<< $(eval "ls -1 $arch_dir/*{$disk_exts,}")
```
joolswills: Why eval ?

I would think `find` would be better choice, and you wouldn't need to change the shell opts. I suspect this may have problems with filenames with spaces ?

scriptmodules/archivefuncs.sh
```diff
@@ -0,0 +1,59 @@
...
+    local src_file=$1
+    local disk_exts=$2
+
+    # clean temp directory if needed
+    archiveCleanup
+    mkdir "$arch_dir"
+
+    local ext="${src_file##*.}"    
+    if [[ ${ext,,} == "zip" ]]; then
+        __zipExtract $src_file $arch_dir
```
joolswills: need to quote them.

scriptmodules/helpers.sh
```diff
@@ -899,6 +902,21 @@ function loadModuleConfig() {
...
+function getSystemExtensions() {
+    local platforms="$scriptdir/platforms.cfg"
+    local platforms_alt="$configdir/all/platforms.cfg"
+    [[ -f "$platform_alt" ]] && platforms="$platforms_alt"
+    
+    iniConfig "=" '"' "$platforms"
+    iniGet "$1_exts"
+
+    function join_by { local IFS="$1"; shift; echo "$*"; }
+    join_by , $ini_value
```
joolswills: a bash replace to replace the spaces with tabs would be simpler - but why not just return them as a string with space separated - can easily be converted to an array, to be processed later. (Also I can then use it for the addSystem function)

Also something like this would seem better

```
local config
for config in "$configdir/all/platforms.cfg" "$scriptdir/platforms.cfg"; do
     [[ -f "$config" ]] && break
done
```

would seem better.

scriptmodules/emulators/vice.sh
```diff
@@ -45,15 +45,44 @@ function install_vice() {
...
+
+romdir="\${ROM%/*}"
+ext="\${ROM##*.}"
+source "$rootdir/lib/archivefuncs.sh"
+
+archiveExtract \$ROM "$c64_exts"
+
+# check successful extraction and if we have at least one file
+if [[ \$? == 0 ]]; then
+    ROM=\${arch_files[0]}
```
joolswills: quotes

scriptmodules/archivefuncs.sh
```diff
@@ -0,0 +1,59 @@
...
+## @fn archiveExtract()
+## @param src_file Full path to archive file to extract
+## @param disk_exts Comma seperated list of supported disk extensios
+## @brief Extracts an archive to a temporary directory
+## @details After calling this the variable arch_dir will contain the directory that was used 
+## for extraction. Also the variable arch_files will contain an array of filenames that are 
+## considered as game disks according to parameter disk_exts.
+function archiveExtract() {
+    local src_file=$1
+    local disk_exts=$2
```
joolswills: quotes

scriptmodules/archivefuncs.sh
```diff
@@ -0,0 +1,59 @@
...
+
+## @file archivefuncs.sh
+## @brief RetroPie archivefuncs library
+## @copyright GPLv3
+
+readonly arch_dir="/tmp/retropie_archive"
+
+function __zipExtract() {
+    unzip "$1" -d "$2"
+}
```
joolswills: It's likely to be a one liner per archive format or not much more, so no need for functions - just do a case statement for checking the extension and put it in there.

scriptmodules/archivefuncs.sh
```diff
@@ -0,0 +1,59 @@
...
+    archiveCleanup
+    mkdir "$arch_dir"
+
+    local ext="${src_file##*.}"    
+    if [[ ${ext,,} == "zip" ]]; then
+        __zipExtract $src_file $arch_dir
+    else
+        echo "Unsupported archive: $src_file"
+        return 1
+    fi
```
joolswills: use a case statement.

joolswills:
realise you are aware of some of the quoting, but I highlighted some other issues.

scriptmodules/archivefuncs.sh
joolswills: ah - you use eval because you are relying on globbing for the extensions. If using find can do as `find -iregex ".*.\(ext1\|ext2\)$"` or build the commandline with multiple `-iname`

I wouldn't use `<<<` either

something like `IFS=$'\n' read -d '' -r -a arch_files < <(find command)`

verybadsoldier:
Well, things are getting a bit embarrassing :) I again read a lot about bash and tried to incorporate your suggestions to get things right this time. Many thanks for the input and your patience, it helped alot!

scriptmodules/archivefuncs.sh
```diff
@@ -0,0 +1,59 @@
...
+            unzip "$src_file" -d "$arch_dir"
+            ;;
+        *)
+            echo "Unsupported archive: $src_file"
+            return 1
+            ;;
+    esac
+
+    # build a regex portion from the passed extensions
+    regexPart=$(echo ${disk_exts} | sed 's/ /\\|/g')
```
joolswills: Please use the bash builtins, no camelCase vars please and use local vars

`local regex="${disk_exts// /\\|}"`

scriptmodules/emulators/vice.sh
```diff
@@ -45,15 +45,44 @@ function install_vice() {
 }
 
 function configure_vice() {
+    # get a list of supported extensions
+    c64_exts=$(getSystemExtensions c64)
```
joolswills: quotes please.

scriptmodules/helpers.sh
```diff
@@ -901,6 +904,22 @@ function loadModuleConfig() {
...
+function getSystemExtensions() {
+    local config
+    for config in "$configdir/all/platforms.cfg" "$scriptdir/platforms.cfg"; do
+         [[ -f "$config" ]] && break
+    done
+    
+    iniConfig "=" '"' "$config"
+    iniGet "$1_exts"
+
+    echo $ini_value
```
joolswills: consistency/styleguide - functionality I guess it's the same.

joolswills: depends what you are printing of course - but in other cases it would matter

```
test="    hello world"
echo $test
```

in this case it won't matter - but it's good practise to avoid problems

joolswills: (Updated previous post as incomplete)

verybadsoldier:
Me again, sorry for wasting your time... :(
I left it as a separate commit for now. I thought it might be easier for you to review. I can squash at any time if needed.

scriptmodules/archivefuncs.sh
```diff
@@ -0,0 +1,59 @@
...
+        *)
+            echo "Unsupported archive: $src_file"
+            return 1
+            ;;
+    esac
+
+    # build a regex portion from the passed extensions
+    local regex="${disk_exts// /\\|}"
+
+    IFS=$'\n' read -d '' -r -a arch_files < <(find "$arch_dir" -iregex ".*.\(${regex}\)$")
```
joolswills: Another thought - you probably want a `| sort` in there, as find does not return files alphabetically

joolswills:
Note there are plenty of places where we miss quotes around the code - I clean up now and again, so it's not a massive deal - especially when it's cases that will never break from not being quoted (some cases are more important). Some are more important than others, but I just noticed when reviewing. So don't worry.

verybadsoldier:
Well, I am usually quite picky about code and all those unneeded mistakes make me just angry with myself :/ Added that sort though :)

joolswills:
Thanks - Looks good to me - might be worth giving it a final test etc but I'm happy to merge if you are.

verybadsoldier:
Let's better wait a bit if you are ok with this version now. So I can give it some tests in the next 2-3 days.
Thanks again for your patience and guidance.

joolswills:
NP. Thanks for contributing :)

verybadsoldier:
So, I checked it again and I think it should be fine. So this is good to go from my side! :)

joolswills:
thanks.

jimbo27:
Today I tried this the new feature. Using multi-disk-zip-files will straighten my list of roms. Great! :)
I have got a question regarding save disks: If I use a multi-file-zip and all disk files will be extracted to a temp folder, what will happen to the save disk after I save a game? Will this save disk be put again into the zip file? Or will I loose my save game because those temporary disk files will be deleted?

verybadsoldier:
When using ZIPs then the working directory will be set to `/tmp/retropie-archive`. So if not specified otherwise then new files will be saved there and indeed wiped after the emulator exits.
Are we speaking about snapshot images or about ingame saves?

jimbo27:
I was talking about ingame saves respectively saves on a disk. Like saving your whole party on a disk while playing Bard's Tale.
But I have not thought of the possibility of using snapshots for saving a game. Using snapshots I can avoid the problem of saving a game into a temp directoy I guess?

verybadsoldier:
I see that it is not perfect at the moment. Too bad that VICE does not allow to specify a directory to load disks from and the default save directory independently (AFAIK).
But it should be possible to use "Create new disk image" in the "Drive" settings to create a new disk solely for savegames in a non-temp directory. That way you would also not have to modify your original disk images. But I did not test it yet :P

PR #1742, opened 2016-11-10
Runcommand changes

joolswills:
* make global vars uppercase - was getting confusing and will make it easier when refactoring / removing
 * rework mode switching
 * rework defaults loading/saving to remove globals
 * readable keys in configs/all/videomodes.cfg and configs/all/emulators.cfg instead of md5hashes

### DIFF
--- a/scriptmodules/admin/docs.sh
+++ b/scriptmodules/admin/docs.sh
@@ -14,7 +14,7 @@ rp_module_desc="Generate developer documentation"
 rp_module_section=""
 
 function depends_docs() {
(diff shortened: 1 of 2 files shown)


joolswills:
This is about ready now - if anyone has some time to test it I would be most appreciative in case I missed anything.

Mostly making sure when switching settings on a per rom basis, everything gets correctly saved, and that video modes switch / restore as they should.

joolswills:
I tested it as much as I could - going to merge.

PR #1744, opened 2016-11-14
es theme modern addition

HerbFargus:


HerbFargus:
... once rook adds it to github.... we'll leave this til he does.

joolswills:
I merged before reading comment. Sorry :-)

joolswills:
Well, doesn't matter. Do you want me to revert it or leave it for now?

HerbFargus:
eh we can leave it, I'll chat with rook when he's around

How I figure is we can just find a swanky default image and then if people want to modify it we'll have a tutorial for it on the forum thread or something

PR #1746, opened 2016-11-18
Fix make execution

JuanMiguelBG:
Fix make execution.


joolswills:
Thanks - Please can you be a little more descriptive when creating a PR in the future - eg the reasons for the change (In this case because SDLPOP changed their repository structure). Also please squash PR's when you have minor followup corrections.

PR #1749, opened 2016-11-19
Attractmode

joolswills:
This adds an attractmode module - should build on X11 and RPI.

Added support for multiple frontends when generating configs / moved ES config generation to ES script

Added initial configuration / romlist generation for attractmode

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -471,10 +471,15 @@ function compareVersions() {
 
 ## @fn dirIsEmpty()
 ## @param path path to directory
+## @param files_only set to 1 to ignore sub directories
 ## @retval 0 if the directory is empty
 ## @retval 1 if the directory is not empty
 function dirIsEmpty() {
-    [[ -z "$(ls -A "$1")" ]] && return 0
+    if [[ "$2" -eq 1 ]]; then
+        [[ -z "$(ls -lA1 "$1" | grep "^-")" ]] && return 0
+    else
+        [[ -z "$(ls -A "$1")" ]] && return 0
+    fi
     return 1
 }
 
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -766,55 +766,10 @@ function iniFileEditor() {
 ## @param theme name of theme to use
 ## @brief Adds a system entry for Emulation Station (to /etc/emulationstation/es_systems.cfg).
 function setESSystem() {
-    local fullname=$1
-    local name=$2
-    local path=$3
-    local extension=$4
-    local command=$5
-    local platform=$6
-    local theme=$7
-
-    local conf="/etc/emulationstation/es_systems.cfg"
-    mkdir -p "/etc/emulationstation"
-    if [[ ! -f "$conf" ]]; then
-        echo "<systemList />" >"$conf"
-    fi
-
-    cp "$conf" "$conf.bak"
-    if [[ $(xmlstarlet sel -t -v "count(/systemList/system[name='$name'])" "$conf") -eq 0 ]]; then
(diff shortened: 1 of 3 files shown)


joolswills:
Still a WIP - needs more testing.

joolswills:
OK. I think this mostly works now. Might be a few bugs.

Currently when installing it reconfigures all installed packages so as to get an initial list for attractmode. I intend to move this to a GUI function.

joolswills:
Should be ready for testing now.

robertybob:
Does this include the menu that Floob included in his Image? The one that includes launching Emulationstation and RetroPie-Setup?

robertybob:
@joolswills  Also, is there a way of bundling FFMPEG with MMAL support? It took me over an hour to compile it from source, but it allows much higher quality video snaps (and therefore may prevent users complaining that videos stop and turn grey if they've encoded them above 240p). If you install AM and then FFMPEG you have to reinstall AM (making sure you first remove the 'develop' folder) for the option to show up.

Instructions:

http://pastebin.com/5ZbBexNc

joolswills:
@robertybob no menu. I don't plan to include any optimised ffmpeg for now - but at some point I might include it for both retroarch and this but no plans currently.

I'm not really looking for feature requests - this PR is here for testing.

amigagamer:
A few things ive noticed trying out attractmode over the last few weeks (may be useful info)

Attractmodes scraper (using thegamesdb) downloads the same boxart images for "snap" as sselphs scraper for emulationstation (tested with no intro named sets).  Only diference being the files are renamed for emulationstation 
Romname (region).jpg on attract
Rom (region)-image.jpg on ES
If the naming was consistent across both frontends it would open up the possibility of sharing a boxart folder .

I've found attractmode isnt as accurate at scraping as emulationstation or sselphs scraper, its mostly minor issues such as NES banners instead of SNES- any issues ive run into have been on games that are multiplatform - i guess this is searching by game name where emulationstation uses platform id as well?

Some custom themes have a screensaver mode that shows 4+ videos at once. This really pushes a pi3 hard on software decoding so its something to look for if bundling themes without mmal support in ffmpg

joolswills:
Thanks for the info.

amigagamer:
A few hours testing on this since it was commited to master.

no problems noticed with the compiling and configuring alongside my preinstalled HW ffmpeg version at /usr/local/bin (ffmpeg in system paths)
auto-generated configs work perfectly (although i did move my entire .attract folder for a clean install beforehand)

Copying previously scraped flyers marquees snaps(inc videos) and wheel arts to the default folders .attract/scraper/PLATFORM all work perfectly (even though the paths are not explicitly defined in the autogenerated configs per emulator)

Theres scraper issues with the generated platform names (thegamesdb)
for example - nintendo gameboy
auto generated config sets  "system  gb"  - this returns no results off the scraper
changing this to the "system nintendo-gameboy" works fine as the listed Alias in http://wiki.thegamesdb.net/index.php/GetPlatformsList

attract.cfg plays well with adding extra joystick actions per single command- eg
left   Left   
left  Joy0 Left
left  PovXneg
with this example i have tried a few different controllers, one with analogue+ digital, a hori arcade stick that only works as Pov , and a retrobit that sends analogue.

defining a "configure" button for joypads is useful (keyboard TAB) ,although id imagine itd be difficult to autoconfig in retropie due to different button layouts per controller

Missing intro video generates an error to console on startup.  Doesnt seem to be causing any issues or delay,however it can be disabled in attract.cfg

joolswills:
Scraper names is a known issue - it's meant as an initial config for users to change currently and not fully automatic - we don't have the gamesdb names in retropie.

Note I have also just changed from short names to long names for the configs - as it uses the name of the "Emulator" in the GUI.

PR #1751, opened 2016-11-20
allow running in elementary OS

pl402:


joolswills:
Thanks. Please provide a bit more detail when creating PR - also you seem to have an extra commit.

I will hold on to this for now as I have some other os compatibility parts to change which will affect this - I want to map Ubuntu / Debian based OS's against an upstream version number to deal with some package name issues.

Will add support to elementary OS once I have done this (also steam os)

LodanZark:
The elementary OS loki 0.4 it's based on Ubuntu 16.04 LTS, I think the biggest diffrence it's the usage of the Pantheon Desktop Environment which is built on top of the GNOME software base, so probably won't be any issues.

joolswills:
Was implemented in https://github.com/RetroPie/RetroPie-Setup/pull/1801

PR #1754, opened 2016-11-20
fs-uae - added integration into emulationstation. Supporting ZIPs containing multiple disks

verybadsoldier:
So, me again ;)

Another try on integrating fs-uae into ES. Now taking the ZIP functions into account (so no multi-disk-discovery needed anymore).

Some explanations on some points (does not make them necessarily good decisions ofc):

-  I thought about adding Amiga to platforms.cfg and to store the supported extensions there. But the extensions are depending on the emulator. So fs-uae supports IPF, ADF etc. (and now ZIP) while uae4all seems to be ADF only. I don't know if platforms.cfg is supposed to handle this so I added the extensions as a variable to the emulator script. Actually I wanted to have it (FSUAE_EXTS) readonly but when I declare it readonly it is always empty later in the configure function. I could not find out why.

- I changed the config directory symbolic link from `~/.config/fs-uae` to `~/Documents/FS-UAE/Configuration` since this is the directory that fs-uae creates on first run to store and read the config from.

- I added a hopefully sensible default config for fs-uae to have it run in fullscreen by default. Also some other parameters are in there so they can be changed easily by the user.

Looking forward to your input!

### DIFF
--- a/scriptmodules/emulators/fs-uae.sh
+++ b/scriptmodules/emulators/fs-uae.sh
@@ -11,7 +11,7 @@
 
 rp_module_id="fs-uae"
 rp_module_desc="Amiga emulator - FS-UAE integrates the most accurate Amiga emulation code available from WinUAE"
-rp_module_help="ROM Extension: .adf"
+rp_module_help="ROM Extension: .adf  .adz .dms .ipf .zip\n\nCopy your Amiga games to $romdir/amiga\n\nCopy a required BIOS file (e.g. kick13.rom) to $biosdir"
 rp_module_section="exp"
 rp_module_flags="!arm"
 
@@ -47,16 +47,28 @@ function remove_fs-uae() {
 function configure_fs-uae() {
     mkRomDir "amiga"
 
-    mkUserDir "$home/.config"
+    # copy configuring start script
+    mkdir "$md_inst/bin"
+    cp "$md_data/fs-uae.sh" "$md_inst/bin"
+    chmod +x "$md_inst/bin/fs-uae.sh"
+
     mkUserDir "$md_conf_root/amiga"
-    moveConfigDir "$home/.config/fs-uae" "$md_conf_root/amiga/fs-uae"
+    mkUserDir "$home/Documents/FS-UAE"
+    mkUserDir "$home/Documents/FS-UAE/Configurations"
+    moveConfigDir "$home/Documents/FS-UAE/Configurations" "$md_conf_root/amiga/fs-uae"
 
-    cat > "$romdir/amiga/+Start FS-UAE.sh" << _EOF_
-#!/bin/bash
-fs-uae-launcher
-_EOF_
-    chmod a+x "$romdir/amiga/+Start FS-UAE.sh"
-    chown $user:$user "$romdir/amiga/+Start FS-UAE.sh"
-
-    addSystem 1 "$md_id" "amiga" "bash $romdir/amiga/+Start\ FS-UAE.sh" "Amiga" ".sh"
+    # copy default config file
+    local config="$(mktemp)"
+    iniConfig " = " "" "$config"
+    iniSet "fullscreen" "1"
+    iniSet "keep_aspect" "1"
+    iniSet "zoom" "full"
+    iniSet "fsaa" "0"
+    iniSet "scanlines" "0"
+    iniSet "floppy_drive_speed" "100"
+    copyDefaultConfig "$config" "$md_conf_root/amiga/fs-uae/Default.fs-uae"
+    rm "$config"
+    
+    local exts=$(getSystemExtensions amiga)
+    addSystem 1 "$md_id" "amiga" "bash $md_inst/bin/fs-uae.sh '$exts' %ROM%"
 }
--- a/scriptmodules/emulators/fs-uae/fs-uae.sh
+++ b/scriptmodules/emulators/fs-uae/fs-uae.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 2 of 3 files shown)


scriptmodules/emulators/fs-uae.sh
```diff
@@ -9,9 +9,11 @@
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
 #
 
+FSUAE_EXTS=(.adf .ipf .dms .adz .zip)
```
joolswills: no globals here please.

scriptmodules/emulators/fs-uae.sh
```diff
@@ -9,9 +9,11 @@
...
+
 rp_module_id="fs-uae"
 rp_module_desc="Amiga emulator - FS-UAE integrates the most accurate Amiga emulation code available from WinUAE"
-rp_module_help="ROM Extension: .adf"
+rp_module_help="ROM Extension: ${FSUAE_EXTS[*]}\n\nCopy your Amiga games to $romdir/amiga\n\nCopy a required BIOS file (e.g. kick13.rom) to $biosdir"
```
joolswills: I think I would prefer just putting the extensions in platforms.cfg - the emulators themselves document what they support and we have this in other cases where some emulators don't support all formats.

scriptmodules/emulators/fs-uae.sh
```diff
@@ -47,16 +49,18 @@ function remove_fs-uae() {
 function configure_fs-uae() {
     mkRomDir "amiga"
 
-    mkUserDir "$home/.config"
+    # copy configuring start script
+    mkdir "$md_inst/bin/"
```
joolswills: minor, but no need for /

scriptmodules/emulators/fs-uae.sh
```diff
@@ -47,16 +49,18 @@ function remove_fs-uae() {
...
+    cp "$scriptdir/scriptmodules/$md_type/$md_id/fs-uae.sh" "$md_inst/bin/"
```
joolswills: use `$md_data`

scriptmodules/emulators/fs-uae.sh
```diff
@@ -47,16 +49,18 @@ function remove_fs-uae() {
...
+    moveConfigDir "$home/Documents/FS-UAE/Configurations" "$md_conf_root/amiga"
 
-    cat > "$romdir/amiga/+Start FS-UAE.sh" << _EOF_
-#!/bin/bash
-fs-uae-launcher
-_EOF_
-    chmod a+x "$romdir/amiga/+Start FS-UAE.sh"
-    chown $user:$user "$romdir/amiga/+Start FS-UAE.sh"
+    # copy default config file
+    cp -v "$scriptdir/scriptmodules/$md_type/$md_id/Default.fs-uae" "$md_conf_root/amiga"
```
joolswills: tiny config - can use iniSet - also make sure we don't overwrite existing configs on upgrade. (configure should not overwrite existing configs).

scriptmodules/emulators/fs-uae/fs-uae.sh
```diff
@@ -0,0 +1,49 @@
...
+
+exts="$1"
+rom="$2"
+
+rootdir="/opt/retropie"
+configdir="$rootdir/configs"
+config="$configdir/amiga/Default.fs-uae"
+
+user="$SUDO_USER"
+[[ -z "$user" ]] && user="$(id -un)"
```
joolswills: no need for this - the script is not run under sudo.

scriptmodules/emulators/fs-uae/fs-uae.sh
```diff
@@ -0,0 +1,49 @@
...
+home="$(eval echo ~$user)"
```
joolswills: can just use `$HOME` in this case

joolswills:
>I changed the config directory symbolic link from ~/.config/fs-uae to ~/Documents/FS-UAE/Configuration since this is the directory that fs-uae creates on first run to store and read the config from.

are you saying that `~/.config/fs-uae` is not used or created by fs-uae ?

joolswills:
thanks - made a few comments above.

verybadsoldier:
Ok, thanks for your hints, next iteration is here :)

Regarding config path:
> are you saying that ~/.config/fs-uae is not used or created by fs-uae ?

fs-uae reads from `Documents/FS-UAE/Configuration` but also from `~/.config/fs-uae`. The latter is the "Hidden Default Configuration File" as documented here: https://fs-uae.net/configuration-files
As I understand it, the main purpose of the hidden config is to use the parameter `base_dir` to redirect the whole fs-uae path to something custom.
So I think using `Documents` is the more straightforward way but I can also change it to `.config` if you prefer it.

I am still not really happy with the handling of the extensions. I have added them to platforms.cfg but also entered them in `rp_module_help`. Also the start script needs them at runtime. So I pass the extensions as parameter to the start script (read from platforms.cfg), because I think the start script cannot access platforms.cfg (and because platforms.cfg is not specific for the current emulator).
(In general I think it is a bit dirty for emulator scripts to use the extensions information from platforms.cfg because they are not specific for the emulator and also might change later when new emulators are added.)
I can live with it but if you have ideas how to improve it I would be happy to implement them.

Also when invoking fs-uae I removed the parameter which passes a specific configfile so I rely on fs-uae finding the default config (Default.fs-uae) on its own, which should be fine (hopefully). It simplified some lines of code in the start script.

joolswills:
I'm happy with the extensions being in platforms.cfg.

Instead of passing by argument, you could include them directly in the launch script - by adding them in the same way we add extensions in the asplashscreen init.d script as it is installed. or put the script in as a heredoc as with vice, and put the extensions in the script there.

joolswills:
Anyway, it's also fine as it is now, so I will merge - I may tweak a few things later.

verybadsoldier:
Ok, great, again thanks!

PR #1760, opened 2016-11-22
added mehstation to experimental

joolswills:
Has a problem currently, in that mehstation loads etc, but once mehstation has scanned for roms, and saved back to the database, it will crash. Not sure the cause - bug in mehstation or something caused by the way we manually "set up" the db structure ourselves, so we can set up each system etc.

### DIFF
--- a/scriptmodules/supplementary/mehstation.sh
+++ b/scriptmodules/supplementary/mehstation.sh
@@ -36,6 +36,17 @@ function _add_system_mehstation() {
     NAME="$fullname" COMMAND="$command" DIR="$path" EXTS="$extensions" "/opt/retropie/supplementary/mehstation/bin/mehtadata" -db="$db" -new-platform
 }
 
+function _del_system_mehstation() {
+    local fullname="$1"
+    local name="$2"
+
+    local db="$(_get_database_mehstation)"
+
+    [[ ! -f "$db" ]] && return 0
+
+    PLATFORM_NAME="$fullname" "/opt/retropie/supplementary/mehstation/bin/mehtadata" -db="$db" -del-platform
+}
+
 function _add_rom_mehstation() {
     local system_name="$1"
     local system_fullname="$2"
@@ -47,6 +58,8 @@ function _add_rom_mehstation() {
     local db="$(_get_database_mehstation)"
 
     NAME="$4" FILEPATH="$path" PLATFORM_NAME="$system_fullname" DESCRIPTION="$desc" "/opt/retropie/supplementary/mehstation/bin/mehtadata" -db="$db" -new-exec
+
+    RESOURCE="$image" FILEPATH="$path" PLATFORM_NAME="$system_fullname" TYPE="cover" "/opt/retropie/supplementary/mehstation/bin/mehtadata" -db="$db" -new-res
 }
 
 function depends_mehstation() {
--- a/scriptmodules/supplementary/mehstation.sh
+++ b/scriptmodules/supplementary/mehstation.sh
@@ -14,6 +14,10 @@ rp_module_desc="mehstation emulator frontend"
 rp_module_section="exp"
 rp_module_flags="frontend"
 
+function _get_database_mehstation() {
+    echo "$configdir/all/mehstation/database.db"
+}
+
 function _add_system_mehstation() {
     local fullname="$1"
     local name="$2"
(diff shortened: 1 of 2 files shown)


joolswills:
@remeh continuing here as requested

>About the command-line flags, I'm not sure to understand what would be the purpose: completely bypass the database to only work with auto scanning?

It would be useful to be able to add *platforms* to mehstation without having to manually set up the database etc. EG when installing mehstation, I then create the sqlite3 database and add platforms to it manually via sqlite3 commandline. if mehstation hvia a mehstation commandline - currently I create the sqlite3 database and then use sqlite3 via commandline to do this from the RetroPie-Setup scripts. Or it would be ad an "init" function to create the db and quit, I could skip the db creation part.

remeh:
OK I understand and the good news is: when I've started to think about an integration in RetroPie, I thought about that and added this feature.
It isn't in `mehstation` but in the `mehtadata` command line tool https://github.com/remeh/mehtadata (yeah sorry, another binary..)

For example using `mehtadata`, you can create a platform using:
```
NAME="SNES" COMMAND="/usr/bin/retroarch -L /usr/lib/libretro/dalib.so %exec%" DIR="/home/pi/snes" EXTS="zip" ./mehtadata -meh-db=/home/pi/.config/mehstation/database.db -new-platform
```
I should document all the methods offered by `mehtadata` but for now, it documents itself if you call the commands without enough parameters.
What I should add is hiding the platform if no roms are found.

As for the crash, I'm looking into it. I've already pushed a change in mehstation master fixing a crash when trying to show "Last started executable"

joolswills:
Awesome - thanks. I will add your mehdata tool to the module, and use that :-)

so mehdata can also create an initial schema if none exists ?

remeh:
Arf, you're right, it currently doesn't have a method to init the schema. I add this asap!

remeh:
I've added a command to `mehtadata` to play the sql files shipped with mehstation in order to initiate the database. 

This way, you can do:
```
SCHEMA=/path/to/mehstation/1.sql mehtadata -db=/path/to/database.db -init
SCHEMA=/path/to/mehstation/2.sql mehtadata -db=/path/to/database.db -init
```

in order to not ship SQL in mehstation.sh

joolswills:
Thanks - I didn't actually notice that mehstation included the schema creation/updates as separate files - would have saved me dumping the schema via sqlite3. Will add mehtadata to the module and switch to using that.

joolswills:
I switched to mehtadata for db init / platform adding now. I still have some sql in the script though as I need to check if a platform already exists before adding it - perhaps mehtadata could do this check? (Unless there is a case where someone might want multiple platforms with the same name).

remeh:
I've pushed a change to mehtadata avoiding creating many platforms with the same name.
The return code is still 0 in this case as it's kind of a no-op.

joolswills:
thanks! :)

joolswills:
Regarding the adding platform change to check first, could it be changed to updates the command/extensions rather than just being a no-op. I will need to "update" platforms with retropie-setup using the "name" as a key. Not a problem if you don't want it to work like this, but then I will have to use sql directly.

I need to do the same for executables too - adding if it doesn't exist, or updating an existing one (based on path). I don't think your commandline does this. I can use sql again, but I thought i'd check with you first. The RetroPie-Setup script will use this to set up items in the "RetroPie" platform (which we use for accessing RetroPie configurations etc)

remeh:
No problem I'll do the modification for the 'update' which makes a lot of sense.
Same for executables, I'll do it asap. It'll be a good addition to `mehtadata` which intend to be the CLI equivalent of mehstation-config.

remeh:
OK it's pushed in mehtadata. See flags `-new-exec`. It supports upsert (based on filepath) and so does `-new-platform` (based on platform name).

joolswills:
Thanks very much. I will add the code so the "RetroPie" menu can be generated (via `-new-exec`)

joolswills:
Another useful facility (which would be used by retropie) would be the ability to delete a system (and possibility an executable). RetroPie would be then able to remove a system from mehstation if all emulators for that system had been removed.

joolswills:
PR is updated to include functionality so the RetroPie menu is correctly added with descriptions via the `-new-exec` functionality. Please could we also have the ability to add resources ? :) I would like to add the images to go with the retropie entries.

remeh:
Roger that! I'll add these features during the weekend.

joolswills:
Thanks very much for your support.

remeh:
Here you go, I've added:
* `-new-res` to create resources
* `-del-platform` to delete a platform (automatically deleting all executables of this platform)
* `-del-exec` to delete one executable

I'm currently thinking on hiding platform without any configured executable.  A bit complicated for platforms in "discover" mode but I'll try something.

Thank you for the work on this PR in RetroPie.

joolswills:
Many thanks. Hiding would be useful and would match the Emulation Station behaviour and would mean we can add all platforms and not worry about it from RetroPie (currently the mehstation module only adds a platform if the folder is not empty)

remeh:
Hi @joolswills 
I've added this feature in the master of mehstation: you can now freely add all platforms in discover mode using mehtadata and they won't appear if no executable is available.
They'll automatically appear after a restart of mehstation if there is executables appearing in the directory. Successfully tested on a rasp3. (I'll improve it later to not even need the restart of mehstation)

PR #1765, opened 2016-11-24
update gliden64 config version

gizmo98:
https://github.com/gonetz/GLideN64/blob/master/src/Config.h#L7


gizmo98:
Without the right config version custom settings will be overwritten.

joolswills:
Thanks. Will merge and rebuild binaries

PR #1766, opened 2016-11-25
hatari - disable quit confirmation, disable statusbar. use fullscreen or x11

verybadsoldier:
I would like to propose these config changes for hatari:
- disable statusbar (permanently visible otherwise)
- disable quit confirmation (needs an extra click otherwise to quit the emu. Other emulators also quit instantly on SIGTERM)
- added fullscreen option for x11


joolswills:
thanks.

PR #1770, opened 2016-11-29
fix snap path for attractmode

amigagamer:
using $path/$snap is defaulting to the rom directory as $snap is not
defined.


joolswills:
That's not the correct fix. it broke when I did the "singular" path change as I forgot about the snap variable. I've done another fix though.

amigagamer:
Many thanks for the fix and apologies- i should have reviewed the previous commits before submitting a quick fix- glad its resolved

PR #1776, opened 2016-12-04
Add files via upload

CrescendoFang:
icon for RetroPie-Manager in the RetroPie ES menu


joolswills:
The RetroPie-Manager doesn't need to be on the RetroPie menu imho. Please give more information if wanting to submit changes to the project.

CrescendoFang:
It doesn't need to be in ES > RetroPie menu, however IT IS there if you install the PetroPie-Manager, without icon.....

So the rest is good with its respective icons, only the RetroPie-Manager....

You can check it yourself, the script specifically creating retropie-manager.rp

joolswills:
ok thanks. I'll remove that bit of code - this was a contributed module originally and I missed that.

CrescendoFang:
errr actually leave it there, its really useful to those who want to run the manager only when you need it.

like me... I hate to leave it run all the time..... I only need it to run whenever I need it (like when copying the games over network from office to home raspberry)....

And I can edit the icon to remove the recalbox word there if you wish...

joolswills:
You can add it manually if you want to by creating a `retropie-manager.rp` file yourself, but I don't want it added automatically (and the code that did it, did it in a way that was not compatible with some other recent changes).

CrescendoFang:
ah, OK then..... 
Thanks for the clarification.....

PR #1780, opened 2016-12-07
Add Minivmac

HerbFargus:
This is just a preliminary PR for review, 

So as a preface,from the minivmac [site](http://www.gryphel.com/c/minivmac/options.html#option_t):

 By default, Mini vMac emulates a Macintosh Plus. But it also can be compiled to emulate a few other computers, with one of these lines:

-m 128K { Macintosh 128K }
-m 512Ke { Macintosh 512Ke }
-m Plus { (default) Macintosh Plus }
-m SE { Macintosh SE }
-m Classic { Macintosh Classic }
-m SEFDHD { Macintosh SE FDHD }
-m II { Macintosh II * }

This PR only handles the Macintosh Plus (some of the options available may also be emulated with Basilisk II)

There are some SDL issues with the mouse pointer with the default compile options through the main minivmac source but our good friend vanfanel has patched them up in his fork which this PR uses (though I don't know which compile options he used for his build, I just know it works)

Just as a quick reference, minivmac has a diff building method, you set up the build environment from minivmac, then you compile it on the device you're using. 

So if I were to build from the main source on my pi, I'd download the ARM binary http://www.gryphel.com/c/minivmac/download.html Then I'd add the required vMac.ROM bios, then the System Tools.dsk for the OS, then I'd take the extracted binary, and start that up and add the compilation flags, so If I wanted to compile for the pi on the Pi I'd add

`-t larm -api sd2`

There are a lot more build options listed [here](http://www.gryphel.com/c/minivmac/options.html)

After compilation it shoots out the build environment to be compiled in the `out` folder, and then you can compile it from there.

This first pass will just be for the mac plus but its possible we could expand it to all the mac variants once we sort the sdl2 issues. 

As far as the modules is concerned I'm not sure the best way to integrate the dsk files, its like basilisk in that you need the disk image and the OS, but I don't necessarily think the start script is needed as we can just make the assumption the user will provide the bios/OS and then can just launch disks like they would games if those are there. idk. open to ideas and thoughts.

### DIFF
--- a/scriptmodules/emulators/minivmac.sh
+++ b/scriptmodules/emulators/minivmac.sh
@@ -11,7 +11,7 @@
 
 rp_module_id="minivmac"
 rp_module_desc="Macintosh Plus Emulator"
-rp_module_help="ROM Extensions: .dsk \n\nCopy your Macintosh Plus disks to $romdir/macintoshplus \n\n You need to copy the Macintosh bios files vMac.ROM and System Tools.dsk to $biosdir"
+rp_module_help="ROM Extensions: .dsk \n\nCopy your Macintosh Plus disks to $romdir/macintoshplus \n\n You need to copy the Macintosh bios file vMac.ROM into "$biosdir" and System Tools.dsk to $romdir"
 rp_module_section="opt"
 rp_module_flags="!mali"
 
@@ -36,7 +36,6 @@ function configure_minivmac() {
     mkUserDir "$md_conf_root/macintoshplus"
 
     ln -sf "$biosdir/vMac.ROM" "$md_inst/vMac.ROM"
-#   ln -sf "$biosdir/vMac.ROM" "$md_inst/vMac.ROM"
 
     addSystem 1 "$md_id" "macintoshplus" "pushd $md_inst; $md_inst/minivmac $romdir/macintoshplus/System\ Tools.dsk %ROM%; popd" "Apple Macintosh Plus" ".txt .dsk"
 }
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Will give it a spin.

joolswills:
Looks good - the addSystem line references a disk - should that be removed or ?

HerbFargus:
Depends. If people want to launch a game directly it preloads the OS disk first through the launch script assuming it's there, if it's not it obviously won't work

joolswills:
Sorry, I didn't read the last paragraph carefully. So it needs a system disk by default - that can be documented. Do you think it's worth keeping it in experimental for a short time or ?

joolswills:
(I've not tested it well yet - but if stable etc we can leave it in opt)

joolswills:
(updated previous comment)

HerbFargus:
I'd be keen to keep it in experimental until we can come up with a more elegant solution along with having more testing etc

joolswills:
are you ok to make that change, and I will squash/merge?

HerbFargus:
Should be good. When I have time maybe I'll look at improvements but don't know when that will be. Will be good to get feedback from anyone who wishes to test it.

joolswills:
thanks.

Mu0n:
I just acquired my first raspberry pi and as a long time user of mini-vMac for the PC, I had to try it there as well. I already had loads of bootable disks (mostly in .hfv form, but they're easily converted to .dsk). The thing I'm missing most is to tell the emulator that you're mounting a new disk while the emulator is running. Many pieces of software fit on multiple disks and oftentimes, only the first one was bootable. Some can be combined in an artificially larger .dsk that fits everything, some can't, they expect multiple disks. The PC port has a file menu that you can use outside of the emulated machine to virtually insert new disks at runtime. It could be a challenge to port this effectively. Unless there are already shortcuts in place I'm not aware of?

joolswills:
@Mu0n use the retropie forum for support.

PR #1781, opened 2016-12-07
launching image per game

meleu:
Show a launching image for a specific game if "Launch menu art" is enabled **and** there's a file named `RetroPie/roms/SYSTEM/images/RomName-launching.png` (or `.jpg`).

Inspired by [this user request](https://retropie.org.uk/forum/topic/4611/runcommand-system-splashscreens/100).

I tested with some of the amazing lilbud and Rookerviks images from [retropie-splashscreens-extra](https://github.com/HerbFargus/retropie-splashscreens-extra) (final fight, castlevania, TMNT, megaman, etc.) and it looks really nice.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -846,6 +846,7 @@ function show_launch() {
     if [[ "$USE_ART" -eq 1 ]]; then
         # if using art look for images in paths for es art.
         images+=(
+            "$HOME/RetroPie/roms/$SYSTEM/images/${ROM_BN}-launching"
             "$HOME/RetroPie/roms/$SYSTEM/images/${ROM_BN}-image"
             "$HOME/.emulationstation/downloaded_images/$SYSTEM/${ROM_BN}-image"
         )


joolswills:
The other launching images don't require use_art to be on, so I think the logic should be the same, but it feels a bit overkill. I don't mind adding it, but would like the logic changed.

joolswills:
I'm not too happy with the current logic either, regarding the launching images vs the use_art setting - it's all getting a bis confusing and it feels like the launching stuff was just hacked on (which it sort of was). I might have a rethink about the implementation.

joolswills:
I think I'm going to change it so all the launching images require the `use_art` to be on. this is a logic change but seems the simplest without adding more settings (which would be more overhead). this would also mean we have less to check if not using art.

Would you be ok with this?

joolswills:
Of course then, someone might want to use the launch art, but not show the game art - so it's not ideal either.

meleu:
> I think I'm going to change it so all the launching images require the `use_art` to be on.

My only concern is about the preference of the searching order. My personal taste is:

1. specific game launching image (the feature of this PR).
2. system launching image.
3. scraped art (which I don't like at all).

But I'm not sure if it's the preference of every user...

I like the way it is implemented today is fine because there's a way to choose the order of searching for the images (but I agree that it isn't very clear to the users).

Well, if you are going to change the logic, may I do a feature request? :-)
Implement a way to choose the searching order.

joolswills:
After more thought, I will keep the logic the same for now so if use_art is on and there is an image, it will take priority.

```diff
diff --git a/scriptmodules/supplementary/runcommand/runcommand.sh b/scriptmodules/supplementary/runcommand/runcommand.sh
index 6d7f073..4de098b 100755
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -852,11 +852,14 @@ function show_launch() {
     fi
 
     # look for custom launching images
-    [[ $IS_SYS -eq 1 ]] && images+=("$CONF_ROOT/launching")
+    if [[ $IS_SYS -eq 1 ]]; then
+        images+=(
+            "$HOME/RetroPie/roms/$SYSTEM/images/${ROM_BN}-launching"
+            "$CONF_ROOT/launching"
+        )
+    fi
     [[ $IS_PORT -eq 1 ]] && images+=("$CONFIGDIR/ports/launching")
-    images+=(
-        "$CONFIGDIR/all/launching"
-    )
+    images+=("$CONFIGDIR/all/launching")
 
     local image
     local path
```

meleu:
By the way, I've realized that all that logic description I put on runcommand's wiki can be simplified by this sentence: "**be aware that if "Launch Menu Art" is enabled, the scaped box art image takes precedence.**". :-)

joolswills:
:-)

PR #1788, opened 2016-12-13
Add standalone option to install Cave Story

RetroPieNerd:
Add any changes you want to the script I made... They are probably some mistakes made (not sure...).


joolswills:
Binary only? This won't work on other platforms - you would need to exclude them. Sorry, but I dont think this is needed, but you can submit it to zerojays repo perhaps.

PR #1789, opened 2016-12-15
Add improved standalone option to install Cave Story

RetroPieNerd:
I'm back with a monster of a script... I took your suggestions, @joolswills , and improved the script with it!
Thanks.
(By the way, binary is RPI and x86 only.)


joolswills:
Thanks, and I appreciate you have gone away to improve on this - but going back to my earlier reply on the other PR, I don't think it's something I want right now. I would recommend to  submit it to zerojays repo or just make it available yourself. There are some issues with the module/functions also - but I don't have time to do a review.

PR #1797, opened 2016-12-21
joy2key - support reading from multiple joysticks

joolswills:
* if device parameter is /dev/input/jsX it will read from all joysticks
 * if a joystick is removed or another reconnected it will be picked up
 * default runcommand and helpers to use all joypads

if anyone has a chance to test this it would be appreciated. Thanks.


meleu:
Great!
Tested here with my two 8bitdo controllers (NES30 and Zero) and both can control the dialogs! :-)

joolswills:
Thanks. Will merge.

PR #1803, opened 2016-12-26
fix ti99 paths so bios actually goes in bios folder

HerbFargus:
see also:

https://retropie.org.uk/forum/topic/3033/getting-ti99-emulator-to-work

https://retropie.org.uk/forum/topic/5649/ti-99-emulator-not-working

https://retropie.org.uk/forum/topic/4299/unable-to-locate-console-roms-on-ti99sim

there is also a 1.4 update as of June http://www.mrousseau.org/programs/ti99sim/archives/ not sure if we want to update to that as well and perhaps incorporate some platform specific building with armhf etc. also not sure how much of a priority this system is ;)

### DIFF
--- a/scriptmodules/emulators/ti99sim.sh
+++ b/scriptmodules/emulators/ti99sim.sh
@@ -36,7 +36,7 @@ function install_ti99sim() {
 function configure_ti99sim() {
     mkRomDir "ti99"
     moveConfigDir "$home/.ti99sim" "$md_conf_root/ti99/"
+    ln -sf "$biosdir/TI-994A.ctg" "$md_inst/TI-994A.ctg"
 
-    addSystem 1 "$md_id" "ti99" "pushd $romdir/ti99; $md_inst/ti99sim-sdl -f %ROM%; popd" "TI99" ".ctg .CTG"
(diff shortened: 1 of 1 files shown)


scriptmodules/emulators/ti99sim.sh
```diff
@@ -36,7 +36,7 @@ function install_ti99sim() {
 function configure_ti99sim() {
     mkRomDir "ti99"
     moveConfigDir "$home/.ti99sim" "$md_conf_root/ti99/"
+    ln -sf "$biosdir/TI-994A.ctg" "$md_inst/TI-994A.ctg"
```
joolswills: you should use "moveConfigFile" to symlink the file (not just for configs, but for other things too).

HerbFargus: guess I'm still stuck on legacy functions (eg. gba script as well has it etc.) will fix

joolswills: yeah, that can be changed too. actually, dont worry, Ill merge as is my brain is still not functioning correctly after high fever for 3 days and I will revisit when I can think clearly (still have a fever - should be in bed!)

HerbFargus: it should at least function for now in the interim. I'm sure there will be another legacy code cleanup and we can address the inadequacies of this script at that point. In the meantime get better!

PR #1806, opened 2016-12-28
remove some entries from retropiemenu for non-rpi installations

meleu:
After installing RetroPie on a x86 I noticed raspiconfig on RetroPie menu (with no icon).
These changes remove unnecessary entries on RetroPie menu for non-rpi platforms.


joolswills:
Was due to some code reordering but this doesn't look like the right fix (removing after).

joolswills:
I will need to look over it again. Thanks though.

joolswills:
Fixed in 60e3f67a6868a49a2b3eb758d3562bd4088823e5 (which needed 0e4a45ddd0d4998b1be827d8ba875ce992aad695 also for mehstation)

PR #1807, opened 2016-12-28
usbromservice - support mounting of usb stick over ~/RetroPie

joolswills:
* add a new mount.d script to allow mounting of a usb stick directly over ~/RetroPie
 * script will check for existance of a retropie-mount folder in the root. If it exists and is empty, the existing ~/RetroPie folder will be copied to it. Then it will be mounted.

Thought I would go for a simple option for allowing users to have their ~/RetroPie folder on a usb stick using the existing usbromservice module.

as mentioned above - you just need a folder in the root called "retropie-mount" and then that folder will be automatically mounted over `~/RetroPie/`

can add some simple gui later to save people having to do this by having a "prep usb stick" function or so, but simple enough to do this on a pc.

Only possible issue is if for some reason ES loads before it is mounted, but it didn't in my tests. If so we might need to add a delay to ES loading if using this or something.

### DIFF
--- a/scriptmodules/supplementary/usbromservice.sh
+++ b/scriptmodules/supplementary/usbromservice.sh
@@ -39,9 +39,15 @@ function install_usbromservice() {
 }
 
 function enable_usbromservice() {
-    cp -v "$md_data/01_retropie_copyroms" /etc/usbmount/mount.d/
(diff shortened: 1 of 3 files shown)


HerbFargus:
seems to be working on initial tests, still haven't tested too thoroughly though

joolswills:
thanks.

Issue #1809, opened 2016-12-29
Pi3 with experimental bootcode (BRANCH=next rpi-update) __platform variable not set in system.sh


petersteger:
in system.sh
get_platform wants BCM2709 to set the variable to pi3.
But with the newest bootfiles (i did an unpdate to boot it from USB) it is BCM2835
please include in select case.

joolswills:
Unfortunately it will require more than just that - https://github.com/raspberrypi/firmware/issues/705

It's no quick fix - need to parse the Revision number.

PR #1812, opened 2016-12-30
Add different rpi3 hardware name

gizmo98:
https://github.com/RetroPie/RetroPie-Setup/issues/1809

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -194,6 +194,13 @@ function get_platform() {
                     __platform="rpi2"
                 fi
                 ;;
+            BCM2835)
+                if [[ "$architecture" == "aarch64" ]]; then
+                    __platform="rpi3-64"
+                else
+                    __platform="rpi3"
+                fi
+                ;;
             ODROIDC)
                 __platform="odroid-c1"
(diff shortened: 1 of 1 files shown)


joolswills:
I will need to confirm this - also it seems strange as - https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2835/README.md

joolswills:
Also why like this and not added as part of the previous `BCM2709|BCM2835)` etc ?

joolswills:
Sorry I think I misread the above link - thought it meant it was only used in zero/1 etc. I still want to confirm this first and I think the change should be added as I mentioned above if it is needed, rather than an addition block.

joolswills:
https://github.com/raspberrypi/firmware/issues/705

joolswills:
Actually I did read the link ok - RPI2 uses a BCM2836, so not sure why it would report this for a RPI3. Doesn't sound right.

joolswills:
Looks like this will be returned for rpi zero and rpi1 also so no good - I will close this and come up with something based on what response I get from this ticket - https://github.com/raspberrypi/firmware/issues/705

PR #1814, opened 2016-12-31
Split addSystem to addEmulator / addSystem

joolswills:
Currently we have just addSystem which is used to add emulators and systems all at the same time. This is fine, except in some cases where we have multiple calls - eg for fuse where we have a 48k and 128k "emulator" (calling fuse with different parameters etc). Right now this causes the frontend system to be updated twice. 

Now we have other frontends, it would be better to just try and add a system once as needed. Also since the addSystem call handled the removal, we would also try and "remove" it multiple times when there was more than one emulator entry for that module.

By splitting it, we can now just add multiple emulators, and then add the system afterwards (it has to be in the order due to when in removal mode - system must be removed last).

As well as avoiding adding/updating a system multiple times it makes the code cleaner as the functions do what the name suggests, rather than one function that does multiple tasks.

Note backward compatibility is included for 3rd party repos like @zerojay 's - but they should be updated soon as it may be removed in the future.

This is quite a large patchset and there may be bugs, so I will give it some more testing before merging - feel free to test.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -43,6 +43,9 @@ coleco_fullname="ColecoVision"
 c64_exts=".crt .d64 .g64 .prg .t64 .tap .x64 .zip .vsf"
 c64_fullname="Commodore 64"
 
+daphne_exts=".daphne"
+daphne_fullname="Daphne"
+
 dragon32_exts=".cas .wav .bas .asc .dmk .jvc .os9 .dsk .vdk .rom .ccc .sna"
 dragon32_fullname="Dragon 32"
 
@@ -70,6 +73,12 @@ gbc_fullname="Game Boy Color"
 
 intellivision_exts=".int .bin"
 intellivision_fullname="Intellivision"
+
+macintosh_exts=".txt"
+macintosh_fullname="Apple Macintosh"
+
+macintoshplus_exts=".txt .dsk"
+macintosh_fullname="Apple Macintosh Plus"
 
 mame-advmame_exts=".zip"
 mame-advmame_fullname="Multiple Arcade Machine Emulator"
@@ -121,6 +130,10 @@ gc_fullname="Nintendo Gamecube"
 wii_exts=".iso"
 wii_fullname="Nintendo Wii"
 
+ports_exts=".sh"
+ports_fullname="Ports"
+ports_platform="pc"
+
 pc_exts=".bat .com .exe .sh"
 pc_fullname="PC"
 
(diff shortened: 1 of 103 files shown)


joolswills:
Oh, and there is another change to use the platforms.cfg for the arcade "platforms/theme" - this simplifies the calls as we had cases where we were passing "system platform theme" as a single parameter to the old addSystem.

joolswills:
I forgot to move some full name/extensions to the addsystem call so some more work needs doing on this

scriptmodules/emulators/oricutron.sh
```diff
@@ -59,8 +59,9 @@ function configure_oricutron() {
     for machine in atmos oric1 o16k telestrat pravetz; do
         default=0
         [[ "$machine" == "atmos" ]] && default=1
-        addSystem 1 "$md_id-$machine" "oric" "pushd $md_inst; $md_inst/oricutron --machine $machine %ROM% --fullscreen; popd" "Oric 1/Atmos" ".dsk .tap"
+        addEmulator 1 "$md_id-$machine" "oric" "pushd $md_inst; $md_inst/oricutron --machine $machine %ROM% --fullscreen; popd" "Oric 1/Atmos" ".dsk .tap"
```
HerbFargus: needs to be added to the platform.cfg

joolswills:
Yep - I've not finished yet :)

joolswills:
This *should* be ready now - but there may be some bugs. I'm doing some final testing so if you notice any typos mistakes please do let me know. I just fixed a couple of typos, and there may be more.

platforms.cfg
```diff
@@ -67,19 +74,31 @@ gb_fullname="Game Boy"
...
+love_fullname="Love"
+
 intellivision_exts=".int .bin"
 intellivision_fullname="Intellivision"
 
+macintosh_exts=".txt"
+macintosh_fullname="Apple Macintosh"
+
+macintoshplus_exts=".txt .dsk"
+macintosh_fullname="Apple Macintosh Plus"
```
HerbFargus: macintoshplus_fullname ?

scriptmodules/libretrocores/lr-picodrive.sh
```diff
@@ -42,17 +42,11 @@ function install_lr-picodrive() {
...
-    ensureSystemretroconfig "mastersystem"
-    ensureSystemretroconfig "segacd"
-    ensureSystemretroconfig "sega32x"
-
-    addSystem 1 "$md_id" "mastersystem" "$md_inst/picodrive_libretro.so"
-    addSystem 1 "$md_id" "megadrive" "$md_inst/picodrive_libretro.so"
-    addSystem 1 "$md_id" "segacd" "$md_inst/picodrive_libretro.so"
-    addSystem 1 "$md_id" "sega32x" "$md_inst/picodrive_libretro.so"
+    local system
+    for system in megadrrive mastersystem segacd sega32x; do
```
HerbFargus: megadrive (only with one R ;) )

PR #1822, opened 2017-01-05
Add Amiga file extension ".uae" to platforms.cfg

dumell:
In addition to the existing disk image extensions such as .adf and dms, the emulator configuration file extension .uae/.UAE is necessary to boot into hard disk images and whdload packages. This way the emulator can be launched with a .uae configuration file containing the path to the hard disk image and any special configurations that image may require.

Whdload packages are often fixed to have less strict hardware/bios requirements and removes the need to swap disks as all files are located in one hard disk image.

An example where .uae is required is the whdload booter for RetroPie: 
http://www.ultimateamiga.co.uk/index.php?topic=9930.0

Adding .uae to the emulators default extensions does not automatically make whdload booting work but makes it simpler - without adverse effects for those using traditional disk images such as .adf and .dms.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -1,7 +1,7 @@
 3do_exts=".iso"
 3do_fullname="3DO Interactive Multiplayer"
 
-amiga_exts=".adf .adz .dms .ipf .zip"
+amiga_exts=".adf .adz .dms .ipf .uae .zip"
 amiga_fullname="Commodore Amiga"
 
 amstradcpc_exts=".cdt .cpc .dsk"


joolswills:
I would prefer to add the .uae extension once it is supported by out launch script (which is something that is planned). In the meantime, you can override it in a `/opt/retropie/configs/all/platforms.cfg`

dumell:
Thank you for the consideration.

Are you aware that there is a a working whdload launch solution for uae4arm on RetroPie? This might be of interest to you:
https://www.facebook.com/groups/1854320841462593/ (Unfortunately the group is still a closed one so you will have to ask to be accepted).

Hopefully a solution like this will one day be part of the default installation but for the time being, having .uae as a default extension would make using this and similar solutions easier.

HoraceAndTheSpider:
Just to confirm this is correct - launching of uae files direct from uae4arm launch is in fact more reliable using the "-config=" option than the "permitted" .adf and .ipf support which are now included as recognised Amiga files

joolswills:
I would like to do more than just pass the ROM name to `-config` hence why I am not adding the extension yet - eg I would like to add support directly for zips, and launching adfs etc directly.

HoraceAndTheSpider:
Supporting ADFs is nonsense though - you will have no way of regulating compatibility , since uae can emulate a vast number of Amiga (even kickstart 1.2 vs 1.3 can have issues, never mind AGA vs ECS)

The .uae config support resolves such issues by including the necessary settings to improve both compatibility and playability. 

The Amiga whd RetroPie group has over 400 members and this group specifically targets .uae usage this way, for the reasons outline by myself and compatriots above. Unfortunately all of those users have to presently go through the task of manually adjusting the emulator support , which just adds an unnecessary complexity to the emulation process.

Please consider repriortising the .uae support which is already working over the .adf and .zip options which offer far less benefit.

joolswills:
Also as i wrote above, you need a single line in `/opt/retropie/configs/all/platforms.cfg` with `amiga_exts="whatever"` if you want to override the default extensions. Since the launch parameters need to be done manually also, it seems of no benefit to add the uae extension until it is supported out of the box.

HoraceAndTheSpider:
Yes changes are already being made manually to do this as explained above. 

You are saying that only I f the uae4arm devs commit a change to support launching " uae4arm game.uae"  then will it be added? I will ask midwan to commit such a change to resolve this.

Default a500/a1200 spec still won't cover it an adf launch. What about playing say Foundation with a gfx card? How will you detect AGA requirements from a disk name? 

Adf is fine for the genuine Amiga emulation requirement, but with RetroPie people want a "click and play" console style response. This is exactly what the Facebook group is looking to achieve and would seriously recommend you join and see the good work being done there and how much it is only looking to benefit a large number of users - there is no want to be difficult here but I feel that an unnecessary difficulty is being created by a drive to prioritise the wrong items (adf/zip etc support)

joolswills:
I am already working on it as I said. 

I know a default a500/a1200 config won't cover everything, but it will be fine for the majority of games for those that don't want to make a .uae file. I didn't say adf support will be INSTEAD of .uae support.

I am not prioritising adf/zip support. It's the same launch code that will handle it - it makes sense to do both at the same time.

joolswills:
I am not saying anything about uae4arm support for .uae files. I am referring to the RetroPie launch script which will handle it.

joolswills:
I would also add, your page you linked earlier references a 1.9gb archive of copyrighted Amiga games. I assume the closed facebook group is also sharing games. Please do not post sites which have links to copyrighted games here (or on the RetroPie forum). It damages the project.

dumell:
My apologies for the inappropriate links.

I look forward to the additions you are working on. Thank you.

PR #1828, opened 2017-01-09
Added Mayhem 2 to ports

martinohanlon:
This is to add Mayhem 2 (a port and extension of an old Amiga game) to the optional ports packages.

Find more out about Mayhem 2 at https://github.com/martinohanlon/mayhem-pi

### DIFF
--- a/scriptmodules/ports/mayhem2.sh
+++ b/scriptmodules/ports/mayhem2.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="mayhem2"
+rp_module_desc="Mayhem 2"
+rp_module_section="opt"
+rp_module_flags="!mali"
+
+function depends_mayhem2() {
+    getDepends libdumb1 libvorbisfile3 libopenal1
+}
+
+function install_bin_mayhem2() {
+    gitPullOrClone "$md_inst" https://github.com/martinohanlon/mayhem-pi
+}
+
+function configure_mayhem2() {
+    addPort "$md_id" "mayhem2" "Mayhem 2" "pushd $md_inst; $md_inst/mayhem2-pi; popd"
+
+    chmod +x "$md_inst/mayhem2-pi"
+}


joolswills:
Thanks.

If source is available it should build from source rather than installing a binary - also it needs to exclude other platforms if no support exists. Alternatively you could submit it to https://github.com/zerojay/RetroPie-Extra 

dependencies should be the -dev packages also (once switched to a source install)

martinohanlon:
I didn't want to compile from source as it will require also cloning and compiling of allegro 5 a process which takes considerable time and resource, as it is not available via apt.

Can you provide some advice re how I exclude the install from other platforms?

joolswills:
`rp_module_flags="!x11 !mali"` for example

As a binary only install for the rpi, this isn't something I want to include with retropie currently sorry. It may be more suitable for the repository I mentioned above. thanks.

martinohanlon:
Ok..

I can provide an install to compile from source. If I do so, will you accept?

joolswills:
It depends on the module tbh - we have a lot of things in retropie and it's a lot of work to maintain it all (and there have been modules submitted by other people that I have had to spend a lot of time fixing up/maintaining later). I hadn't realised about allegro5, but if it needs it, then that would need to be built first (and not installed to /usr/local but either statically linked, or installed to same folder etc).

I can still provide binaries for rpi, built from the retropie module, but it would be good to support other retropie platforms. Does this code work on X86/X11 etc or does it have rpi specific code in ?

I see in source history, it was ported by you from allegro4 (as well as other changes etc)  - did allegro 5 offer advantages on the rpi ?

If you are keen to have this in retropie-setup soon for users, the other repo may still be your best bet. thanks.

martinohanlon:
Ok, so its on me to create a good installation limiting maintenance, Im pretty sure I can achieve this and if not, Im comfortable it wont be included.

There is a specific version which works on X86, so I could provide a script to install on either platform, my preference would be to get this working on the Pi first.

Allegro 4 is effectively dead, there are numerous improvements in moving to allegro 5 regardless of platform, for the Pi moving to an event based system significantly improved performance and also meant that I can guarantee some longevity to the project.

So I think I have 2 options for allegro 5:

1. clone and compile allegro 5 as part of the build process and statically link libraries when compiling the mayhem 2 executable, this is what I do when building mayhem 2 anyway
2. include specific allegro.so with the mayhem 2 source code, and statically link these when compiling 

Thoughts?

joolswills:
I prefer option 1 even though it's more work - the only binary only stuff we have left in retropie is due to projects being closed source etc so it would be nice to keep it like this. Also makes it easier to expand it to linux/x86 later I should think.

zerojay:
Hey guys, I already had a mayhem installation script in my repo and just recently discovered the switch made to Allegro5 and had to start working on fixing my script. Well, it appears that allegro5 is now included in Jessie so there's no longer a need to compile it anymore.

I'm working on updating my install script however when the game launches, it tells me it couldn't initialize the pulse backend, but hey, if the inclusion of Allegro5 binaries in Jessie means that you guys can work this out and add it to this repo, all the better.

zerojay:
I decided to take this over to https://github.com/martinohanlon/mayhem-pi/issues/29 for further discussion so as not to bother anyone.

PR #1829, opened 2017-01-09
kill "feh" when invoking runcommand menu

meleu:
When running under X and showing the launching image with `feh`, if the user "press a button" to access the runcommand menu, the image stays in front of the terminal. This change kills the `feh` if the user "press a button".


joolswills:
Thanks

PR #1838, opened 2017-01-13
add support for Devuan (fully Debian compatible)

AlisterAmo:
Devuan 1 is a fully compliant Debian 8 Jessie, with the only modification of the init system (it gets rid of systemD).
100% working at the time of submitting.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -87,6 +87,12 @@ function get_os_version() {
                 __platform_flags+=" xbian"
             fi
             ;;
+         Devuan)
+            # Devuan 1 === Debian 8 Jessie, with the only modification
+            # of a different init system (it gets rid of systemD)
+            # so any Devuan version meets the requeriments
+            foo=bar
+            ;;
         LinuxMint)
             if compareVersions "$__os_release" lt 17; then
                 error="You need Linux Mint 17 or newer"


scriptmodules/system.sh
```diff
@@ -87,6 +87,12 @@ function get_os_version() {
                 __platform_flags+=" xbian"
             fi
             ;;
+         Devuan)
+            # Devuan 1 === Debian 8 Jessie, with the only modification
+            # of a different init system (it gets rid of systemD)
+            # so any Devuan version meets the requeriments
+            foo=bar
```
joolswills: is this here as a "noop" ? This PR needs a little work - what is `__os_debian_ver` set to in this case ?

joolswills: Sorry, i mean `__os_release`

AlisterAmo:
Hi again and sorry for the delay.
* You mean the foo=bar? A noop, yes.
* devuan __os_release is "1", hence the comments I wrote expecting to clarify that. Devuan 1 == Debian 8

Cheers!

joolswills:
It was `1.0` not 1 which is why I was clarifying but I didn't use it in the end.

joolswills:
ps. can use `:` for a no-op rather than setting a global variable.

joolswills:
(I don't think it actually needed a no-op in this case anyway)

AlisterAmo:
Fine!
My intention was more to show that possibility of adding support to devuan with no effort, maybe a small comment or issue somewhere else had been better idea instead of documenting that in a request. And sorry for the bad habits writing shell code, too! My bad altogether! :-P
Cheers, and thanks for the great job.
Alex

joolswills:
:) Thanks for the heads up - I didn't test it much on Devuan - there may be an issue with a few of things that are linked to the init.d system (eg the autostart functionality).

PR #1839, opened 2017-01-15
Add .car files to Atari800

ChoccyHobNob:
.car files are Atari 8-bit cart images created from .bin files by the
Atari800 Emulator itself. They add a metadata header telling Atari800
what type of cart it is and saving the user from entering it manually.
TOSEC managed to scramble some carts by just renaming the .car files to
.bin in a lot of cases, not realising there was a difference. In
retropie, using a .car file in preference to a .bin file will save a
step for the user.


joolswills:
thanks!

PR #1844, opened 2017-01-18
Hook ODROID-C1 into lr-mupen64plus.sh properly

hemi426:
This fix builds lr-mupen64plus properly on ODROID-C1. I have a C2 on the way, and it looks like the fix will be the same for it. If there's a better way to fix this, please let me know and I'll roll it properly, then submit another pull request.


joolswills:
Thanks.

PR #1848, opened 2017-01-21
Add DraStic DS Emulator

PokeEngineer:
The person who did the RPI port recently put out the link to the test port on [this forum post](https://pyra-handheld.com/boards/threads/drastic-nintendo-ds-emulator.73431/page-83#post-1403469), and it's actually playable!

Tested:
- using a Pi 3
- using the Super Mario 64 DS (USA version) ROM
- using the Pokémon HeartGold (USA version) ROM

Results:
- Game speed is usually around 100%
- Smooth, continuous sound

Bugs:
- After you exit the emulator, you can't start anymore applications or else the Pi freezes... The only way around this (at least right now) is to restart the Pi.

### DIFF
--- a/scriptmodules/emulators/drastic.sh
+++ b/scriptmodules/emulators/drastic.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

PR #1849, opened 2017-01-21
Bug fixes for the DraStic script

PokeEngineer:


joolswills:
Couple of questions as I didn't check the archive before:

What bios files are these? If they contain any 3rd party copyrighted code they will need to be removed.

I don't want files copied into the home folder - they won't get cleaned up etc. They could be symlinked though via the "moveConfigFile" functionality.

joolswills:
Checked the website faq and I can see they are custom bios files. So that's ok. But the location issue needs to be sorted still.

joolswills:
scrap what I said about moveConfigFile - a manual `ln -sf` should be enough - but the module needs a remove_ function to clean them up on removal.

joolswills:
I had a look into this further - I will sort this as it requires `moveConfigDir` and a copy rather than a move etc.

joolswills:
the code also makes folders in the current directory, so we need to make sure change folder on launch etc.

joolswills:
Required a bunch of changes - possibly will need more. this is a bit of a workaround due to the fact the game expects to write stuff to the current folder

* fix permissions on game_database.xml
 * create config structure for bios symlinks
 * symlink default install files from config folder
 * pushd to config folder before launch

PR #1852, opened 2017-01-22
Fix UnicodeEncodeError when enumerating devices

mat128:
Currently, enumerating devices with non-ascii characters results in a UnicodeEncodeError.

Simply ignore non-ascii characters to ensure devices can be listed.


joolswills:
Thanks

PR #1855, opened 2017-01-23
Powerblock

joolswills:
@petrockblog made some changes to match this up against other code. Please can you check it still works for you ok. more details in commit messages.

It didn't seem to work for me building / copying to md_inst then trying to install from there - because the previous location was saved in the cmake cache I believe. In this case for something where we need to keep the files around for enabling/disabling/uninstalling, it's best just to checkout to md_inst.

you will need to remove and reinstall though since the install location is now the git checkout.

### DIFF
--- a/scriptmodules/supplementary/powerblock.sh
+++ b/scriptmodules/supplementary/powerblock.sh
@@ -35,7 +35,7 @@ function build_powerblock() {
 }
 
 function install_powerblock() {
-    # then install from there to system folders
+    # install from there to system folders
     cd "$md_inst/build"
     make install
 }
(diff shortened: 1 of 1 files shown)


joolswills:
does the powerblock work on other hardware other than the RPI ? if not, it will require some flags with "!mali !x86" for example for now, and there is no need for the rpi check in the depends_ function for installing libraspberrypi-dev

joolswills:
fixed a heap of typos in original PR text - must have been asleep ;-)

joolswills:
the snesdev module needs some similar work - so once this is sorted, I will update snesdev in a similar manner. I guess that module was used as a base for this one - but unfortunately snesdev is one part of retropie that hasn't been updated much.

joolswills:
Will merge this if there are no objections.

petrockblog:
Great, thanks for that support. I was thinking more complicated and was looking into CPack these minutes. Can merge from my side!

PR #1858, opened 2017-01-25
Fix multiple matching device filenames 

arcane47:
$ cat /sys/class/input/js0/device/name
> Microsoft X-Box 360 pad

$ grep --exclude=*.bak -rl "/opt/retropie/configs/all/retroarch-joypads/" -e "Microsoft X-Box 360 pad"
> /opt/retropie/configs/all/retroarch-joypads/udev/Microsoft_X-Box_360_pad.cfg
> /opt/retropie/configs/all/retroarch-joypads/Microsoft X-Box 360 pad.cfg
> /opt/retropie/configs/all/retroarch-joypads/android/Microsoft_XBOX_360_Controller_USB.cfg

-sort and awk ensures breadth first as to comply with joypad config

### DIFF
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -113,7 +113,7 @@ function remap() {
         bind=""
         for device_num in "${!devices[@]}"; do
             # get name of retroarch auto config file
-            file=$(grep --exclude=*.bak -rl "$configdir/all/retroarch-joypads/" -e "\"${devices[$device_num]}\"")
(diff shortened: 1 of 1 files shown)


joolswills:
Why do you have subfolders with multiple configs in there ? We don't have that structure in retropie.

joolswills:
(Please give me an example of the output with your changes too - it looks like it could be simplified - although I'm not sure it's actually needed as we don't include android/udev folders etc).

joolswills:
The grep could be simplified though so just something like

```
$(grep -lF "\"${devices[$device_num]}\"" "$configdir/all/retroarch-joypads/"*.cfg)
```

as there is no need for it to be recursive (our joypad configs sit in the top level folder).

arcane47:
I initially assumed the grep recursion existed for when the joypad autoconf (top level) was not previously generated before running the script.  I just adhered to previous recursion but ensured top level first, and limit config file return to one.  Can be simpler as you provided

arcane47:
As for the reason with other subfolders, I am not sure.  I installed all from setup script.

joolswills:
You must have had some old retroarch install lying around perhaps or some old auto configs. I will commit my change separately rather than merging/squashing this PR. thanks.

PR #1863, opened 2017-01-28
add quasi88, a NEC PC-8801 emulator

nakatamaho:
add quasi88, a NEC PC-8801 emulator

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -164,8 +164,8 @@ psx_fullname="PlayStation"
 ps2_exts=".iso .img .bin .mdf .z .z2 .bz2 .dump .cso .ima .gz"
 ps2_fullname="PlayStation 2"
 
-quasi88_exts=".d88 .D88 .cmt .CMT .t88 .T88"
-quasi_fullname="PC-8801"
+quasi88_exts=".d88 .cmt .t88"
+quasi88_fullname="PC-8801"
 
 sega32x_exts=".32x .smd .bin .md .zip"
 sega32x_fullname="Sega 32X"
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -164,6 +164,9 @@ psx_fullname="PlayStation"
 ps2_exts=".iso .img .bin .mdf .z .z2 .bz2 .dump .cso .ima .gz"
 ps2_fullname="PlayStation 2"
 
+quasi88_exts=".d88 .D88 .cmt .CMT .t88 .T88"
+quasi_fullname="PC-8801"
+
 sega32x_exts=".32x .smd .bin .md .zip"
 sega32x_fullname="Sega 32X"
 
--- a/scriptmodules/emulators/quasi88.sh
+++ b/scriptmodules/emulators/quasi88.sh
@@ -22,16 +22,15 @@ function depends_quasi88() {
 function sources_quasi88() {
     wget -q -O- "http://www.eonet.ne.jp/~showtime/quasi88/release/quasi88-0.6.4.tgz" | tar -xvz --strip-components=1
     applyPatch "$md_data/01_Makefile.diff"
-    sed -i -e "s|%%md_inst%%|$md_inst|g" Makefile
 }
 
 function build_quasi88() {
     make clean
-    make SOUND_SDL=1 USE_OLD_MAME_SOUND=1 USE_FMGEN=1 ROMDIR="$biosdir/quasi88" DISKDIR="$romdir/pc88" TAPEDIR="$romdir/pc88"
+    make X11_VERSION='' SDL_VERSION=1 ARCH=linux SOUND_SDL=1 USE_OLD_MAME_SOUND=1 USE_FMGEN=1 ROMDIR="$biosdir/quasi88" DISKDIR="$romdir/pc88" TAPEDIR="$romdir/pc88" 
 }
 
 function install_quasi88() {
-    make install
+    make X11_VERSION='' SDL_VERSION=1 BINDIR="$md_inst" install
 }
 
 function configure_quasi88() {
--- a/scriptmodules/emulators/quasi88.sh
+++ b/scriptmodules/emulators/quasi88.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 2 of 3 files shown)


platforms.cfg
```diff
@@ -164,6 +164,9 @@ psx_fullname="PlayStation"
 ps2_exts=".iso .img .bin .mdf .z .z2 .bz2 .dump .cso .ima .gz"
 ps2_fullname="PlayStation 2"
 
+quasi88_exts=".d88 .D88 .cmt .CMT .t88 .T88"
+quasi_fullname="PC-8801"
```
joolswills: should be quasi88_fullname

scriptmodules/emulators/quasi88.sh
```diff
@@ -0,0 +1,97 @@
...
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="quasi88"
+rp_module_desc="NEC PC-8801 emulator"
+rp_module_help="ROM Extensions: .d88 .D88 .cmt .CMT .t88 .T88."
+rp_module_section="exp"
+version="0.6.4"
```
joolswills: this should be removed

scriptmodules/emulators/quasi88.sh
```diff
@@ -0,0 +1,97 @@
...
+
+function depends_quasi88() {
+    local depends=(libsdl1.2-dev)
+    getDepends "${depends[@]}"
```
joolswills: you can just do `getDepends libsdl1.2-dev` in this case, but it's not a problem.

scriptmodules/emulators/quasi88.sh
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function sources_quasi88() {
+    wget -q -O- "http://www.eonet.ne.jp/~showtime/quasi88/release/quasi88-${version}.tgz" | tar -xvz --strip-components=1
+    cat > retropie_fixes.diff <<\_EOF_
```
joolswills: for inline patches use applyPatch

use unified diff format.

scriptmodules/emulators/quasi88.sh
```diff
@@ -0,0 +1,97 @@
...
+    make
+}
+
+function install_quasi88() {
+    make install
+}
+
+function configure_quasi88() {
+    mkRomDir "quasi88"
+    mkUserDir "$md_conf_root/quasi88"
```
joolswills: is this folder needed ? if the emulator has a config folder in the home directly it needs to be symlinked with "moveConfigDir"

platforms.cfg
joolswills: also lowercase extensions only here.

scriptmodules/emulators/quasi88.sh
joolswills: also needs lowercase extensions.

scriptmodules/emulators/quasi88.sh
joolswills: local version can be included here or just hardcode it.

scriptmodules/emulators/quasi88/01_Makefile.diff
```diff
@@ -0,0 +1,98 @@
...
++SDL_VERSION	= 1
+ # GTK_VERSION	= 1
+ 
+ 
+@@ -32,7 +32,7 @@
+ #	�ǥ��쥯�ȥ꤬���Ѥ���ޤ���
+ #	 ~/ �ϡ�QUASI88�ε�ư���˥ۡ���ǥ��쥯�ȥ��$HOME�ˤ�Ÿ������ޤ���
+ 
+-ROMDIR	= ~/quasi88/rom/
++ROMDIR	= /home/pi/RetroPie/BIOS/quasi88
```
joolswills: This shouldn't be hardcoded as RetroPie can be used for any user.

You can put the patch in-line and use romdir/BIOS/quasi88, or put some string here and sed the patch before applying.

scriptmodules/emulators/quasi88/01_Makefile.diff
```diff
@@ -0,0 +1,98 @@
...
+ 
+ 
+ # DISK�ѥǥ��쥯�ȥ�����ꤷ�ޤ�
+@@ -42,7 +42,7 @@
+ #	�ǥ��쥯�ȥ꤬���Ѥ���ޤ���
+ #	 ~/ �ϡ�QUASI88�ε�ư���˥ۡ���ǥ��쥯�ȥ��$HOME�ˤ�Ÿ������ޤ���
+ 
+-DISKDIR	= ~/quasi88/disk/
++DISKDIR	= /home/pi/RetroPie/roms/pc88
+
```
joolswills: as above

joolswills:
actually, there is no need to patch the Makefile - it should be possible to override these makefile variables when calling MAKE - which would simplify it a lot.

joolswills:
eg `make VARIABLE=` (you can combine this with a smaller patch if needed - but it can be used for the PATHS etc)

joolswills:
eg something like (just an example - untested)

```
make X11_VERSION=0 SDL_VERSION=1 ROMDIR="$romdir/BIOS/quasi88" DISKDIR="$romdir/pc88" TAPEDIR="$romdir/pc88" USE_OLD_MAME_SOUND=1 BINDIR="$md_inst"
```

nakatamaho:
* fixed issues mentiond avobe. avoid hardcoded paths.
* somehow make VARIABLE= didn't work. only ROM_DIR, DISK_DIR and TAPE_DIR work.
following is quote from Makefile.
CFLAGS += -DROM_DIR='"$(ROMDIR)"' -DDISK_DIR='"$(DISKDIR)"' \
          -DTAPE_DIR='"$(TAPEDIR)"' $(USEINLINE) -DCLIB_DECL=

joolswills:
Which variable didnt work ? BINDIR should work ok for example.

nakatamaho:
thanks for your comment. now I understand how to use command line options.
could you please review again?
Best regards.

joolswills:
I added this with some changes:

* More help info
* Allow building for X11 but exclude Mali
* Rename patch and include fix for psg.cpp on later gcc
* Change platforms.cfg and configure_ to use pc88 (system name rather than emulator name)
* Change config location and launch commandline

joolswills:
Cheers.

nakatamaho:
Thank you very very much! you're kind person!

Issue #1871, opened 2017-02-01
Add Amiga emulator libretro cores


skaba:
There are multiple amiga emulators WIP ported to Libretro with different maturity levels.
- FS-UAE
- P-UAE
- uae4arm

Request to add them to experimental for now until they reach completion state.

joolswills:
So long as they are all functional that would be ok.

skaba:
I couldnt get uae4arm and fs-uae compile (I tried binaries from Gamestarter kodi plugin and got them to work at a level)

Managed to get p-uae compile from source, games work, savestates doesn't work for me maybe you can, see my pull request

a7mag3ddon:
uae4arm is already in retropie? i use it all the time works great, or is it a different version now? I am using 0.5 midwan / amiberry version.

joolswills:
the standalone uae4arm is in retropie - the wip libretro core is not yet.

midwan:
You might want to include Amiberry in the list of Amiga emulators of RetroPie. It's a version forked from uae4arm-rpi, but with several improvements and bug fixes. I just pushed out a binary release of v2.1, which was tested on Raspbian-based distros (including RetroPie): https://github.com/midwan/amiberry/releases

I'm currently working on an SDL2 version also, with the final aim of getting everything up to the WinUAE/FS-UAE standards and then get a libretro version out as well. But there's still work to be done in that part, so for now the current release is still based on SDL1.2.

I'll notify you when the SDL2 version is ready for release.

joolswills:
@midwan Will look into it - any reason why another fork rather than not working with the existing uae4arm developer ?

a7mag3ddon:
Thus new fork seems to be updated very regularly the old one hasn't been looked at in a long time even tho I and others logged verified bugs in it.

joolswills:
@a7mag3ddon uae4arm is still developed - last commit was only a month ago. You were asked for more information on at least one of your tickets which you didn't provide btw.

My question was for midwan though as it's a shame to fragment the already very fragmented uae ports further. But I understand in some cases this is unavoidable.

midwan:
@joolswills 
I did try to reach out to Chips from the very beginning, several times in fact. Unfortunately he wasn't interested in merging in all the updates I have been adding to his version. He has selectively merged in some parts (like the Shutdown button) but not most of them.

[...]

PR #1872, opened 2017-02-01
Add lr-puae #1871

skaba:
Initial pull request for code review, notes:
- No directory setup for kickstart or rom folders, I set that up with RetroPie assistant template and I can share that too if wanted
- Adding .uae extension (which can also be done by retropieassistant)
- Not verified with other arhitecture hence the restriction
- Save states doesn't work for me (Need to check with upstream maybe)

### DIFF
--- a/scriptmodules/libretrocores/lr-puae.sh
+++ b/scriptmodules/libretrocores/lr-puae.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Will give it a test.

joolswills:
This builds/runs fine on x86 - Please remove the flag and squash and I will merge. thanks.

skaba:
Thanks, removed and squashed.

PR #1875, opened 2017-02-05
Incorporate Herb's mkdocs work into a module

joolswills:
Renamed existing docs module to apidocs.
Added this one as wikidocs.

```
sudo ./retropie_packages.sh wikidocs
```

to generate

```
sudo ./retropie_packages.sh wikidocs upload
```

to push to the site (requires your ssh public key in retropie account on our server - we can sort that if you want to test).

docs end up here

https://retropie.org.uk/wikidocs/

thanks for your work on this

### DIFF
--- a/scriptmodules/admin/apidocs.sh
+++ b/scriptmodules/admin/apidocs.sh
@@ -9,19 +9,19 @@
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
 #
 
-rp_module_id="docs"
+rp_module_id="apidocs"
 rp_module_desc="Generate developer documentation"
 rp_module_section=""
 
-function depends_docs() {
+function depends_apidocs() {
     getDepends doxygen graphviz
 }
 
-function sources_docs() {
+function sources_apidocs() {
     gitPullOrClone "$md_build" https://github.com/Anvil/bash-doxygen.git
 }
 
-function build_docs() {
+function build_apidocs() {
     local config="Doxyfile"
     rm -f "$config"
     doxygen -g "$config" >/dev/null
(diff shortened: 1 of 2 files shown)


joolswills:
Still uses github for the assets - was thinking we could download all the assets and rewrite the markdown or something so the assets are hosted with the docs etc.

joolswills:
That can be done later of course - just an idea.

HerbFargus:
It's looking good so far and the modules are working

But it seems that it isn't changing with updates:

for example I added a line about exiting emulationstation before manually editing gamelist.xmls right above the following heading:

https://github.com/RetroPie/RetroPie-Setup/wiki/Scraper#emulationstation-built-in-scraper

I built the site with

    sudo ./retropie_packages.sh wikidocs

and in `~/RetroPie-Setup/tmp/wikidocs/Scraper` the edits show  but on deploying with

    sudo ./retropie_packages.sh wikidocs upload

It doesn't show the latest edits so I'm not sure if the sync or copy is working or perhaps it's cached. I ran it under a private window in firefox but still no change

HerbFargus:
And yes down the road might be worth considering hosting out own assets once the pages are cleaned up and we have our assets sorted.

HerbFargus:
Oh and one other idea that may be worth looking at is perhaps a webhook of sorts that automatically generates the wiki at end of day or listens for changes in the wiki and updates anytime an edit is made.

joolswills:
Not sure about the updating - but can't test current as the yml needs updating with the page name changes. Please can you provide me with an updated/working yml and I will look into that issue.

joolswills:
Aah you have done the changes on the other repo? I will change the script to pull from there as discussed.

joolswills:
I changed the script to use the other repo - building and uploading was fine and the changes were shown up so I can't reproduce the issue. Please test though and if you still have the issue post a log.

HerbFargus:
I ran a few tests last night after our discussion and was able to get it working and pushed online. I'll test again this evening but as far as I can see it functions as intended.

joolswills:
thanks.

PR #1879, opened 2017-02-06
Make lr-mame2010 experimental.

dankcushions:
Reflects the wiki: https://github.com/retropie/retropie-setup/wiki/MAME#lr-mame2010-mame-2010

I think it's right to be experimental as this one is missing some crucial features (<2 player, for example)

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2010.sh
+++ b/scriptmodules/libretrocores/lr-mame2010.sh
@@ -12,7 +12,7 @@
 rp_module_id="lr-mame2010"
 rp_module_desc="Arcade emu - MAME 0.139 port for libretro"
 rp_module_help="ROM Extension: .zip\n\nCopy your MAME roms to either $romdir/mame-libretro or\n$romdir/arcade"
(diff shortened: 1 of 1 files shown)


joolswills:
It's likely that was written before it moved into optional which I think was done as it takes a long time to build on the rpi (experimental stuff is source only), and yet some people were saying it was useful for some games.

Please can you explain <2 player feature missing ?

dankcushions:
It only supports 2 players. There's a few other issues but that's the big one for me. It's probably easy to fix but there's so many mame cores...

If it's useful for some games (I think some sega games have better sound in this) then that's probably a good enough reason to keep it in 'optional'?

joolswills:
I'm ok leaving it in optional for now at least - despite limitations. Note that the experimental section often reflects the age of the retropie-setup module also (It's not always that the emulator is experimental, but the retropie module is experimental).

dankcushions:
I have updated the wiki :) Sorry for the spam!

PR #1881, opened 2017-02-09
generate launching images based on ES themes

meleu:
Here is the scriptmodule we talked about [in the forum](https://retropie.org.uk/forum/topic/8029/where-s-the-retropie-setup-api-docs/3).

### DIFF
--- a/scriptmodules/supplementary/launchingimages.sh
+++ b/scriptmodules/supplementary/launchingimages.sh
@@ -80,8 +80,7 @@ function _dialog_menu_launchingimages() {
         ((i++))
     done
     choice=$(dialog --backtitle "$__backtitle" --menu "$text" 22 86 16 "${options[@]}" 2>&1 >/dev/tty) || return
-    i=$[ 2 * choice - 1 ]
-    echo "${options[$i]}"
+    echo "${options[choice*2-1]}"
 }
 
 function _set_theme_launchingimages() {
--- a/scriptmodules/supplementary/launchingimages.sh
+++ b/scriptmodules/supplementary/launchingimages.sh
@@ -87,9 +87,7 @@ function _dialog_menu_launchingimages() {
 }
 
 function _set_theme_launchingimages() {
-    _dialog_menu_launchingimages \
-        "List of available themes" \
-        $("$md_inst/generate-launching-images.sh" --list-themes) \
+    _dialog_menu_launchingimages "List of available themes" $("$md_inst/generate-launching-images.sh" --list-themes) \
     || echo "\$theme"
 }
 
--- a/scriptmodules/supplementary/launchingimages.sh
+++ b/scriptmodules/supplementary/launchingimages.sh
@@ -16,7 +16,8 @@ rp_module_section="exp"
 rp_module_flags="noinstclean"
 
 function depends_launchingimages() {
-    local depends=(imagemagick librsvg2-bin)
+    #local depends=(imagemagick librsvg2-bin)
+    local depends=(imagemagick)
     if isPlatform "x11"; then
         depends+=(feh)
     else
@@ -67,8 +68,16 @@ function _show_images_launchingimages() {
 function _dialog_menu_launchingimages() {
     local text="$1"
     shift
+    local options=( "$@" )
+
     [[ "$#" -eq 0 ]] && return 1
-    dialog --backtitle "$__backtitle" --no-items --menu "$text" 22 86 16 "$@" 2>&1 >/dev/tty
+
+    if [[ "$#" -eq 1 ]]; then
+        options+=( "" )
+        dialog --backtitle "$__backtitle" --menu "$text" 22 86 16 "${options[@]}" 2>&1 >/dev/tty
+    else
+        dialog --backtitle "$__backtitle" --no-items --menu "$text" 22 86 16 "${options[@]}" 2>&1 >/dev/tty
+    fi
 }
 
 function _set_theme_launchingimages() {
@@ -396,12 +405,12 @@ function gui_launchingimages() {
 
                 4) # View the launching image of a specific system
                     while true; do
-                        local img_list=$(_get_all_launchingimages)
-                        if [[ -z "$img_list" ]]; then
+                        local img_list=( $(_get_all_launchingimages) )
+                        if [[ "${#img_list[@]}" -eq 0 ]]; then
                             printMsgs "dialog" "No launching image found."
                             break
                         fi
-                        choice=$(_dialog_menu_launchingimages "Choose the system" "$img_list")
+                        choice=$(_dialog_menu_launchingimages "Choose the system" "${img_list[@]}")
                         [[ -z "$choice" ]] && break
                         _show_images_launchingimages "$choice"
                     done
--- a/scriptmodules/supplementary/launchingimages.sh
+++ b/scriptmodules/supplementary/launchingimages.sh
@@ -77,12 +77,21 @@ function _set_theme_launchingimages() {
 }
 
 function _set_system_launchingimages() {
-    local choice=$(
+    local options=()
+    local choice
+
+    options=( all )
+    options+=( $("$md_inst/generate-launching-images.sh" --list-systems) )
+    choice=$(
         _dialog_menu_launchingimages \
-            "List of available systems.\n\nSelect the system you want to generate a launching image or cancel to generate for all systems." \
-            $("$md_inst/generate-launching-images.sh" --list-systems)
+            "List of available systems.\n\nSelect the system you want to generate a launching image or \"all\" to generate for all systems." \
+            "${options[@]}"
     )
-    [[ -n "$choice" ]] && echo "--system $choice"
+    case "$choice" in
+        "all")  echo "" ;;
+        "")     echo "$system" ;;
+        *)      echo "--system $choice" ;;
+    esac
 }
 
 function _set_extension_launchingimages() {
@@ -212,12 +221,12 @@ function _load_config_launchingimages() {
 }
 
 function _settings_launchingimages() {
-    local cmd=(dialog --backtitle "$__backtitle" --title " SETTINGS " --menu "runcommand launching images generation settings." 22 86 16)
+    local cmd=(dialog --backtitle "$__backtitle" --title " SETTINGS " --cancel-label "Back" --menu "runcommand launching images generation settings." 22 86 16)
     local options
     local choice
     local config_file
 
-    iniConfig ' = ' '"' "$md_inst/.current.cfg"
+    iniConfig ' = ' '"' "$current_cfg"
 
     while true; do
         eval $(_load_config_launchingimages)
@@ -269,7 +278,6 @@ function _settings_launchingimages() {
 function _manage_config_file_launchingimages() {
     local choice
     local config_file
-    local current_cfg="$md_inst/.current.cfg"
 
     eval $(_load_config_launchingimages)
 
@@ -353,7 +361,9 @@ function gui_launchingimages() {
     local cmd=()
     local options=()
     local choice
+    local current_cfg="$md_inst/.current.cfg"
 
+    rm -f "$current_cfg"
     while true; do
         cmd=(dialog --backtitle "$__backtitle" --title " runcommand launching images generation " --menu "Choose an option." 22 86 16)
         options=( 
@@ -364,7 +374,7 @@ function gui_launchingimages() {
         )
         choice=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
         if [[ -n "$choice" ]]; then
-            case $choice in
+            case "$choice" in
                 1) # Image generation settings
                     _settings_launchingimages
                     ;;
@@ -379,14 +389,19 @@ function gui_launchingimages() {
                     if [[ -s "$file" ]]; then
                         _show_images_launchingimages 1 "$file"
                     else
-                        printMsgs "dialog" "There are no launching images found on your system."
+                        printMsgs "dialog" "No launching image found."
                     fi
                     rm -f "$file"
                     ;;
 
                 4) # View the launching image of a specific system
                     while true; do
-                        choice=$(_dialog_menu_launchingimages "Choose the system" $(_get_all_launchingimages))
+                        local img_list=$(_get_all_launchingimages)
+                        if [[ -z "$img_list" ]]; then
+                            printMsgs "dialog" "No launching image found."
+                            break
+                        fi
+                        choice=$(_dialog_menu_launchingimages "Choose the system" "$img_list")
                         [[ -z "$choice" ]] && break
                         _show_images_launchingimages "$choice"
                     done
@@ -397,5 +412,5 @@ function gui_launchingimages() {
             break
         fi
     done
-    rm -f "$md_inst/.current.cfg"
+    rm -f "$current_cfg"
 }
--- a/scriptmodules/supplementary/launchingimages.sh
+++ b/scriptmodules/supplementary/launchingimages.sh
@@ -88,7 +88,7 @@ function _dialog_menu_launchingimages() {
 
 function _set_theme_launchingimages() {
     _dialog_menu_launchingimages "List of available themes" $("$md_inst/generate-launching-images.sh" --list-themes) \
-    || echo "\$theme"
+    || echo "$theme"
 }
 
 function _set_system_launchingimages() {
@@ -102,14 +102,14 @@ function _set_system_launchingimages() {
 
 function _set_extension_launchingimages() {
     _dialog_menu_launchingimages "Choose the file extension of the final launching image." png jpg \
-    || echo "\$extension"
+    || echo "$extension"
 }
 
 function _set_show_timeout_launchingimages() {
     _dialog_menu_launchingimages \
         "Set how long the image will be displayed before asking if you accept (in seconds)" \
         1 2 3 4 5 6 7 8 9 10 \
-    || echo "\$show_timeout"
+    || echo "$show_timeout"
 }
 
 function _set_loading_text_launchingimages() {
@@ -119,7 +119,7 @@ function _set_loading_text_launchingimages() {
         0 70 \
         "NOW LOADING" \
         2>&1 >/dev/tty \
-    || echo "\$loading_text"
+    || echo "$loading_text"
 }
 
 function _set_press_button_text_launchingimages() {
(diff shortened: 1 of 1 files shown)


joolswills:
thanks. Few general comments before a full review - there are a lot of multi-line statements - although it's useful to keep lines short, I think perhaps it's a little too much in some cases here. Also please use 4 spaces for indentation (even for multiline).

eg

```
    dialog \
        --backtitle "$__backtitle" \
        --no-items \
        --menu "$text" \
        22 86 16 \
        $@ \
        2>&1 >/dev/tty
```

but in this case I don't think it needs splitting - and it's enough to just do it as follow - note that "$@" should be quoted.

```
    dialog --backtitle "$__backtitle" --no-items --menu "$text" 22 86 16 "$@"  2>&1 >/dev/tty
```

when you have an array like

```
    options=(
        'something' \
        'something' \
    )
```

you can leave out the `\` characters.

joolswills:
also function names all need to end with the module name - so they are guaranteed unique.

meleu:
OK. I'll try to do as you said right now.

meleu:
From [line 300](https://github.com/meleu/RetroPie-Setup/blob/31991e9b5a9cac6d464834ea0aea458c62dcb15d/scriptmodules/supplementary/launchingimages.sh/#L300) to [360](https://github.com/meleu/RetroPie-Setup/blob/31991e9b5a9cac6d464834ea0aea458c62dcb15d/scriptmodules/supplementary/launchingimages.sh/#L360) the code has a deep indentation. I would like to use nested if-elif (rather than case) to avoid 4 spaces of ugliness. Are you OK with it?

joolswills:
I would probably avoid it completely, by separating things out a bit into modular functions. I would use a case statement though still.

joolswills:
BTW check out the scraper and _load_config_scraper function - for how we load settings there .

the multiple settings loading / saving is probably overkill - a single saved config should be enough - but up to you - but `loadModuleConfig` should be useful here and allow you to split out the code a bit more.

meleu:
Yeah, I tried to think in a modular function logic to avoid it, but I need access to several local variables that isn't accessible globally (I'm assuming that a scriptmodule can't declare global variables).
I'm working on the style first...
P.S.: OK. I'll look at the scraper function.

joolswills:
That's why you can use loadModuleConfig and have a function that does that to load the variables so you can split it out a bit (means a bit more reloading, but it's not noticable). note with local variables also in bash that functions below the function they are defined in can read them

joolswills:
in the case of the scraper, we reload them in each function, as from commandline each function can be called independently - eg to automate and so on.

joolswills:
Looking at your code I would split out these

```
1 "Image generation settings"
2 "Generate launching images"
3 "View slideshow of all current launching images"
4 "View a specific system's launching image"
```

into separate functions that can be called independently (so they would just reload the current config as needed).

meleu:
I will take a look at this loadModuleConfig stuff. Thanks for the tips!

meleu:
Sorry if I'm making this conversation a bash class, but I think I misunderstood what you said here:

> note with local variables also in bash that functions below the function they are defined in can read them

To check it I've made this test:
```
[prompt]$ cat test.sh
#!/bin/bash

function func_a() {
    local var_a="this is declared in func_a()"
    echo "func_a(): var_a=$var_a"
}

function func_b() {
    echo "func_b(): var_a=$var_a"
}

func_a
func_b
echo "no functions: var_a=$var_a"

[prompt]$ ./test.sh
func_a(): var_a=this is declared in func_a()
func_b(): var_a=
no functions: var_a=
```

Maybe you meant variables declared in a function but **not** declared explicitly as `local`, right? If yes, is it an acceptable practice for scriptmodules?

(I'm studying the `loadModuleConfig` and wanna use it, I'm asking it just to learn more and more. :) )

joolswills:
Your code is as expected. When I said below I mean when one function is called from another.

```
#!/bin/bash

function func_a() {
    local var_a="this is declared in func_a()"
    echo "func_a(): var_a=$var_a"
    func_b
}

function func_b() {
    echo "func_b(): var_a=$var_a"
}

func_a
echo "no functions: var_a=$var_a"
```

meleu:
Oh, I've got it. Thanks for the clarification.

I'm working on the scriptmodule right now, and I'm making available the multiple config files because there are some known good configs for different themes.

I'm thinking in putting some optimized config files for some themes in the [generate-launching-images repository](https://github.com/meleu/generate-launching-images), to make things easier to the user.

meleu:
I think it's ready for your review. Thanks for all the tips.
Cheers!

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,413 @@
...
+    getDepends "${depends[@]}"
+}
+
+function sources_launchingimages() {
+    gitPullOrClone "$md_inst" "https://github.com/meleu/generate-launching-images.git"
+}
+
+function install_launchingimages() {
+    cd "$md_inst"
+    chmod a+x generate-launching-images.sh
```
joolswills: as this is a single script, you could probably just have an install_bin function

joolswills: the file permissions could also be committed to git, which would remove the need for the chmod

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,413 @@
...
+}
+
+function _show_images_launchingimages() {
+    [[ -f "$1" ]] || return 1
+
+    local image="$1"
+    local timeout=5
+    local is_list=0
+
+    [[ "$(file -i $1)" =~ text/plain ]] && is_list=1
```
joolswills: This seems odd - why do you need to check the type of file ? could be unreliable - can you not just use a parameter if it is to process a list 

no need to do a regular expression for a match like this either - can use *text/plan*

scriptmodules/supplementary/launchingimages.sh
meleu: good catch! Should I commit these changes or will you do?

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,413 @@
...
+            --autozoom \
+            $([[ "$is_list" -eq 1 ]] && echo --list) \
+            "$image" </dev/tty &>/dev/null
+    fi
+}
+
+function _dialog_menu_launchingimages() {
+    local text="$1"
+    shift
+    [[ -z "$@" ]] && return 1
```
joolswills: You use "$#" below - would be a better test here also.

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,413 @@
...
+    fi
+
+    dialog --backtitle "$__backtitle" --no-items --menu "$text" 22 86 16 "$@" 2>&1 >/dev/tty
+}
+
+function _set_theme_launchingimages() {
+    _dialog_menu_launchingimages \
+        "List of available themes" \
+        $("$md_inst/generate-launching-images.sh" --list-themes) \
+    || echo "\$theme"
```
joolswills: I would have thought the last line should be indented too, if splitting onto lots of lines.

joolswills: Not a big deal, just doesn't really match any other code style in retropie

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,413 @@
...
+        choice=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
+
+        [[ -z "$choice" ]] && break
+
+        if [[ "$choice" = "config_file" ]]; then
+            config_file=$(_manage_config_file_launchingimages)
+            continue
+        fi
+
+        eval iniSet "$choice" \"$(_set_${choice}_launchingimages)\"
```
joolswills: why eval ?

joolswills: oh. you are returning a variable name in a called function - wouldn't it be best if the functions just echoed the variable value - this seems more logical.

meleu: it wasn't an iniSet before. I'm changing...

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,413 @@
...
+            )"
+            extension ".$extension"
+            loading_text "\"$loading_text\""
+            press_button_text "\"$press_button_text\""
+            loading_text_color "$loading_text_color"
+            press_button_text_color "$press_button_text_color"
+            show_timeout "$( [[ -n "$no_ask" ]] && echo "don't show (see no_ask)" || echo "$show_timeout seconds")"
+            no_ask "$( [[ -n "$no_ask" ]] && echo true || echo false)"
+            no_logo "$( [[ -n "$no_logo" ]] && echo true || echo false)"
+            logo_belt "$( [[ -n "$logo_belt" ]] && echo true || echo false)"
```
joolswills: I'm not sure I like this for readability over how we build options lists in other modules, and requiring commandline tools like echo etc for each option. You don't need to change it, just mentioning.

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,413 @@
...
+    local config_file
+
+    iniConfig ' = ' '"' "$md_inst/.current.cfg"
+
+    while true; do
+        eval $(_load_config_launchingimages)
+
+        options=( 
+            config_file "$(
+               [[ "$config_file" =~ $theme\.cfg$ ]] && echo "$(basename "$config_file")"
```
joolswills: can you avoid a re here ? eg something like

`[[ "$config_file" == *"$theme.cfg" ]] && echo "$(basename "$config_file")"`

joolswills:
Thanks for your work on this - I will give it a test and merge when you have finished with the changes. Please can you squash the changes when you are ready. Cheers.

scriptmodules/supplementary/launchingimages.sh
joolswills: I see the logic in this now - you are indenting the parameters for the function, but not the `|| echo` - so feel free to leave this as is. It's just due to this module having a lot more line splitting making it it look quite different from other retropie code.

scriptmodules/supplementary/launchingimages.sh
meleu: Done: https://github.com/RetroPie/RetroPie-Setup/pull/1881/commits/73e07f28483f075a50580bda2857eb702291dd8f

scriptmodules/supplementary/launchingimages.sh
meleu: Done: https://github.com/RetroPie/RetroPie-Setup/pull/1881/commits/fd3ce817a0fb7b92f983095245fe01b4ec1daae6

scriptmodules/supplementary/launchingimages.sh
meleu: Done: https://github.com/RetroPie/RetroPie-Setup/pull/1881/commits/67b3bf54360f3ec7cc089c3478f46c10dac881d9

scriptmodules/supplementary/launchingimages.sh
meleu: Done: https://github.com/RetroPie/RetroPie-Setup/pull/1881/commits/f757c39ca030d3846e511e61568d8c869b8134f3

scriptmodules/supplementary/launchingimages.sh
meleu: Done: https://github.com/RetroPie/RetroPie-Setup/pull/1881/commits/d935c6c8fd0e5df3b14ab0b26c4cf3c7c32c2ae9

scriptmodules/supplementary/launchingimages.sh
meleu: Fixing: https://github.com/RetroPie/RetroPie-Setup/pull/1881/commits/07a695735429f6c2b83a91f5fd19e729249c8db1

Thanks to your tip here: https://github.com/RetroPie/RetroPie-Setup/pull/1881#issuecomment-279080194
Thanks!

scriptmodules/supplementary/launchingimages.sh
meleu: I'm planning to add some optimized `.cfg` files for some themes.
Examples:
- material theme: use `--solid-bg-color` and `--press-button-text-color white`.
- clean-look theme: use `--logo-belt`

Is the `install_bin` approach good for this case also?

meleu: Done: https://github.com/RetroPie/RetroPie-Setup/pull/1881/commits/cdac6bbf949041668cb9ca71a280b43a1db4eeb4

meleu:
I think it's ready for your tests.

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -0,0 +1,410 @@
...
+rp_module_section="exp"
+rp_module_flags="noinstclean"
+
+function depends_launchingimages() {
+    local depends=(imagemagick librsvg2-bin)
+    if isPlatform "x11"; then
+        getDepends feh
+    else
+        getDepends fbi
+    fi
```
joolswills: these should be depends+=(feh) otherwise apt will be called twice.

joolswills:
Spotted the above when testing - please can you squash the changes.

joolswills:
the system selection logic if different from other dialogs in retropie. cancel reverts it back to "all". I htink all should be the first option, and cancel reverts it back to what it was before.

joolswills:
if choosing option 4 - when no images are present, it does nothing - perhaps a dialog ?

When only one theme is installed, it says "list of available themes: carbon" and has a yes / no which seems odd - why not just a menu selection with a single item ?

If choosing generate and no theme was selected it shows an error. perhaps this should be done at the "image generation settings" stage ?

the image generation settings "cancel" button should be renamed I think - since it doesn't actually cancel anything that has been entered  perhaps "return" or "exit"

joolswills:
I see in the code there is some reason for the yes no dialog instead of a menu - looks unneeded - just provide an empty string as the item ?

You could also just have a numbered menu which is what we do in other dialogs.

meleu:
I'm quite busy currently. But I'll do the changes you pointed as soon as possible! ;-)

joolswills:
No rush :-) Apart from the points I mentioned above it all looks good, and works very well - we should be good to go once that's sorted. Thanks.

meleu:
@joolswills I found some time to do what you suggest.

> if choosing option 4 - when no images are present, it does nothing - perhaps a dialog ?

Done.


> When only one theme is installed, it says "list of available themes: carbon" and has a yes / no which seems odd - why not just a menu selection with a single item ?

Done. The problem is that the Raspbian's dialog version is 1.2-20140911 and I wrote the scriptmodule in Linux Mint 18 (dialog version 1.3-20160209). The dialog version on Raspbian has a small bug when using `--menu` option with `--no-items` and only one tag.

I need the `--no-items` option because I wanna get exactly what dialog puts in the stdout, (this is simpler than getting the tag number and then get the corresponding item).

> If choosing generate and no theme was selected it shows an error. perhaps this should be done at the "image generation settings" stage ?

I did this way thinking in a curious user navigating through the dialogs. If the user enter in the "image generation settings" and then cancel without a theme selected (to view the current images or just exit the script), everything is fine, there's no problem with this action. The really problem happens if the user chose to create the images with no theme selected.

If it's not a big problem, I prefer to leave this way...


> the image generation settings "cancel" button should be renamed I think - since it doesn't actually cancel anything that has been entered perhaps "return" or "exit"

I changed to "Back".


> the system selection logic is different from other dialogs in retropie. cancel reverts it back to "all". I think all should be the first option, and cancel reverts it back to what it was before.

Done.

meleu:
Oh! I forgot to mention about the `librsvg2-bin` dependency. It is needed to convert the logo `.svg` files.

- in my RetroPie on Linux Mint 18 it **is** a dependency.
- when testing in my 4.1 on Raspi installation, it's unnecessary (imagemagick is enough).
- when testing in a 4.0 on Raspi of a friend of mine, `librsvg2-bin` was a dependency and installed a bunch of X/Wayland libraries (can't remember exactly, but it was 80+ MB of packages). Maybe an `apt-get update/upgrade` would avoid it, but I don't know the best way to do it in a scriptmodule.

Could you please sort this last thing?
(As if the normal problems weren't enough, my youngest son has pneumonia...)

joolswills:
>Done. The problem is that the Raspbian's dialog version is 1.2-20140911 and I wrote the scriptmodule in Linux Mint 18 (dialog version 1.3-20160209). The dialog version on Raspbian has a small bug when using --menu option with --no-items and only one tag.

I would do as I mentioned before and not use no-items - just use numbered items. You can just get the value from the array. We don't use no-items anywhere else, so it is more fitting with the retropie menu styling.

If you are busy I can make this change for you.

regarding the dependencies - it should be the same for retropie 4 and 4.1 - if testing installing manually for testing - make sure you do `apt-get install --no-install-recommends` or else more will be installed than is needed. getDepends does this automatically.

Sorry to hear about your son - best wishes for a quick recovery!

meleu:
OK. Now the dialog menus have numbers.
A touch of algebra helped to make it simple. :-)

Regarding the `librsvg2-bin` depedency, I'm pretty sure I don't have it installed on my raspi/RetroPie 4.1 and I can convert the `.svg` logo files. So it's **not** a dependency in this platform.
On the other hand, if this package isn't installed on a x86 RetroPie, the convert fails when working with `.svg` files.

Summing up: I think the scriptmodule is complete now. Phew!

scriptmodules/supplementary/launchingimages.sh
```diff
@@ -68,16 +67,21 @@ function _show_images_launchingimages() {
...
-        dialog --backtitle "$__backtitle" --no-items --menu "$text" 22 86 16 "${options[@]}" 2>&1 >/dev/tty
-    fi
+    i=1
+    for opt in "$@"; do
+        options+=( "$i" "$opt" )
+        ((i++))
+    done
+    choice=$(dialog --backtitle "$__backtitle" --menu "$text" 22 86 16 "${options[@]}" 2>&1 >/dev/tty) || return
+    i=$[ 2 * choice - 1 ]
+    echo "${options[$i]}"
```
joolswills: can just do `echo "${options[choice*2-1]}"`

meleu: done

joolswills: btw please use (not needed in this case but for reference) `$(( ... ))` for arithmetic expansion

joolswills: to clarify - `$[ ... ]` is obsolete (from old bash I think).

joolswills:
Thanks. I squashed via github interface and merged.

PR #1882, opened 2017-02-09
uae4arm - game launch script / enhancements

joolswills:
* add default a500 / a1200 configs
 * allow launching directly from adf/dms/zip
 * support multiple disks in zip or if named *Disk1.adf *Disk2.adf as separate files
 * choose a1200 config if filename contains AGA or CD32 else use a500 config

### DIFF
--- a/scriptmodules/emulators/uae4arm.sh
+++ b/scriptmodules/emulators/uae4arm.sh
@@ -58,15 +58,40 @@ function configure_uae4arm() {
     fi
     moveConfigDir "$md_inst/kickstarts" "$biosdir"
 
+    local conf="$(mktemp)"
+    iniConfig "=" "" "$conf"
+    iniSet "config_description" "RetroPie A500, 68000, OCS, 512KB Chip + 512KB Slow Fast"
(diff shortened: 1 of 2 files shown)


joolswills:
If anyone has time to test this it would be appreciated - thanks!

Darknior:
Woaw a really great update :D 
@Waal will be very happy with these news features to port our xBox Amiga Pack to Retropie :)
It lacks only the auto whdload now :)

skaba:
Can we make this reusable in other uae forks?

joolswills:
Not unless they use the same commandline parameters and configuration as uae4arm.

skaba:
I don't think I can reuse it directly, I want to port this to lr-puae but addEmulator helper function makes the command retroarch

$emudir/retroarch/bin/retroarch -L $cmd --config $md_conf_root/$system/retroarch.cfg %ROM%

I would need a whole script to run and generate a uae file and pass that as argument instead of %ROM%. I could possibly not name it as lr-puae but would that be the best way?

PR #1887, opened 2017-02-13
Support for Git repositories that use Submodules, drivers module for ControlBlock

petrockblog:
I have added support for Git repository that use Git submodules. Also, I have added a driver module for the ControlBlock hardware, which is similar to the PowerBlock module.


joolswills:
Thanks - A good idea to put the submodule line in the gitPullOrClone function - I have now removed this from a few modules that did it manually. I also wrapped the git calls in the runCmd function to better catch/log errors.

PR #1890, opened 2017-02-17
Mupen64 renames

joolswills:


joolswills:
@gizmo98 upstream has done some renames - if you have a moment I would be happy for any feedback on this change (if I missed anything).

joolswills:
One odd decision about the rename is they renamed glupen64 to location of the previous mupen64plus-libretro, so old retropie-setup will now install glupen64 for mupen64plus.

Would have been better to avoid using the same repo name imho.

gizmo98:
I have no ready to use pi at hand atm. At first glance everything looks good. Is lr-parallel-n64 working? I thought it is a vulkan port of mupen64plus and the angrylion video plugin. All PIs have no vulkan support and are to slow for angrylion LLE video plugin. Could be they replaced lr-mupen64plus with lr-glupen so they have a good OpenGL LLE/HLE core and a good Vulkan LLE core.

joolswills:
parallel-n64 is the same as the previous mupen64-libretro repo. Glupen64 has been forked/renamed to mupen64-libretro

joolswills:
Parallel-n64 includes the other plugins as before

gizmo98:
I would not be surprised if twinaphex removes this code in the near future. The vulkan port does not use old OpenGL video plugins and the name of the vulkan port is parallel. The repo would look much cleaner without the old code.

gizmo98:
https://www.libretro.com/index.php/nintendo-64-vulkan-low-level-emulator-parallel-pre-alpha-release/

joolswills:
Yeah, I've been following the progress. They haven't said anything about removing opengl.

https://www.patreon.com/posts/nintendo-64-core-8140492

>The old 'mupen64plus libretro' core has meanwhile been renamed to 'ParaLLEl N64', and this single core now has both GL and Vulkan renderers integrated! It will also switch on the fly to Vulkan or GL mode depending on which video driver you have selected in RetroArch

If they remove opengl I can just switch it to x11 targets only.

PR #1891, opened 2017-02-18
Scraper patch 1

dragon76n:
Fix an issue that the Scraper was not saving data because gamelist.xml files were missing. This script will create gamelist.xml files in all the roms folders.


joolswills:
Thanks but this is best kept as a 3rd.party script - which scraper are you referring to?

dragon76n:
This was written to fix a problem that I experienced with the Scraper that it built into the Emulation Station.

PR #1899, opened 2017-02-23
Download a modern version of Go

sselph:


joolswills:
thanks.

Are you moving to golang 1.8 for your scraper ?

I'm not keen on installing a systemwide go, unless it is packaged. Alternatively a custom install under the retropie structure would be fine - without messing around with the system version.

sselph:
Yeah just want to move to something newer than 1.3.3 and if we do want to
move forward on adding it to ES we need 1.6. Go has a API promise in that
any Go code written for Go1 will work with any future Go1 version so I just
chose the latest available.

I don't mind installing to retropie it actually makes things simpler and
safer. Do you mind giving some pointers on if I install to $md_inst in the
golang package, what is the correct way to reference that from the scraper
package? I'll make the changes this evening.

joolswills:
You can just make sure the scraper is forced to use it - I'm unfamiliar with go, so I would have to experiment myself :-)

joolswills:
You can reference it from `$rootdir/supplementary/golang` or so

sselph:
Thanks. Yeah the Go portion I know how to do. I just wanted to make sure I
used the correct varaible/method for referencing the go binary from another
package.

sselph:
I switched it to install to retropie and not to the system. Go expects to be in /usr/local/go so had to set GOROOT so the commands are a bit more verbose but my tests seem to work.

scriptmodules/supplementary/golang.sh
```diff
@@ -0,0 +1,45 @@
...
+
+function depends_golang() {
+    getDepends curl tar
+}
+
+function remove_golang() {
+    if [[ -d "${md_inst}/go" ]]; then
+        rm -rf ${md_inst}/go
+    fi
+}
```
joolswills: this should not be needed as `$md_inst` is removed automatically.

scriptmodules/supplementary/golang.sh
```diff
@@ -0,0 +1,45 @@
...
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+rp_module_id="golang"
+rp_module_desc="Golang binary install"
+rp_module_section="opt"
+
+function depends_golang() {
+    getDepends curl tar
```
joolswills: Tar is not needed as a dependency (system depends on it).

Please use wget rather than curl for consistency with other modules (it is also not needed as a dependency).

scriptmodules/supplementary/golang.sh
```diff
@@ -0,0 +1,45 @@
...
+        local version=$(GOROOT=${md_inst}/go ${md_inst}/go/bin/go version | sed -e 's/.*\(go1[^ ]*\).*/\1/')
+    fi
+    echo Current Go version: $version
+    if [[ ! "${version}" < "go1.8" ]]; then
+        return 0
+    fi
+    remove_golang
+    local arch="armv6l"
+    if isPlatform "x86"; then
+        if [[ $(getconf LONG_BIT) == "64" ]]; then
```
joolswills: can use isPlatform "64bit"

scriptmodules/supplementary/golang.sh
```diff
@@ -0,0 +1,45 @@
...
+        rm -rf ${md_inst}/go
+    fi
+}
+
+function install_bin_golang() {
+    local version=""
+    if [[ -e "${md_inst}/go/bin/go" ]]; then
+        local version=$(GOROOT=${md_inst}/go ${md_inst}/go/bin/go version | sed -e 's/.*\(go1[^ ]*\).*/\1/')
+    fi
+    echo Current Go version: $version
```
joolswills: please use printMsgs "console"

scriptmodules/supplementary/golang.sh
```diff
@@ -0,0 +1,45 @@
...
+        if [[ $(getconf LONG_BIT) == "64" ]]; then
+            arch="amd64"
+        else
+            arch="386"
+        fi
+    fi
+    echo Downloading go1.8.linux-$arch.tar.gz
+    curl -o ${md_inst}/golang.tar.gz https://storage.googleapis.com/golang/go1.8.linux-$arch.tar.gz
+    tar -C ${md_inst} -xzf ${md_inst}/golang.tar.gz
+    rm ${md_inst}/golang.tar.gz
```
joolswills: Please quote these and use "$md_inst" rather than "${md_inst}" (we use the latter only when really needed - perhaps the second form is better, but it's not consistent with the rest of the code).

scriptmodules/supplementary/scraper.sh
```diff
@@ -14,15 +14,15 @@ rp_module_desc="Scraper for EmulationStation by Steven Selph"
 rp_module_section="config"
 
 function depends_scraper() {
-    getDepends golang
+    rp_callModule golang
 }
 
 function sources_scraper() {
-    GOPATH="$md_build" go get -u github.com/sselph/scraper
+    GOPATH="$md_build" GOROOT=$rootdir/supplementary/golang/go $rootdir/supplementary/golang/go/bin/go get -u github.com/sselph/scraper
```
joolswills: how about adding `_get_goroot_golang` function and doing
````
local goroot="$(_get_goroot_golang)
GOPATH="$md_build" GOROOT="$goroot/go" "$goroot/go/bin/go get -u github.com/sselph/scraper
```

scriptmodules/supplementary/scraper.sh
```diff
@@ -14,15 +14,15 @@ rp_module_desc="Scraper for EmulationStation by Steven Selph"
...
 }
 
 function build_scraper() {
-    GOPATH="$md_build" go build github.com/sselph/scraper
+    GOPATH="$md_build" GOROOT=$rootdir/supplementary/golang/go $rootdir/supplementary/golang/go/bin/go build github.com/sselph/scraper
```
joolswills: as above.

sselph:
Thanks. I think I addressed all the comments. I had a question about the GOROOT stuff but responded on that comment.

joolswills:
@sselph sorry for the delay - I missed/can't see your goroot comment sorry - what was it?

sselph:
Weird. I mentioned I wasn't sure where you proposed to have the function.
It seemed logical to have it in the golang file but seemed more appropriate
as a constant. Let me know if this looks okay or if you envisioned
something else.

joolswills:
If it's ok I would prefer a function in golang that echos the root - avoids having more globals etc (even though it's unlikely to ever conflict with anything and is set as readonly).

sselph:
converted the constant to the _get_goroot_golang function

joolswills:
Thanks. Will give this a spin in the next day or two and then merge it.

joolswills:
thanks.

joolswills:
One thing I didn't realise until now what the size of a new go install - 280mb on disk and close to 90MB download - so it takes a while to install the scraper now on a rpi. I might move the scraper to "optional" packages, as it's currently set as a "configuration" tool, which is really not for things like this as you can't remove once installed.

I changed some code around - so there is no subfolder go in golang, and some other bits. Hopefully I didn't break anything :-)

sselph:
Sounds good. I checked my pi and it seems to use ~52M more disk space(and similar increase in download) than `sudo apt-get install golang` but yeah scraper is optional. Another thing is that golang is only needed to compile. Once compiled, golang can be removed since Go binaries are statically compiled. I don't know if that type of dependency is already possible but if it isn't it might not be worth any serious changes since the scraper is only temporarily needed too.

PR #1901, opened 2017-02-23
webgui commit

daeks:
Pulling the latest version of the module for commenting. No need to merge it to the repo yet :)

First question from my side: Is it possible to configure the port somehow? So user could decide which port it should run on.

### DIFF
--- a/scriptmodules/supplementary/webgui.sh
+++ b/scriptmodules/supplementary/webgui.sh
@@ -25,9 +25,9 @@ function install_bin_webgui() {
 
 function configure_webgui() {
     killall php
-    php -S "$(hostname -I | xargs):8080" -t "$md_inst" > /dev/null 2>&1 &
+    php -S "$(ip route get 8.8.8.8 | head -1 | cut -d' ' -f8):8080" -t "$md_inst" > /dev/null 2>&1 &
 
-    local config="php -S \"\$\(hostname -I \| xargs\):8080\" -t \"$md_inst\" > /dev/null 2>\&1 \&"
+    local config="php -S \"\$\(ip route get 8.8.8.8 \| head -1 \| cut -d' ' -f8\):8080\" -t \"$md_inst\" > /dev/null 2>\&1 \&"
     sed -i "s|^exit 0$|${config}\\nexit 0|" /etc/rc.local
 }
 
--- a/scriptmodules/supplementary/webgui.sh
+++ b/scriptmodules/supplementary/webgui.sh
@@ -1,0 +1,38 @@
+
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="webgui"
+rp_module_desc="RetroPie WebGUI on port 8080"
+rp_module_section="exp"
+
+function depends_webgui() {
+    getDepends sqlite3 php5 php5-sqlite
+}
+
+function install_bin_webgui() {
+    gitPullOrClone "$md_inst" "https://github.com/daeks/RetroPie-WebGui"
+}
+
+function configure_webgui() {
+    killall php
+    php -S "$(hostname -I | xargs):8080" -t "$md_inst" > /dev/null 2>&1 &
+
+    local config="php -S \"\$\(hostname -I \| xargs\):8080\" -t \"$md_inst\" > /dev/null 2>\&1 \&"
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks.

Yes you would need a gui_webgui function - check out some other modules for examples.

scriptmodules/supplementary/webgui.sh
```diff
@@ -0,0 +1,38 @@
...
+    getDepends sqlite3 php5 php5-sqlite
+}
+
+function install_bin_webgui() {
+    gitPullOrClone "$md_inst" "https://github.com/daeks/RetroPie-WebGui"
+}
+
+function configure_webgui() {
+    killall php
+    php -S "$(hostname -I | xargs):8080" -t "$md_inst" > /dev/null 2>&1 &
```
joolswills: I would not kill/ start the service here like that. I would create a little start script to avoid having to do this and keep it simple in /etc/rc.local - you can also then save the PID for starting/stopping etc

Also `ip route get 8.8.8.8 | head -1 | cut -d' ' -f8` is more reliable for getting the interface ip.

scriptmodules/supplementary/webgui.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="webgui"
+rp_module_desc="RetroPie WebGUI on port 8080"
```
joolswills: As we already have some other "guis" it might be good to include your name here and some more information in a rp_module_help.

daeks: Hmm, I dont care about the name so I have no idea how to name it. Depending on the later purpose like being a default web frontend for retropie I selected a very common name. Any idea here?

joolswills: Daeks' WebGui ? ;-)

Not sure, but just want to avoid confusion between the retropie-manager this, and I believe there was another in development on the forum (not sure if development stopped).

It would be nice to have an official one, just need one that is "feature" complete and well tested - so perhaps that will be yours ? :)

daeks:
Based on your recommendation, I rebranded the whole app with a new name "YARMan Web". Anyhow I like the name itself much better than the old one :)

Regarding moving the php webserver startup to a separate sh file, I am still struggling a bit.

daeks:
Another Question: 

At which point do you consider it as "feature complete"? 
Which features are required to be in for that? 

As YARMan is considered to be modular, there is no real "feature list" instead of a basic framework with some monitoring dashboard. The currently implemented module "Metadata Editor" is more or less the first module from my personal list of module wishes.

joolswills:
Regarding feature complete, it was more what you have planned etc - I guess if this did everything the other gui did and more for example. I've not had a chance to test it yet (sorry). Hopefully others will contribute to it too.

I can probably put something together for you for startup/stopping although it might have to wait some days due to lack of free time currently.

meleu:
Hello friends.

I think it would be valuable to have a script to manage the yarman in [its own repository](https://github.com/daeks/yarman) (just like the [`rpmanager.sh` that I made for RetroPie-Manager](https://github.com/botolo78/RetroPie-Manager/blob/retropie/rpmanager.sh)). And then we can manage yarman through the scriptmodule just like we do in [RetroPie-Manager's one](https://github.com/meleu/RetroPie-Setup/blob/master/scriptmodules/supplementary/retropie-manager.sh). Doing this way, we could have these options in the `gui_yarman` function:
- Start yarman now
- Stop yarman now
- Enable yarman on Boot
- Disable yarman on Boot

@daeks if you agree it's a good idea I can write the script and submit to your repository. And then I can help with this scriptmodule here.

Example of yarman's manager features in a form of a fictitious help:
```
prompt$ yarman.sh --help
Usage: yarman.sh [OPTIONS]

The OPTIONS are:

-h|--help           print this message and exit

--start             start the yarman

--stop              stop the yarman

--isrunning         show if yarman is running and the listening port and exit

The --start and --stop options are, obviously, mutually exclusive. If the
user uses both, only the first works.
```

joolswills:
@meleu Thanks - I agree the start/stop script would be best included with in the repo with the application. Thanks for offering to do that :)

daeks:
I appreciate also your support as I am very bad with bash scripting 😊

daeks:
thinking on your idea other would be good to have a reset option as well. This option should be able to delete all *.sdb files in the data directory which will be introduced with sqlite implementation next week.

meleu:
OK. Maybe it's better to talk about it on your repository, I'll try to submit a PR in a few days.
By the way, did you see this PR?: https://github.com/daeks/yarman/pull/34

daeks:
Closing that PR as @meleu will open a new one for it :)

PR #1902, opened 2017-02-23
Swapping the mapping for L and Z buttons, because having the actual t…

clone206:
…rigger button mapped to Z makes more sense, since on the N64 controller Z is a trigger, whereas L is a shoulder.

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/mupen64plus.sh
@@ -212,6 +212,15 @@ function onend_mupen64plus_joystick() {
         fi
     done
 
+    # If there is no Z Trig try to map the L shoulder
+    # button to it via copying over the existing L Trig
+    # value and deleting it (L Trig) after
+    if ! grep -q "Z Trig" /tmp/mp64tempconfig.cfg ; then
+        iniGet "L Trig"
+        iniSet "Z Trig" "${ini_value}"
+        iniDel "L Trig"
+    fi
+
     echo "; ${DEVICE_NAME}_END " >> /tmp/mp64tempconfig.cfg
     echo "" >> /tmp/mp64tempconfig.cfg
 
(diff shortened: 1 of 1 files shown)


HerbFargus:
what about all the people who use snes controllers that don't have trigger buttons?

clone206:
They could still edit the AutoInputCfg.ini file. Seems to me that doing something nonstandard, like using an snes controller to play n64 games, should require the hack; as opposed to trying to use a controller that more closely approximates the experience of the original and needing to apply a hack to do so.

I suppose it would also be easy enough to allow for a Z/L swap param, similar to how a and b can be swapped in emulationstation. I'm open to trying to implement that as well.

clone206:
Could also add some logic in onend_mupen64plus_joystick() to just handle this the same way mapping the d-pad values works when there's no axis present. If there's no left trigger, then make the left shoulder button be Z.

joolswills:
I don't want to make a decision on this - not convinced myself but as @gizmo98 did the code he may have a reason for the current mapping - but I'll leave the decision to him.

gizmo98:
I had snes controllers in mind. They have only one L/R shoulder button. Z is a much more important button for N64 than L. In my opinion a workaround like "if there is only L1 map Z to L1 else L1 = L and L2 = Z" would be ok.

clone206:
@gizmo98 That does seem like the best solution. Shall I give it a try and make a pull request? Would you rather?

gizmo98:
You can give it a try. I would try to read "L TRIG" and "Z TRIG" with iniGet. If both keys contain values swap them and write them back with iniSet.

clone206:
Ok, did a fair amount of testing with and without trigger buttons present on the input device and InputAutoCfg.ini is generated as expected, and controls work as expected, each time on configuration of the controller via emulationstation.

gizmo98:
Nice. This looks much more elegant.

joolswills:
thanks for this.

PR #1907, opened 2017-03-04
--frames isn't used by retroarch anymore

boarpig:
This error is preventing netplay from working in recent release. --check-frames is new sort of similar(but not the same) option but isn't necessary to be adjusted if it works with default values.


joolswills:
thanks.

joolswills:
I removed this from the retropie-setup configuration also

boarpig:
Nice, I was so concerned with fixing the bug so I could get back to playing that I forgot other consequences

joolswills:
NP. The configuration tools is one of the older parts of retropie-setup and needs some further updating too.

Issue #1880, opened 2017-02-07
Add Amiberry amiga emulator


joolswills:
From @midwan

>You might want to include Amiberry in the list of Amiga emulators of RetroPie. It's a version forked from uae4arm-rpi, but with several improvements and bug fixes. I just pushed out a binary release of v2.1, which was tested on Raspbian-based distros (including RetroPie): https://github.com/midwan/amiberry/releases

>I'm currently working on an SDL2 version also, with the final aim of getting everything up to the WinUAE/FS-UAE standards and then get a libretro version out as well. But there's still work to be done in that part, so for now the current release is still based on SDL1.2.

>I'll notify you when the SDL2 version is ready for release.

PR #1909, opened 2017-03-05
Master

ChoccyHobNob:
This resolves issue Add Amiberry amiga emulator #1880

I used the new installer script for UAE4ARM to install this too, it has the same shell script for launching and same default .uae configs for machine type. I’ve added .uae as a valid rom file type for people with custom settings in whdload and things like that.

While I was in platforms.cfg I added the atari800 filetype .car to Atari 5200 as well, it’s a valid cart type there too.


joolswills:
Thanks, but this needs to be implemented in a manner to avoid duplication of code/scripts.

I have some plans for this, so I would prefer to implement this myself.

ChoccyHobNob:
Is it OK if I issue a separate pull request for the .car file format in atari800 and the .uae filetype in amiga? They got killed in the rejection of this amiberry change.

Is there anything I can do to help? I'm keen to help improve all of the classic 8/16 bit computers if I can

joolswills:
Thanks - don't worry about a pr. I can add the .car extension. for .uae I may do it differently so it's activated only if uae4arm is installed.

regarding help - I can't think of anything specific right now. is there something you wanted to see improved ?

ChoccyHobNob:
OK I won't send a pr then :)

Things I want to change in the emulators... button mapping for controllers in the computer emulators is the biggie for me, I want to pick a game from a list have it map any keyboard controls required to an available button on the controller and quit with a button press when I'm done, trying to make the computer emulators feel more consoleish when 20foot gaming. 

To do that there're changes to the actual emulators required and some changes to retropie. If I can get controller mapping added to the emulators (atari800/vice/uae4arm/fuse particularly) a generic config file would fit most games for each emulator, most games used similar buttons and the oddities could have a custom config file. Then retropie would need to be able to load config files as well (I see an issue with duplicates if there is a cfg and an atr file for the same game). 

I see you are going down the same road with uae4arm.sh. the reason for the amiberry pr was because amiberry now support a quit button and an extra controller button and is being actively developed so they are heading the same way.

Any thoughts on being able to override the ES systems order in a file like platforms.cfg does for filetypes/logos etc.? 

Single line at the end requiring a massive change lol, don't you love when users do that!

joolswills:
You are meaning to get upstream support for controller mapping in the emulators? vice already has some controller mapping configs afair, but we don't autogenerate configs for it.

Sounds like a lot of work.

I have considered some order gui, but it's a low priority (people can manually re-order as they like and maintain their own config if they want it in a custom order).

ChoccyHobNob:
Yes I am. I mean to try and help add that if I can, now that emus like hatari and vice have gone fully sdl2 in the last updates, atari800 is slowly moving that way, so is amiberry. It should be possible to do in a pretty standard way across multiple emulators (this would help with libretro cores for these emulators too) 

Very little that is worthwhile is easy or quick!

Agree about the low priority order gui. I ask because I support a few retropie users in a group on facebook and one of the common problems we get is users modifying es_systems and then not understanding why the newly added emulator wont appear after they add it. nds support recently caused a huge kerfuffle with people who can barely use a shell trying to use nano to add a new system to es_systems or moaning about losing all their customizations when they went back to the default!

joolswills:
SDL2 brings with it it's own issues of course (We have a fair number of patches maintained in our custom SDL2, and it seems difficult to get stuff upstream with tickets with no replies etc).

Although I am happy for usability improvements, I'm also not wanting RetroPie to become something where everything is hidden from the user. I like the fact it is customisable, and I don't want to dumb it down (TBH it has already been dumbed down far too much and with that all the problems we have with people selling it, and so on).

If someone wants to reorder their systems list, but struggled to copy a file, they might do better to not re-order it, or better still, spend a little time in the docs etc so they understand the tools they are using.

ChoccyHobNob:
I've had the same problem with SD2 upstream, they have a fairly active forum that I've had some luck getting things changed by discussing it there. 

I agree about the usability of Retropie, the number of users we support that come to the group after buying a ready made system or downloading a pre loaded image file and needing support because it's somehow broken far outweighs the number we support who build from scratch. It's nice to see them learning slowly, realise how crap that makes the experience and take the leap to a custom build. That's what makes it worthwhile for me. 

You could argue that's an elitist attitude (and that argument happens regularly in my group LOL) that what we are saying is that only people who know linux should be able to play old games.

joolswills:
I don't think it's elitist anyway - the Raspberry Pi was originally designed as a learning tool - and we are happy to help people. You don't need to know shell scripting to use RetroPie. There are guis for the main things, and most configurations can be accessed remotely.

However emulation is not always straight forward. Reading the documentation and spending some time familiarising yourself is a good idea. eg. learning about the versions of MAME etc.

The problem is some people don't do any reading - even the *read me first post* on the forum.

There are lots of alternatives to RetroPie too - so those that are not happy, are free to try another solution.

ChoccyHobNob:
We have 5 rules in a pinned post in our group, I still ban people all the time for breaking them who then message me and say they didn't know. I've been asked to write a FAQ, I honestly don't think there is much point, every answer will be a link to the retropie wiki anyway and nobody will read the FAQ let alone follow the links to the wiki.

I build MAME mainline for the Raspberry Pi (and mac) every month btw, and I wrote the demystifying mame roms doc that everyone links now.

PR #1910, opened 2017-03-06
added amiberry - fixes #1880

### LINKED ISSUE #1880: Add Amiberry amiga emulator

joolswills:
@ChoccyHobNob - i've tested this yet - but your PR at least motivated me to get this done :) (and actually did help, as I could see amiberry still used the same commandline parameters etc for now)

### DIFF
--- a/scriptmodules/emulators/amiberry.sh
+++ b/scriptmodules/emulators/amiberry.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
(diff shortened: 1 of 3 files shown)


ChoccyHobNob:
I'll give it a test, cheers!

joolswills:
it will always use the .uae file if it exists, so you can override any settings for the adf. In the case of using .uae files (which are not needed now for basic usage), you may want to put the adf files elsewhere, so they don't show in the list.

I may refactor to make the start script get removed properly etc (or at least not re-added).

joolswills:
thanks for testing.

PR #1913, opened 2017-03-07
Fix _del_system_emulationstation

diogomsantos:
Should be $name instead of $system. There is no $system.


diogomsantos:
Well, it works like it is because, you define $system earlier, but if you use the function externally it doesn't work. The purposed seems more logical.

joolswills:
Thanks - yes, must have broken this during the splitting out of the frontend functions. Appreciated.

PR #1916, opened 2017-03-10
platforms.cfg - added 7zip extensions to some systems that should support it via retroarch.

joolswills:
This needs some testing - I may have missed some also. Some cores do their own loading so I suspect 7zip wouldn't work for them.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -13,19 +13,19 @@ apple2_fullname="Apple II"
 arcade_exts=".fba .zip"
 arcade_fullname="Arcade"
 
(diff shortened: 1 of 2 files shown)


HerbFargus:
Unrelated but while we are at it we should alphabetise it to quell my OCD ;)

joolswills:
It mostly it, just a few out of place - intellivision/wii etc.

I will do that once the extensions have been sorted :)

avuton:
Thanks! I was in the process of checking every platform I use, they all work great so far.

PR #1922, opened 2017-03-14
mupen64plus: use config version of branch "flicker fix"

gizmo98:
Use version 15 until we can use upstream again.
https://github.com/joolswills/GLideN64/blob/flicker_fix/src/Config.h#L7


joolswills:
thanks.

PR #1923, opened 2017-03-14
Minivmac changes

joolswills:
@HerbFargus - does this look ok - just changed system and added dependency. Also removed !mali as sdl2 stuff should be fine on mali.


joolswills:
herb msged looked ok - merging.

PR #1932, opened 2017-03-22
mupen64plus: cleanup and add GlideN64 upstream again

gizmo98:
-remove graphics plugin videocore.
-change some GlideN64 settings which produced glitches.
-enable FB emulationen for all games. Increases compatibility. May be a little bit slower but solves misalignment problems for now.
-use GlideN64 upstream again.


joolswills:
Thanks!

joolswills:
note I also rebuilt binaries.

PR #1938, opened 2017-03-26
scriptmodule for joystick-selection tool

meleu:
It's a scriptmodule to install the joystick-selection tool. This way it'll be easier for users to update it.


joolswills:
There is some functionality overlap with work I did in the configuration editor. Not sure i want this as is in main RetroPie - at least as a 3rd party maintained script. Sorry.

PR #1948, opened 2017-04-11
Automatically place the lr-fbalpha hiscore.dat in the appropriate pla…

dankcushions:
…ces.

Similar to https://github.com/RetroPie/RetroPie-Setup/commit/de59962c2ef7db8b223e7c264244bc27ba2b0147.

Note, lr-fbalpha needs it to be in the SAVE_DIRECTORY (`/roms/fba/` or `/roms/arcade/` or `/roms/neogeo/` directories for us), rather than the SYSTEM_DIRECTORY (`/BIOS/` for us). This ends up with 3 copies of the same 231KB file. Potentially you could have it in one place and symlink, but maybe that's over-engineering it. Also I seem to recall symlinks don't work properly in FAT32 (USB drives, etc)?

...or we could see if lr-fbalpha would first accept a PR to move the hiscore path to the SYSTEM_DIRECTORY path.

### DIFF
--- a/scriptmodules/libretrocores/lr-fbalpha.sh
+++ b/scriptmodules/libretrocores/lr-fbalpha.sh
@@ -37,6 +37,7 @@ function install_lr-fbalpha() {
         'gamelist.txt'
         'whatsnew.html'
         'preset-example.zip'
+        'metadata'
     )
 }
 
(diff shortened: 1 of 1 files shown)


joolswills:
I think it would be worth seeing if upstream would be willing to change this (for consistency).

dankcushions:
Agreed! Done, and accepted: https://github.com/libretro/fbalpha/pull/129

Updated and squashed this PR to suit.

joolswills:
thanks.

PR #1951, opened 2017-04-14
add Sonic Robo Blast 2

HerbFargus:
This is a working module for Sonic Robo Blast 2 built on the doom legacy engine. I have tested it on both my pi3 and my ubuntu 16.10 mate laptop.

Tagging @alama @Logan-A and @Yukitty from Sonic Team Junior: Brilliant game I love it! I want to run this by you guys first to ask if we can include an installer in RetroPie for srb2 and if so: can we pull from the assets on your site or would you prefer we host them ourselves?

### DIFF
--- a/scriptmodules/ports/srb2.sh
+++ b/scriptmodules/ports/srb2.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
I implemented this manually with a change to host the assets ourselves. Thanks.

Yukitty:
srb2.org has not historically been a reliable host. I would recommend rehosting assets yourself or finding other mirrors to use if only to avoid arbitrary unexpected downtimes.

PR #1954, opened 2017-04-19
iniProcess returns 0 if not editing retroarch.cfg

meleu:
The test `[[ "$file" =~ retroarch\.cfg$ ]]` in the end of iniProcess was making the function always return a non-zero (fail) when not editing a retroarch.cfg file, even if everything was fine in iniProcess.

This change makes iniProcess return zero after that test.


joolswills:
thanks.

PR #1958, opened 2017-04-22
joy2key starters get button mappings from retroarch-joypads cfgs

meleu:
@joolswills 
Not sure if you'll want to merge it the way I'm doing here, but I'm submitting this PR so we can talk about the method I used to achiev what we were talking [in this forum topic](https://retropie.org.uk/forum/post/70225)...

Summarizing what the function `joy2key_input_config()` in runcommand.sh does:

1. gets the name of `/dev/input/jsN` device.
2. checks if there's a `retroarch-joypads/JOYSTICK.cfg` for the respective device.
3. gets the **A** and **B** button number.
4. echoes the parameters needed for joy2key.py setting the **A** for `<enter>` and **B** for `<tab>`.
  4.1. if the global retroarch.cfg has `menu_swap_ok_cancel_buttons = true`, **A** is set to `<tab>` and **B** to `<enter>`.
5. If the function fails in any of the steps above the default behavior is button 0 = `<enter>` and button 1 = `<tab>`.

**Some notes:**

- The main thing to note is that in `start_joy2key()` I'm launching one instance of joy2key.py for each `/dev/input/jsN` device, but I'm keeping track of all the PIDs to kill them in `stop_joy2key()`.

- The function `joy2keyInputConfig()` in helpers.sh works similarly, but using `<space>` rather than `<tab>`.

- The function `joy2keyStart()` in helpers.sh has a feature that let the caller explicitly set the parameters for joy2key, I'm being careful to not break this feature. When the caller explicitly sets those params, `joy2keyStart()` does **NOT** get the retroarch joypad configs. (AFAIK the wikiview.sh is the only `joy2keyStart`'s client that use it this way.)

- I tested both retropie_setup and runcommand here with 3 different controllers, tried different mappings in `retroarch-joypads/JOYSTICK.cfg` for each controller, tested the `menu_swap_ok_cancel_buttons` thing, all of it several times, etc. and everything seems to be fine. **But I didn't test it with those bluetooth joysticks with weird button numbers yet** (don't have one at hand currently).

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -937,19 +937,40 @@ _EOF_
 ## see: http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html
 function joy2keyStart() {
     local params=("$@")
-    if [[ "${#params[@]}" -eq 0 ]]; then
-        params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20)
-    fi
 
-    # get the first joystick device (if not already set)
-    [[ -c "$__joy2key_dev" ]] || __joy2key_dev="/dev/input/jsX"
+    # get all joystick devices, if not already set
+    if [[ ! -c "$__joy2key_dev" ]]; then
+        # if the function client passed parameters explicitly, use the generic "jsX"
+        if [[ "${#params[@]}" -ne 0 ]]; then
+            __joy2key_dev="/dev/input/jsX"
+        else
+            __joy2key_dev=$(find /dev/input -name "js*")
+        fi
+    fi
 
     # if no joystick device, or joy2key is already running exit
     [[ -z "$__joy2key_dev" || -n "$SSH_TTY" ]] || pgrep -f joy2key.py >/dev/null && return 1
 
-    # if joy2key.py is installed run it with cursor keys for axis, and enter + space for buttons 0 and 1
+    # if NOT using the generic "jsX" joystick, get the the custom mappings
+    if [[ "$__joy2key_dev" != "/dev/input/jsX" ]]; then
+        local dev
+        iniConfig " = " '"'
+        for dev in $__joy2key_dev ; do
+            params=( $(joy2keyInputConfig) )
+            [[ -z "$params" ]] && params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20)
+
+            # if joy2key.py is installed run it with cursor keys for axis, and enter + space for buttons A and B
+            if "$scriptdir/scriptmodules/supplementary/runcommand/joy2key.py" "$dev" "${params[@]}" & 2>/dev/null; then
+                __joy2key_pid+=($!)
+            fi
+        done
+        [[ "${#__joy2key_pid[@]}" -gt 0 ]]
+        return $?
+    fi
+
+    # the code below executes only if the function client passed parameters explicitly
     if "$scriptdir/scriptmodules/supplementary/runcommand/joy2key.py" "$__joy2key_dev" "${params[@]}" & 2>/dev/null; then
-        __joy2key_pid=$!
+        __joy2key_pid=($!)
         return 0
     fi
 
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks. Unfortunately you must have misunderstood on the forum as I was not planning on putting code in retropie-setup to change the way parameters to joy2key, but intending on improving the joy2key tool itself so it manages mappings.

meleu:
> Unfortunately I was not planning on putting code in retropie-setup

No problem. I've made this just as a kind of "proof of concept". It's just a way to say to myself "well, it's feasible". :-)

>  improving the joy2key tool itself so it manages mappings.

Yeah. Implementing that feature directly in  joy2key.py is the best solution. No doubt about it. But unfortunately I'm not able to code using python (currently :-) )

On the forum you said:

> If you want to create a input configuration script to output a config for it which it can read, or make it use the retroarch configs directly, I would accept a PR for that.

I'm not sure if you meant "a **python** input configuration script" or if a bash script would be acceptable. As I suck at python coding, then I've made it in bash.

I'm closing this PR, since it won't be merged. But I'd like to continue the conversation in the comments here, if it doesn't bother you.

joolswills:
I meant an input configuration script to be triggered from ES (bash) to generate configs which joy2key will read, or have joy2key read the retroarch configs directly.

Passing specific devices to joy2key is considered deprecated, as the idea is that it handles all joysticks via the jsX parameter (it's just got the old behaviour for backwards compatibility). 

Running multiple joy2key and having more duplicate code between runcommand and retropie-setup would be a backwards step imho.

meleu:
I totally agree.

Let's forget the changes of this PR and think about another approach:

Here is a small bash script with a function able to get a `/dev/inpu/jsN` string as input and generate joy2key button mappings params based on retroarch-joypads configs as output: https://github.com/meleu/share/blob/master/joy2key_btn_mappings.sh

Usage examples:

```
$ # js0 is my 8Bitdo NES30 Gamepad
$ # checking the current configs
$ grep 'input_[ab]_btn' /opt/retropie/configs/all/retroarch-joypads/8Bitdo\ NES30\ GamePad.cfg 
input_a_btn = "3"
input_b_btn = "4"
$ grep menu_swap_ok_cancel /opt/retropie/configs/all/retroarch.cfg 
menu_swap_ok_cancel_buttons = false
$ 
$ # EXAMPLE 1: getting mappings from retroarch-joypads config
$ ./joy2key_btn_mapping.sh /dev/input/js0 
'' '' '' 0x0a 0x09
$ 
$ 
$ # changed menu_swap_ok_cancel_buttons to true
$ grep menu_swap_ok_cancel /opt/retropie/configs/all/retroarch.cfg 
menu_swap_ok_cancel_buttons = true
$
$ # EXAMPLE 2: similar to EXAMPLE 1, but "menu_swap_ok_cancel_buttons = true"
$ ./joy2key_btn_mapping.sh /dev/input/js0 
'' '' '' 0x0a 0x09
$
$ 
$ # deleting the joystick config file to see the default output
$ rm /opt/retropie/configs/all/retroarch-joypads/8Bitdo\ NES30\ GamePad.cfg 
$ 
$ # EXAMPLE 3: no retroarch-joypads config for this device
$ ./joy2key_btn_mapping.sh /dev/input/js0 
0x0a 0x09
```
Maybe you could make joy2key.py call this script to assign its `button_codes`. What do you think?

joolswills:
Sorry - I want to implement it in joy2key not as a separate script.

meleu:
And I'm sure you'll do far better than what I'm trying to do here. :-)

Thanks!

meleu:
@joolswills 
> I want to implement it in joy2key

I've learned at least the python syntax and tried a bit more. But I stopped when I realized that in the current joy2key there's only one button config for every joystick. My python knowledge is so poor that I can not figure out a way to make each joystick have its own `button_codes`.

But at least I managed to write a python function that takes the `/dev/input/jsN` as input and returns the list of button_codes according to what is in the retroarch-joypads cfg files.

If you want to take a look at it, here is the python script: https://github.com/meleu/share/blob/master/joy2key_get_button_codes.py

Summarizing what the function `get_button_codes()` does:

1. gets the name of `/dev/input/jsN` as argument.
2. checks if there's a `retroarch-joypads/JOYSTICK.cfg` for the respective device.
3. gets the **A** and **B** button number.
4. build and return a list of button codes where the item 0 is the code for button 0, item 1 for button 1, and so on. The button number defined for **A** receives the code for `<enter>`, the one for **B** receives the `<tab>`.
  4.1. if the global retroarch.cfg has `menu_swap_ok_cancel_buttons = true`, **A** is set to `<tab>` and **B** to `<enter>`.
5. If the function fails in any of the steps above the default list returned is button 0 = `<enter>` and button 1 = `<tab>`.

PR #1964, opened 2017-04-30
Request to add anthonycaccese's crt themes

anthonycaccese:
Seeing if it's possible to submit my themes to the official list.  They both currently support video previews and are in a stable state for use.

Details of both themes can be found here: https://retropie.org.uk/forum/topic/7022/new-theme-crt-with-support-for-video-preview.


joolswills:
It is. thanks.

anthonycaccese:
Thank you!

PR #1968, opened 2017-05-01
Change Eudora Repo to one being maintained

ChoccyHobNob:
Pull requests are not being merged, deleted or rejected by AmadhiX so
I’ve forked his theme and made the updates and bug fixes required.

### DIFF
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -55,9 +55,9 @@ function gui_esthemes() {
         'ehettervik luminous'
         'ehettervik minilumi'
         'ehettervik workbench'
-        'AmadhiX eudora'
-        'AmadhiX eudora-bigshot'
-        'AmadhiX eudora-concise'
+        'ChoccyHobNob eudora'
+        'ChoccyHobNob eudora-bigshot'
+        'ChoccyHobNob eudora-concise'
         'InsecureSpike retroplay-clean-canela'
         'InsecureSpike retroplay-clean-detail-canela'
         'RetroPie turtle-pi'
@@ -165,4 +165,3 @@ function gui_esthemes() {
         fi
     done
 }
-


joolswills:
Would like to get a reply from @AmadhiX before doing anything (if they don't reply, I will merge).

ChoccyHobNob:
There has been zero activity on his account since he pushed the themes in Nov 2015, I submitted a pull request last year that he never actioned, there are 2 pulls waiting for simple "add system" requests. I doubt you'll get any reply.

joolswills:
Perhaps not, but I will still give the author the opportunity.

ChoccyHobNob:
I think it's just easier if I try this again :/

AmadhiX:
Hi all!  Apologies for the extended leave of absence, and I'm glad you're enjoying the theme enough to contribute!  I'd like to become active on this again, but can't guarantee that I'll be around as much as would be useful.  So, I've gone ahead and invited @ChoccyHobNob to become a collaborator on the original repos.  If that's easier than working in the forks, have at it.  I'm not an expert at github so let me know if there's something else I could do to make it easier for others to contribute, short of making it entirely public.

AmadhiX:
On second look, it appears ChoccyHobNob's fork may already be in use.  That's fine by me.  Thanks joolswills for attempting to reach out.  And thanks Choccy for being respectful with the credits, etc., as well as for the updates to Eudora.

ChoccyHobNob:
Hi AmadhiX,
I’l go ahead and push my updates to your original repos then remove my updated version from retropie. It might take a few days, I’m in the middle of other things, but I will get to it. Nice to have you back!

joolswills:
Those that have the current repo, will need to remove then re-install the theme to get any updates if it's switched.This wasn't an issue before as we used to reinstall from scratch before, to work around a bug, this code isn't removed, so updating will just do a git pull, but won't change the repo.

PR #1980, opened 2017-05-13
joy2key can get button maps from retroarch configs

meleu:
Here we go with another attempt to implement custom button mappings on joy2key...

## Some notes:

- ~~**IMPORTANT:** this change on joy2key does **NOT** change the current behavior of `joy2keyStart()` in helpers.sh and `start_joy2key()` in runcommand.sh (I've submitted this PR to let you review the code).~~
  - ~~this `joy2key.py` only gets the custom configs from retroarch-joypads files if it's called **without those hexadecimal codes for buttons**~~
  - ~~i.e. in order to make `retropie_setup.sh` and `runcommand.sh` take advantage of the custom button mappings, we need to delete those hexadecimal codes for buttons on joy2key starters (`runcommand.sh` and `helpers.sh`).~~

- the package `python-pyudev` is a dependency. Needed to get `/dev/input/jsX` names. But I'm not sure where to `getDepends` for this (**update**: added to [system.sh](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/system.sh)?)

- I kind of translated to python what I tried in #1958, but avoiding launching one instance of joy2key for every controller. Now only one instance is enough.

- The function `get_button_codes()`:
1. gets the name of `/dev/input/jsN` device.
2. checks if there's a `retroarch-joypads/JOYSTICK.cfg` for the respective device.
3. gets the **A B X Y** button numbers.
4. build and return a list of button codes where the item 0 is the code for button 0, item 1 for button 1, and so on. The index with the button number defined for **A** receives the code for `<enter>`, the one for **B** receives the `<tab>` (**update**: it changed a little, see comments below).
  4.1. if the global retroarch.cfg has `menu_swap_ok_cancel_buttons = true`, **A** is set to `<tab>` and **B** to `<enter>`.
5. If the function fails in any of the steps above the default behavior is button 0 = `<enter>` and button 1 = `<tab>`.

- I tested both retropie_setup and runcommand with 3 different controllers (one generic USB controller, one USB adapter to PlayStation controllers, and one 8Bitdo NES30), tried different mappings in `retroarch-joypads/JOYSTICK.cfg` for each controller, tested the `menu_swap_ok_cancel_buttons` thing, all of it several times, etc. and **everything seems to be fine.**

- I didn't test it with those bluetooth joysticks with weird button numbers yet (e.g.: 8Bitdo Zero). I don't have one at hand right now...

### DIFF
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -10,7 +10,8 @@
 #
 
 import os, sys, struct, time, fcntl, termios, signal
-import curses, errno
+import curses, errno, re
+from pyudev import Context
 
 
 #    struct js_event {
@@ -29,6 +30,82 @@ JS_THRESH = 0.75
 JS_EVENT_BUTTON = 0x01
 JS_EVENT_AXIS = 0x02
 JS_EVENT_INIT = 0x80
+
+CONFIG_DIR = '/opt/retropie/configs/'
+RETROARCH_CFG = CONFIG_DIR + 'all/retroarch.cfg'
+
+def ini_get(key, cfg_file):
+    pattern = r'[ |\t]*' + key + r'[ |\t]*=[ |\t]*'
+    value_m = r'"*([^"\|\r]*)"*'
+    value = ''
+    with open(cfg_file, 'r') as ini_file:
+        for line in ini_file:
+            if re.match(pattern, line):
+                value = re.sub(pattern + value_m + '.*\n', r'\1', line)
+                break
+    return value
+
+def get_btn_num(btn, cfg):
+    num = ini_get('input_' + btn + '_btn', cfg)
+    if num: return num
+    num = ini_get('input_player1_' + btn + '_btn', cfg)
+    if num: return num
+    return ''
+
+def get_button_codes(dev_path):
+    js_cfg_dir = CONFIG_DIR + 'all/retroarch-joypads/'
+    js_cfg = ''
+    dev_name = ''
+
+    # getting joystick name
+    for device in Context().list_devices(DEVNAME=dev_path):
+        dev_name_file = device.get('DEVPATH')
+        dev_name_file = '/sys' + os.path.dirname(dev_name_file) + '/name'
+        for line in open(dev_name_file, 'r'):
+            dev_name = line.rstrip('\n')
+            break
+    if not dev_name:
+        return default_button_codes
+
+    # getting retroarch config file for joystick
+    for f in os.listdir(js_cfg_dir):
+        if f.endswith('.cfg'):
+            if ini_get('input_device', js_cfg_dir + f) == dev_name:
+                js_cfg = js_cfg_dir + f
+                break
+    if not js_cfg:
+        js_cfg = RETROARCH_CFG
+
+    # getting configs for buttons A, B, X and Y
+    btn_num = {}
+    biggest_num = 0
+    i = 0
+    for btn in 'a', 'b', 'x', 'y':
+        i += 1
+        if i > len(default_button_codes):
+            break
+        btn_num[btn] = get_btn_num(btn, js_cfg)
+        try:
+            btn_num[btn] = int(btn_num[btn])
+        except ValueError:
+            return default_button_codes
+        if btn_num[btn] > biggest_num:
+            biggest_num = btn_num[btn]
+
+    # building the button codes list
+    btn_codes = [''] * (biggest_num + 1)
+    i = 0
+    for btn in 'a', 'b', 'x', 'y':
+        btn_codes[btn_num[btn]] = default_button_codes[i]
+        i += 1
+        if i >= len(default_button_codes): break
+
+    # if button A is <enter> and menu_swap_ok_cancel_buttons is true, swap buttons A and B functions
+    if btn_codes[btn_num['a']] == '\n' and ini_get('menu_swap_ok_cancel_buttons', RETROARCH_CFG) == 'true':
+        btn_codes[btn_num['a']] = btn_codes[btn_num['b']]
+        btn_codes[btn_num['b']] = '\n'
+
+    return btn_codes
 
 def signal_handler(signum, frame):
     close_fds(js_fds)
@@ -58,6 +135,7 @@ def open_devices():
     for dev in devs:
         try:
             fds.append(os.open(dev, os.O_RDONLY | os.O_NONBLOCK ))
+            js_button_codes[fds[-1]] = get_button_codes(dev)
         except:
             pass
 
@@ -114,7 +192,9 @@ def process_event(event):
 
 signal.signal(signal.SIGINT, signal_handler)
 
+js_button_codes = {}
 button_codes = []
+default_button_codes = []
 axis_codes = []
 
 curses.setupterm()
@@ -125,7 +205,7 @@ for arg in sys.argv[2:]:
     if i < 4:
         axis_codes.append(chars)
     else:
-        button_codes.append(chars)
+        default_button_codes.append(chars)
     i += 1
 
 event_format = 'IhBB'
@@ -157,6 +237,10 @@ while True:
             event = read_event(fd)
             if event:
                 if time.time() - js_last[i] > JS_REP:
+                    if fd in js_button_codes:
+                        button_codes = js_button_codes[fd]
+                    else:
+                        button_codes = default_button_codes
                     if process_event(event):
                         js_last[i] = time.time()
             elif event == False:
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -198,7 +198,7 @@ function get_retropie_depends() {
         echo "deb http://archive.raspberrypi.org/debian/ $__os_codename main" >>$config
     fi
 
-    local depends=(git dialog wget gcc g++ build-essential unzip xmlstarlet)
+    local depends=(git dialog wget gcc g++ build-essential unzip xmlstarlet python-pyudev)
     if [[ -n "$__default_gcc_version" ]]; then
         depends+=(gcc-$__default_gcc_version g++-$__default_gcc_version)
     fi


scriptmodules/supplementary/runcommand/joy2key.py
```diff
@@ -30,6 +31,76 @@
 JS_EVENT_AXIS = 0x02
 JS_EVENT_INIT = 0x80
 
+RETROARCH_CFG = '/opt/retropie/configs/all/retroarch.cfg'
+DEFAULT_BTN_CODES = ['\n', '\t']
```
meleu: I used `\n` and `\t` directly to avoid the need to `.decode('hex')`.

scriptmodules/supplementary/runcommand/joy2key.py
```diff
@@ -30,6 +31,76 @@
...
+
+def ini_get(key, cfg_file):
```
meleu: Pretty much the same Regular Expressions found in inifuncs.sh with double-quotes `"` hardcoded.

joolswills: for python use ConfigParser

joolswills:
Please use ConfigParser as mentioned above, and remove any key hardcoding - it can still take the parameters as before (just map to A B X Y etc).

joolswills:
Please also hardcode `/opt/retropie` only once. You may be able to rid of that altogether by just using "$HOME/.retroarch` since we symlink that.

scriptmodules/supplementary/runcommand/joy2key.py
```diff
@@ -114,8 +187,10 @@ def process_event(event):
 
 signal.signal(signal.SIGINT, signal_handler)
 
+js_button_codes = {}
```
meleu: This dictionary will be used to store button codes for each connected controller using. The dictionary keys will be the file descriptors obtained in `open_devices()`.

meleu:
> Please use ConfigParser

got it. Will do.

> remove any key hardcoding - it can still take the parameters as before (just map to A B X Y etc).

Sorry. I didn't get what you mean. Could you point with comments in the code?

> Please also hardcode `/opt/retropie` only once.

Do you mean something like this?:
```
CONFIG_DIR = '/opt/retropie/configs'
RETROARCH_CFG = CONFIG_DIR + '/all/retroarch.cfg'
js_cfg_dir = CONFIG_DIR + '/all/retroarch-joypads/
```

> You may be able to rid of that altogether by just using `$HOME/.retroarch` since we symlink that.

Sorry, didn't get this too.

scriptmodules/supplementary/runcommand/joy2key.py
meleu: OK. I've read [ConfigParser docs](https://docs.python.org/2/library/configparser.html),and some examples and then tried here. It expects `[Sections]` in the config files.

I googled about how to use ConfigParser with sectionless files and the solutions I found look very ugly. IMHO much uglier than implementing `ini_get()` here.

Do you know some way to use ConfigParser with sectionless files?

meleu:
Using ConfigParser seems a bit overkill for this small application. [More comments here.](https://github.com/RetroPie/RetroPie-Setup/pull/1980#discussion_r116357306)

meleu:
@joolswills I noticed that helpers.sh sets button 1 to `<space>` ([line 941](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/helpers.sh#L941)), while runcommand.sh sets it to `<tab>` ([line 119](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/supplementary/runcommand/runcommand.sh#L119)).

Is that `<space>` in helpers.sh really necessary or it can be a `<tab>`? The joy2key of my PR sets **B** button always to `<tab>`, and it worked just fine in my tests navigating through retropie_setup.sh menus.

meleu:
I think I've just found the answer. And it is: **yes, we need the `<space>` in helper.sh**.

**Reason:** [scraper.sh](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/supplementary/scraper.sh) uses `dialog --checklist` and we need `<space>` to check one or more options...

Need to sort this out...

meleu:
@joolswills are you OK if I define the **X** button as `<space>`?

meleu:
@joolswills 
> remove any key hardcoding - it can still take the parameters as before (just map to A B X Y etc).

Oh! I think I've got what you meant here.

The arguments for joy2key currently are:
`up down left right button0 button1 [buttonN ...]`

You want it to be:
`up down left right A B X Y`

Am I right?

joolswills:
Yes :)

Regarding the comment about $HOME - ignore that, it was just thinking to avoid hardcoding of /opt/retropie but that is needed for the `all/retroarch.cfg` for now. So that's fine.

joolswills:
But can do as you mentioned, by putting `/opt/retropie` in a single variable such as ROOTDIR, to avoid using it twice.

meleu:
Great. Will work on this! Thanks.

meleu:
I think it's ready! :-)

scriptmodules/supplementary/runcommand/joy2key.py
```diff
@@ -30,6 +31,82 @@
...
+    if num: return num
+    return ''
+
+def get_button_codes(dev_path):
+    js_cfg_dir = CONFIG_DIR + 'all/retroarch-joypads/'
+    js_cfg = ''
+    dev_name = ''
+
+    # getting joystick name
+    for device in Context().list_devices(DEVNAME=dev_path):
```
meleu: The package python-pyudev is a dependency to make this trick to get the device name.
It's a very small package:
```
The following NEW packages will be installed:
  python-pyudev
(...)
Need to get 31.6 kB of archives.
After this operation, 193 kB of additional disk space will be used.
```

scriptmodules/supplementary/runcommand.sh
```diff
@@ -19,7 +19,7 @@ function _update_hook_runcommand() {
 }
 
 function depends_runcommand() {
-    local depends=()
+    local depends=(python-pyudev)
```
joolswills: as we use joy2key for retropie-setup navigation also, the dependency may be better to put in the main retropie-setup script. I was also thinking we could do it without the module perhaps - noticed http://stackoverflow.com/questions/43599936/pyudev-how-to-get-the-devices-friendly-name while I was googling ;-)

scriptmodules/supplementary/runcommand/joy2key.py
```diff
@@ -30,6 +31,82 @@
...
+        if btn_num[btn] > biggest_num:
+            biggest_num = btn_num[btn]
+
+    # building the button codes list
+    btn_codes = [''] * (biggest_num + 1)
+    for i in range(len(default_button_codes)):
+        if   i == 0: btn_codes[btn_num['a']] = default_button_codes[0]
+        elif i == 1: btn_codes[btn_num['b']] = default_button_codes[1]
+        elif i == 2: btn_codes[btn_num['x']] = default_button_codes[2]
+        elif i == 3: btn_codes[btn_num['y']] = default_button_codes[3]
```
joolswills: Wondering if the logic could be simplified here by looping though a list of a b x y, and incrementing i as an index, rather than lots of if/elif (and breaking out of the loop if len(default_button_codes) is reached.

meleu: I did this way to accomplish what you requested.

`default_button_codes` is a list with the button codes given as arguments to joy2key.

We need to loop through `default_button_codes` because we want to make `default_button[0]` be the code for **A** button, `default_button[1]` for **B** button, etc.

As joy2key don't know the length of `default_button_codes` beforehand, looping through a list of a b x y can launch an `IndexError` exception (e.g.: runcommand gives only two button codes, one for **A**, one for **B**).

As an improvement here I see an `else: break`, which will be executed only if the joy2key caller give more than 4 button codes (and the 5th code onwards will be ignored).

scriptmodules/supplementary/runcommand.sh
meleu: > the dependency may be better to put in the main retropie-setup script.

The dependencies for RetroPie-Setup are coded in which file?

> I was also thinking we could do it without the module perhaps.

Well, the package is really small:
```
The following NEW packages will be installed:
  python-pyudev
(...)
Need to get 31.6 kB of archives.
After this operation, 193 kB of additional disk space will be used.
```

I think it's doable without pyudev using `os.system()` (or other python function to execute shell commands) and executing those commands you used in [configedit.sh](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/supplementary/configedit.sh#L114). But IMHO it would be ugly and non-pythonic - and, as an aspiring python programmer, I need to be proud to do things in a pythonic way :-)

scriptmodules/supplementary/runcommand/joy2key.py
joolswills: Not sure I follow - Surely len(default_button_codes) gives the length ?

joolswills: eg something like (untested)
```
i = 0
for btn in 'a', 'b', 'x', 'y':
    btn_codes[btn_num[btn]] = default_button_codes[i]
    i += 1
    if i >= len(default_button_codes): break
```

meleu: Good catch! I think it will work. I will test when I get home. ;-)

joolswills: you can then have a single list of the button mappings then for both loops (we could add perhaps more in the future).

joolswills: I didn't look at the previous loop - maybe it's possible to combine then ? not sure.

scriptmodules/supplementary/runcommand.sh
joolswills: global dependencies are in system.sh - but the issue is not the size, but the delay to check each dependency (it's already a noticeable startup delay due to this. But it can be added sure.

joolswills: I may change the startup dependency checks to be a "quick check" without using dpkg.

joolswills: I don't have a problem if it did it without using the code - might be worth looking how the module does it since your only use a single function of it anyway and then access sysfs manually - we handle ini manually, and the actual joystick handling is low level rather than via pygame etc.

meleu: For now I'll remove it from runcommand.sh and put in system.sh. And later we can check how the module does to implement it directly. Are you OK with it?

scriptmodules/supplementary/runcommand/joy2key.py
meleu: > I didn't look at the previous loop - maybe it's possible to combine them ?

yes, it is. Will do it this evening.

meleu: I changed the logic as you suggested. It works perfectly. https://github.com/RetroPie/RetroPie-Setup/pull/1980/commits/ac33c990fba5692e2a844bc7f17d476367686ef2

But combining the loops would bring an unnecessary complexity, since we don't know the biggest number until the first loop end.

When initializing `btn_codes` list to have the length of `biggest_num` we can simplify the list manipulation simply using `list[n] = n`. There's no need to worry with indexes and list methods (`.append()` etc).

meleu:
added python-pyudev package to system.sh and removed from runcommand dependencies. https://github.com/RetroPie/RetroPie-Setup/pull/1980/commits/c9c7b15727319e17dd287ada36b8fab2d863746a

joolswills:
Please can you squash the commits and I will merge. thanks.

meleu:
Sorry my ignorance, I don't know how to squash commits. I'm googling about it and it seems that there are several ways to do it...

I'm thinking about closing this PR and submitting another with only one commit...

HerbFargus:
quick summary on rebasing/squashing

`git rebase -i <hash of commit before first commit>`

then should open a file with commits like:

```
pick hash1
pick hash2
pick hash3
```
edit to:
```
pick hash1
squash hash2
squash hash3
```
then it will squash all commits into hash1

then it will open up you commit messages for all commits. usually I just delet them all and keep the first and modify it to summarise all commits

then:

`git push --force origin branchname`

that should be it.

meleu:
Thanks herb. Will do it as soon as I get home.

meleu:
**Wow!** It really works!! Thanks @HerbFargus .

@joolswills squash commit has been done.

Cheers!

joolswills:
Thanks!

meleu:
@joolswills would you mind to bump the version to 4.2.4? Then I can tell forum users that the "press a button" behaviour has changed since 4.2.4.

PR #1989, opened 2017-05-27
Putting drastic back to RetroPie

meleu:
This is the very same code as it was before being removed.

DraStic creator answered the @mediamogul message [here](http://www.drastic-ds.com/viewtopic.php?f=5&t=4307&sid=7bc928f81c7633d7c1577ad1bb1d3b19&start=10). This part of the message makes me believe that there's no problem to add DraStic to RetroPie-Setup again:

> I think at this point there's no real harm keeping it in there, if you don't have any misgivings doing so. I feel bad seeing it be released like this then taken away, and it's otherwise been out there so there's nothing to really hold back now.

### DIFF
--- a/scriptmodules/emulators/drastic.sh
+++ b/scriptmodules/emulators/drastic.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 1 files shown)


joolswills:
Is this a revert or re-add? Would prefer to just revert the commit

meleu:
A re-add.
My intention is only tell you about his answer in a more practical way. :-)
Could you close this PR and revert that commit as want, please?

Cheers!

joolswills:
No worries. Thanks. I saw the post earlier but have been at a BBQ all day. Will revert it when next free

PR #1990, opened 2017-05-30
retroarch: switch to xmb menu driver & disable xmb icon shadows

psyke83:
* Switch back to retroarch's new default xmb menu driver
* Disable xmb icon shadows (this was causing performance issues)
* Install retroarch assets by default (needed for xmb)

Things to consider for pull request:
* The xmb driver menu has very choppy scrolling with default settings on RPi3, but the menu icon shadow setting appears to have been the sole culprit of the performance issue. With this setting disabled (in this patch), the menu scrolling is now 100% smooth for me, but I can't say the same for older Pi revisions.
* Retropie assets (needed for xmb) occupy ~148MB in /opt/retropie/configs/all/retroarch/assets.
* I've seen reports in an old pull request that the xmb driver was causing segfaults for N64 cores when threaded video is enabled, but I can't reproduce the issue on my system (tested lr-mupen64plus and lr-parallel-n64). See: https://github.com/RetroPie/RetroPie-Setup/pull/1674
* One cosmetic "issue" with the xmb driver is that the background does not become transparent unless you disable "Pause when menu activated". However, I submitted a patch for this issue which has been merged upstream, so the behaviour will change in later retroarch releases: https://github.com/libretro/RetroArch/commit/57503051ea33c619a92d9f795708e710897a338b

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -110,6 +110,9 @@ function configure_retroarch() {
     # install shaders by default
     update_shaders_retroarch
 
+    # install assets by default (needed for xmb menu driver)
+    update_assets_retroarch
+
     local config="$(mktemp)"
 
     cp "$md_inst/retroarch.cfg" "$config"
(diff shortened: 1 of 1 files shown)


joolswills:
I don't want to move to xmb by default currently I'm afraid - keeps the image size down. I'm happy to make it easier to switch to though as an option.

joolswills:
(There are other reasons to keep the default, certainly it could be a discussion to have, but I would want to have that on the forum before making a change like this to the interface). Need to do testing with older rpi's also.

psyke83:
Yes - understood. How about I open a new pull request just to add the xmb icon shadow disable option to the default configuration? That way, people who manually switch to xmb can see the full performance.

joolswills:
I'm happy to include that. Thanks.

psyke83:
joolswills,

I've opened a new thread on the forums regarding the idea of making it the default selection: https://retropie.org.uk/forum/topic/10972/retroarch-xmb-performance-on-raspberry-pi-1-2-zero/4

Still waiting to hear back on performance for older Pi models, but something to keep in mind is that we only need 6MB of assets to make the default icon theme work - you can check yourself by deleting everything in /opt/retropie/configs/all/retroarch/assets except xmb/monochrome. We could package the assets along with the binary, or maintain a small .zip archive that will not interfere with installation of the full assets.

PR #1991, opened 2017-05-30
retroarch: disable xmb menu icon shadows

psyke83:
This causes choppy scrolling performance even on a RPi3, so disable it by
default to benefit users who manually switch to the xmb menu driver.


joolswills:
Thanks

PR #1992, opened 2017-05-30
emulationstation: sync A/B swap configuration to retroarch config

psyke83:
When configuring "Swap A/B buttons in ES", users will most likely
need to enable the equivalent option (menu_swap_ok_cancel_buttons)
in RetroArch. Do this automatically to minimize user confusion.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -114,6 +114,10 @@ function configure_retroarch() {
 
     cp "$md_inst/retroarch.cfg" "$config"
 
+    # query ES A/B key swap configuration
+    local es_swap="false"
+    getAutoConf "es_swap_a_b" && es_swap="true"
+
     # configure default options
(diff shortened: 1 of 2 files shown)


psyke83:
For your consideration, two changes to help keep the A/B key swap setting of ES synchronized with RetroArch.

emulationstation.sh: modify the retroarch.cfg when the user changes the "Swap A/B Buttons in ES" configuration.
retroarch.sh: query the equivalent setting in autoconf.cfg & set in retroarch configuration to keep behaviour consistent during (re)install or upgrade.

joolswills:
Sorry, other merges I did first have created conflicts so this needs rebasing. Thanks

psyke83:
It's rebased now.

Just a little note on retroarch.sh: the getAutoConf helper function clobbers the existing iniConfig configuration - that's why I invoked it earlier in the script.

joolswills:
Thanks

PR #1993, opened 2017-05-30
mupen64plus: add some improvements

gizmo98:
-use new CRC optimization if platform is not rpi3. rpi3 has a fast hardware crc32 implementation.
-use texture packs if available.
-remove GLideN64 system lib fix.
-use VEC4 OPT again and enable more NEON optimizations.


dankcushions:
I tested a source build on RPI3 - worked fine :) no performance improvements that i could see in my quick test but no regressions either.

joolswills:
Thanks :-)

PR #1995, opened 2017-06-01
Mupen64plus: fix target rpi1

gizmo98:
-There was a build error if mupen64plus was build with `sudo __platform=rpi1 ./RetroPie-Setup`. -DARMv5_ONLY will be only used if HOST_CPU is armv5 or armv6.
-Don't mix neon and videocore options. Only rpi platforms need VC=1.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -66,8 +66,9 @@ function build_mupen64plus() {
         if [[ -f "$dir/projects/unix/Makefile" ]]; then
             make -C "$dir/projects/unix" clean
             params=()
-            isPlatform "rpi1" && params+=("VC=1" "VFP=1" "VFP_HARD=1" "HOST_CPU=armv6")
-            isPlatform "neon" && params+=("VC=1" "NEON=1")
+            isPlatform "rpi1" && params+=("VFP=1" "VFP_HARD=1" "HOST_CPU=armv6")
+            isPlatform "rpi" && params+=("VC=1")
+            isPlatform "neon" && params+=("NEON=1")
             isPlatform "x11" && params+=("OSD=1" "PIE=1")
             isPlatform "x86" && params+=("SSE=SSSE3")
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

joolswills:
We should probably also do

```
isPlatform "armv6" && params+=("VFP=1" "VFP_HARD=1" "HOST_CPU=armv6")
```

instead of

```
isPlatform "rpi1" && params+=("VFP=1" "VFP_HARD=1" "HOST_CPU=armv6")
```

(We don't support any other armv6 platforms, but it makes sense to use the armv6 platform flag).

joolswills:
What was the error without using "HOST_CPU=armv6" btw ? Is that a new thing, as I'm sure I built the binaries previously for rpi1 without. Also - do we actually need `"VFP=1" "VFP_HARD=1"` (since we pass in our CFLAGS anyway - or does it try and use softfloats without).

gizmo98:
I tried to build without neon optimizations and used __platform=rpi1. Without -DARMv5_ONLY mupen64plus tries to build with armv7 asm. -DARMv5_ONLY  will be only set if HOST_CPU is armv6 or armv5. There is a build error because -march=armv6j does not support armv7 instructions.

This commit should have introduced the issue:
https://github.com/mupen64plus/mupen64plus-core/commit/1accde74ff65e747d15168bec954b5b6dbe136d8

Hardfloat CFLAGS do not replace VFP=1 because VFP=1 option adds VFP source code and enables VFP_OPT defines:
https://github.com/ricrpi/mupen64plus-video-gles2n64/blob/master/projects/unix/Makefile#L116
https://github.com/ricrpi/mupen64plus-video-gles2n64/blob/master/projects/unix/Makefile#L390

VFP_HARD=1 could be removed. I need to think about it.

joolswills:
That's ok - we can leave it. Are you happy for the rpi1 -> armv6 change for isPlatform ?

gizmo98:
The rpi1 -> armv6 change is good!

Issue #1947, opened 2017-04-10
[Request] cdogs-sdl


Omnija:
I've compiled and tested the game and has been working. I think it would a nice addition to the ports games and hopefully gets incorporated. Has controller and multiplayer support.

Source
https://github.com/cxong/cdogs-sdl

Prerequisites
https://github.com/cxong/cdogs-sdl/wiki/Developer-Getting-Started:-Linux

joolswills:
We prefer requests to go to the forum first as detailed when opening an issue

>For ideas and enhancements, please also post to the forum first, to see if it is something we would consider and to see if other users would find it useful.

>Please remove this text before posting.

zerojay:
With this merged, shouldn't this issue be closed?

joolswills:
yep. Thanks.

PR #1998, opened 2017-06-04
add C-Dogs SDL

HerbFargus:
addresses https://github.com/RetroPie/RetroPie-Setup/issues/1947

not too bad of a game.


joolswills:
Looks good. Can it go in experimental and I will test on Mali and build binaries then move to opt?

PR #2000, opened 2017-06-04
Update esthemes.sh

lilbud:
Add Switch-Light to setup script


joolswills:
Your branch may be out of date as this PR has conflicts. I'll do this manually with the other theme PRs. Thanks

joolswills:
Also for future reference the names should exclude the es-theme prefix as that is added automatically.

PR #2002, opened 2017-06-06
Mupen64plus improvements

gizmo98:
-fix bomberman 64: disable legacy blending.
-fix resident evil 2: use AUDIO-SDL. AUDIO-OMX crashs emulator at startup.
-fix pokemon stadium 2: use AUDIO-SDL. AUDIO-OMX crashs emulator at startup.


joolswills:
Thanks

PR #2004, opened 2017-06-10
mupen64plus: add default AUTO option

gizmo98:
New default option "mupen64plus-auto" sets video plugins and resolution
automatically to get best performance and visual quality.


joolswills:
thanks.

PR #2007, opened 2017-06-13
retroarch: Install minimal assets (6MB) needed for xmb menu driver

psyke83:
For your consideration, two changes to enable fetching of minimal assets required for the xmb menu driver to function. The download and total occupied space is a mere 6MB for assets and a few extra megabytes (2MB packed) due to the additional subversion dependency.

"svn export" allows fetching of specific data from a folder of a git repository without downloading or storing the entire .git repository. This seems ideal for our purposes.

Note: "git archive" could probably achieve a similar result, but unfortunately, the github servers don't support this functionality over the https protocol.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -15,7 +15,7 @@ rp_module_licence="GPL3 https://raw.githubusercontent.com/libretro/RetroArch/mas
 rp_module_section="core"
 
 function depends_retroarch() {
-    local depends=(libudev-dev libxkbcommon-dev libsdl2-dev libasound2-dev libusb-1.0-0-dev)
+    local depends=(libudev-dev libxkbcommon-dev libsdl2-dev libasound2-dev libusb-1.0-0-dev subversion)
     isPlatform "rpi" && depends+=(libraspberrypi-dev)
     isPlatform "mali" && depends+=(mali-fbdev)
     isPlatform "x11" && depends+=(libx11-xcb-dev libpulse-dev libavcodec-dev libavformat-dev libavdevice-dev)
@@ -86,6 +86,17 @@ function update_assets_retroarch() {
     chown -R $user:$user "$dir"
 }
 
+function update_xmbassets_retroarch() {
+    local dir="$configdir/all/retroarch/assets"
+    local path="xmb/monochrome"
+    # export files only if full assets are missing
+    if [ ! -d "$dir/.git" ]; then
+        rm -rf "$dir/$path"
+        svnExport "$dir/$path" https://github.com/libretro/retroarch-assets.git "$path"
+        chown -R $user:$user "$dir"
+    fi
+}
+
 function configure_retroarch() {
     [[ "$md_mode" == "remove" ]] && return
 
@@ -103,6 +114,9 @@ function configure_retroarch() {
 
     # install shaders by default
     update_shaders_retroarch
+
+    # install minimal xmb assets by default
+    update_xmbassets_retroarch
 
     local config="$(mktemp)"
 
@@ -287,6 +301,7 @@ function gui_retroarch() {
                         ;;
                     2)
                         rm -rf "$configdir/all/retroarch/$dir"
+                        [[ ! -d "$configdir/all/retroarch/assets" ]] && update_xmbassets_retroarch
                         ;;
                     *)
                         continue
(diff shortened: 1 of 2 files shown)


joolswills:
Not sure it's worth installing/using subversion just for this (a helper function isn't needed as we don't use it elsewhere yet). It is easier to just archive up the minimal assets ourselves. I would prefer to be involved in  discussions before submitting PRs with changes like this (You only mentioned it 12 hours ago on github and I haven't had a chance to comment.)

joolswills:
I'll have a think anyway.

psyke83:
I didn't think this was a major change due to the low footprint compared to our earlier expectations; subversion fetches just 2mb of debs. I added svnExport in case you could envision other uses to selectively fetch data for other modules.

Anyhow, apologies, I didn't think this part was a major change requiring forum discussion, as the patch doesn't actually set xmb to the default.

psyke83:
Also to note, whether you accept the PR or package assets manually, the icon assets seem duplicated in PNG and svg format. You might be able to trim down the size further by excluding one if you think 6mb is still too much.

joolswills:
NP -  I was mostly thinking it would be best to discuss first when you brought it up in the other PR as it could have saved some coding time etc.

scriptmodules/emulators/retroarch.sh
```diff
@@ -86,6 +86,17 @@ function update_assets_retroarch() {
     chown -R $user:$user "$dir"
 }
 
+function update_xmbassets_retroarch() {
+    local dir="$configdir/all/retroarch/assets"
+    local path="xmb/monochrome"
+    # export files only if full assets are missing
+    if [ ! -d "$dir/.git" ]; then
```
joolswills: Just FYI we use bash `[[` rather than `[`

scriptmodules/helpers.sh
```diff
@@ -347,6 +347,22 @@ function gitPullOrClone() {
...
+## @param branch branch to export from (optional)
+## @brief Export a working copy of files from svn or git repository
+function svnExport() {
+    local dir="$1"
+    local repo="$2"
+    local path="$3"
+    local branch="$4"
+    [[ -z "$branch" ]] && branch="trunk"
+
+    runCmd svn export $repo/$branch/$path $dir/
```
joolswills: these would need quoting.

joolswills:
I made a couple of code comments, but I'm going to close this and think about implementing it differently (if I decide to include some assets by default - those who want to run xmb probably want to install the full assets anyway so I'm not convinced there is a big advantage of installing a minimal amount). I may add a quick option to switch to xmb, which would install assets and switch the config.

psyke83:
No worries. In case you change your mind about implementing it yourself, I'll keep my branch alive for the time being so if you reopen, I can implement the changes you specified. 

The xmb menu driver seems to function perfectly well with just the icon theme assets; the problem with the full assets is that the .git storage is an additional 100MB (but of course, the advantage is that it can be fully updated with minimal bandwidth usage).

I didn't look too closely, but if you're already packaging other raw files from repositories on files.retropie.org.uk for other modules, svnExport might help you save on your bandwidth costs. ;)

joolswills:
Thanks.

We don't generally package stuff up from github repositories, but we do package some assets and sourcecode (non github stuff). We are not billed based on bandwidth usage currently so it's not a problem to include some stuff on our server. We also host all the RPI binaries, so it wouldn't make much of a difference anyway :)

joolswills:
BTW since SVN operates on URLs, there is no need to have in that function a split between $repo/$branch/$path etc. Just a url and a destination would suffice.

psyke83:
Yep, I just split that up so that it could support similar params to gitPullorClone to keep things consistent. "svn export" is a bit odd in that you can't specify the master branch (git master maps to "trunk"), and branches need to be specified via "branches/{branchname}".

joolswills:
SVN doesn't see a difference between branches / tags etc (tags and branches are the same - both are really just copies). github just choose to map it that way for their git repositories. We don't use svn enough in retropie to need it yet anyway - I prefer to keep the API as simple as posible and only add things that are used in multiple places etc.

joolswills:
Quite clever how github implement both git and svn support.

psyke83:
It is, but it would have been a lot easier if their https transport supported "git archive"; then we could do the same without relying on subversion mapping at all.

https://git-scm.com/docs/git-archive
https://github.com/isaacs/github/issues/554

PR #2008, opened 2017-06-14
retroarch - install minimal assets by default

joolswills:
@psyke83 please give this a go. It's much the same as your PR, but without using svn and just grabbing a tar.gz from our own server.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -86,6 +86,14 @@ function update_assets_retroarch() {
     chown -R $user:$user "$dir"
 }
 
+function install_xmb_monochrome_assets_retroarch() {
+    local dir="$configdir/all/retroarch/assets"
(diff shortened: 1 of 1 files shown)


psyke83:
Working perfectly - same behaviour as my patch. When installing full assets, update_assets_retroarch will delete the assets folder and avoid causing issues, and uninstalling full assets (whether installed or not) will trigger redownload of the minimal assets.

One extremely nitpicky observation (happened in my patch too) is that the filenames during untar will be visible after choosing the "uninstall assets" menu option. If that bothers you, perhaps a redirect to /dev/null might be in order for that specific function call.

psyke83:
No, sorry. A potential oversight in your patch is that you didn't chmod the assets. They've got root privileges now.

joolswills:
I don't mind the output - it gets logged etc which can be useful. I'll fix the permissions

joolswills:
Thanks.

PR #2009, opened 2017-06-15
added runcommand-onmenu feature

meleu:
`runcommand-onmenu.sh` is a script to be launched only if user invoke runcommand menu and choose the option to launch it.

I can see some use cases for this.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -462,6 +462,10 @@ function main_menu() {
             options+=(Z "Launch with netplay enabled")
         fi
 
+        if [[ -f "$CONFIGDIR/all/runcommand-onmenu.sh" ]]; then
+            options+=(M "Execute runcommand-onmenu.sh")
+        fi
+        
         options+=(Q "Exit (without launching)")
 
         local temp_mode
@@ -533,6 +537,9 @@ function main_menu() {
             L)
                 COMMAND+=" --verbose"
                 return 0
+                ;;
+            M)
+                user_script "runcommand-onmenu.sh"
                 ;;
             Q)
                 return 1


meleu:
@HerbFargus here is the PR I was talking about [in the forum](https://retropie.org.uk/forum/post/85841).

joolswills:
When opening a PR please summarise the use case as it saves me time from having to read though a forum thread etc.

I'm not keen on this sorry - at some point RPI retroarch will be built against ffmpeg (when we move to debian stretch). When that happens I may add a menu entry (one could be added now, but it would need to only show if video recording is enabled).

joolswills:
If there are more use cases for custom launch options from a menu, I might consider a user menu, but rather than a single entry it could be constructed from a folder with user scripts perhaps. However, it's not something I want to have implemented at this time.

meleu:
How I would like to use such feature at this moment (things to execute only if the user explicitly launch it via runcommand menu):

- launch my joystick-selection tool directly in a system dialog that is being launched (e.g. configure input for nes when launching Contra).
- add the current ROM to a custom ES system folder [like we are talking here](https://retropie.org.uk/forum/topic/10849/create-a-custom-es-system-able-to-launch-games-for-many-systems)

joolswills:
What's the purpose of the rom copying - to work as a favourites feature ? We will likely get some favourites support at some point in the near future in ES.

Certainly if there are more use cases, having a single user script that appears as a generic entry doesn't seem to re-usable so I would want to implement it differently. EG you will be limited to one functionality. If other users are going to create custom things to be launched from the menu you would want them to be easily installable and combined.

For now you will have to just use a custom runcommand for yourself though.

meleu:
> What's the purpose of the rom copying - to work as a favourites feature ?

No, it's not copying, but creating a symlink. It's not for favouries (only), but for any custom ES system (like the konami "system" I'm talking [in the forum](https://retropie.org.uk/forum/topic/10849/create-a-custom-es-system-able-to-launch-games-for-many-systems)).

> If other users are going to create custom things to be launched from the menu you would want them to be easily installable and combined.

Fully agree. And the way to make them combined would be make another menu in `runcommand-onmenu.sh`, but it would be the user responsibility.

backstander:
@meleu could this be used for Doom (and other source ports) for launching a custom map or mod?
[There has been some discussion here about it](https://retropie.org.uk/forum/topic/10602/doom-and-other-game-launcher-menu).  And example with Doom is:
* pick your engine (lr-prboom, ZDoom, etc.)
* pick your IWAD (doom.wad, doom2.wad, tnt, plutonia, freedoom, etc.)
* pick your PWAD (The Darkest Hour, Brutal Doom, etc.)

meleu:
I'm not sure what that is, but the feature I tried to implement here were implemenyed a way better in #2012

PR #2010, opened 2017-06-15
Add video and ArcadeItalia support for scraper

sselph:
There is a potential issue with this PR. Only ScreenScraper and ArcadeItalia support video so people may expect it to work and it won't. Should I see if I can alert if they request video on a non-video source?


joolswills:
I'll merge for now - it can be documented in the wiki perhaps, but if you want to add some additional warning to the module feel free.

joolswills:
thanks.

PR #2011, opened 2017-06-15
Mupen64plus more opts

gizmo98:
-GLideN64: make fps counter grey so it is less noticeable
-add defaut keyboard config if InputAutoCfg.ini is missing
-replace platform rpi1 with armv6


joolswills:
Thanks

PR #2012, opened 2017-06-15
runcommand - add user menu

joolswills:
* scripts must be called "name.sh" and live in /opt/retropie/configs/all/runcommand-menu
 * scripts are passed the same parameters as onstart / onend ("$SYSTEM" "$EMULATOR" "$ROM" "$COMMAND")
 * exit code of 0 returns to user menu, 1 exits without launching, and 2 exits and launches rom

@meleu despite saying I didn't want to implement the runcommand menu stuff yet, I decided to throw this together as it should provide the ability to do what you want but also allow scripts to be "dropped in" easily by users. It's had limited testing though and may have errors.

The other thing I thought was important was to handle what happens on exit - see the return codes above.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -412,6 +412,9 @@ function main_menu() {
     local cmd
     local choice
 
+    local user_menu=0
+    [[ -d "$CONFIGDIR/all/runcommand-menu" && -n "$(find "$CONFIGDIR/all/runcommand-menu" -maxdepth 1 -name "*.sh")" ]] && user_menu=1
+
     [[ -z "$ROM_BN" ]] && ROM_BN="game/rom"
     [[ -z "$SYSTEM" ]] && SYSTEM="emulator/port"
 
@@ -460,6 +463,10 @@ function main_menu() {
         if [[ "$COMMAND" =~ retroarch ]]; then
             options+=(L "Launch with verbose logging")
             options+=(Z "Launch with netplay enabled")
+        fi
+
+        if [[ "$user_menu" -eq 1 ]]; then
+            options+=(U "User Menu")
         fi
 
         options+=(Q "Exit (without launching)")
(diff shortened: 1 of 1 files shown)


meleu:
**OMFG!!** You're awesome!!! A way better than I tried.

Will test it as soon as I get home! Thanks!

scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -412,6 +412,9 @@ function main_menu() {
     local cmd
     local choice
 
+    local user_menu=0
+    [[ -d "$CONFIGDIR/all/runcommand-menu" && -n "$(ls -A "$CONFIGDIR/all/runcommand-menu")" ]] && user_menu=1
```
meleu: suggestion to avoid empty/non-sh files:
```
[[ -n "$(find "$CONFIGDIR/all/runcommand-menu" -type f -name "*.sh" ! -size 0 2>/dev/null)" ]] && user_menu=1
```

scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -677,6 +690,35 @@ function choose_fb_res() {
...
+    local scripts=()
+    local script
+    local i=1
+    while read -r script; do
+        script="${script##*/}"
+        script="${script%.*}"
+        scripts+=("$script")
+        options+=($i "${scripts[i-1]}")
+        ((i++))
+    done < <(find "$CONFIGDIR/all/runcommand-menu" -type f -name "*.sh" | sort)
```
meleu: suggestion, use the same find as I said in the comment above to avoid empty/non-sh files.

```
find "$CONFIGDIR/all/runcommand-menu" -type f -name "*.sh" ! -size 0 2>/dev/null
```

joolswills:
I'll tweak the first case. Don't think there is a need to exclude empty .sh files. If someone makes an unwanted empty .sh file somehow they can delete it.

joolswills:
PR is updated. I check for .sh files specifically now - but empty files will show.

meleu:
Testing here and trying to break it to catch bugs.

I confirm your conditions:

- exit status 0, returns to user menu (script list).
- exit status 1, exits runcommand without launching the game.
- exit status 2, launch the game

But there's another condition:

- exit status different than 0, 1, and 2, goes back to the runcommand's main_menu.

It's not a big worry but I feel like the expected behavior would be to return to the user menu (same as exit status 0).

Apart from that everything runs fine.

meleu:
awesome work! Thanks!

PR #2013, opened 2017-06-19
adds an experimental package for supacec

superterran:
I'm very new to this community but so far I'm loving RetroPie and was searching for the ability to switch inputs using HDMI-CEC. After some googling, I didn't really find anything that fit the bill and ended up writing something for myself. I'm not sure how to distribute this (even to my other RetroPie), but thought it would be useful as an experimental package. Please let me know your thoughts and accept my PR if you agree this is a valid way to proceed. Again, I'm very new and welcome feedback. Thanks!


joolswills:
It's not sorry. You can share tips and scripts via the forum. Thanks.

PR #2014, opened 2017-06-19
adds experimental package supacec

superterran:
Sorry for spamming, realized I committed a symlink before. This is what I meant to submit...

This adds an experimental package that will trigger a CEC command to switch the TV source when a new input device is detected. This is fairly rudimentary, but it works, has a low footprint, and enables  me to not fumble with a remote.


joolswills:
Sorry but this isn't suitable for a module at this time.

joolswills:
Would recommend sharing via the forum etc. Thanks.

PR #2025, opened 2017-07-01
added option to update all installed ES themes

meleu:
The first commit is just removing rendundant tests.
The second one adds the option to update all installed ES themes, as [requested in the forums](https://retropie.org.uk/forum/topic/11319/updating-all-installed-themes).


joolswills:
thanks.

PR #2034, opened 2017-07-07
lr-bluemsx - added colecovision support - fixes #2032

### LINKED ISSUE #2032: added support for colecovision in libretro blue-msx
Hi
FIY colecovision support have been added in libretro:
https://github.com/libretro/blueMSX-libretro/pull/41#event-1148586688

joolswills:
* document msx/colecovision bios / rom locations
 * added to coleco system
 * use a separate retroarch-core-options.cfg so as to force the colecovision system

This needs some testing. I think the retroarch input configuration might need switching/changing to work also with this.

### DIFF
--- a/scriptmodules/libretrocores/lr-bluemsx.sh
+++ b/scriptmodules/libretrocores/lr-bluemsx.sh
@@ -10,8 +10,8 @@
 #
 
 rp_module_id="lr-bluemsx"
-rp_module_desc="MSX/MSX2 emu - blueMSX port for libretro"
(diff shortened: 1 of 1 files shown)


joolswills:
controls seem to be ok.

HerbFargus:
I can confirm this works (after updating bluemsx from source).

I've also updated the docs on the msx and colecovision wiki pages. I still need to rebuild the pretty docs but I have a few things I need to sort out with some other pages first.

joolswills:
Thanks for testing :-)

PR #2037, opened 2017-07-10
Add Dmmarti's "Maximus Pie" Theme

robertybob:
Hi @dmmarti , as promised! Have you posted this in the RP forums?


dmmarti:
Here's the post in the forum for MaximusPie theme.

https://retropie.org.uk/forum/topic/11490/new-theme-maximus-pie-retropie-version-of-maximus-arcade

PR #2038, opened 2017-07-10
Added common screenshot directory (~/RetroPie/screenshots/) for mupen…

dankcushions:
…64plus and libretro (lr-) emulators.

* This solves the annoyance of having screenshots saved all over the place, and libretro ones being stored in the hidden `/home/pi/.config` path.
* I've only done this for cores I use/can test. Are there others that store screenshots in customizable locations?
* It might also make sense to save them in the `/rom/<system>/` directory, but for this isn't possible via changing the retroarch.cfg file (I think you'd have to add `screenshot_directory = "/home/pi/RetroPie/roms/<system>` to every single `/<system>/retroarch.cfg`). I'm open to that, though, if that is the preference?
* unrelated, but mupen64plus.sh defines the directory paths internally. I wonder if there's a way of getting it to use those in retropie_packages.sh?

### DIFF
--- a/retropie_packages.sh
+++ b/retropie_packages.sh
@@ -25,6 +25,7 @@ biosdir="$datadir/BIOS"
 romdir="$datadir/roms"
 emudir="$rootdir/emulators"
 configdir="$rootdir/configs"
+screenshotsdir="$datadir/screenshots"
 
 scriptdir="$(dirname "$0")"
 scriptdir="$(cd "$scriptdir" && pwd)"
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -23,6 +23,7 @@ config="$configdir/n64/mupen64plus.cfg"
 inputconfig="$configdir/n64/InputAutoCfg.ini"
 datadir="$HOME/RetroPie"
 romdir="$datadir/roms"
+screenshotsdir="$datadir/screenshots"
 
 source "$rootdir/lib/inifuncs.sh"
 
@@ -361,7 +362,7 @@ if ! grep -q "\[Core\]" "$config"; then
     echo "Version = 1.010000" >> "$config"
 fi
 iniConfig " = " "\"" "$config"
-iniSet "ScreenshotPath" "$romdir/n64"
+iniSet "ScreenshotPath" "$screenshotsdir"
 iniSet "SaveStatePath" "$romdir/n64"
 iniSet "SaveSRAMPath" "$romdir/n64"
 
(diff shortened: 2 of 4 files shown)


joolswills:
>stored in the hidden /home/pi/.config path.

It's not hidden - we use this folder now and it's symlinked to `/opt/retropie/configs/all/retroarch` so screenshots should correctly go do `/opt/retropie/configs/all/retroarch/screenshots`. So I'm happy with the current behaviour (I have actually been migrated stuff back to the default retroarch folders - for example the joypad configs now live there again).

Regarding mupen64plus script - I did plan at some point to make the paths so they are generated on install as we do with the splashscreen etc. However, it's not a big deal for now.

joolswills:
I realise this means screenshots don't all go to the same place for things, but I'm not sure I want another folder in ~/RetroPie/. Will think about it - perhaps we should just use the retroarch folder for all screenshots for example.

dankcushions:
> It's not hidden

doesn't a .prefix mean hidden in linux? i may be wrong in my understanding, but I thought the folder didn't show up for me in FTP etc, unless I show hidden folders.

but yeah, maybe the root directory isn't the place to have it, if anywhere. I think the /roms/<system>/ folder makes sense (same place as save files), but i wonder if that's possible to configure in an elegant way?

joolswills:
Yes, but we symlink `$HOME/.config/retroarch` to `/opt/retropie/configs/all/retroarch` which is the actual location.

Also:

* most emulators people use are retroarch - and not all standalone emulators even allow screenshots.
* the changes above wouldn't change old configs, so I'm reluctant to do a change like this. You end up with some people getting screenshots in one place and some another.

joolswills:
(updated previous post).

HerbFargus:
I have to agree with Jools on this one, I'm hesitant to modify anything from the default configs with retroarch as it can confuse things, and for those that want to modify it on their own they can but I think some of the standalone emulators may also not have options to modify where screenshots end up which is something that may need to be sorted upstream if that functionality is desired

dankcushions:
For me it doesn't make sense for screenshots to go in /opt/ as they are user-generated, but that might semantics :) If we can default them to a more obvious location, I think we should. Even if this doesn't cover all standalone emulators it would make the situation no 'worse' than it is now.

> the changes above wouldn't affect existing configs, so I'm reluctant to do a change like this. You end up with some people getting screenshots in one place and some another.

yes this is true. however i don't think we have documented where screenshots are kept anywhere. there's [this](https://retropie.org.uk/docs/Take-and-Scrape-Your-Own-Screenshots/) but that has instructions to override the settings anyway.

this is definitely just a quality of life thing for me, and I guess the others doing things like the MAME ROW. If there's any resistance I don't feel strongly enough about it so we can close :)

joolswills:
Sorry, but I don't want to move the retroarch screenshots from their current retroarch folder for the reasons I gave above.

There is lots of user generated and user content in `/opt/retropie/configs` btw, that area is designed to be user changeable. We also know it's on a native FS so can control permissions, and use symlinks etc whereas `~/RetroPie` can be on fat32, ntfs etc

PR #2039, opened 2017-07-11
fix for PlayStation DS4 controller lagging

slajerek:
Fix for PlayStation DS4 controller lagging in runcommand due to a badly placed time.sleep in joy2key and thus long time of processing large number of accelerometer events.

Tested and confirmed working/fixed on my Raspberry Pi3 setup.

Forum post:
https://retropie.org.uk/forum/topic/10315/ds4-in-textmode-menus

### DIFF
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -220,6 +220,7 @@ except:
 js_fds = []
 rescan_time = time.time()
 while True:
+    do_sleep = True
     if not js_fds:
         js_devs, js_fds = open_devices()
         if js_fds:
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Makes sense - but will test on a few of my controllers before merging (feedback from others is welcome too)

I had a feeling I forced a delay due to over sensitive analogue sticks or something. We should probably try to ignore accelerometers - I think we have a workaround for ignoring more than the first two axis.

Cheers

slajerek:
Yes, the workaround for ignoring axis is there, but in processing_event. Sleep is performed always, regardless if the event has been processed or not. In overall it is not good practice to always do sleep after reading each event, and only sleep when queue is empty to not pile up events.

I tested this only in my DS4 setup, so of course please confirm how this behaves with other controllers.

joolswills:
Tested and working well. Thanks.

PR #2040, opened 2017-07-12
Add initial support for Pegasus frontend

mmatyas:
This patch adds initial/experimental support for the [pegasus-frontend](http://github.com/mmatyas/pegasus-frontend). The script downloads and extracts the [latest automatic build](https://github.com/mmatyas/pegasus-frontend/releases/tag/continuous) for the current platform (there are RPi and X11 builds). At the moment it uses the ES2 config files, so probably no additional configuration steps are necessary.

### DIFF
--- a/scriptmodules/supplementary/pegasus-fe.sh
+++ b/scriptmodules/supplementary/pegasus-fe.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="pegasus-fe"
+rp_module_desc="Pegasus: A cross platform, customizable graphical frontend (latest alpha release)"
+rp_module_licence="GPL3+ https://raw.githubusercontent.com/mmatyas/pegasus-frontend/master/LICENSE.md"
+rp_module_section="exp"
+rp_module_flags="!mali frontend"
+
+function depends_pegasus-fe() {
+    local depends=(
+        fontconfig
+        gstreamer1.0-alsa
+        gstreamer1.0-libav
+        gstreamer1.0-plugins-good
+        jq
+        policykit-1
+    )
+
+    getDepends "${depends[@]}"
+}
+
+function install_bin_pegasus-fe() {
+    # get all asset urls for the latest continuous release
+    local all_assets
+    all_assets="$(wget -q -O - https://api.github.com/repos/mmatyas/pegasus-frontend/releases/tags/continuous)" || return
+    all_assets="$(echo "${all_assets}" | jq -r '.assets[] | .browser_download_url')"
+
+    printMsgs "console" "Available releases:"
+    printMsgs "console" "${all_assets}"
+
+    # find out which platform's package we'll need
+    local platform
+    isPlatform "x11" && platform="x11"
+    if isPlatform "rpi"; then
+        if isPlatform "armv6"; then
+            platform="rpi1"
+        else
+            platform="rpi2"
+        fi
+    fi
+    if [[ -z "${platform}" ]]; then
+        md_ret_errors+=("Sorry, Pegasus is not yet available for this platform. Consider reporting this on the forum!")
+        return
+    fi
+
+    printMsgs "console" "Package platform: ${platform}"
+
+    # select the url for the platform
+    local asset_url
+    asset_url="$(echo "${all_assets}" | grep ${platform})"
+
+    if [[ -z "${asset_url}" ]]; then
+        md_ret_errors+=("Looks like the latest Pegasus release is not yet available for this platform. This happens when the build is so fresh it's being uploaded right now, or when there's a technical problem on the download server. Either way, this is a temporary problem, so please try again in 1-2 minutes. If the problem persists, consider reporting it on the forum!")
+        return
+    fi
+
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Sorry for the delay but I have no had a chance to test yet - any testing from others is also welcome.

HerbFargus:
I can confirm it installs on my pi 3 and I am very impressed with the speed. It is a bit crashy but I've been doing a bunch of stuff on my pi and I know this is an alpha build so I won't fault it too much. I'll defer to Jools on the coding style of the module and whatnot but for all intents and purposes it installs and functions

scriptmodules/supplementary/pegasus-fe.sh
```diff
@@ -0,0 +1,83 @@
...
+rp_module_section="exp"
+rp_module_flags="!mali frontend"
+
+function depends_pegasus-fe() {
+    local depends=(
+        gstreamer1.0-alsa
+        gstreamer1.0-plugins-good
+        jq
+        libinput10
+        libsdl2-2.0-0
```
joolswills: Although this is a binary only install we use the dev packages for dependencies - this is due to the names of binary packages changing - so please just include libsdl2-dev etc.

This doesn't currently work on my x86_64 box as the binary it installs is linked against libpng12 which isn't available on Ubuntu 17.04. Perhaps if distributing a binary only, you can statically link the png library ?

joolswills:
Ideally on x86 this would be a source install but I realise the requisites are quite big etc. Are the required libraries available on Ubuntu 16.04 + ?

scriptmodules/supplementary/pegasus-fe.sh
mmatyas: I see, it seems libpng 1.2 was fully replaced with 1.6 in 17.04. I'll bundle it with the binary then.

mmatyas:
Unfortunately no; the minimal required Qt version is 5.7, which is partially available first in 16.10 (and on Debian stretch), but still has some packages missing, eg. gamepad support. Also on ARM the official releases seem to depend on X11.

There is a [third party PPA](https://launchpad.net/~beineri/+archive/ubuntu/opt-qt591-xenial) with all the packages, but that requires running a shell script on every new session to make the Qt libs available for the programs. It's intended for testing, and only provides builds for desktop.

joolswills:
thanks for the info.

mmatyas:
Ok, I've fixed the dependency names.

Also I plan to add some new files to the package in the following days, I'll update the script again when it's done.

joolswills:
Thanks!

scriptmodules/supplementary/pegasus-fe.sh
```diff
@@ -0,0 +1,85 @@
...
+            platform="rpi1"
+        else
+            platform="rpi2"
+        fi
+    fi
+    if [[ -z "${platform}" ]]; then
+        printMsgs "dialog" "Sorry, Pegasus is not yet available for this platform. Consider reporting this on the forum!"
+        return
+    fi
+
```
joolswills: There should be no dialogs in the code in this function ideally - please use `md_ret_errors+=("whatever")` then return

scriptmodules/supplementary/pegasus-fe.sh
```diff
@@ -0,0 +1,85 @@
...
+    getDepends "${depends[@]}"
+}
+
+function install_bin_pegasus-fe() {
+    # get all asset urls for the latest continuous release
+    local all_assets
+    all_assets="$(curl -s -XGET https://api.github.com/repos/mmatyas/pegasus-frontend/releases/tags/continuous)" || return
+    all_assets="$(echo "${all_assets}" | jq -r '.assets[] | .browser_download_url')"
+
```
joolswills: is it possible to use wget as we use it elsewhere for our downloading (saves a dependency) ?

scriptmodules/supplementary/pegasus-fe.sh
```diff
@@ -0,0 +1,85 @@
...
+        printMsgs "dialog" "Looks like the latest Pegasus release is not yet available for this platform. This happens when the build is so fresh it's being uploaded right now, or when there's a technical problem on the download server. Either way, this is a temporary problem, so please try again in 1-2 minutes. If the problem persists, consider reporting it on the forum!"
+        return
+    fi
+
+    printMsgs "console" "Download URL: ${asset_url}"
+
+    # download and extract the package
+    wget "${asset_url}" -O "$md_inst/pegasus-fe.zip" || return
+    unzip -n "$md_inst/pegasus-fe.zip" -d "$md_inst" || return
+    rm "$md_inst/pegasus-fe.zip"
```
joolswills: I have added a new function - you can now use `downloadAndExtract "${asset_url}" "$md_inst"`

scriptmodules/supplementary/pegasus-fe.sh
```diff
@@ -0,0 +1,85 @@
...
+}
+
+function remove_pegasus-fe() {
+    rm -f "$md_inst/pegasus-fe"
+}
```
joolswills: This function should not be needed.

mmatyas:
Thanks, and sorry for the lack of updates; I was experimenting with the dependencies and managed to remove some of them. I've also tested the gstreamer packages: `gstreamer-omx` (that is, hardware accelerated video rendering) causes problems at the moment, but a combination of `gstreamer-libav` and `gstreamer-plugins-good` seems to play most formats. Also I've listed `fontconfig` as a dependency, in case it's not available by default.
As for the additional new files I've mentioned last time, I think I can keep them all in the same directory, so the script will still work fine.

mmatyas:
Merged and rebased to master, with new changes -- I'll do a test on a fresh image just in case.

joolswills:
Cheers.

mmatyas:
Fixed a typo, works fine now.

joolswills:
Thanks :)

PR #2041, opened 2017-07-15
Amiga: remove emulator ROM scripts correctly on uninstall

psyke83:
Applies to uae4all, uae4arm & amiberry scripts.


joolswills:
Will merge, but I think I will make a function for launch script generation, and then automatically remove the files on uninstall in that function (in the same way we handle it for addEmulator etc)

PR #2043, opened 2017-07-16
Bluetooth: enhance sixad compatibility

psyke83:
Re-enable bluetoooth stack in Bluetooth dialog menu, and restart
sixad daemon upon exit.

This allows standard BT device pair/unpair management with DS3
controller fully working in-menu. The only drawback is that
your PS3 controller will turn off upon exiting the menu due to the
daemon restart.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -377,6 +377,8 @@ function gui_bluetooth() {
 
         local choice=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
         if [[ -n "$choice" ]]; then
+            # temporarily restore Bluetooth stack (if needed)
+            service sixad status >/dev/null && sixad -r
             case $choice in
                 R)
                     register_bluetooth
@@ -402,6 +404,8 @@ function gui_bluetooth() {
                     ;;
             esac
         else
+            # restart sixad (if running)
+            service sixad status >/dev/null && service sixad restart
             break
         fi
     done


scriptmodules/supplementary/bluetooth.sh
```diff
@@ -402,6 +404,8 @@ function gui_bluetooth() {
                     ;;
             esac
         else
+            # restart sixad (if running)
+            service sixad status >/dev/null && service sixad restart
```
psyke83: I'm not exactly sure if restarting the daemon is even necessary. I've tried omitting this line and I can successfully turn off/on & repair my DS3 controller after using the Bluetooth menu.

psyke83: Yes, maybe it is. If I reboot the system and invoke "sudo sixaxis -r" before pairing, the device will not pair correctly. It's probably better to restart the daemon so the stack doesn't interfere.

psyke83:
Closing this as I found a way to pair DS3/4 with the native bluez stack. Will send a new PR later.

psyke83:
Please see this topic: https://retropie.org.uk/forum/topic/11638/dualshock-3-4-testers-needed

Recent BlueZ versions has native sixaxis support, but it has some big disadvantages as far as I can see. For one thing, the driver doesn't seem to support rumble, and the controller will no longer turn off on idle (with the wrapper, it turns off after 10 mins of inactivity). I leave the Pi powered 24/7, so this would drain my controller if I forgot to manually turn it off.

Considering the disadvantages of the native driver, I'm reopening this PR in order to make the ps3controller wrapper more useful. If there's a simple way to get rumble and inactivity timeout working, the native driver could be a suitable replacement.

joolswills:
thanks.

PR #2044, opened 2017-07-16
Removing .git directories from ES themes

meleu:
The files in `/etc/emulationstation/themes/$theme/.git` are unnecessary and can occupy dozens of MB.

Example:
```
[prompt] $ ls /etc/emulationstation/themes/
carbon/  ComicBook/  nes-mini/  pixel/  Retrorama/
[prompt] $ du -hs /etc/emulationstation/themes/
257M    /etc/emulationstation/themes/

# went to RetroPie-Setup -> esthemes and updated all installed themes:

[prompt] $ du -hs /etc/emulationstation/themes/
138M    /etc/emulationstation/themes/
```

### DIFF
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -33,6 +33,7 @@ function install_theme_esthemes() {
     fi
     mkdir -p "/etc/emulationstation/themes"
     gitPullOrClone "/etc/emulationstation/themes/$theme" "https://github.com/$repo/es-theme-$theme.git"
+    rm -rf "/etc/emulationstation/themes/$theme/.git"
 }
 
 function uninstall_theme_esthemes() {


meleu:
You're right. I updated only once and thought everything was fine. :(

joolswills:
Also if we do it so we don't use git, the larger themes would require a large download each time to update. Since you can only use a single theme at a time, I don't see this as an issue. It's only the graphically heavy themes which are large. Carbon is under 10MB for example.

dankcushions:
you could do a shallow clone with `git --depth 1`. that only gets the latest commit, if you don't care about the history. see https://www.atlassian.com/blog/git/handle-big-repositories-git

joolswills:
We already do. However you still get everything then in a single commit worth of changes.

PR #2045, opened 2017-07-17
Launch Art Patch

Nam-Ereh-Won:
Adds code for a 5th arg to specify launch art. Due to how the script verifies the files are there, I have it trim any extension from the image file path. May act up in the event that you have a path with periods in other places. Originally created this patch as a way to get custom launch art when triggering from, and specified by, an external script.


joolswills:
Sorry but I don't want to add more parameters to runcommand.

joolswills:
You may be able to do what you want via the onstart script. If calling runcommand from another script you could copy files in place as needed. Thanks.

PR #2050, opened 2017-07-20
joy2key: implement key verification & dpad support

psyke83:
* Allow dpad control in addition to axis for cursor keys
* Verify existence of keys before mapping

The cursor keys will now be mapped to both the axis and dpad, but if
the dpad is not present, only the axis mappings will be used.

This should also make it possible to add default mappings for X/Y
buttons without having to worry about pad compatibility.


joolswills:
thanks.

PR #2058, opened 2017-07-28
Fix Black Screen when using fbi

Nam-Ereh-Won:
Fixes fbi only showing Launch Art for $IMAGE_DELAY seconds and then a black screen until emulator launch.


joolswills:
We can't do this sorry. Breaks some emulators (FBI closing after). Info in commit history and probably a ticket or two.

PR #2059, opened 2017-07-28
joy2key: fix key verification for multiple devices

psyke83:
* Do verification on copy of default_button_codes for each device
  (so missing buttons on one device are not pruned from all others)
* Add exception handling for a/b swap to avoid missing buttons blocking
  script execution


joolswills:
Thanks

PR #2073, opened 2017-08-06
My ES Themes

Omnija:


robertybob:
Nice you see you back and making themes again :)

joolswills:
Just to note - you have a merge commit - you probably should reset your master branch, and use a new branch for changes to avoid this instead of using master. I will squash and merge this time.

Please also use a descriptive commit message rather than "ES Themes" - see commit history for examples.

PR #2074, opened 2017-08-07
Updated Commander Genius

gerstrong:
Version 1.9.9.2 might be beta but it is usable enough and has great features and bug fixes.

### DIFF
--- a/scriptmodules/ports/cgenius.sh
+++ b/scriptmodules/ports/cgenius.sh
@@ -19,7 +19,7 @@ function depends_cgenius() {
 }
 
 function sources_cgenius() {
-    wget -O- -q "https://github.com/gerstrong/Commander-Genius/archive/v180release.tar.gz" | tar -xvz --strip-components=1 -C "$md_build"
+    wget -O- -q "https://github.com/gerstrong/Commander-Genius/archive/v1992beta.tar.gz" | tar -xvz --strip-components=1 -C "$md_build"
 }
 
 function build_cgenius() {


joolswills:
thanks.

HerbFargus:
I'll have to try this out. The reason I had used an older version in the first place was the newer versions had issues building so hopefully those have been resolved

gerstrong:
Oh okay. I remember someone reported that. I tried it recently and it worked. Let me know if there are problems...Am 07.08.2017 9:21 nachm. schrieb HerbFargus <notifications@github.com>:I'll have to try this out. The reason I had used an older version in the first place was the newer versions had issues building so hopefully those have been resolved

joolswills:
I didn't check sorry. Can roll back if needed

HerbFargus:
@gerstrong hey sorry, I just got to testing this on my pi3. It failed for two reasons, one was some folder changes which are easily fixed through the module (due to the vfsroot changes with the downloader) 

the other issue is a compilation issue which I don't currently know the cause. Here is the build log:

https://pastebin.com/bhYcYfAe

If you prefer I can add this to the tracker on the cgenius repo if you feel thats a better place for compilation discussion

gerstrong:
Hi HerbFargus, thanks for compiling the CG sources. Not sure why the compiler likes to segfault at this part, but I saw it before somewhere else and same method it was. Did you observe how much memory the build process consumes?

Anyways, I did some modifications on that part, cleaning up a bit. There is new Tag called "v1995beta". Can you try it with that one. I also threw out a lot of files that weren't used anymore.

I can't say for sure, but you might have different results now. Let me know if it works better now...

Because I want to bring up a stable nice version 2.0, I'm in the process of cleaning up the code a lot lately and fixing bugs. 

P.S.: Compilation log can be kept at pastebin :-)

HerbFargus:
I'm not sure of the memory but I've increased the swap and it still crashes at the same point regardless.

I've tested v1995beta and unfortunately got the same results.

log for reference:

https://pastebin.com/LxtqZMM6

I suppose I could test forcing 02 optimisations in the script module as a workaround til we update to stretch.

mmatyas:
You can avoid the segfault by renaming the two `serialize` functions to something else. Looks like it's a bug in GCC 4.9.2; updating to 4.9.3 should fix it too.

gerstrong:
I don't understand how renaming would change the Situation. Let me try to simplify the save system by also changing these methods around a bit...

mmatyas:
This is not your fault @gerstrong. I assume the word `serialize` triggered some incorrect behavior in this particular GCC version. Unfortunately it'd take about a day to finish compiling the whole game on my Pi 1, but with renaming the functions I did made it past the segfault. Someone with a faster Pi could check if there are any other problems further along the way.

HerbFargus:
@gerstrong  I wouldn't worry too much about patching for an old version of GCC. We can just do a local patch in the module for older versions as we will be moving toward stretch which I just tested and it compiles fine on. We can consider this resolved. I'll submit a PR an update for the folder changes though and will tweak it when get further along with stretch.

gerstrong:
Well thanks, I agree completely. Still I might change these methods because I want so as being for the topic of general cleanups. The code design requires a lot of improvements also at that part. So, sooner or later this will be fixed the one or the other way...

Thanks for contributing and best wishes for a nice RPI Application

mmatyas:
Using Clang (eg. 3.9) instead of GCC also fixes the problem, without modifying the source code.

Jondark:
Hi;

I'm brand new to Retropie--and am a complete newbie and not a computer programmer--

Is there an "easy" way to get the Commander Genius working?  I'm a big CKeen fan--as are my grown kids, and would love to get this up & running...

If anyone can give "Simplified instructions for installing Commander Genious on Retropie v4.2 for Idiots", that would be great...

:-)

mmatyas:
I've just sent a patch which should temporarily fix the problem. If it gets accepted, you should be able to install the game as usual after updating the RetroPie script.

Jondark:
That would be awesome... I'll anxiously await update notification :-) 

Would this port also be able to play other similar old Apogee games--like Cosmo, Secret Agent, etc?

mmatyas:
My patch has landed, and now the game builds fine. There's still an error during the installation step, but #2099 fixes that.

PR #2081, opened 2017-08-11
Quick fix for non ascii bluetooth device name

lukeart:
Bluetooth devices that have names with non-ascii characters won't connect. I experienced this with the Xiaomi Bluetooth game controller.

[The fix](https://retropie.org.uk/forum/topic/4804/xiaomi-bluetooth-gamepad-after-4-0-6-update/4) in the forum solves it, so I though it might as well become part of the code base.


joolswills:
There was a previous workaround for this from someone else - so that didn't work ? https://github.com/lukeart/RetroPie-Setup/commit/d09a77f4134de98b5d1fd7922540946d13550ac1

lukeart:
You're right; I thought it was the latest version, but it clearly wasn't. It's working as expected now :+1:

PR #2084, opened 2017-08-12
retroarch: prune unnecessary menu options

psyke83:
common: hide online/core updater menu options
xmb: prune history, music, image & import content menu tabs

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -181,6 +181,16 @@ function configure_retroarch() {
     # rgui by default
     iniSet "menu_driver" "rgui"
 
+    # hide online updater menu options
+    iniSet "menu_show_core_updater" "false"
+    iniSet "menu_show_online_updater" "false"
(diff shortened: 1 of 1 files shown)


psyke83:
Not sure if this is something you'd consider appropriate, but I think the core/online update menu items are out of place for RetroPie considering that the cores are supposed to be managed via your script. Also, xmb presents a lot of unnecessary menus for those that change from the rgui default.

dankcushions:
I 👍  this. I have seen users getting into trouble or not having the desired effect when attempting to update within RGUI. I didn't realise you could disable them.

psyke83:
What about the xmb menus? Would people miss the image/music tabs? Maybe I'm overlooking an obvious use case (e.g. for playing custom music during games)...

Darknior:
Good idea 👍 
I'm sure some other menu can be removed like in the SAVESTATE menu, you can activate an option to take screenshot of the save, but it doesn't work on PI.

PR #2086, opened 2017-08-13
Replaced old version of Space with 2.0 (video support)

robertybob:
I've done some testing on my spare RPi2 with GB,GBC and SNES. Seems to work ok!

### DIFF
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -73,7 +73,7 @@ function gui_esthemes() {
         'mattrixk io'
         'mattrixk metapixel'
         'mattrixk spare'
-        'robertybob space'
+        'robertybob space-2.0'
         'robertybob simplebigart'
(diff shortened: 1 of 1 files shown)


joolswills:
This won't allow people to update or remove the old space - it will be a new install - if it's just the addition of video support it might be worth using the same repo so people can update ?

robertybob:
Fair comment, I'll just rename the new branch the same as this

joolswills:
If you mean deleting the old repo and renaming the new one that will break updating also. The only way would be to update the old repo with the new code.

robertybob:
Hm, so giving it the same name won't work? Is that due to the way the theme installer/updater works?

joolswills:
No - you will need to update the old repository - which would also seem the logical way if adding video support. It's because it does a `git pull` to update.

robertybob:
Ok I think I've done it correctly, should be sorted out now :)

PR #2089, opened 2017-08-16
added - lr-dosbox to experimental

joolswills:


psyke83:
lr-dosbox automatically mounts c:\ to the current same directoroy as the executables/batch files launched. Maybe it would be a good idea to set up a launch script that starts the core without any content, so that there's a way to mount c:\ to roms/pc for the purpose of installing new programs?

joolswills:
Possible there was still a key repeating issue as I just saw this - https://github.com/libretro/dosbox-libretro/commit/06ca21e82aeca71403f6d59ee6d2852c75cf760a

Will look into sorting the C:\ folder location.

joolswills:
I will merge as is for now - but later it could be improved as you mentioned. Perhaps via a default config - https://wiki.libretro.com/index.php?title=DOSBox

psyke83:
It looks like you didn't merge before deleting/closing?

Re: the key repeating issue, I saw no issue during testing, but it could be a transient problem that only affects certain keyboards or layouts... who knows. I was testing on an old DELL UK keyboard without changing the default keymap and never saw the issue.

Maybe I'll give the core a try on my laptop either via Windows or Ubuntu via the official RetroArch install to see if the bug triggers there.

joolswills:
Looks like I may have pressed the wrong button :-) good thing I didn't remove the branch on my local machine and github. Except I did ;-)

joolswills:
Luckily github/git keeps the refs. Will check and commit

psyke83:
Thanks. Github will also allow you to restore a deleted branch (it appears in the place of the the "delete branch" button after deleting).

joolswills:
Thanks. Will do when on pc next.

joolswills:
(I also have daily system backups going back some time - thanks to borgbackup - the best deduplicating backup system ever). / Spam

psyke83:
Best not outsource to me for additional capacity, as my current storage solution involves a salvaged laptop hard drive with many bad sectors - a range of several megabytes manually blacklisted by injecting a custom badblocks file to the ext4's metadata - permanently attached to my Pi 3. A good way to recycle an otherwise useless drive for a cheap media centre, but I wouldn't trust anything important on it, LOL.

PR #2090, opened 2017-08-16
Could not change save path for save game to own directory

Ley21:
Could not chnage directory for savegames, screenshots,... in cfg because on each startup the script replace them again.
Add a fix for this with the auto configuration.


joolswills:
Thanks.

PR #2099, opened 2017-08-27
update cgenius

HerbFargus:
This is an update to keep up with the latest changes. the high quality packs aren't needed as they are maintained as separate packs through the built in downloader now.

this still won't build on the stock raspbian Jessie due to the compiler bugs per discussion here:

https://github.com/RetroPie/RetroPie-Setup/pull/2074

So don't know if it's worth patching in the module til we update to stretch (compiles fine on stretch from my testing) or if we just let it sit in its current state til then. I'm fine either way.

### DIFF
--- a/scriptmodules/ports/cgenius.sh
+++ b/scriptmodules/ports/cgenius.sh
@@ -15,11 +15,11 @@ rp_module_licence="GPL2 https://raw.githubusercontent.com/gerstrong/Commander-Ge
 rp_module_section="exp"
 
 function depends_cgenius() {
-    getDepends build-essential libvorbis-dev libogg-dev libsdl2-dev libsdl2-image-dev libboost-dev
+    getDepends build-essential cmake libcurl4-openssl-dev libvorbis-dev libogg-dev libsdl2-dev libsdl2-image-dev libboost-dev
 }
 
 function sources_cgenius() {
-    downloadAndExtract "https://github.com/gerstrong/Commander-Genius/archive/v1992beta.tar.gz" "$md_build" 1
+    downloadAndExtract "https://github.com/gerstrong/Commander-Genius/archive/v1995beta.tar.gz" "$md_build" 1
 }
 
 function build_cgenius() {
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. I will take a look at this when I have a chance - taking into account the comments on the issue

mmatyas:
This will require a little rebasing now, but together with #2107 the whole cgenius installation process seems to work fine (tested on Pi 3, jessie; haven't tried the gameplay).

Jondark:
Would there be a set of simplified instructions for how to install from command prompt?  The patch looks like a file to edit, but I am not sure that the CG stuff successfully downloaded to my RP3... and I can't find it using file manger... This is my fault, and not the fault of directions given--I am unfortunately totally unfamiliar with Ubuntu.

Any way to type up the command steps specifically? Please assume I can read and type in a command prompt, but that's about it on this platform...

Then, if the patch is the modification of a particular file, is this something I can edit simply on the "desktop" (graphical user interface on RP3)?

Then, if file is modified, what to type to get it completed?

Thank you!
--a Commander Keen fan--

scriptmodules/ports/cgenius.sh
```diff
@@ -31,23 +31,16 @@ function build_cgenius() {
...
+    mkUserDir "$home/.CommanderGenius"
+    
     moveConfigDir "$home/.CommanderGenius"  "$md_conf_root/$md_id"
 
-    mv "$md_inst/games" "$romdir/ports/$md_id/"
-    mv "$md_inst/hqp" "$romdir/ports/$md_id/"
-
-    ln -snf "$romdir/ports/$md_id/games" "$md_inst"
-
-    chown -R $user:$user "$romdir/ports/$md_id"
```
joolswills: @HerbFargus I will manually include your changes - but just to check - before we copied the games to the roms/ports dir and symlinked it - this is no longer required or ?

HerbFargus: The latest update includes a downloader, if someone is adding their own commander keen files they may have to add it to the config folder which I believe is where the downloaded game files end up. So it just depends on if the users are adding their own or using content from the cgenius downloader. I haven't really tested it with my own games with the latest changes

HerbFargus: The hqp files are built into the game files now though so those are definitely not necessary

joolswills: will test and see if anything needs to be done in regards to symlinking or config to make downloaded files go to the roms/ports etc

HerbFargus: ok thanks, sorry I haven't had more time to test. Just for reference the path to the game files is `~/.CommanderGenius/games` which per the moveconfig function symlinks to `/opt/retropie/configs/ports/cgenius/games` so if we want the user to be able to use their own game files we can symlink that folder as well to the `~/RetroPie/roms/ports/cgenius`  for them to be able to add them otherwise we can let them add them to the configs folder. I'm fine either way, I'll leave that up to you

Jondark:
Is there a simplified set of instructions for installing the files?  My retropie fails to install at all--I have tried the set of commands listed for ubuntu installation on the Commander Genius main page.  Without installation, the files to modify are--not there.

Again, any specifics including commands to even get to the point where fills can be modified (they aren't copied to SD) would be great!

PR #2100, opened 2017-08-27
Allow multiple instances of the same token in an emulator command.

steverobb:
This allows a token to be reused more than once in an emulator command, e.g. attaching a disk image and a snapshot in VICE:

```
vice-x64-snap-disk = "/opt/retropie/emulators/vice/bin/x86 -8 %ROM%.d64 %ROM%"
```

Where there are a pair of files in the ROM directory:

```
romname.vsf
romname.vsf.d64
```


joolswills:
Thanks.

PR #2101, opened 2017-08-27
runcommand: don't run fbi/feh if already running

meleu:
This little change gives some flexibility for users who want to show other custom image via `runcommand-onstart` or something similar.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -937,11 +937,15 @@ function show_launch() {
     if [[ -n "$image" ]]; then
         # if we are running under X use feh otherwise try and use fbi
         if [[ -n "$DISPLAY" ]]; then
-            feh -F -N -Z -Y -q "$image" & &>/dev/null
-            IMG_PID=$!
-            sleep "$IMAGE_DELAY"
+            if ! pgrep feh >/dev/null; then
+                feh -F -N -Z -Y -q "$image" & &>/dev/null
+                IMG_PID=$!
+                sleep "$IMAGE_DELAY"
+            fi
         else
-            fbi -1 -t "$IMAGE_DELAY" -noverbose -a "$image" </dev/tty &>/dev/null
+            if ! pgrep fbi >/dev/null; then
+                fbi -1 -t "$IMAGE_DELAY" -noverbose -a "$image" </dev/tty &>/dev/null
+            fi
         fi
     elif [[ "$DISABLE_MENU" -ne 1 && "$USE_ART" -ne 1 ]]; then
         local launch_name


joolswills:
Can't they just disable launch art ?

meleu:
Yes, they can. But I would like to add some logic before show that "other custom image", and if it fails, runcommand should show the default "launching image".

The usecase that gave me the idea to make the change in this PR is that I want is to show a custom launching image for a "custom ES system".

For example: I have a custom collection named Mega Man, where I have Mega Man games for GameBoy, GameBoy Color, NES and SNES. Currently when I launch Mega Man X from the "Mega Man system" it shows the launching image for SNES, when I launch Mega Man 6 it shows the launching image for NES.

I would like to show a custom Mega Man launching image when I launch a game from "Mega Man system" in ES. But if I launch the same game from the original system, I want to see the launching image for the respective system.

joolswills:
I'm not convinced sorry. You can always just show the other artwork in your custom script.  Your use case seems outside of anything common.

meleu:
> You can always just show the other artwork in your custom script.

Yeah, I tried that. My other artwork is shown, but the fbi from `runcommand` is also executed.

meleu:
By the way, TMNTturtlguy has the same usecase and is achieving that with an ugly, messy and confusing approach (he detailed [here](https://retropie.org.uk/forum/post/103262)).

In the same thread cyperghost suggested another approach but it is also messy and can lead to confusion...

The little changes in this PR make the solution very simple.

TMNTturtleguy:
Yes, i know there a few of us trying to achieve showing a defined launching image for custom systems.  It would be great to have a cleaner way to do it than what i have suggested, however my approach is not messy.  I apologize if my instructions are not clear.  All i do is create a custom copy of the runcommand for each custom system.  It is a simple process.  It is not ideal, and i need to make sure that everything is set up exactly as I want it prior to copying the runcommand.sh file, but it works well, as a work around.  I would prefer not to need to modify my runcommand files, but I have taken precautions to not touch the original runcommand.  A cleaner approach would be appreciated.

meleu:
@TMNTturtleguy 
> however my approach is not messy

Forgive my words bro. I have no intention of offending or diminishing your effort.

I just would like to highlight that using a copy of runcommand for each custom system may be an immediate solution, but it is not the ideal long-term solution. If the official runcommand gets an update, your custom copies of runcommand will be outdated different...

meleu:
@joolswills here is a video of what we want to do:
https://www.youtube.com/watch?v=MfEUukBJn7s

although this is not the only use case I see for this change...

TMNTturtleguy:
@meleu 

I agree with you, copies of the runcommand are not a good solution, and as stated, i have to make sure that i have everything set exactly as I want.  Not a good or ideal solution at all.  A solution to make this work within the runcommand file itself would be very nice and a great addition.  Thanks for your efforts.

joolswills:
I have already made a decision. You can disable artwork and use your own script to do all displaying.

TMNTturtleguy:
@meleu Thanks for your efforts on this, i do appreciate it.  I do have a sub-optimal solution that works for my situation, i am not sure trying to disable all artwork and handling all the standard systems as well as all the custom systems is worth the effort, and that could get messy as well when adding, deleting or changing systems and custom images.

meleu:
@TMNTturtleguy I have a plan B. Let's go back to the forums... :-)

joolswills:
Now that I am on my PC, to clarify, there are other use cases where this would interfere.

Someone might be using FBI to display an image on another framebuffer device.

Someone might be using FEH to display another image on their system.

It's outside of the scope of runcommand to check for any other feh/fbi process and refuse to do stuff. It's simple enough if you want some launch images in some situation, just to do that also in your custom code and disable all image output from runcommand.

meleu:
@joolswills 
> It's simple enough if you want some launch images in some situation, just to do that also in your custom code and disable all image output from runcommand.

This is exactly what I have in mind. Thanks for the suggestion.

PR #2103, opened 2017-08-29
runcommand: only look for art if delay is not zero

meleu:
@joolswills sorry for bothering again, but I've just noticed that there's no option to disable launching image in runcommand setup. The "Launch menu art" option is related to scraped art only.

Currently to disable launching image we **must** delete the `launching.png` file from `/opt/retropie/configs/$SYSTEM/`, which would be unwanted.

One solution I've found is to set the "Launch image delay in seconds" to zero.

The diff of this PR make it looks like I've made a lot of changes but it's because of indentation. I only added an `if [[ "$IMAGE_DELAY" -ne 0 ]]` before the logic that looks for the image file (and a closing `fi` at the end of the same logic).

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -904,35 +904,37 @@ function get_sys_command() {
 function show_launch() {
     local images=()
 
-    if [[ "$IS_SYS" -eq 1 && "$USE_ART" -eq 1 ]]; then
-        # if using art look for images in paths for es art.
-        images+=(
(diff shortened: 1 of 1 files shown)


joolswills:
Just don't have any launching images if you don't want to use them in your very specific use case (put them somewhere else and display them from your custom script).

I don't like the code structure here. I will consider adding a way to disable launch images perhaps, but please leave this for now - runcommand development is not based around your own niche customisations.

joolswills:
BTW the way to disable launch images is to not install launch images. That was by design when we did all the discussion about logic previously when they were added.

PR #2104, opened 2017-09-03
Added Minimal Theme

lilbud:


lilbud:
I added the minimal theme to the estheme file

PR #2105, opened 2017-09-03
Update esthemes.sh

lilbud:
Added Minimal theme


joolswills:
PR has whitespace errors - I will add manually when I have a chance.

PR #2106, opened 2017-09-08
TV Theme

robertybob:
Supports 37 systems and video snaps. 

Posted it to the forum the other week, had no complaints. I haven't been able to test every single system, however it's been a straight copy & paste job so hopefully there won't be any issues.


joolswills:
This should be added underneath your other themes, rather than at the bottom of the list.

PR #2107, opened 2017-09-09
Workaround for cgenius build error

mmatyas:
As discussed in #2074 and #2099, building Commander Genius using the stock Debian GCC currently fails due to an internal compiler error. As a workaround, this patch installs and uses Clang 3.9 to build the program. The bug doesn't seem to occur with later GCC releases, so these changes can be rolled out after the upgrade to Stretch.


joolswills:
Lowering the optimisation level to -O2 also avoids the crash with gcc 4.9 - I will switch to doing this as saves installing clang. We use -O2 as default optimisation level anyway - on some things -O3 is slower on arm (due to things like unrolling vs smaller cpu caches)

PR #2113, opened 2017-09-13
Adding ability to override user during setup

tstone2077:
instead of assuming retropie will be setup under the SUDO_USER account,
there is now a way to override it using the RETORPIE_USER environment
variable.

For example:
sudo RETROPIE_USER=pi bash retropie_packages.sh setup basic_install

### DIFF
--- a/retropie_packages.sh
+++ b/retropie_packages.sh
@@ -17,9 +17,14 @@ __version="4.2.20"
 rootdir="/opt/retropie"
 
 user="$SUDO_USER"
+[[ -n "$RETROPIE_USER" ]] && user=$RETROPIE_USER
 [[ -z "$user" ]] && user="$(id -un)"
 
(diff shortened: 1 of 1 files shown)


joolswills:
thanks. Not against adding this facility, but I'm not sure about the check to see if the user is valid. I will consider adding it in a different way.

joolswills:
I will add it, but it will use the `__user` variable - just because we already use this scheme for other variables. I actually intend to change configuration variables to be UPPERCASE - perhaps with some prefix, but for now I will leave it like this.

tstone2077:
That works. Yeah, the check probably doesn't make as much sense. The reason I did that was because I have an ansible script that installs it before creating the user... but there's no reason I couldn't make the user first.

Thanks! This fix definitely is required with the ansible script since ansible is connecting via a different user (required because of other software installed on the machine).

PR #2119, opened 2017-09-17
Add console font scriptmodule

psyke83:
Allows convenient configuration of certain preset font type/sizes
appropriate for viewing on large screens.

### DIFF
--- a/scriptmodules/supplementary/consolefont.sh
+++ b/scriptmodules/supplementary/consolefont.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="consolefont"
+rp_module_desc="Configure default console font size/type"
+rp_module_section="config"
+rp_module_flags="!x11"
+
+function set_consolefont() {
+    iniConfig "=" '"' "/etc/default/console-setup"
+    iniSet "FONTFACE" "$1"
+    iniSet "FONTSIZE" "$2"
+    service console-setup restart
+    # force font configuration update if running from a pseudo-terminal
+    [[ "$(tty | egrep '/dev/tty[1-6]')" == "" ]] && setupcon -f --force
+}
+
+function check_consolefont() {
+    local fontface
+    local fontsize
+
+    iniConfig "=" '"' "/etc/default/console-setup"
+    iniGet "FONTFACE"
+    fontface="$ini_value"
+    iniGet "FONTSIZE"
+    fontsize="$ini_value"
+    echo "$fontface" "$fontsize"
+}
+
+function gui_consolefont() {
+    local cmd
+    local options
+    local choices
+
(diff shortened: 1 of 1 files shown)


psyke83:
@joolswills,

I made the changes you requested on the forum, with the caveat that I decided to move the lines to check the current font into a new function in order for the GUI to report them twice - for the menu subtitle, and once a selection has been made.

joolswills:
Looks good - thanks.

joolswills:
I will need to add another option - the Fixed font is not the default (VGA is the default, but `/etc/default/console-setup` doesn't include anything for that)

psyke83:
During the process of writing the module, I was experimenting with inserting values via "echo ... | debconf-set-selections". I ultimately decided that it was making things overly-complex, so I scrapped the idea and instead opted for direct config editing (since you have already functions intended for this kind of purpose).

I purge-removed console-setup and re-installed console-setup in order to check the font size/type default, and that results in Fixed 8x16 being filled. Does Raspbian change the package default when building images?

joolswills:
I have no idea, but the default Raspbian image doesn't have that as the default. Perhaps it's due to reinstalling the package. I can confirm that behaviour - so understand the confusion. The fixed font is thinner than the default Kernel font (which basically looks like the VGA 8x16 with minor differences) and less readable on my screen. Anyway - I made the changes now - so the Fixed is still available but you can switch to use the Default Kernel font (reboot is needed).

joolswills:
default Raspbian console-setup config -

```
# CONFIGURATION FILE FOR SETUPCON

# Consult the console-setup(5) manual page.

ACTIVE_CONSOLES="/dev/tty[1-6]"

CHARMAP="UTF-8"

CODESET="guess"
FONTFACE=""
FONTSIZE=""

VIDEOMODE=

# The following is an example how to use a braille font
# FONT='lat9w-08.psf.gz brl-8x8.psf'
```

after purging and reinstalling console-setup

```
# CONFIGURATION FILE FOR SETUPCON

# Consult the console-setup(5) manual page.

ACTIVE_CONSOLES="/dev/tty[1-6]"

CHARMAP="UTF-8"

CODESET="Lat15"
FONTFACE="Fixed"
FONTSIZE="8x16"

VIDEOMODE=

# The following is an example how to use a braille font
# FONT='lat9w-08.psf.gz brl-8x8.psf'
```

psyke83:
Ah... thanks - I neglected to backup the original file before I started messing around. A way to avoid the reboot for changes to take effect would be to set the font twice; i.e:

```
            5)
                 set_consolefont "Fixed" "8x16"
                 set_consolefont "" ""

                 ;;
```

The check_consolefont could be cleaned up slightly to report empty results as "defaults" or whatever. Will I send a PR to add that to your changes or are you happy to leave as-is?

joolswills:
Your code above would leave it as "Fixed". If the fields are cleared, you don't get the Kernel font without a reboot. the VGA 8x16 is virtually the same, but not quite (hard to tell though).

joolswills:
i'm happy to leave it as is.

joolswills:
(apart from perhaps adding some more choices).

psyke83:
I edited Fixed -> VGA moments before your reply, but I missed your explanation that it's slightly diff. to the kernel font, so it's not exactly true that it avoids the need for reboot anyway.

Sorry for the mess. Yes, let's wait and see if users have a need for other sizes, perhaps. I'm restricted to testing on a 720p capable TV, so "Medium" is the only realistically usable size for me.

joolswills:
Aah. Could do a switch to VGA before, but it's not a big deal I guess.

PR #2125, opened 2017-09-20
runcommand: fix default emulator dialog behaviour

psyke83:
When a system's default emulator is not yet selected, the automatic invocation of choose_emulator function can be erratic (most notably with the joy2key process exiting prematurely). Avoid this issue by disabling the cancel button when invoked in this way.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -575,6 +575,7 @@ function choose_mode() {
 function choose_emulator() {
     local mode="$1"
     local default="$2"
+    local cancel="$3"
 
     local default
     local default_id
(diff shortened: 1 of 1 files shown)


psyke83:
I'm not sure why the mapping breaks randomly (scrolling down the list, before selecting), but the issue can be reproduced 100% by choosing the "Back" dialog button. This will cause the choose_emulator function to execute again without joy2key running.

I'm testing with a DualShock 3 (via BT) and a DELL UK USB keyboard connect at the same time. When choosing back & mapping breaks, standard keyboard input still works OK, so it's not an issue with the previous contents of the terminal being stuck on screen with nothing running.

joolswills:
The error is due to the cancel not being handled properly (the dialog gets re-triggered - and I guess we quit joy2key or dont restart it correctly), but I don't see a reason really to have the ability to exit I guess from here, so I will just merge this. thanks.

PR #2126, opened 2017-09-22
retropie-setup - rename "update all packages" to "update" and improve…

joolswills:
… help text

 * this should help avoid confusion people have regarding the correct way to update an installed system (even though it's documented)

Feedback welcome.


meleu:
My suggestion:

"Updates RetroPie-Setup and all currently installed RetroPie based packages - also allows to update OS packages. Where binaries aren't available packages will be built from source."

PR #2136, opened 2017-10-04
splashscreen: convert to systemd service

psyke83:
* Switch to systemd service to allow finer control of start sequence
* Add insserv to depends (not installed in Raspbian stretch lite image).

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -27,21 +27,36 @@ function _video_exts_splashscreen() {
 }
 
 function depends_splashscreen() {
-    getDepends fbi omxplayer
+    getDepends fbi omxplayer insserv
 }
 
 function install_bin_splashscreen() {
-    cp "$md_data/asplashscreen" "/etc/init.d/"
+    cat > "/etc/systemd/system/asplashscreen.service" << _EOF_
+[Unit]
+Description=Show custom splashscreen
+DefaultDependencies=no
+Before=local-fs-pre.target
+Wants=local-fs-pre.target
+ConditionPathExists=$md_inst/asplashscreen.sh
 
-    iniConfig "=" '"' /etc/init.d/asplashscreen
+[Service]
+Type=oneshot
+ExecStart=$md_inst/asplashscreen.sh
+RemainAfterExit=yes
+
+[Install]
+WantedBy=sysinit.target
+_EOF_
+
+    gitPullOrClone "$md_inst" https://github.com/RetroPie/retropie-splashscreens.git
+
+    cp "$md_data/asplashscreen.sh" "$md_inst"
+
+    iniConfig "=" '"' "$md_inst/asplashscreen.sh"
     iniSet "ROOTDIR" "$rootdir"
     iniSet "DATADIR" "$datadir"
     iniSet "REGEX_IMAGE" "$(_image_exts_splashscreen)"
     iniSet "REGEX_VIDEO" "$(_video_exts_splashscreen)"
-
-    chmod +x /etc/init.d/asplashscreen
-
-    gitPullOrClone "$md_inst" https://github.com/RetroPie/retropie-splashscreens.git
 
     mkUserDir "$datadir/splashscreens"
     echo "Place your own splashscreens in here." >"$datadir/splashscreens/README.txt"
@@ -67,15 +82,19 @@ function default_splashscreen() {
 }
 
 function enable_splashscreen() {
-    insserv asplashscreen
+    systemctl enable asplashscreen
 }
 
 function disable_splashscreen() {
-    insserv -r asplashscreen
+    systemctl disable asplashscreen
 }
 
 function configure_splashscreen() {
     [[ "$md_mode" == "remove" ]] && return
+
+    # remove legacy service
+    [[ -f "/etc/init.d/asplashscreen" ]] && insserv -r asplashscreen && rm -f /etc/init.d/asplashscreen
+
     disable_plymouth_splashscreen
     enable_splashscreen
     [[ ! -f /etc/splashscreen.list ]] && default_splashscreen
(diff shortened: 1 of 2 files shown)


psyke83:
Apologies for not discussing before submitting, but I've been re-testing against stretch in lieu of the pkgconfig change that's necessary to make building easier.

I noticed that on stretch, sysvinit scripts no longer work on runlevel S. This means that the splashscreen service won't work unless you change it to higher runlevels, but that comes at the cost of it executing later than wanted.

So, I decided to convert it to a systemd service. I tested against both jessie and stretch, and it seems to works fine on both.

I made sure that the legacy /etc/init.d/asplashscreen is cleanly removed during install of the new service, and the new remove function will clean up all files related to the service correctly.

joolswills:
No worries and thanks. This has needed switching to a systemd service for some time. I have a few minor changes though.

scriptmodules/supplementary/splashscreen.sh
```diff
@@ -27,20 +27,41 @@ function _video_exts_splashscreen() {
 }
 
 function depends_splashscreen() {
-    getDepends fbi omxplayer
+    getDepends fbi omxplayer insserv
 }
 
 function install_bin_splashscreen() {
-    cp "$md_data/asplashscreen" "/etc/init.d/"
+    cp -a "$md_data/asplashscreen.sh" "$rootdir/lib/"
```
joolswills: `"$md_inst"` would be a more suitable location (actual install folder for the module)

scriptmodules/supplementary/splashscreen.sh
```diff
@@ -84,7 +105,7 @@ function configure_splashscreen() {
 function remove_splashscreen() {
     enable_plymouth_splashscreen
     disable_splashscreen
-    rm -f /etc/splashscreen.list
+    rm -f /etc/splashscreen.list /etc/systemd/system/asplashscreen.service "$rootdir/lib/asplashscreen.sh"
```
joolswills: removing splashscreen.sh won't be required when it's installed to `"$md_inst"`

scriptmodules/supplementary/splashscreen.sh
```diff
@@ -27,20 +27,41 @@ function _video_exts_splashscreen() {
...
+    getDepends fbi omxplayer insserv
 }
 
 function install_bin_splashscreen() {
-    cp "$md_data/asplashscreen" "/etc/init.d/"
+    cp -a "$md_data/asplashscreen.sh" "$rootdir/lib/"
 
-    iniConfig "=" '"' /etc/init.d/asplashscreen
+    # remove legacy service
+    [[ -f "/etc/init.d/asplashscreen" ]] && insserv -r asplashscreen && rm -f /etc/init.d/asplashscreen
```
joolswills: This would probably be better done at the configure stage (to match other module behaviour).

psyke83:
I've made the changes you requested. One thing to note: I moved the legacy service removal to the configure function, but it can't run in the "remove" phae, otherwise systemd will try to synchronize the old script to the new service and will prevent it from being removed correctly.

joolswills:
Thanks. Will look at tomorrow.

psyke83:
No rush. I overlooked the reason why I originally didn't choose "$md_inst" - asplashscreen.sh would be included in the menu to select the splash media. I refreshed the patch to add .sh files to the the regex exclusion.

scriptmodules/supplementary/splashscreen.sh
```diff
@@ -27,22 +27,39 @@ function _video_exts_splashscreen() {
...
+
+[Install]
+WantedBy=sysinit.target
+_EOF_
+
+    gitPullOrClone "$md_inst" https://github.com/RetroPie/retropie-splashscreens.git
+
+    cp -a "$md_data/asplashscreen.sh" "$md_inst"
+    chown root:root "/etc/systemd/system/asplashscreen.service" "$md_inst/asplashscreen.sh"
+    chmod +x "/etc/systemd/system/asplashscreen.service" "$md_inst/asplashscreen.sh"
```
joolswills: Is the chown needed ? (removing the `-a` should remove the need for the chown, and would still preserve any executable permissions of the .sh script. Also the service doesn't need to be executable agaik.

scriptmodules/supplementary/splashscreen.sh
```diff
@@ -137,7 +159,7 @@ function choose_splashscreen() {
             options+=("$i" "$splashdir")
             ((i++))
         fi
-    done < <(find "$path" -type f ! -regex ".*/\..*" ! -regex ".*LICENSE" ! -regex ".*README.*" | sort)
+    done < <(find "$path" -type f ! -regex ".*/\..*" ! -regex ".*LICENSE" ! -regex ".*README.*" ! -regex ".*.sh" | sort)
```
joolswills: it should be `".*\.sh"` as your regexp would match a file called "whoosh" for example (works anyway, but technically wrong).

scriptmodules/supplementary/splashscreen.sh
psyke83: You're right about the executable permissions for the .service file - I'll remove that. As for the chown, it's not necessary, but it could be a security issue to allow a system service to be modified by a user account. I think that git preserves file permissions but not user/group ones.

joolswills: if you dont use `cp -a` the files will be installed as root. cp doesn't copy user ownership by default.

psyke83:
PR has been updated with the permissions and regex issue fixed.

Are there other targets for retropie that use a distribution without systemd installed? Perhaps it's not a good idea to merge if that's the case.

joolswills:
There are, but none that use the splashscreen.

joolswills:
BTW did you test with a splash video also ? Forgot to ask that.

psyke83:
No, but I just tested this moment and it's working. The video persists for a second or so after ES loads, but I'm fairly sure that already happened.

joolswills:
yep.

joolswills:
depending on length of video vs startup time

joolswills:
thanks.

PR #2137, opened 2017-10-04
Mupen64plus x86 cleanup

gizmo98:
Fix ubuntu >= 16.10 audio issue (no sound) and remove old emulator entries.


joolswills:
Thanks

PR #2138, opened 2017-10-04
Add platform flags & pkgconfig

psyke83:
* Add "mesa kms" flags when vc4 module is inserted.
* Add "dispmanx" flag if /sys/bus/platform/devices/*.firmwarekms is enumerated (fake kms).
* Fallback to "videocore dispmanx" if neither is detected.
* Put a hard block on script execution if VC4 driver is enabled.
* Set environment variable for PKGCONFIG (coming in next stretch firmware).
* Use fallback copy of pkgconfig if needed; useful for current stretch firmware that lacks pkgconfig & legacy jessie compatibility. Separate commit so that it can be cleanly reverted in future.

### DIFF
--- a/pkgconfig/bcm_host.pc
+++ b/pkgconfig/bcm_host.pc
@@ -1,0 +1,10 @@
+prefix=/opt/vc
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+Name: bcm_host
+Description: Broadcom VideoCore host API library
+Version: 1
+Libs: -L${libdir} -lbcm_host -lvcos -lvchiq_arm -pthread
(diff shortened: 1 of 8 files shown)


joolswills:
Thanks for your contributions. Looks good but on phone so will check/merge tomorrow. Cheers

psyke83:
Updated. Tried to simplify fake kms detection slightly, too. The only difference I can see when the fkms DT overlay active is the presence of the symlink /sys/bus/platform/drivers/vc4_firmware_kms/3f600000.firmwarekms, which points to /sys/devices/platform/soc/3f600000.firmwarekms. Full KMS enumerates /sys/bus/platform/drivers/vc4_firmware_kms/ but not the .firmwarekms symlink or actual soc node.

joolswills:
thanks.

PR #2140, opened 2017-10-05
Re-enabled online updater in RGUI (for cheats)

dankcushions:
Many reports of this:
https://www.reddit.com/r/RetroPie/comments/74fpcj/online_updater_missing_from_retroarch/
https://www.reddit.com/r/RetroPie/comments/6x9515/where_is_the_online_updater/
https://www.reddit.com/r/RetroPie/comments/72trxn/cheats_online_update_and_retropie_43/
https://www.reddit.com/r/RetroPie/comments/740ea1/no_online_updater_option_in_rgui/

and in our docs:
https://retropie.org.uk/docs/Cheats/

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -181,9 +181,11 @@ function configure_retroarch() {
     # rgui by default
     iniSet "menu_driver" "rgui"
 
-    # hide online updater menu options
+    # hide core updater menu option
     iniSet "menu_show_core_updater" "false"
-    iniSet "menu_show_online_updater" "false"
+    
+    # show online updater menu option (for cheats)
+    iniSet "menu_show_online_updater" "true"
 
     # disable unnecessary xmb menu tabs
     iniSet "xmb_show_add" "false"


joolswills:
I don't really want to re-enable it just for cheats - The other options in the menu conflict with retropie (we dont manage other things from there - eg shaders). I may add cheat downloading to the retropie gui though.

joolswills:
I will remove the section from the docs for now until something is in place - was probably user contributed.

joolswills:
(actually it was created by Herb, but in any case, the online updater can cause confusion as people previously tried to use it to download core updates which didn't work etc). I will add cheats to the RetroPie GUI or as you have suggested depending which is easiest.

HerbFargus:
>probably user contributed. 

this implies that I actually use retropie ;) anyways no qualms with me to have an option in the retropie gui if that eliminates confusion with retroarch's online updater.

ROBYER1:
Good ###'ing don't update documentation after making a change on stable release developer moves...
https://docs.libretro.com/guides/install-windows/

PR #2146, opened 2017-10-07
Implement patch_rpi_video function & restrict pkgconfig

psyke83:
As requested, a list of the projects that have been converted to the new libraries:

``` 
PRs merged:
emulationstation (master only, stable unaffected)
pisnes
mame4all-pi
mupen64plus
pifba
retroarch (master, so v1.6.7 is unaffected)

PRs pending:
lr-ppsspp
lr-parallel-n64
```

This patch will "fix" all of the above projects so that they link against the old libraries on jessie. When this is applied, it should be safe to continue sending PRs to other projects & building against them. When you feel confident that enough projects have been switched over, you can revert this patch.

I recommend keeping the 2.0.5 packages, but keep the rpi-2.0.6 branch for when you want to switch. 
The 2.0.6 branch could be used, but would need these environment variables set:

```
SDL_VIDEO_EGL_DRIVER=/opt/vc/lib/libEGL.so
SDL_VIDEO_GL_DRIVER=/opt/vc/lib/libGLESv2.so
```


psyke83:
I found a small issue that needs resolving. Closing until I've resolved it & done a full battery of tests against *all* packages to ensure no false positives.

joolswills:
I thought you meant a changeset to patch up software to use the new library names, not the old.

I would like to resolve this with flags and then env variables so we can gradually move everything over. Ideally things like mupen64plus wouldn't have been changed until things like the new retroarch are out.

It might be better to hold on, as I would like to look into this next week - since only a few sdl2 programs have been changed so far, I think I will come up with a flag system, but also fork/revert the mupen64 changes for now.

joolswills:
btw as I guess you are aware - pisnes, mame4all-pi and pifba can remain updated since they don't use sdl2

psyke83:
Yes, I was going to try that, but I don't recommend using a flag system.

I'm guessing that if there's an active EGL context held by libEGL.so, it doesn't get communicated to libbrcmEGL.so correctly (and vice versa). I started seeing intermittant issues such as emulationstation (using libEGL.so) freezing after returning from an application using the new library name.

I'm going to do a thorough test to 100% ensure no false positives. Here's how I'm testing:

```
diff --git a/scriptmodules/packages.sh b/scriptmodules/packages.sh
index d184137..5797706 100644
--- a/scriptmodules/packages.sh
+++ b/scriptmodules/packages.sh
@@ -91,7 +91,7 @@ function rp_callModule() {

     # automatically build/install module if no parameters are given
     if [[ -z "$mode" ]]; then
-        for mode in depends sources build install configure clean; do
+        for mode in depends sources; do
             rp_callModule "$md_idx" "$mode" || return 1
         done
         return 0
```
Then I install all packages from source (which in reality just fetches the source and processes everything via the patch_rpi_video). Then I check the build log, and manually check the makefiles in the tmp/build/ to ensure no false positives.

The current issue with the PR is that it processes the RetroPie-Setup script itself. I've already fixed that case on my local copy with which I'm testing this build.

joolswills:
I would prefer a manual flag system over trying to patch stuff up like this. I would like to look at this myself.

psyke83:
Just let me work out the kinks before looking at it, please. I genuinely believe this is the safest way forward to ensure no breakage.

When working as intended, we will be able to freely send PRs to projects to change the library name, but RetroPie users building from source will not be negatively affected. 

For you or I who needs to test the new library and/or PRs, we simply disable the patch_rpi_video function on our local builds.

joolswills:
But first, I would like to fork/revert any changed upstream code (only a few so far we need to handle), as right now, there are users reporting issues (installing mupen64 from source). I am in no rush to move to the new library names, and will at least wait for the next retroarch to be released.

psyke83:
I don't mean to sound like I'm bossing you around, but I'd just suggest that you avoid reverting anything until you take a proper look at this PR in its finished state. If I can guarantee no false positives or ill effects, this seems like the perfect solution to ease the transition process.

I know you wanted to take a break for the weekend, sorry :). I'm certain the PR will be ready within a few hours, and certainly for you to look at on Monday with fresh eyes.

joolswills:
I appreciate your view, and your help, but I think you rushing forward with all this is causing problems. The upstream library changes were done too early (as was my sdl2 update).

joolswills:
I can only reiterate I don't want to take your approach currently -so you may well be wasting your time.

psyke83:
That's fine, it's at your discretion. I'll update the PR for you to look at without any expectation of it being merged, and if you want to revert the upstream patches, I won't object.

I agree re: SDL 2.0.6 being premature. In case I wasn't clear, this PR is intended to be used with the stable SDL 2.0.5.

joolswills:
right now there is only mupen64plus that needs reverting - it's a quick fix and I prefer doing that and then planning properly the move to the new names.

joolswills:
(and ppsspp if they accept the PR). but we can always switch to the last stable for that.

joolswills:
I want to move to the new SDL 2.0.6 - but with compatibility with the older library names also.

psyke83:
Damn it, I forgot that force-pushing before re-opening screws up github. Anyway, with the latest patch here: https://github.com/RetroPie/RetroPie-Setup/compare/master...psyke83:vendor_workaround

I added an exception to SDL2 due to the file needed to be changed won't match the restrictive regex:
```
diff --git a/scriptmodules/supplementary/sdl2.sh b/scriptmodules/supplementary/sdl2.sh
index 8a7253d..1d29dcb 100644
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -51,6 +51,7 @@ function sources_sdl2() {
     gitPullOrClone "$md_build/$pkg_ver" https://github.com/RetroPie/SDL-mirror.git "$branch"
     cd "$pkg_ver"
     DEBEMAIL="Jools Wills <buzz@exotica.org.uk>" dch -v "$pkg_ver" "SDL $ver configured for the $__platform"
+    isPlatform "rpi" && patch_rpi_video "src/video/SDL_egl.c"
 }

 function build_sdl2() {
```


Watch what happens:
```
/home/pi/RetroPie-Setup/tmp/build/sdl2 /home/pi/RetroPie-Setup

Getting sources for 'sdl2' : SDL (Simple DirectMedia Layer) v2.x

git clone --recursive --depth 1 --branch rpi-2.0.6 "https://github.com/RetroPie/SDL-mirror.git" "/home/pi/RetroPie-Setup/tmp/build/sdl2/2.0.6+1rpi"
Cloning into '/home/pi/RetroPie-Setup/tmp/build/sdl2/2.0.6+1rpi'...
remote: Counting objects: 1445, done.
remote: Compressing objects: 100% (1304/1304), done.
remote: Total 1445 (delta 385), reused 412 (delta 64), pack-reused 0
Receiving objects: 100% (1445/1445), 4.71 MiB | 775.00 KiB/s, done.
Resolving deltas: 100% (385/385), done.
Checking connectivity... done.
Applying Raspberry Pi legacy graphics patch: src/video/SDL_egl.c
/home/pi/RetroPie-Setup
```

Now looking at the patched file:
```
pi@retropie:~/RetroPie-Setup/tmp/build/sdl2/2.0.6+1rpi $ git diff src/video/SDL_egl.c
diff --git a/src/video/SDL_egl.c b/src/video/SDL_egl.c
index 9e10c2b..98c670b 100644
--- a/src/video/SDL_egl.c
+++ b/src/video/SDL_egl.c
@@ -44,12 +44,12 @@

 #if SDL_VIDEO_DRIVER_RPI
 /* Raspbian places the OpenGL ES/EGL binaries in a non standard path */
-#define DEFAULT_EGL ( vc4 ? "libEGL.so" : "/opt/vc/lib/libbrcmEGL.so" )
-#define DEFAULT_OGL_ES2 ( vc4 ? "libGLESv2.so.2" : "/opt/vc/lib/libbrcmGLESv2.so" )
+#define DEFAULT_EGL ( vc4 ? "libEGL.so" : "/opt/vc/lib/libEGL.so" )
+#define DEFAULT_OGL_ES2 ( vc4 ? "libGLESv2.so.2" : "/opt/vc/lib/libGLESv2.so" )
 #define ALT_EGL ( vc4 ? "libEGL.so" : "/opt/vc/lib/libEGL.so" )
 #define ALT_OGL_ES2 ( vc4 ? "libGLESv2.so.2" : "/opt/vc/lib/libGLESv2.so" )
-#define DEFAULT_OGL_ES_PVR ( vc4 ? "libGLES_CM.so.1" : "/opt/vc/lib/libbrcmGLESv2.so" )
-#define DEFAULT_OGL_ES ( vc4 ? "libGLESv1_CM.so.1" : "/opt/vc/lib/libbrcmGLESv2.so" )
+#define DEFAULT_OGL_ES_PVR ( vc4 ? "libGLES_CM.so.1" : "/opt/vc/lib/libGLESv2.so" )
+#define DEFAULT_OGL_ES ( vc4 ? "libGLESv1_CM.so.1" : "/opt/vc/lib/libGLESv2.so" )

 #elif SDL_VIDEO_DRIVER_ANDROID || SDL_VIDEO_DRIVER_VIVANTE
 /* Android */
```

So, ironically, this function would even fix SDL 2.0.6.

PR #2148, opened 2017-10-09
 helpers: joy2keyStop: wait for successful SIGINT

psyke83:
Under certain scenarios, joy2key.py does not receive the INT signal
which leads to the process staying resident. Avoid this situation by
sending the signal repeatedly until a successful exit code is received.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1025,8 +1025,9 @@ function joy2keyStart() {
 ## @brief Stop previously started joy2key.py process.
 function joy2keyStop() {
     if [[ -n $__joy2key_pid ]]; then
-        kill -INT $__joy2key_pid 2>/dev/null
-        sleep 1
+        while kill -INT $__joy2key_pid 2>/dev/null; do
+            sleep 0.1
+        done
     fi
 }
 


psyke83:
Hmm, I can't replicate Ctrl+C triggering this but any more. Sometimes I was seeing an errant joy2key process after exiting from the script. Controller input would interfere with the bash session...

I'll close until I can trigger it properly again and know if this really helps or not.

psyke83:
Discovered the condition in which to trigger the issue:

1. Ensure that a joypad is connected.
2.
```
sudo ./retropie_packages test123
```
(or a valid command that completes quickly, such as runcommand)

This always results in a joy2key process being left over. After some research, it seems that python's signal handler may not reliably receive the signal every time. Sending the signal repeatedly seems to resolve the issue. Adding debug, it seems to successfully trigger the signal handler after the second or sometimes third iteration of this loop.

joolswills:
will test.

psyke83:
Cool. I added some debug to joy2key's main thread and signal_handler. What's happening is that the first joy2keyStop is being issued before the joy2key.py's main thread or handler fully executes, most likely due to the overhead of python to initialize & parse the script.

psyke83:
FYI: I was able to reproduce the same issue running on Debian 9's live image - the PR fixes it on this setup too.

joolswills:
I think this could cause an infinite loop in some scenarios - eg if for some reason joy2key fails. I will look into it further when I have time.

psyke83:
My desc of the patch was a bit misleading. The loop stops once the INT signal fails to be sent. Here's what happens without the PR:

* joy2keyStart runs joy2key.py, backgrounded
* joy2keyStop triggers "kill -INT x" 
* kill -INT reports successful exit code (python receives the signal but hasn't started interpreting the script)
* joy2key.py registers signal_handler, but missed the signal

So the loop actually stop once kill receives an *unsuccessful* exit code. If joy2key is killed prematurely, it won't get stuck in the loop.

An additional safety check could be added to the existing PR to check if the process still exists (but it's not necessary).

```
    if [[ -n $__joy2key_pid ]] && kill -0 $joy2key_pid; then
```

But I guess joy2key itself might get stuck, as sys.exit() can trigger an exception under certain circumstances... so maybe leave this for now, I guess. :/

joolswills:
aah sorry. I see. I think I'll accept as is for now then thanks - if we get reports of any issues, we can always adjust.

PR #2149, opened 2017-10-10
Unblank framebuffer on setup, runcommand and joypad inputs

psyke83:
Raspbian stretch uses a very short framebuffer timeout. While good to avoid
burn-in when users leave an active console displayed, this interferes with
script and runcommand visibility. Solve the issue by unblanking on first
invocation of scripts, and also trigger unblank on every joypad input.

* Unblank framebuffer on RetroPie-Setup and runcommand launch.
* Avoid unnecessary unblank invocation on pseudoterminals.
* Modify joy2key to send ANSI setterm code ESC [13] (unblank framebuffer)
  to simulate regular wakeup on key press.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -486,6 +486,7 @@ function reboot_setup()
 # retropie-setup main menu
 function gui_setup() {
     depends_setup
+    [[ "$TERM" == "linux" ]] && setterm --blank poke
     joy2keyStart
     local default
     while true; do
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -31,6 +31,8 @@ JS_EVENT_BUTTON = 0x01
 JS_EVENT_AXIS = 0x02
 JS_EVENT_INIT = 0x80
 
+FB_UNBLANK = '\033[13]'
+
 CONFIG_DIR = '/opt/retropie/configs/'
 RETROARCH_CFG = CONFIG_DIR + 'all/retroarch.cfg'
 
@@ -194,6 +196,9 @@ def process_event(event):
     if hex_chars:
         for c in hex_chars:
             fcntl.ioctl(tty_fd, termios.TIOCSTI, c)
+        if os.getenv('TERM') == 'linux':
+            sys.stdout.write(FB_UNBLANK)
+            sys.stdout.flush()
         return True
 
     return False
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -1025,9 +1025,10 @@ function runcommand() {
         exit 1
     fi
 
-    # turn off cursor and clear screen
+    # turn off cursor, clear screen and unblank
     tput civis
     clear
+    [[ "$TERM" == "linux" ]] && setterm --blank poke
 
     rm -f "$LOG"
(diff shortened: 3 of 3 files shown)


psyke83:
To be specific, the console blank timeout on stretch is 10 minutes:
```
pi@retropie-test:~/RetroPie-Setup $ cat /sys/module/kernel/parameters/consoleblank
600
```

If you reboot, wait >10 minutes, and then launch an item or the script, you will only see a black screen. This can be alleviated by pressing a key on a connected keyboard, but this is obviously not much help if your primary control method is a controller.

If you want a faster way to test, use the first command on the real VT or the second via ssh:

```
setterm --blank 1 (1 min)
sudo bash -c "echo 1 >/sys/devices/platform/soc/soc:fb/graphics/fb0/blank" (immediately blank)
```

You could disable the FB entirely to fix this, but I'd be concerned about burn-in if I forgot to turn off the TV, etc.

joolswills:
Will test this out - we usually disable the console blanking on retropie btw (It's disabled on the images).

joolswills:
instead of using `$TERM` or just redirecting stderr, just check `$SSH_TTY`

joolswills:
wrong ticket. sorry

joolswills:
I got the tickets mixed up - this one you can use what I wrote in the other ticket:

`[[ "$(tty) != /dev/tty* ]]`

psyke83:
Still thinking about the ssh issue, but just for this PR, what if I simply change to
```
echo -ne '\033[13]'
```

This doesn't do any harm on any terminal client, including ssh, as it's a standard ANSI code. I was only adding the conditional for setterm to avoid unnecessary warnings if not on a real tty.

psyke83:
It seems that github is having issues. I'm seeing notifications for your replies via email but they don't appear here (on the PR page). I also posted a reply but it seems to have disappeared and I'm not sure if I forgot to click send, or it's the same issue (so, also sorry for spam if that's the case ;))

I wrote that maybe it's OK for this PR just to unconditionally print the ANSI characters, as it doesn't harm ssh, gnome-terminal, etc.
```
echo -ne '\033[13]'
```

joolswills:
https://status.github.com/messages

>October 11, 2017
16:51 BST
We're currently experiencing high replication delay on our primary database cluster. Recently performed site actions may not appear in the UI.

joolswills:
if you are sure the echo is harmless that's fine.

psyke83:
I don't have access to my regular linux installation, but give me a little time to boot up a Debian livecd and run the script with the change to be 100% sure. I'm trying to fix bugs rather than adding to them :)

psyke83:
On Debian it spews some invalid characters on gnome-terminal, so it's better not to merge this yet (as joy2key may cause the same issue). I'll have to add an exception to joy2key as well. Would you object to changing the scripts so that joy2key's first argument declares whether to emulate the unblank?

joolswills:
I'd rather not change the arguments - some people use joy2key in custom scripts etc. I'm not convinced that it's joy2key's job to do this or that we need it. Since we disable blanking in the official retropie image and have an option to disable it in raspbian tools.

We could disable blanking on launch of retropie-setup also.

joolswills:
I think I'd rather just not include this than spend more time discussing back and forth at this time. I have other things I would like to work on currently. I may revisit it if I think it really is needed.

psyke83:
Sure, no problem. I'll close this but if I find a good solution worth your attention, I'll reopen.

psyke83:
Just updated the patch to restrict joy2key to TERM=linux to avoid the issue with unsupported ANSI codes on Ubuntu, etc. It probably works OK now, but will close again as it's not a big issue if you still plan to disable blanking by default.

PR #2150, opened 2017-10-10
joy2key: don't execute on pseudoterminals

psyke83:
Avoid joy2key invocation on pseudoterminals (ssh via xterm, tmux, etc).

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1001,6 +1001,8 @@ _EOF_
 ## @details Arguments are curses capability names or hex values starting with '0x'
 ## see: http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html
 function joy2keyStart() {
+    [[ "$TERM" != "linux" ]] && return
+
     local params=("$@")
     if [[ "${#params[@]}" -eq 0 ]]; then
         params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20)
@@ -1024,6 +1026,8 @@ function joy2keyStart() {
 ## @fn joy2keyStop()
 ## @brief Stop previously started joy2key.py process.
 function joy2keyStop() {
+    [[ "$TERM" != "linux" ]] && return
+
     if [[ -n $__joy2key_pid ]]; then
         kill -INT $__joy2key_pid 2>/dev/null
         sleep 1


psyke83:
This change seems to make sense. Useful if you're using the script via ssh to install/configure/test something, then you will be able to use EmulationStation without having to worry about the script in the ssh terminal accidentally selecting random items.

joolswills:
I'm pretty sure this will break joy2key on Ubuntu etc. I don't really see this as an issue - as people can just avoid using the joystick when they are doing admin over ssh etc.

dankcushions:
I have definitely been caught out by this a few times before :blush:

psyke83:
This would probably be safe for any linux distro:
```
[[ ! "$(tty) | grep /dev/pts" ]] && ...
```

For the other PR, I could just switch from setterm to:

```
echo -ne '\033[13]'
```
Which shouldn't even need a conditional like the above.

joolswills:
I was just coming here to suggest you use the output of `tty` :-) but you don't need to use grep - eg `[[ "$(tty) != /dev/tty* ]]`

joolswills:
Sorry, I got mixed up - ignore previous message.

for this PR, you can check `$SSH_TTY` to avoid doing it on SSH sessions. you still want joy2key to run on linux terminals on Ubuntu etc, and tty will be `/dev/pts/*` for them.

joolswills:
I don't want the logic in the functions themselves - as the functions should always do what they claim imho - better to do the logic in the caller. I'll just make the change as it will be quicker.

joolswills:
Actually, I will put it in the functions as we call the function in multiple places - seems simpler actually.

joolswills:
Sorry for spam - actually SSH_TTY isn't any good since the variable is lost due to sudo being used I think. Will need another solution.

joolswills:
maybe something like `[[ "$(who -m)" == *\(* ]] && echo "SSH"` (check for bracket in output as last field contains ip/hostname in brackets over ssh)

or `[[ -n $(who -m | awk -F " " '{print $NF}') ]] && echo "SSH"` using awk rather than the bracket matching.

or can check parent processes for ssh I guess. Unless you have a better solution ?

joolswills:
Please can you test

```diff
diff --git a/scriptmodules/helpers.sh b/scriptmodules/helpers.sh
index 1e4e1b54..c808b385 100644
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1001,6 +1001,10 @@ _EOF_
 ## @details Arguments are curses capability names or hex values starting with '0x'
 ## see: http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html
 function joy2keyStart() {
+    # don't start on SSH sessions
+    # (check for bracket in output - ip/name in brackets over an SSH connection)
+    [[ "$(who -m)" == *\(* ]] && return
+
     local params=("$@")
     if [[ "${#params[@]}" -eq 0 ]]; then
         params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20)
```

No logic is needed in the stop function, since it is a no-op if no joy2key was started.

On the terminal built into my editor Geany, it outputs `jools    pts/3        2017-10-11 17:11 (:0.0)` which means joy2key isn't started, but seems ok from gnome-terminal and my xfce-terminal  and on the local terminal on the RPI, so should be good enough unless there is a simpler solution.

psyke83:
I can confirm it works fine over ssh - also when embedded in tmux, as it nests brackets like so:

```
pi       pts/1        2017-10-11 17:58 (tmux(4197).%4)
```

I won't have physical access to test if the real VT is working OK until tonight. Will let you know ASAP.

PR #2155, opened 2017-10-12
bluetooth: fix bluez 5.43 compatibility

psyke83:
It's necessary to read more characters per line to match the "[bluetooth]" string due to commit:
https://git.kernel.org/pub/scm/bluetooth/bluez.git/commit/?id=6fd2662dc96ccf5fc194c53d5c6f90431ba9f8a5

Fixes bluetooth on stretch.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -56,7 +56,7 @@ function list_available_bluetooth() {
         mkfifo "$fifo"
         exec 3<>"$fifo"
         local line
-        while read -r -n12 line; do
+        while read -r -n18 line; do
             if [[ "$line" == *"[bluetooth]"* ]]; then
                 echo "scan on" >&3
                 read -r line


psyke83:
On stretch, the script needs to match against the POWER_ON macro in the client/main.c which includes control codes to set terminal colours. The current limit of 12 characters is not enough, leading to the "Searching" dialog to run indefinitely.

The ps3controller driver is also having issues with daemon re/start. I'll send a PR for that once I've got it fixed and verified not to break jessie, etc.

joolswills:
Thanks. Everything still ok on Jessie after this? Will test when I'm next free. Cheers

psyke83:
Jessie is fine - just double-checked now.

Without this PR, jessie may send the "scan on" command prematurely due to the prompt appearing before the daemon connection is made (PROMPT_OFF). It's actually the PROMPT_ON that signifies it to be ready. We could try to make jessie more reliable by matching against the [bluetooth] with the colour codes, but that may make the script more brittle in case they change PROMPT_ON in future...

joolswills:
thanks.

PR #2156, opened 2017-10-13
Enable PowerBlock and ControlBlock drivers with installation

petrockblog:
Some users stumbled over the current modules implementation that the ControlBlock and PowerBlock  drivers are not enabled with the installation, but needed to be enabled in another step after the installation. The drivers are enabled with the installation now.

### DIFF
--- a/scriptmodules/supplementary/controlblock.sh
+++ b/scriptmodules/supplementary/controlblock.sh
@@ -39,6 +39,7 @@ function install_controlblock() {
     # install from there to system folders
     cd "$md_inst/build"
     make install
+    make installservice
 }
 
 function gui_controlblock() {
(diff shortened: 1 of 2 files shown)


joolswills:
One problem with this is that if someone disables the service and then updates, it will get re-enabled. Can use a flag/file in $md_inst (eg touch "$md_inst/disabled") to workaround this as we do currently in usbromservice, or perhaps just remove the GUI, and have it so it's just enabled when installed and disabled when removed.

joolswills:
I made the changes I mentioned with the flag. When testing the powerblock module (without having a powerblock on the test rpi), I noticed that when installing the service, I lost connection to my RPI - does it reboot by default when installing the service? (I couldn't say for sure since the RPI had no screen connected) - Ideally we would avoid rebooting without informing the user in retropie-setup (and the powerblock tries to start  on boot even if the hardware doesn't exist with a 5m timeout - is there a way to avoid that in your code ?)

petrockblog:
The reboot is indeed a problem I could have thought about myself, sorry. I just created a pull request that basically reverts the last changes and adds some more information about the need to manually enable the service.

PR #2158, opened 2017-10-13
ps3controller: update for new systemd package

psyke83:
The systemd package will automatically cleanup damage from the old package and enable/start the new system service, so remove it from here.

### DIFF
--- a/scriptmodules/supplementary/ps3controller.sh
+++ b/scriptmodules/supplementary/ps3controller.sh
@@ -16,7 +16,7 @@ rp_module_section="driver"
 
 function depends_ps3controller() {
     depends_bluetooth
-    local depends=(checkinstall libusb-dev libbluetooth-dev joystick insserv)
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/ps3controller.sh
```diff
@@ -47,27 +47,19 @@ function install_ps3controller() {
...
-    # Start sixad daemon
-    /etc/init.d/sixad start
 }
 
 function remove_ps3controller() {
-    /etc/init.d/sixad stop
-    insserv -r sixad
     dpkg --purge sixad
     rm -f /etc/udev/rules.d/99-sixpair.rules
     rm -f /etc/udev/rules.d/10-local.rules
```
psyke83: Not sure if we need to delete these udev rules? The sixad package doesn't provide them...

joolswills: Probably can now - looks like something that was needed from a couple of years ago.

PR #2161, opened 2017-10-17
Stretch fixes

psyke83:
Both seem to be working OK on both jessie and stretch.

darkplaces-quake uses SDL2, so the newbrcmlibs flag has been added (even though it seemed to work OK without it).


joolswills:
no newbrcmlibs flag for atari800 ?

psyke83:
No, it doesn't use SDL2.

joolswills:
Thanks. Must have been thinking about hatari.

PR #2163, opened 2017-10-17
Update README.md

NerdOfLinux:
Added alternate way of downloading scripts that does not require the installation of git.


joolswills:
RetroPie has a dependency on git uses it updating etc. I don't see any benefit from updating this way and it will cause problems.

Issue #1798, opened 2016-12-22
Cannot configure 2 different controllers which have the same name


danmcgoo:
First of all, I am not sure if this issue is really related to the Retropie-Setup itself.
It is actually related to the configuration of controllers in EmulationStation and retroarch.

Currently the controllers are identified by their name (name attribute given by udev).
However, I happen to have 2 different types of gamepads which firmware is made by the same company (while the brand of the gamepads are different), so their udev name is the same ("USB, 2-axis 8-button gamepad  ").

However, the physical disposition of the buttons is different, which means that if I configure one of the gamepad, the other one will have the buttons A swapped with B and X swapped with Y.

I tried to fix the issue with the following methods:
- Remap the buttons of one controller using jscal and jscal-store. However, since EmulationStation and Retroarch relies on udev, this solution does not work (I tried to change the input driver from udev to linuxraw, but the result was the same)
- Try to change the name of the devices with a udev rule (what a fool I am). I realized this is impossible...

But I do see one solution, which would be to use the Product and Vendor ID or GUID in addition of the device name.
This solution currently works in EmulationStation. If I add the DeviceGUID in addition of the device name, I can have 2 different settings for controllers having the same name.
So it would be nice that the DeviceGUIS would not be stripped off in the es_input.cfg.

The remaining task would be to make Retroarch compatible with another way to identify the controller.

joolswills:
I can probably add the GUID to the ES config, but if If retroarch doesn't have support then it will need to be put in as a request on their bugtracker.

Did you try adding the product/vendor id to the controller configs for retroarch ?

```
input_vendor_id = XX
input_product_id = XX
```

This is a minor issue imho, as this is the first I have heard of someone with two devices, with the same name but different layouts. The easiest solution would just be to purchase another controller (vs the effort of sorting all this out).

danmcgoo:
Indeed by using the product/vendor id, it works in retroarch.

Then I guess adding those information to the generated cfg files would be welcome too ;)

By the way, I found this repository which is quite interesting and may help many people to avoid configuring their gamepad:
https://github.com/libretro/retroarch-joypad-autoconfig/tree/master/udev

I agree that this is not a common issue, but it was driving me mad because those 2 gamepads are the one I may use the most frequently (both for different reasons) and I did not want to reset the inputs everytime I switch the gamepads. It would be more annoying if I use both of them at the same time.

dankcushions:
i think this might be the same issue:
[...]

PR #2164, opened 2017-10-18
Fix for joy2key.py related to issue #1798.

danmcgoo:
This is a quick fix to issue #1798 for joy2key.py.
It could be improved if all the retroarch joypad *.cfg files include the vendor ID and the product ID.
It would be also nice to improve the retroarch joypad *.cfg file creation by including those IDs in the filename, like this:
<vendor_id>:<product_id>:<name>.cfg.
Since I have only few involvement with this project, I don't want to take the initiative to implement such a change. This is just an idea that might be helpful.

### DIFF
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -52,26 +52,39 @@ def get_btn_num(btn, cfg):
     if num: return num
     return ''
 
+def sysdev_get(key, sysdev_path):
+    value = ''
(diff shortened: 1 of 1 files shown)


joolswills:
ES would need changes to pass those details onto our code that generates the configs, or we could do it from the generation script which would work (so long as there was only one of the joysticks plugged in at once when configuring). Not a high priority for me as it's not that common a problem imho.

joolswills:
Thanks for the PR.

PR #2169, opened 2017-10-23
Autostart only on configured user

mioux:
Actually, the 10-autostart.sh script in profiles.d unconditionnaly runs kodi.

This versions tests the username which must be the same as the one used when installing RetroPie preventing kodi to start when switching to root, or emulationstation to start when switching user.

TODO: don't use username but uid, because if you rename your user, then you have to reinstall the script, as the `if` statement whould always be false.


mioux:
BTW (if needed), all the code I write is shared without conditions.

joolswills:
I see your point, but I would prefer to do it differently via the /etc/profile.d/10-retropie.sh - I will do a changeset for this.

PR #2170, opened 2017-10-23
scriptmodules: RPI: move optional SDL2 applications to new library names

psyke83:
The following optional scriptmodules use SDL2 but do not link GLES or EGL
libraries:

* advmame
* eduke32
* hatari
* love
* scummvm
* sdlpop
* solarus
* stella
* tyrquake
* vice
* zdoom

These packages are safe to switch to the new library names on jessie,
with the added benefit of fixing stretch support.

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -14,7 +14,7 @@ rp_module_desc="AdvanceMAME v3.5"
 rp_module_help="ROM Extension: .zip\n\nCopy your AdvanceMAME roms to either $romdir/mame-advmame or\n$romdir/arcade"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/amadvance/advancemame/master/COPYING"
 rp_module_section="opt"
-rp_module_flags="!mali"
+rp_module_flags="!mali newbrcmlibs"
 
 function _update_hook_advmame() {
     # if the non split advmame is installed, make directories for 0.94 / 1.4 so they will be updated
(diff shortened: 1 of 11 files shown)


psyke83:
None of the above link to the EGL/GLES libs, so the "newbrcmlibs" flag should be safe for jessie. This allows these modules to run correctly on stretch.

If you'd prefer me to split the PR into separate commits for each scriptmodule, please let me know.

OT: A few more projects in the optional category require -Wno-narrowing in order to build successfully on gcc 6 (pisnes, snes9x and zdoom). Will send PRs after I've tested a bit more.

joolswills:
They will work without the flag too right ? My plan was to put the flag in only for those that need it, then remove them when everything is done - so we only need the flag for stuff that actually requires it.

joolswills:
so if none of them link with gles, we should do nothing.

psyke83:
They will work without the flag on jessie, but not on stretch.

e.g.
zdoom errors out with:
"Could not create new screen (640 x 480)"
eduke32 stays on a black screen with 100% CPU usage.

They don't link with GLES or EGL, but the newbrcmlibs flag still affects it. For example, zdoom and eduke32 don't link to GLES, but clearly use GLES; it's initialized through libSDL2-2.0.so.0 (which is affected by the environment variables).

psyke83:
OK, if you intend to completely remove the flags later on, I'll just remove the commit adding the flags from my testing branch.

I can close this. You're still interested in adding gcc 6 compatibility for the few projects that are affected by the narrowing error?

joolswills:
Happy for the gcc6 patches. Thanks. (Unless they can be fed upstream).

joolswills:
I will consider removing the flag asap to fix squeeze. I believe I have rebuilt all the binaries now for Jessie.

psyke83:
The projects that require -Wno-narrowing are usually down to us using outdated forks - case in point, your retropie fork of zdoom. Upstream zdoom already builds OK (and includes your patch related to _FPU_GETCW), but the latest master also has an unrelated issue that causes it to fail to work. 

For zdoom, I plan to investigate the issue with the upstream master so that once it's working, you can consider switching to it (or rebasing your fork)... but I'm not sure whether you prefer to deliberately stay on a specific older revision.

joolswills:
Zdoom upstream is broken in other ways. Game engine issues and upstream dev was stopped so we rolled back.

PR #2172, opened 2017-10-24
Add more common filetypes for Sharp X68000

zerojay:
Including hard disk image support (hdf).

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -212,7 +212,7 @@ wonderswan_fullname="Wonderswan"
 wonderswancolor_exts=".7z .wsc .zip"
 wonderswancolor_fullname="Wonderswan Color"
 
-x68000_exts=".dim"
+x68000_exts=".dim .hdf .2hd .d88"
 x68000_fullname="X68000"
 
 zmachine_exts=".dat .zip .z1 .z2 .z3 .z4 .z5 .z6 .z7 .z8"


zerojay:
Might be best to wait until this gets resolved first: https://github.com/libretro/px68k-libretro/issues/38

zerojay:
@gingerbeardman If it interferes with user expectations and trying to run other games, of course it's an issue which is why this was filled. No user is going to expect that when they quit an emulator, any sort of state carries over to the next time they launch it, especially if it's going to make games they expect to work to stop working. No floppies should be "in the drive" when you are just starting up the machine if they aren't specified on the command line.

gingerbeardman:
I hear you, but it's complicated. 

right now if I switch of my physical machine with a floppy disk in it, said disk is still inserted when i come to power it up again. so the current behaviour is actually accurate to real life.

if all disks are ejected when the core is restarted then you'd have to reinsert the discs after a restart which in some cases would be annoying.

maybe the answer is ejecting all disks when the core is fully quit? would this break anything?

gingerbeardman:
i've changed this, watch out for it soon

zerojay:
@gingerbeardman I wasn't actually aware that you were a developer on the core. Thanks for hearing me out.

gingerbeardman:
You're welcome. Also, you might want to add more file extensions for PX68K. There's a nice list at https://www.reddit.com/r/miniSNESmods/comments/7tccrf/list_of_cores_from_kmfdmanic_set_their_command/

PR #2173, opened 2017-10-24
Fix mk_arcade_joystick module building for the wrong kernel

d-rez:
At least with the retropie image, the actual kernel isn't e.g. 4.9.35-v7+, but rather 4.9.35+. The line I changed never used the correct kernel name in use, hence it never worked in the first place and required a manual rebuild and installation every time

### DIFF
--- a/scriptmodules/supplementary/mkarcadejoystick.sh
+++ b/scriptmodules/supplementary/mkarcadejoystick.sh
@@ -34,7 +34,7 @@ function build_mkarcadejoystick() {
     if dkms status | grep -q "^mk_arcade_joystick"; then
         _dkms_remove_mkarcadejoystick
     fi
-    dkms install -m mk_arcade_joystick_rpi -v 0.1.5 -k "$(ls -1 /lib/modules | tail -n -1)"
+    dkms install -m mk_arcade_joystick_rpi -v 0.1.5 -k "$(uname -r)"
     if dkms status | grep -q "^mk_arcade_joystick"; then
         md_ret_error+=("Failed to install $md_id")
         return 1


joolswills:
Worked here last time I checked and my kernel seems to be `4.9.35-v7+` on my RetroPie image. I didn't use `uname -r` intentionally, as otherwise it get the wrong kernel for testing in my chroot dev environment.

d-rez:
To be honest I've had this issue with this particular kernel module build from retropie_setup ever since I discovered mk_arcade_joystick. Which was... half a year ago? I did a bunch of kernel update in between and every time neeeded a manual build. Weird.

I'm positive my initial image was an official retropie build.

Maybe there's a better way to get the kernel?

joolswills:
possibly, but in your case, are you sure you don't just have multiple kernel modules installed (so it's getting the wrong one) ? Maybe you can remove some old module dir ?

d-rez:
I'm positive. I've removed them with dkms already (tools/custombuild does it anyway). Just added another paragraph to my previous comment: 

The build process does display a warning about chroot / kernel version being off when I build it like that, but without it I can't load the module at all (modprobe says it can't find it)

joolswills:
I will do some more testing from a fresh image.

d-rez:
Thanks! Just for the record after the custom build the module appears in the list

```
pi@retropie:~/scripts $ dkms status
mk_arcade_joystick_rpi, 0.1.5, 4.9.35+, armv6l: installed
xpad, 0.4, 4.9.35-v7+, armv6l: installed (original_module exists)
```

d-rez:
@joolswills nice! Just tested by updating the module from GUI and it removed itself properly and then built and installed properly as well. 

Thanks for checking that and glad I could help :)

joolswills:
Thanks.

PR #2175, opened 2017-10-26
dxx-rebirth: use master branch

psyke83:
The unification/master branch is no longer recommended by upstream,
and the master branch has support for the new vendor libraries (as
well as the experimental Mesa driver).

### DIFF
--- a/scriptmodules/ports/dxx-rebirth.sh
+++ b/scriptmodules/ports/dxx-rebirth.sh
@@ -22,7 +22,7 @@ function depends_dxx-rebirth() {
 }
 
 function sources_dxx-rebirth() {
-    gitPullOrClone "$md_build" https://github.com/dxx-rebirth/dxx-rebirth "unification/master"
+    gitPullOrClone "$md_build" https://github.com/dxx-rebirth/dxx-rebirth "master"
 }
 
 function build_dxx-rebirth() {


joolswills:
Thanks. This is all tested? (ie master branch works and has same layout as the unification branch ?)

psyke83:
See this comment: See this comment: https://github.com/dxx-rebirth/dxx-rebirth/pull/351#issuecomment-339546493

I've built against the master branch on both jessie and stretch - it seems to run with no issues on my Pi 3.

If you're not comfortable with the mesa addition yet (even though it's still hard blocked), I can take that out.

joolswills:
Might be cleaner to add the mesa stuff later, as a new branch etc. But it's not a major issue.

psyke83:
I tested D1 & D2, albeit only using the shareware data that's installed by default. Works fine on both distros.

As for layout, the scons parameters are identical. The commit that was merged to master keeps the legacy option of "raspberrypi=1/true/yes" to build against the vendor driver, and "raspberrypi=mesa" to target VC4. The master branch still uses SDL1, so there's no complications in that regard.

psyke83:
I'll remove the mesa stuff, no prob. Just a moment.

vLKp:
> (ie master branch works and has same layout as the unification branch ?)

Branch `unification/master` was the main line of development for several years, but development moved to branch `master` since most people expect that to be the active line.  Branch `master` is directly descended from `unification/master`, which is now largely frozen.  All the big reorganization is long-since completed.  I don't anticipate making any backward-incompatible changes in the build system on either branch.

joolswills:
Thanks.

PR #2181, opened 2017-11-01
sdl1: allow building against distro-specific sources

psyke83:
Use different source versions for jessie/stretch, and automatically
bump package version with rpiX and os codename appended.

ex:
* 1.2.15-10 will bump to 1.2.15-11~rpi1~jessie
* 1.2.15+dfsg1-4 will bump to 1.2.15+dfsg1-5~rpi1~stretch

### DIFF
--- a/scriptmodules/supplementary/sdl1.sh
+++ b/scriptmodules/supplementary/sdl1.sh
@@ -16,7 +16,27 @@ rp_module_section=""
 rp_module_flags="!mali !x86"
 
 function get_pkg_ver_sdl1() {
-    echo "1.2.15-12rpi"
+    local basever
+    local revision
+    local rpi_revision
+
+    if compareVersions "$__os_release" ge 9; then
+        basever="1.2.15+dfsg1"
+        revision="4"
+        rpi_revision="rpi1"
+    else
+        basever="1.2.15"
+        revision="10"
+        rpi_revision="rpi1"
+    fi
+
+    if [[ "$1" == "source" ]]; then
+        echo "$basever-$revision"
+    elif [[ "$1" == "base" ]]; then
+        echo "$basever"
+    else
+        echo "$basever-$(expr $revision + 1)~$rpi_revision~$__os_codename"
+    fi
 }
 
 function depends_sdl1() {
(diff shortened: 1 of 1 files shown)


psyke83:
I asked about whether you'd want this a few days ago but I don't think you answered. Sending just in case it would be useful, otherwise it'll get overlooked.

The current rpi-related patch applies cleanly to both jessie and stretch,, so both packages build and appear to function normally. That could change in future, but I'm guessing it's fairly unlikely since SDL1 is in a legacy state.

joolswills:
Sorry I forgot. Will check and merge but may have a change or two. Thanks.

joolswills:
We might want to just use the stretch package on Jessie if there are no major dependency issues. Not we need to switch to the raspbian mirror for this over the upstream Debian packages though. I Would like to look into this myself, to see if there is a solution that requires less version wrangling.

joolswills:
I'll merge this for now, but may just switch to a single source version later if it backports easily enough. thanks.

psyke83:
I imagine the stretch source will work ok on jessie without breaking dependencies in other packages, since the base package version is the same, but I'd need to double-check to say that for sure.

By the way, have you noticed that stretch received a new firmware update? It was recently bumped to version 1.20171029 (which now has the pkgconfig in /opt/vc/lib/pkgconfig). The reason why I mention is that jessie's firmware has been neglected for several months. If the Foundation aren't committed to keeping their jessie repo maintained alongside stretch, perhaps your energy is better dedicated towards exclusively supporting stretch in future for RetroPie? The Raspbian repository itself is still churning out security updates, though, and stretch still presents some headaches when dealing with binary-only software that expect the original library names.

joolswills:
It's a shame if they are not going to update Jessie - especially as newer kernel updates include security patches etc.

RetroPie is aiming to move to Stretch as our default image btw (But we will continue to support Jessie).  Hence I will be rebuilding bins and doing some testing soon (I had some other things I wanted to finish first though like the tinker board stuff).

joolswills:
we should probably ask if they will be doing updates for Jessie. They didn't update the Jessie Kodi packages either - only stretch. I don't think it's a good idea for them to drop Jessie support right away, but of course, it's their choice.

psyke83:
Agreed, but I'm only speculating re: their intentions due to the delay in jessie firmware updates.

The idea of having a transition phase for projects to switch to the new library names also didn't really make sense, considering that nobody was aware of the issue until stretch suddenly stopped providing the libraries. If it's really necessary, it's possible to obtain the old library versions for the latest firmware versions via their https://github.com/raspberrypi/firmware/commits/master repo. Using "rpi-update" would also give those libraries on stretch, but it seems like a very messy solution.

joolswills:
Not sure I understand why we would need to do that?

psyke83:
Sorry, wasn't being clear (a bit exhausted lately). We shouldn't need to do anything with the old libraries, but some software you include via the script doesn't include source, like rpix86. If they're never updated, then stretch will not work unless we implement some kind of compatibility hacks.

Using symlinks might be ok or creating a dummy library that loads the brcm versions; it's just that ldconfig doesn't determine the library name from the filename, but maybe that won't matter.

joolswills:
aah ok. Yeah we will need to implement a workaround for them for stretch unless the authors rebuild them against the new libs.

PR #2183, opened 2017-11-02
Joy2Key-related fixes

psyke83:
Although you've re-arranged the start/stop invocations to avoid the process remaining in most cases, I thought you might want to consider switching to the USR1 signal to solve the orphaned process issue fully.

Additionally, I identified a few cases in which scripts will display a dialog without joy2key running, leaving users without a keyboard stuck. I considered adapting the scripts like so:

```
scriptdir=$scriptdir (needed for joy2keyStart to function; actual name should be replaced by sed or through substitution when creating new files)
source "$scriptdir/scriptmodules/helpers.sh"
joy2keyStart
dialog [...]
joy2keyStop
```
...but I think it might be a bit overkill, so I figured that changing these to pause dialogs with a 15 second timeout might work better.

Edit: the reicast warning is not visible with or without this PR. A different issue that needs checking, but it doesn't stall indefinitely with this PR, at least.

### DIFF
--- a/scriptmodules/emulators/fs-uae/fs-uae.sh
+++ b/scriptmodules/emulators/fs-uae/fs-uae.sh
@@ -21,7 +21,7 @@ kickfile="$biosdir/kick13.rom"
 source "$rootdir/lib/archivefuncs.sh"
 
 if [[ ! -f "$kickfile" ]]; then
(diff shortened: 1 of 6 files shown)


joolswills:
thanks.

meleu:
Can the changes of this PR be related with the issue reporte by @chewvodka (aka hansolo77) [in this forum post](https://retropie.org.uk/forum/post/120094)?

joolswills:
I don't see how.

PR #2184, opened 2017-11-02
Tinker board

joolswills:
initial support for the tinker board using SDL2 with drmkms video driver. Works both from X and from console. SDL1 apps are currently disabled, but could work under X.

Testing / feedback from devs welcome.

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -47,6 +47,7 @@ function sources_sdl2() {
     local branch="release-$ver"
     isPlatform "rpi" && branch="rpi-$ver"
     isPlatform "mali" && branch="mali-$ver"
+    isPlatform "kms" && branch="kms-$ver"
 
     gitPullOrClone "$md_build/$pkg_ver" https://github.com/RetroPie/SDL-mirror.git "$branch"
     cd "$pkg_ver"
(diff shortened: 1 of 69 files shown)


scriptmodules/supplementary/sdl2.sh
```diff
@@ -47,6 +47,7 @@ function sources_sdl2() {
     local branch="release-$ver"
     isPlatform "rpi" && branch="rpi-$ver"
     isPlatform "mali" && branch="mali-$ver"
+    isPlatform "kms" && branch="kms-$ver"
```
psyke83: No big deal yet, but if/when you add rpi vc4 support, considering that the rpi-2.0.x branches will have support for both rpi and kmsdrm, this change may cause problems *unless* you also plan to import all the rpi patches onto this branch as well.

joolswills: I will likely switch to the other branch - but we may well have the newer SDL by the time we add rpi vc4 support (I'm in no hurry for that), but not sure I understand why it will cause problems ? the kms flag for RPI will only be set if in kms mode, so then the RPI will get kms only version of SDL2 for now (but if switching back it would then get the rpi version)

psyke83: If VC4 becomes viable (when hardware acceleration, tvservice, HDMI-CEC support etc. is compatible), then this commit might not be optimal if you want to support the vendor/kms in one package for Pi, that's all.

When this patch is merged, the kms-2.0.x branch will be selected for Pi when the vc4 driver is enabled, but the current rpi-2.0.6 branch is more useful (as the kmsdrm and rpi drivers can coexist). I was suggesting only to re-arrange the kms line to come before rpi so that Pi always uses the rpi branch, and won't affect the Tinker build.

But, we don't know when, if ever, the vc4 driver will be in a usable state to be considered for default use, nor am I certain exactly what patches you intend to merge onto your kms branch (I'm just assuming that you'd want to avoid the dispmanx stuff). Never mind! ;)

joolswills: I may well switch it at that point - so it's not a issue for now. Other packages will also need to be switched so it's not a big deal if how we do it. I'll make a decision closer to the time.

PR #2192, opened 2017-11-07
Updating Dolphin extensions for GC and Wii

raelgc:
Like described in https://dolphin-emu.org/docs/faq/#what-dump-formats-are-supported-dolphin. I didn't include `.wbfs` for Gamecube because I believe it's a Wii only format (while not specified in the Wiki).


joolswills:
Thanks

Issue #2051, opened 2017-07-20
dosbox + fluidsynth


fogobogo:
This is mainly for documentation purposes since Retropie currently doesn't maintain a dosbox fork.

This patch for dosbox is derived from an earlier one here: [https://sourceforge.net/p/dosbox/patches/107/](https://sourceforge.net/p/dosbox/patches/107/)

Rationale:

The Raspberry Pi has somewhat sorry audio capabilities. Theres no support for dmix, there is no line in or mic and it all goes off via pulse wave modulation on top.
Playing multiple sound sources at once via alsa is almost impossible without a software mixer or more dedicated hardware.
Typically `pulseaudio` is used to provide software mixing but it introduces a fair amount of lag ( >100ms from the feel of it. ) and is quite resource intense.

What this patch does is adding fluidsynth as a dependency to dosbox and uses dosbox's internal mixing to provide midi output. It does so by adding it's own channel to the dosbox mixer.

This eliminates the need to have pulseaudio (as a software mixing daemon) and fluidsynth (as a midi software synthesizer daemon) both running and greatly reduces overhead.

Basically it allows midi + soundblaster output at the same time with just alsa

The patch is quite dirty and could need some work but is functional with a bit of fuzzing. I tested it on the current svn version (r4025 as of time of writing).

dosbox.conf must be changed to make use of it:

```
[midi]

mpu401=intelligent
mididevice=synth
midiconfig=/path/to/soundfont/example.sf2
```

NOTE:
There is also a samplerate setting but it seems to have little to no effect. The default seems to be using whatever dosbox is set to. Use`midisamplerate=44100` for example.

It also seems a bit picky about the kind of soundfont it accepts. Someone probably should try different ones.

[dosbox-fsynth.txt](https://github.com/RetroPie/RetroPie-Setup/files/1162695/dosbox-fsynth.txt)

fogobogo:
Addenum:
There is another more recent fluidsynth patch ([http://www.vogons.org/viewtopic.php?f=32&t=27831&start=20](http://www.vogons.org/viewtopic.php?f=32&t=27831&start=20)) for dosbox but this one doesn't integrate with dosbox's mixer. 
In consequence you don't have to run the fluidsynth daemon anymore but still need pulseaudio for the mixing with poor results.

psyke83:
[...]

PR #2203, opened 2017-11-15
dosbox: implement software MIDI synth support

psyke83:
* Configure DOSbox to use TiMidity/FluidSynth ALSA port for MIDI if a running
  service is detected during installation or if the target is a Pi.
* On Pi, explicitly spawn/kill a TiMidity daemon - but only if an existing
  TiMidity/FluidSynth instance is not already running.

Using timidity is safe on the Pi, as the bcm2835 audio codec supports hardware
mixing on up to 8 substreams. It may be usable on other targets, but we
need to ensure that either hardware mixing or dmix (ALSA software mixer)
is available to avoid sound blocking.

Alternative resolution to #2051.

### DIFF
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -17,7 +17,9 @@ rp_module_section="opt"
 rp_module_flags="dispmanx !mali !kms"
 
 function depends_dosbox() {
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

PR #2208, opened 2017-11-21
Retroarch v1.6.8 bump

LodanZark:
Retroarch v1.6.8 bump


joolswills:
Thanks, but we have some other things to do at the same time as bumping.

(I also like to test before, so prefer to handle retroarch updates myself).

joolswills:
(I also rebuild binaries before committing, so it's easier to do changes like this myself).

Issue #2091, opened 2017-08-17
Add Raspbian Stretch support


joolswills:
Notes for myself:

* Test gcc 6 compatibility
* rework SDL2 / ES and other GLES apps for new GLES library locations
* Rebuild all binaries for stretch
* Test new image based on stretch.
* Build retroarch against ffmpeg.

psyke83:
Good luck! I upgraded an image to stretch a day or so ago and had a lot of trouble getting emulationstation to compile properly (linker errors related to GL functions). Comparing the CMakeCache.txt, it seems that on stretch it wasn't using libraries and includes from /opt/vc related to GLES (files like /opt/vc/lib/libGLESv2_CM.so and its v1 symlink went missing after the upgrade).

I got it compiled eventually after installing various packages GLEW packages that weren't installed/necessary on jessie, but it still failed to launch with an SDL library initialization error.

joolswills:
yeah they no longer include non broadcom libs in libraspberrypi-dev so the ES cmake search paths need changing etc. the SDL2 library probably also needs changes. I will have to rework a bunch of stuff I guess.

gizmo98:
This will be fun. Without dispmanx support we can remove gpsp and enable display resolution switiching for rpi1/zero again. 

I tried to compile SDL2 with kms video backend so we do not need X11. It's a relatively new SDL2 feature. Commit is 17 days old. SDL2 compiles fine but EmulationStation fails at startup.
https://github.com/RetroPie/RetroPie-Setup/compare/master...gizmo98:rpi-kms?expand=1
https://github.com/spurious/SDL-mirror/compare/master...gizmo98:retropie-kms-2.0.5?expand=1

SDL2 kms docu:
https://github.com/spurious/SDL-mirror/commit/9afeb8f34f1877985e59f4506b05c276e72b5aff
https://github.com/sigmaris/SDL-mirror/commits/kmsdrm
https://forums.libsdl.org/viewtopic.php?p=51389

gizmo98:
I downloaded raspbian stretch today. /opt/vc libs is there and the old driver seems to be enabled by default.

gizmo98:
Ok. Emulationstation runs now from console. I can build Retroarch with kms support if i rename /opt/vc.

haggistech:
@gizmo98 did you use the lite version or the desktop version

gizmo98:
Desktop version.

psyke83:
[...]

PR #2209, opened 2017-11-21
retroarch - update to new version including modules that require new brcm library names - #2091

joolswills:
* include patches for lr-mupen64plus and lr-parallel-n64 until upstream accepts the PRs - https://github.com/libretro/mupen64plus-libretro/pull/39 / https://github.com/libretro/parallel-n64/pull/469

@psyke83 tagging you for review.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -33,7 +33,7 @@ function depends_retroarch() {
 }
 
 function sources_retroarch() {
-    gitPullOrClone "$md_build" https://github.com/libretro/RetroArch.git v1.6.7
+    gitPullOrClone "$md_build" https://github.com/libretro/RetroArch.git v1.6.9
     applyPatch "$md_data/01_hotkey_hack.diff"
     applyPatch "$md_data/02_disable_search.diff"
 }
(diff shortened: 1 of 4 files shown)


psyke83:
For now, I can confirm that retroarch, lr-parallel-n64 and lr-mupen64plus compile OK. Will be able to actually test later tonight.

Keep in mind that lr-mupen64plus uses audio/video plugins that use SDL2. Please don't forget to cherry-pick my patch to the rpi-2.0.7 branch to make sure jessie is stable.

psyke83:
Update: tested retroarch with lr-mupen64plus, lr-parallel-n64, and some other cores that aren't changed by this PR, and all seems to be working OK.

Maybe it's time to get rid of commit ee41a3f6138aca92ad63cb3179a3634298f1759a, as long as the SDL2 commit is added (as modules not using the flag will no longer have the library names force)? I'm not sure that it's preventing breakage any longer. If you don't want to get rid of that commit yet, RetroArch itself has an SDL2 video option which, although not the default, would necessitate the RP-NEWBRCMLIBS flag to be added to the retroarch scriptmodule.

joolswills:
Yep. Cheers. I plan to remove the flag soon.

joolswills:
Soon= probably right after I bud the bins and merge this.

joolswills:
Build even!

joolswills:
Quick question - why do you think your sdl2 patch is required to remove the NEWBRCMLIBS flag ?

PR #2211, opened 2017-11-22
retropiemenu: quit joy2key for RetroArch case

psyke83:
P.S. The small interval between start/stop should no longer matter now that the python is killed before script processing completes, via the USR1 signal. I've tested several times and never saw an orphaned process before exiting retroarch, but even so, the second invocation of joy2keyStop after exiting retroarch would probably catch it.


joolswills:
Thanks. :-) including your joy2key improvements.

PR #2212, opened 2017-11-23
Atari800 libretro core

joolswills:


joolswills:
Seems to work well from my initial testing and autoloaded a disk perfectly. Am placing directly in main, as the 8bit atari system is popular (and the libretro core is easier to configure than the standalone emulator - just needs the BIOS files installed).

zerojay:
Where do the BIOS files need to be for this? I had previously installed Atari800 and it was working fine with 5200.rom in the BIOS directory however it appears that lr-atari800 isn't seeing it.

PR #2214, opened 2017-11-27
Implement configKernelModule helper

psyke83:
Apologies for dumping this all of a sudden. @joolswills and I briefly discussed the possibility of adding software midi daemon function to runcommand for use with AGS (and dosbox), but I don't think he was happy with the idea :). Instead of doing that, I can just add the daemon to a script for AGS, just as DOSbox currently does. The problem, however, is that AGS requires a raw midi device in order to function, so a non-standard kernel module needs to be loaded in order for it to work.

Since AGS will need a kernel to be loaded with custom parameters for MIDI support, I've written a new helper function to manage install & configuration of kernel modules. I tried to keep it as concise as possible, but I've commented the important lines to help illustrate its function.

It adds some extra bloat to the code, but makes module management more consistent between the projects, has better systemd support, and does a better job at cleaning up on uninstall. I'm only sending the PR to see if you would find this acceptable for inclusion; depending on your decision, my AGS PR (which I still need time to work on) will either utilize this function, or else will have to re-use the same code that's duplicated between gamecondriver/xpad etc. for configuring the required module (snd-virmidi).

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -115,6 +115,7 @@ function addLineToFile() {
     if [[ -f "$2" ]]; then
         cp -p "$2" "$2.bak"
     else
+        touch "$2"
         sed -i --follow-symlinks '$a\' "$2"
     fi
 
@@ -129,6 +130,57 @@ function editFile() {
     local cmd=(dialog --backtitle "$__backtitle" --editbox "$file" 22 76)
     local choice=$("${cmd[@]}" 2>&1 >/dev/tty)
     [[ -n "$choice" ]] && echo "$choice" >"$file"
+}
+
+## @fn configKernelModule()
+## @brief mode configuration mode (install or remove)
+## @brief module module name to be added to /etc/modules-load.d file
+## @param prefix prefix name of .conf files in /etc/modprobe.d & /etc/modules-load.d
+## @param parameter module parameter to be added to /etc/modprobe.d file
+function configKernelModule() {
+    local mode="$1"
+    local module="$2"
+    local prefix="$3"
+    local parameter="$4"
+
+    [[ -z "$prefix" ]] && prefix="$module"
+
+    local parameter_file="/etc/modprobe.d/$prefix.conf"
+    local module_file="/etc/modules-load.d/$prefix.conf"
+
+    # SysVinit legacy compatibility
+    [[ ! -d "/etc/modules-load.d/" ]] && module_file="/etc/modules"
+
+    if [[ "$mode" == "remove" ]]; then
+        # remove all entries matching module name from configs
+        touch "$module_file" "$parameter_file"
+        sed -i --follow-symlinks "/$module\b/d" "$module_file" "$parameter_file"
+        # delete empty config files (if necessary)
+        [[ ! -s "$module_file" && -d "/etc/modules-load.d/" ]] && rm -f "$module_file"
+        [[ ! -s "$parameter_file" ]] && rm -f "$parameter_file"
+        # rmmod module (if necessary)
+        [[ -n "$(lsmod | grep $module)" ]] && rmmod "$module"
+        return
+    fi
+
+    # add parameter to config
+    [[ -n "$parameter" ]] && if ! grep -qs "$parameter" "$parameter_file"; then
+        # if a new parameter is specified, remove all lines referencing this module
+        [[ -f "$parameter_file" ]] && sed -i --follow-symlinks "/$module\b/d" "$parameter_file"
+        addLineToFile "$parameter" "$parameter_file"
+    fi
+
+    # remove legacy entries from /etc/modules (only if using systemd)
+    [[ -d "/etc/modules-load.d/" ]] && sed -i --follow-symlinks "/$module\b/d" /etc/modules
+
+    # add module to config
+    if ! grep -qs "$module" "$module_file"; then
+        addLineToFile "$module" "$module_file"
+    fi
+
+    # rmmod (if needed) & modprobe
+    [[ -n "$(lsmod | grep $module)" ]] && rmmod "$module"
+    modprobe "$module"
 }
 
 ## @fn hasPackage()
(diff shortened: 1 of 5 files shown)


scriptmodules/helpers.sh
```diff
@@ -131,6 +131,56 @@ function editFile() {
...
+
+    local parameter_file="/etc/modprobe.d/$prefix.conf"
+    local module_file="/etc/modules-load.d/$prefix.conf"
+
+    # SysVinit legacy compatibility
+    [[ ! -d "/etc/modules-load.d/" ]] && module_file="/etc/modules"
+
+    if [[ "$mode" == "remove" ]]; then
+        # remove all entries matching module name from configs
+        sed -i "/$module\b/d" "$module_file" "$parameter_file" 2>/dev/null
```
joolswills: do we need to redirect stderr ?

scriptmodules/helpers.sh
```diff
@@ -131,6 +131,56 @@ function editFile() {
...
+        # delete empty config files (if necessary)
+        [[ ! -s "$module_file" && -d "/etc/modules-load.d/" ]] && rm -f "$module_file"
+        [[ ! -s "$parameter_file" ]] && rm -f "$parameter_file"
+        # rmmod module (if necessary)
+        [[ -n "$(lsmod | grep $module)" ]] && rmmod "$module"
+        return
+    fi
+
+    # add parameter to config
+    if ! grep -qs "$parameter" "$parameter_file" && [[ -n "$parameter" ]]; then
```
joolswills: if `[[ -n parameter` would be better first. else you do the grep even if the parameter is empty.

scriptmodules/helpers.sh
```diff
@@ -131,6 +131,56 @@ function editFile() {
...
+        # if a new parameter is specified, remove all lines referencing this module
+        sed -i "/$module\b/d" "$parameter_file" 2>/dev/null
+        addLineToFile "$parameter" "$parameter_file" 2>/dev/null
```
joolswills: as above regarding redirection.

scriptmodules/helpers.sh
```diff
@@ -131,6 +131,56 @@ function editFile() {
...
+## @fn configKernelModule()
+## @brief mode configuration mode (install or remove)
+## @brief module module name to be added to /etc/modules-load.d file
+## @param parameter module parameter to be added to /etc/modprobe.d file
+## @param prefix prefix name of .conf files in /etc/modprobe.d & /etc/modules-load.d
+function configKernelModule() {
+    local mode="$1"
+    local module="$2"
+    local parameter="$3"
+    local prefix="$4"
```
joolswills: perhaps parameter would be better last.

scriptmodules/supplementary/mkarcadejoystick.sh
```diff
@@ -48,24 +48,16 @@ function build_mkarcadejoystick() {
 }
 
 function remove_mkarcadejoystick() {
-    [[ -n "$(lsmod | grep mk_arcade_joystick_rpi)" ]] && rmmod mk_arcade_joystick_rpi
+    configKernelModule "remove" "mk_arcade_joystick_rpi" "" "mk_arcade_joystick_rpi"
```
joolswills: are the last two parameters needed in this case ?

scriptmodules/helpers.sh
```diff
@@ -115,6 +115,7 @@ function addLineToFile() {
     if [[ -f "$2" ]]; then
         cp -p "$2" "$2.bak"
     else
+        touch "$2"
         sed -i --follow-symlinks '$a\' "$2"
```
joolswills: I think I may have introduced an error in this function as the sed is only being run if the file doesn't exist - which is wrong. It should be run if it does exist  (to make sure it ends with a LF).

joolswills: I can fix it after the PR anyway - the fix should be done in a separate commit from this function addition etc.

scriptmodules/helpers.sh
```diff
@@ -131,6 +132,57 @@ function editFile() {
...
+
+    if [[ "$mode" == "remove" ]]; then
+        # remove all entries matching module name from configs
+        touch "$module_file" "$parameter_file"
+        sed -i --follow-symlinks "/$module\b/d" "$module_file" "$parameter_file"
+        # delete empty config files (if necessary)
+        [[ ! -s "$module_file" && -d "/etc/modules-load.d/" ]] && rm -f "$module_file"
+        [[ ! -s "$parameter_file" ]] && rm -f "$parameter_file"
+        # rmmod module (if necessary)
+        [[ -n "$(lsmod | grep $module)" ]] && rmmod "$module"
```
joolswills: you can do `lsmod | grep -q "$module" && rmmod "$module"`

scriptmodules/helpers.sh
```diff
@@ -131,6 +132,57 @@ function editFile() {
...
+
+    # SysVinit legacy compatibility
+    [[ ! -d "/etc/modules-load.d/" ]] && module_file="/etc/modules"
+
+    if [[ "$mode" == "remove" ]]; then
+        # remove all entries matching module name from configs
+        touch "$module_file" "$parameter_file"
+        sed -i --follow-symlinks "/$module\b/d" "$module_file" "$parameter_file"
+        # delete empty config files (if necessary)
+        [[ ! -s "$module_file" && -d "/etc/modules-load.d/" ]] && rm -f "$module_file"
```
joolswills: Not sure about this logic - I don't think it's too clear (and checking if the dir exists before removing module file etc)

scriptmodules/helpers.sh
```diff
@@ -131,6 +132,57 @@ function editFile() {
...
+        [[ ! -s "$parameter_file" ]] && rm -f "$parameter_file"
+        # rmmod module (if necessary)
+        [[ -n "$(lsmod | grep $module)" ]] && rmmod "$module"
+        return
+    fi
+
+    # add parameter to config
+    [[ -n "$parameter" ]] && if ! grep -qs "$parameter" "$parameter_file"; then
```
joolswills: I meant `if [[ -n "$parameter" ]] && grep -qs "$parameter" "$parameter_file"; then`

PR #2216, opened 2017-11-30
mupen64plus: Pi: temporarily workaround texture rendering/segfault issue

psyke83:
May be reverted later - see: https://github.com/gonetz/GLideN64/issues/1665


psyke83:
Confirmed to resolve the texture issue (only tested on stretch but I don't imagine issues with jessie).

BTW - I can see a black/semi-transparent(?) FPS counter. It's caused by this:
```
            # Enable FPS Counter. Fixes zelda depth issue
            iniSet "ShowFPS " "True"
            iniSet "fontSize" "14"
            iniSet "fontColor" "1F1F1F"
```
Is this workaround still necessary or can we get rid of it?

gizmo98:
FPS Counter is necessary to fix depth issues with fb emulation.

joolswills:
thanks.

joolswills:
(rebuilding bins)

PR #2220, opened 2017-12-09
usbromservice - support overlayfs of usb stick over ~/RetroPie

bit4bit:
hi, thanks for your work,
this add support overlayfs from usb dir **retropie-overlayfs** over **~/RetroPie** allowing more roms, BIOS, etc.. and keeping sensitive information -saves,history,etc..- in home directory SD.

how works:
  * create on usb stick a directory **retropie-overlayfs**
  * connect usb stick, this auto populate with initial dirs
  * plug to PC copy roms,BIOS etc..
  * plug again to RetroPie and restart emulationstation
  * play


scriptmodules/supplementary/usbromservice.sh
```diff
@@ -41,10 +41,19 @@ function install_bin_usbromservice() {
 
 function install_scripts_usbromservice() {
     # copy our mount.d scripts over
+    local filesrc
     local file
     local dest
-    for file in "$md_data/"*; do
-        dest="/etc/usbmount/mount.d/${file##*/}"
+    for file in "$md_data/"*.mount"; do
+	filesrc=${file%%.mount}
```
joolswills: wrong indent.

scriptmodules/supplementary/usbromservice.sh
```diff
@@ -41,10 +41,19 @@ function install_bin_usbromservice() {
...
+        dest="/etc/usbmount/mount.d/${filesrc##*/}"
+        sed "s/USERTOBECHOSEN/$user/g" "$file" >"$dest"
+        chmod +x "$dest"
+    done
+
+    for file in "$md_data/"*.umount"; do
+	filesrc=${file%%.umount}
+        dest="/etc/usbmount/umount.d/${filesrc##*/}"
```
joolswills: why the renaming to .mount - I would have though a single copy loop would be possible for copying all files ?

scriptmodules/supplementary/usbromservice/20_retropie_overlayfs.mount
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
joolswills: these scripts should use the default shell and not rely on bash.

joolswills: Sorry - ignore this. It's ok to use bash (and we use it already - although possibly we shouldn't for this).

joolswills:
I had a quick look over. Made a few comments. 

I think it might be best to incorporate this into the existing mount script rather than having more duplication of functions etc.

scriptmodules/supplementary/usbromservice.sh
joolswills: I see the reasoning, but it might be better to just use a different structure to simplify. Is there much use in the umount script ?

joolswills:
Closing. May consider in the future but would need implementing a bit differently as mentioned. Thanks.

PR #2222, opened 2017-12-17
Revert "mupen64plus: Pi: temporarily workaround texture rendering/segfault issue"

psyke83:
Reverts RetroPie/RetroPie-Setup#2216

No longer needed due to changes committed to GlideN64 master branch.


johnmanko:
When is the next release?

PR #2223, opened 2017-12-17
Fix to enable compilation on ubuntu 17.10

Neo-Rio:
As newer versions of GCC have stack protection turned on by default, this change is necessary. If stack protection is on by default, the resulting binary will crash with stack protection errors. By setting stack protection off, then advmame will compile and work without issue.

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -37,7 +37,7 @@ function sources_advmame() {
 }
 
 function build_advmame() {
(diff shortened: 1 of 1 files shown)


joolswills:
Ideally this would be submitted upstream. We can include it in RetroPie until it is, but better the fix goes there also. Thanks.

joolswills:
The above change will override our existing cflags btw so it needs to be don't slightly differently.

PR #2226, opened 2017-12-19
ags: add to platforms & enable DIGMID support

psyke83:
Add Adventure Game Studio to platforms configuration and fetch
recommended GUS patch from http://liballeg.org/digmid.html.

Fixes MIDI playback on devices with no hardware MIDI support (such as
Raspberry Pi).

--

It's not easy to determine which games use MIDI sound, so to help test this PR, I recommend downloading this: http://www.g4g.it/2009/11/30/fullyramblomatic-special-editions-pack/. I can confirm that "5 Days Special Edition" plays midi sound on the title menu.

If you'd like to mirror the GUS patch file, let me know and I can update the PR once you've uploaded the file to the retropie server. The current "digmid.dat" file is actually bz2 compressed, which is why "downloadAndExtract" is not possible unless we mirror a properly compressed & named archive.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -1,6 +1,9 @@
 3do_exts=".iso"
 3do_fullname="3DO Interactive Multiplayer"
 
+ags_exts=".exe"
+ags_fullname="Adventure Game Studio"
+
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks!

Issue #1350, opened 2016-03-07
Enhancement for DosBox: SDL2, DynRec


jimbo27:
I found an enhancement for DosBox. There is a patch which provides SDL2 which should give some speed improvements when using SDL2.0.4.
In addition enabling DynRec should give additional speed according to this thread:

https://www.raspberrypi.org/forums/viewtopic.php?f=78&t=137974

> The patch you need is here:
> http://www.vogons.org/viewtopic.php?t=34770
> 
> Go to the last pages for most recent version.
> You caan apply it like this on the dosbox sources dir:
> patch -p0 < ../dosbox_sdl2_20150713.diff
> 
> That will give you SDL2 support. NOTE you must have previously built and installed SDL 2.0.4 since it's not yet available on Raspbian repositories!
> Then you can enable dynarec support by setting these like this in config.h:
> # define C_DYNREC 1
> # define C_TARGETCPU ARMV4LE
> # define C_UNALIGNED_MEMORY 1
> 
> Then, set cpu type to dynamic in your .conf file, and you can easily aim for a 2000 cycles system without worrying about sound dropouts. :)

http://www.vogons.org/viewtopic.php?f=32&t=34770&start=56

Perhaps this is interesting enough to give it a try...

joolswills:
we already have DYNREC enabled.

I don't think there will be any speed improvements using sdl2 - but it would be better to move to it - I will have a look.

amigagamer:
This is an old thread but should be useful.  See post 4 regarding the scaling settings for sdl2
[https://www.raspberrypi.org/forums/viewtopic.php?f=78&t=110957&p=761752#p761752](url)

Looking at my working config file from sdl2 build (r3910 may 2015) i recall lots of time spent trying to get sdl2 to handle the scaling.  To get this working i used the following settings in dosbox.conf, although keep in mind this was a 12 month old svn/version of the patch.  Hopefully the information is useful!

[sdl]
fullscreen=true
vsync=false
fullresolution=0x0
windowresolution=original
output=texture

[render]
frameskip=0
aspect=false
scaler=none

psyke83:
There's now an experimental dosbox-sdl2 scriptmodule - I think we can close this issue.

PR #2231, opened 2017-12-24
Update dosbox & add experimental sdl2-only version

psyke83:
Testing a newer revision of the SDL2 patch as discussed previously: https://github.com/RetroPie/RetroPie-Setup/issues/1350

The new revision of vanilla DOSbox seems to work OK. I decided to update it so that we have an equal baseline for comparison with the SDL2 version. I also added some minor cleanups and a fix to ensure that the midi synth is launched for .sh script content.

I've done some testing of the SDL2 build - performance is almost identical, but there is a noticeable reduction in screen tearing, better scaling to fit the entire screen, and mapper colours are OK. The splash screen, windowed mode and the mapper doesn't scale to the full screen properly, and while SDL_VIDEO_RPI_SCALE_MODE=1 does scale the window, the mouse cursor's bounding appears limited to the unscaled resolution, making it impossible to navigate on parts of the screen, so I left that out. I'm guessing that the dispmanx patch in the SDL2 source would need updating so that the cursor bounds reflect the scaled window size instead of the original.

I'm sending the PR since you already expressed interest in testing SDL2, but I've no expectation for anyone to even so much as take a look until after the holidays :).

### DIFF
--- a/scriptmodules/emulators/dosbox-sdl2.sh
+++ b/scriptmodules/emulators/dosbox-sdl2.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="dosbox-sdl2"
+rp_module_desc="DOS emulator (enhanced DOSBox fork)"
+rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
+rp_module_licence="GPL2 https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/trunk/COPYING"
+rp_module_section="exp"
+rp_module_flags="!mali !kms"
+
+function depends_dosbox-sdl2() {
+    depends_dosbox libsdl2-dev libsdl2-net-dev libfluidsynth-dev fluid-soundfont-gm
+}
+
+function sources_dosbox-sdl2() {
+    gitPullOrClone "$md_build" "https://github.com/duganchen/dosbox.git"
+    # use custom config filename & path to allow coexistence with regular dosbox
+    sed -i "src/misc/cross.cpp" -e 's/~\/.dosbox/~\/.'$md_id'/g' \
+       -e 's/DEFAULT_CONFIG_FILE "dosbox-"/DEFAULT_CONFIG_FILE "'$md_id'-"/g'
+}
+
+function build_dosbox-sdl2() {
+    build_dosbox
+}
+
+function install_dosbox-sdl2() {
+    install_dosbox
+}
+
+function configure_dosbox-sdl2() {
+    configure_dosbox
+    if [[ "$md_mode" == "install" ]]; then
+        local config_path=$(su "$user" -c "\"$md_inst/bin/dosbox\" -printconf")
+        if [[ -f "$config_path" ]]; then
+            iniConfig "=" "" "$config_path"
+            iniSet "fluid.driver" "alsa"
+            iniSet "fluid.soundfont" "/usr/share/sounds/sf2/FluidR3_GM.sf2"
+            iniSet "fullresolution" "desktop"
+            iniSet "fullscreen" "true"
+            iniSet "mididevice" "fluidsynth"
+            iniSet "output" "texture"
+            iniDel "usescancodes"
+        fi
+    fi
+}
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -59,11 +59,17 @@ function install_dosbox() {
 }
 
 function configure_dosbox() {
-    local def="1"
-    local launcher_name="+Start DOSBox.sh"
-    # needs software synth for midi; limit to Pi for now
-    if isPlatform "rpi"; then
-        local needs_synth="1"
+    if [[ "$md_id" == "dosbox-sdl2" ]]; then
+        local def="0"
+        local launcher_name="+Start DOSBox-SDL2.sh"
+        local needs_synth="0"
+    else
+        local def="1"
+        local launcher_name="+Start DOSBox.sh"
+        # needs software synth for midi; limit to Pi for now
+        if isPlatform "rpi"; then
+            local needs_synth="1"
+        fi
     fi
 
     mkRomDir "pc"
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -11,24 +11,31 @@
 
 rp_module_id="dosbox"
 rp_module_desc="DOS emulator"
-rp_module_help="ROM Extensions: .bat .com .exe .sh\n\nCopy your DOS games to $romdir/pc"
+rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
 rp_module_licence="GPL2 https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/trunk/COPYING"
 rp_module_section="opt"
 rp_module_flags="dispmanx !mali !kms"
 
 function depends_dosbox() {
-    local depends=(libsdl1.2-dev libsdl-net1.2-dev libsdl-sound1.2-dev libasound2-dev libpng12-dev automake autoconf zlib1g-dev)
+    local depends=(libsdl1.2-dev libsdl-net1.2-dev libsdl-sound1.2-dev libasound2-dev libpng12-dev automake autoconf zlib1g-dev subversion "$@")
     isPlatform "rpi" && depends+=(timidity freepats)
     getDepends "${depends[@]}"
 }
 
 function sources_dosbox() {
-    downloadAndExtract "$__archive_url/dosbox-r3876.tar.gz" "$md_build" 1
+    local revision="$1"
+    [[ -z "$revision" ]] && revision="4006"
+
+    svn checkout https://svn.code.sf.net/p/dosbox/code-0/dosbox/trunk "$md_build" -r "$revision"
 }
 
 function build_dosbox() {
     local params=()
+
     ! isPlatform "x11" && params+=(--disable-opengl)
+    # add or override params from calling function
+    params+=("$@")
+
     ./autogen.sh
     ./configure --prefix="$md_inst" "${params[@]}"
     if isPlatform "arm"; then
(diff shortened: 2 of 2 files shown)


scriptmodules/emulators/dosbox-sdl2.sh
```diff
@@ -0,0 +1,53 @@
...
+rp_module_desc="DOS emulator"
+rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
+rp_module_licence="GPL2 https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/trunk/COPYING"
+rp_module_section="exp"
+rp_module_flags="dispmanx !mali !kms"
+
+function depends_dosbox-sdl2() {
+    local depends=(libsdl2-dev libsdl2-net-dev libasound2-dev libpng12-dev automake autoconf zlib1g-dev subversion)
+    isPlatform "rpi" && depends+=(timidity freepats)
+    getDepends "${depends[@]}"
```
joolswills: can probably do some depends sharing with main dosbox package ?

scriptmodules/emulators/dosbox-sdl2.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function sources_dosbox-sdl2() {
+    svn checkout https://svn.code.sf.net/p/dosbox/code-0/dosbox/trunk "$md_build" -r 4006
```
joolswills: Also can use the sources_dosbox ?

joolswills:
the patch is very large. I wonder if it might be worth having a fork with the patch applied rather than including it in retropie-setup.

joolswills:
Another note is that we used an archive rather than direct svn access originally due to the instability of sourceforge (it went down a number of times, although has been better recently).

We use sourceforge directly for vice, but thought I would mention it. If they go down again, we will need to switch it back to an archive etc.

scriptmodules/emulators/dosbox-sdl2.sh
```diff
@@ -0,0 +1,53 @@
...
+function configure_dosbox-sdl2() {
+    configure_dosbox
+
+    if [[ "$md_mode" == "install" ]]; then
+        local config_path=$(su "$user" -c "\"$md_inst/bin/dosbox\" -printconf")
+        if [[ -f "$config_path" ]]; then
+            iniConfig "=" "" "$config_path"
+            iniSet "fullscreen" "true"
+            iniSet "fullresolution" "desktop"
+            sed -i "/usescancodes=false/d" "$config_path"
```
joolswills: could use iniDel ?

scriptmodules/emulators/dosbox.sh
```diff
@@ -98,24 +104,25 @@ midi_synth start
...
+        chown $user:$user "$romdir/pc/$launcher_name"
+
+        local config_path=$(su "$user" -c "\"$md_inst/bin/dosbox\" -printconf")
+        if [[ -f "$config_path" ]]; then
+            iniConfig "=" "" "$config_path"
+            iniSet "usescancodes" "false"
+            iniSet "core" "dynamic"
+            iniSet "cycles" "max"
+            iniSet "scaler" "none"
+            if isPlatform "rpi" || [[ -n "$(aconnect -o | grep -e TiMidity -e FluidSynth)" ]]; then
```
joolswills: unrelated to this changeset, but what's the logic here with checking for rpi or the output of aconnect ? Since the depends only installs timidity for the rpi etc.

joolswills: Also the fact there is a check for this with variable needs_synth ?

joolswills: ignore that - that's in the launch script. sorry.

scriptmodules/emulators/dosbox.sh
```diff
@@ -52,14 +52,18 @@ function install_dosbox() {
 }
 
 function configure_dosbox() {
+    local def=1
+    local launcher_name="+Start DOSBox.sh"
     local needs_synth
+    [[ "$md_id" == "dosbox-sdl2" ]] && launcher_name="+Start DOSBox-SDL2.sh" && def=0
```
joolswills: We don't really do multiple variable setting like this elsewhere - would be more readable with an if/fi

joolswills:
Thanks for the changes. Made a few comments - See you the other side of xmas.

scriptmodules/emulators/dosbox-sdl2.sh
psyke83: Yes, but the patch actually strips out SDL1 support, so there would be some unnecessary depends pulled into the mix (not really an issue unless you don't have vanilla dosbox installed).

scriptmodules/emulators/dosbox-sdl2.sh
psyke83: LOL, didn't think of that for some reason. Will change it.

scriptmodules/emulators/dosbox.sh
psyke83: We set our daemon only on Pi for now, as it might be risky to enable on platforms that don't have some kind of working ALSA mixing (dmix or hardware mixing). However, the check above is to configure the ALSA midi port if there is an existing daemon running that's installed by the user (e.g. timidity-daemon).

psyke83:
I'll update the PR with the changes you specified, but I'll leave the subversion stuff as-is until the last step. You can let me know if you want to distribute a tarball and/or start a new repo, and when everything else is sorted, I'll change the PR to point to the new tarball or git repo. It's just easier to point to working (svn) source for the moment since I don't have the means to upload tarballs to your server. Cheers & have a nice holiday :)

scriptmodules/emulators/dosbox-sdl2.sh
psyke83: Actually, I remember why I didn't do it. It might be a bad idea to share the function as it will lock us into r4006 for the vanilla port. We need to stick to that revision for the SDL2 patch to apply cleanly, but newer SVN builds actually works fine on Pi, and we might want to update the vanilla port later.

psyke83:
I've updated the PR with most of your suggestions. I've improved the support for variants by allowing more custom variables such as the build flags, even though dosbox-sdl2 currently doesn't need it. Same idea with sharing of depends and svn revision so that you don't need to stay locked to r4006 on the vanilla scriptmodule.

I've not noticed any issues with SVN connectivity, so perhaps the SourceForge's servers have been improved? Their web UI certainly hasn't, though. I'll leave it to your discretion whether you want to go back to a tarball for the source, store your patch on the server, or start a new repo for the latter. I can do the work of creating the repo if that's what you want. Importing the proper svn history would be tricky; IIRC, svn repos don't store author email addresses and other small details that make it a bit annoying to import into a proper git commit history, but we can just be lazy and commit vanilla r4006 in a big chunk ;). Thanks and no hurry to re-review - svn situation still needs deciding so, the PR may not be final.

joolswills:
It's not general connectivity but sourceforge went down multiple times for long periods in the past which is why we self hosted some projects.

Setting up a git mirror of svn is quite simple btw - I have done it for other projects. I may do this and commit the patch so please don't merge this just yet. thanks.

joolswills:
BTW regarding your original comments in this PR - if you enable dispmanx sdl1 driver for dosbox there is no screen tearing.

The mouse cursor issue may be due to the sdl2 implementation for dosbox rather than the sdl2 rpi driver, but i'd have to check further.

psyke83:
Don't worry, I won't merge something big like a scriptmodule introducing a new port without your consent. Actually, I'll temporarily close the PR, as I want to test some alternative forks. The author of the SDL2-only patch maintains a fork on github that also works fine on the Pi (and has FluidSynth support for MIDI), so it might make sense to use that build rather than creating our own fork based on a snapshot of his work. If it turns out to be a useful build we can contribute fixes back. 

I'll also look at the dispmanx sdl1 mode for vanilla, but I think it doesn't diminish the other possible benefits of having an SDL2 build available alongside the official SVN build. Will reopen/update the PR when I'm satisfied that such a build is worth consideration.

psyke83:
I've updated the PR so that dosbox-sdl2 is sourced from: https://github.com/duganchen/dosbox

This fork is based on r4006 with the main changes being SDL2 + FluidSynth support. I've updated the scriptmodule to pull in the FluidR3 GM soundfont and configure dosbox-sdl2 to use it. The soundfont is a fairly large package, but I'm guessing that anybody who wants to try the build will want to have it, as it has a better range of instruments than freepats via timidity.

P.S. I tested dosbox using the dispmanx driver - it's definitely improved, but IMO, this port has better fullscreen scaling than it.

psyke83:
Oh, and I'm happy to take off the subversion stuff in the original module. I made a test import of the svn repo: https://github.com/psyke83/dosbox

It seemed to import fine, but my concern would be how to rebase against newer changes (i.e. to re-import newer changes) without having to recreate the repo or do a force-push and mess up local copies. I've never bothered with mirroring a svn repo so I'm a bit clueless about this.

joolswills:
can use `git svn` etc.

scriptmodules/emulators/dosbox-sdl2.sh
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function install_dosbox-sdl2() {
+    install_dosbox
+}
+
+function configure_dosbox-sdl2() {
+    local def=0
+    local launcher_name="+Start DOSBox-SDL2.sh"
+    local needs_synth=0
```
joolswills: I would either do the checks in the configure_dosbox to set these variables or add some parameters rather than using local variable scope and then using that in configure_dosbox - it's easier to follow.

psyke83:
Just to note that I decided to keep the string replace, but use the bash builtin regex instead of sed. I removed the passing of local scope variables, and instead use arguments for the depends and build functions. As for the configure function, I moved the code into a conditional block of configure_dosbox 
 as you suggested. Hopefully it's OK now, apart from the decision re: SVN. All I can say is that the sourceforge servers haven't been slow or failed to checkout during my testing.

joolswills:
The sourceforge issue were not related to it being slow/failing occasionally. It was due to long periods of downtime on sourceforge - they have had many problems over the years.

joolswills:
thanks.

zerojay:
By the way, this new version installs a new conf file for the sdl2 version but that new conf file is never used and instead the SDL2 version defaults to the dosbox-SVN.conf instead.

psyke83:
@zerojay,

Are you sure? It looks OK to me.
```
$ strings -f /opt/retropie/emulators/dosbox*/bin/dosbox | grep SVN.conf
/opt/retropie/emulators/dosbox/bin/dosbox: dosbox-SVN.conf
/opt/retropie/emulators/dosbox-sdl2/bin/dosbox: dosbox-sdl2-SVN.conf
```
dosbox saves its config to ~/.dosbox/dosbox-SVN.conf and dosbox-sdl2 saves to ~/.dosbox-sdl2/dosbox-sdl2-SVN.conf. Both of those locations are symlinks to /opt/retropie/configs/pc/.

zerojay:
When I launch dosbox-sdl2, I can see it loads up ~/.dosbox/dosbox-SVN.conf as its config file.

psyke83:
I've just re-tested a fresh install of dosbox-sdl2 on my jessie image, and I can't reproduce the problem. Please post your build log or at least the runcommand.log. Mine is working as expected:
```
Parameters:
Executing: bash "/home/pi/RetroPie/roms/pc/+Start DOSBox-SDL2.sh"
fluidsynth: warning: Requested a period size of 64, got 256 instead
fluidsynth: warning: Failed to set thread to high priority
fluidsynth: warning: Failed to pin the sample data to RAM; swapping is possible.
DOSBox version SVN
Copyright 2002-2015 DOSBox Team, published under GNU GPL.
---
CONFIG:Loading primary settings from config file /home/pi/.dosbox-sdl2/dosbox-sdl2-SVN.conf
SDL: Current window pixel format: SDL_PIXELFORMAT_ABGR8888
MIDI:fluidsynth: loaded soundfont: /usr/share/sounds/sf2/FluidR3_GM.sf2
MIDI: Opened device:fluidsynth
Using driver "opengles2" for renderer
Using driver "opengles2" for renderer
```

hex007:
Can we have prebuild binaries for this as the source doesnt seem to change much and the compile time is too long.

joolswills:
No. I don't provide binaries for experiment packages.

PR #2236, opened 2017-12-26
RPi stretch binary-only compatibility patching

psyke83:
Hi @joolswills,

This seems like the ideal solution for our legacy/binary only items. I've confirmed that giana, rpix86 and coolcv work perfectly on stretch after patching with this PR.

This PR can be expanded to all other binary packages. Can you let me know what other binary-only software is distributed by the script and thus may potentially need patching?

Should help towards resolving #2091.

### DIFF
--- a/scriptmodules/emulators/coolcv.sh
+++ b/scriptmodules/emulators/coolcv.sh
@@ -22,6 +22,7 @@ function depends_coolcv() {
 
 function install_bin_coolcv() {
     downloadAndExtract "$__archive_url/coolcv.tar.gz" "$md_inst" 1
+    patchVendorGraphics "$md_inst/coolcv_pi"
 }
 
 function configure_coolcv() {
--- a/scriptmodules/emulators/drastic.sh
+++ b/scriptmodules/emulators/drastic.sh
@@ -18,6 +18,7 @@ rp_module_flags="!mali !x86 !armv6 !kms"
 
 function install_bin_drastic() {
     downloadAndExtract "http://drastic-ds.com/drastic_rpi.tar.bz2" "$md_inst" 1
+    patchVendorGraphics "$md_inst/drastic"
 }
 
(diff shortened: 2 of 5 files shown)


joolswills:
Can't think of any others currently. Could use a function to de-dupe here a bit, but maybe not worth it if there are only 4 needed (and this is hopefully temporary until the upstream projects rebuild etc).

Please use printMsgs instead of echo.

psyke83:
I think the helper would be a good idea (as it can catch all possible renamed libraries in case we find more software needing patching), but where would t be acceptable for you? Here?

```
diff --git a/scriptmodules/system.sh b/scriptmodules/system.sh
index 6155e240..68f9bd00 100644
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -191,6 +191,20 @@ function get_rpi_video() {
     export PKG_CONFIG_PATH="$pkgconfig"
 }

+function patch_rpi_video() {
+    local filename="$1"
+
+    getDepends patchelf
+    # replace all instances of old graphics library names with new names
+    printMsgs "console" "patching: $filename"
+    patchelf --replace-needed libEGL.so libbrcmEGL.so \
+             --replace-needed libGLES_CM.so libbrcmGLESv2.so \
+             --replace-needed libGLESv1_CM.so libbrcmGLESv2.so \
+             --replace-needed libGLESv2.so libbrcmGLESv2.so \
+             --replace-needed libOpenVG.so libbrcmOpenVG.so \
+             --replace-needed libWFC.so libbrcmWFC.so "$filename"
+}
+
```
I'm not sure it's worth putting in helpers.sh, as runcommand would never need it (and we've already got get_rpi_video right there in system.sh).

joolswills:
I would prefer helpers.sh - thanks.

joolswills:
Thanks!

PR #2250, opened 2018-01-02
Allow advmame to compile on ODROID, fix ODROID device_keyboard error …

zerojay:
…on launch.

### DIFF
--- a/scriptmodules/emulators/advmame-0.94.sh
+++ b/scriptmodules/emulators/advmame-0.94.sh
@@ -14,7 +14,7 @@ rp_module_desc="AdvanceMAME v0.94.0"
 rp_module_help="ROM Extension: .zip\n\nCopy your AdvanceMAME roms to either $romdir/mame-advmame or\n$romdir/arcade"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/amadvance/advancemame/master/COPYING"
 rp_module_section="opt"
-rp_module_flags="!mali !kms"
+rp_module_flags="!kms"
 
 function depends_advmame-0.94() {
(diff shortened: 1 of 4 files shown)


joolswills:
I had segfaults when launching roms when testing previously on the XU4. This won't work on the C1 as it will enable the module but won't set the options.

zerojay:
No segfault problems for me on XU4 with both advmame-0.94/1.4. Haven't tried 3.5/3.6 though.
I purposely limited the device_keyboard error fix for XU3/4 because I don't know if C1 needs it or not.

joolswills:
it was the 3.5/3.6 versions which I tested.

The C1 would need that change also, but it needs to be tested.

joolswills:
Otherwise you would need to just exclude the c1/c2 boards (if enabling only for the XU).

zerojay:
3.6 segfaults on me with the one game I set up for it. I'll do more extensive testing soon.

zerojay:
A question for you. Since 3.5 and 3.6 seem to not function (to this point anyways), can we remove the !mail flag from 0.94 and 1.4 but keep it on the latest (3.5/3.6)? Does having !mali on 3.5/3.6 cause the calls to use the configure function from the 3.5/3.6 script to not get called? (I believe it's the case but can't say for sure.)

joolswills:
Yes. 

BTW you have put an unrelated to jzintv on here. That uses SDL1 as I remember and it never did used to work, since it didn't work directly on the framebuffer before. I would prefer not to have this on this PR. I'm going to close this and look into advmame later when I have some free time.

zerojay:
Uh, that's strange as that was supposed to be a separate pull request.

PR #2252, opened 2018-01-03
customhidsony: add custom hid-sony dkms driver module

psyke83:
Implement custom hid-sony driver patched to fix the
eternal vibrate bug with third-party Shanwan controllers. Currently
compatible with Raspberry Pi only.

Uses the following patch: https://github.com/raspberrypi/linux/commit/492ca83c3d19fba1622164f07cd7b775596a7db2

This module could also be useful if we can fix the Bluetooth idle timeout
not being respected. That would make the official BlueZ plugin a viable
replacement for the sixad wrapper for official and third-party controllers
alike.

---

Please don't merge - this is untested except to confirm successful build & dkms installation. Feedback from anyone with a Shanwan third-party controller that has the vibration bug would be appreciated. The patched modules probably still requires the use of the ps3controller driver for BT pairing, and it also shouldn't interfere with official controllers.

### DIFF
--- a/scriptmodules/supplementary/customhidsony.sh
+++ b/scriptmodules/supplementary/customhidsony.sh
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="customhidsony"
+rp_module_desc="Custom hid-sony driver for Raspberry Pi"
+rp_module_help="Installs patched hid-sony driver to fix Shanwan third-party controllers always vibrating via USB.\n\nNote: this driver should be used in conjunction with ps3controller to ensure Bluetooth compatibility."
+rp_module_section="driver"
+rp_module_flags="noinstclean !x86 !mali !x11"
+
+function _version_customhidsony() {
+    echo "0.1.0"
+}
+
+function _dkms_remove_customhidsony() {
+    dkms remove -m hid-sony -v "$(_version_customhidsony)" --all
+}
+
+function depends_customhidsony() {
+    depends_gamecondriver
+}
+
+function sources_customhidsony() {
+    mkdir -p "$md_inst/"
+    pushd "$md_inst"
+
+    cat > "$md_inst/Makefile" << _EOF_
+obj-m := drivers/hid/hid-sony.o
+_EOF_
+
+    cat > "$md_inst/dkms.conf" << _EOF_
+PACKAGE_NAME="hid-sony"
+PACKAGE_VERSION="$(_version_customhidsony)"
+PRE_BUILD="hidsony_source.sh"
+BUILT_MODULE_LOCATION="drivers/hid"
+BUILT_MODULE_NAME="\$PACKAGE_NAME"
+DEST_MODULE_LOCATION="/updates/dkms"
+AUTOINSTALL="yes"
+_EOF_
+
+    cat > "$md_inst/hidsony_source.sh" << _EOF_
+#!/bin/bash
+rpi_kernel_ver="raspberrypi-kernel_\$(dpkg -s raspberrypi-kernel | grep '^Version:' | awk '{print \$2}')"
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Looks good to me.

pjft:
Thanks. I confirm that at least on my old kernel that this works as intended with my Shanwan controllers. It also doesn't interfere with my original controllers.

Thanks for the quick response here!

psyke83:
@pjft - thanks for the feedback :)

@buzz - I updated the PR just to clarify in the help text that it's intended to be used alongside ps3controller. Any objections if I merge? I'd like to take a look at the ps3controller repo next, and try to roll the shanwan branch into main by properly fixing co-existence between third party and official controllers, as @pjft brought to my attention the issue of blinking LEDs on official controllers.

Might need @pjft to help with testing again if I do that... ;)

pjft:
By all means - more than happy to help with the testing at least! :)

Please let me know at any point in time - here and over hangouts as well!

Thank you!

buzz:
@psyke83 I really love RetroPie. It's awesome. But I think I have nothing to do with this PR, right?

pjft:
@buzz we really use this as a way to get positive feedback from random users. It's always heartwarming :)

-- j/k, sorry for bothering. It should have been @joolswills ;)

Have a great weekend.

psyke83:
Apologies for tagging you, @buzz - even though it *seems* like my fault, I blame @joolswills' for using your username on the RetroPie forums. ;)

joolswills:
Merge away.

tateshep:
Hello, This looks like exactly the solution I need. But I don't know how to implement it. Do I copy and paste that code in somewhere? I'm sorry, newb here.

fernanfng:
Hi @tateshep , ¿could you tell me where the README is?

thank you

tateshep:
@fernanfng https://github.com/RetroPie/RetroPie-Setup/blob/master/README.md

uqs:
So while this was merged, it doesn't seem to work for me. It looks like the device strings can be ... different. I have 2 PS3 controllers and they both show up as: ``ShanWan PS(R) Ga`epad``  (yeah, no kidding about the m to ` flip). Not sure how that is possible, it happens on both controllers and is not a simple bit flip.

Also, from this guide here, https://raspberryblog.de/?p=1870&page=3 that dude's controller shows up exactly as mine.

Now, I'm confused in that there is this quirk here which applies to the proper vendor/device id and talks about operational mode:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/usb/input/hid-core.c?id=4a1a4d8b87389e35c3af04c0d0a95f6a0391b964

and the commit referenced in this PR https://github.com/raspberrypi/linux/commit/492ca83c3d19fba1622164f07cd7b775596a7db2 (which I think is superseded by https://github.com/torvalds/linux/commit/69481059c507b3b5a99aaab0d544e6516034defc ) also talks about "operational mode", but it's in hid-sony, not hid-core and it uses strcmp, not the device/vendor ID, so it will never match for my controller, right?

My RetroPie has kernel 5.10.52-v7+ and 5.10 was branched in late 2020, so it should have all those quirks, right?

I guess it is infeasible to compile a kernel on the Pi, how would I just recompile hid_sony (and where to get the proper sources to patch?)

Should that quirk be changed from a strcmp to being based on vendor/device IDs or would that break proper original controllers (btw, the packaging of my controllers has Sony all over it, no mention of it being not an original)?

uqs:
Anyway, for folks that bump into this thread in the future, the workaround is this patch, that I've just submitted upstream. Make note of the alternative spelling of the device string, this spelling can be found all over the interwebs as well:

```
diff --git drivers/hid/hid-sony.c drivers/hid/hid-sony.c
index b3722c51ec78..a2fef59063a6 100644
--- drivers/hid/hid-sony.c
+++ drivers/hid/hid-sony.c
@@ -2974,7 +2974,8 @@ static int sony_probe(struct hid_device *hdev, const struct hid_device_id *id)
        if (!strcmp(hdev->name, "FutureMax Dance Mat"))
                quirks |= FUTUREMAX_DANCE_MAT;
 
-       if (!strcmp(hdev->name, "SHANWAN PS3 GamePad"))
+       if (!strcmp(hdev->name, "SHANWAN PS3 GamePad") ||
+           !strcmp(hdev->name, "ShanWan PS(R) Ga`epad"))
                quirks |= SHANWAN_GAMEPAD;
 
        sc = devm_kzalloc(&hdev->dev, sizeof(*sc), GFP_KERNEL);

```

PR #2263, opened 2018-01-19
Sixaxis scriptmodule

psyke83:
This is a very simple service to properly configure the native hid-sony and BlueZ sixaxis plugin, intended to completely replace sixad.

Current status: successfully tested on RPI3 (stretch) with official controllers. ~~~Won't yet work on third-party controllers until custom BlueZ 5.48+ packages are compiled, but the code is already in place ready to be used (as well as the means to downgrade back to the respository package versions).~~~ Confirmed as working on Shanwan controllers (which require the BlueZ 5.48 packages).

Bluetooth pairing is achieved by choosing the ```Register and connect``` option in the RetroPie Bluetooth menu, then connecting the controller via USB while the ```Scanning``` prompt is shown (note: must not be connected before scanning begins).
~~~This will probably only work on Pi until an issue (see below) is resolved.~~~

Advantages:
* Uses the native hid-sony driver for both USB and BT (via BlueZ's sixaxis)
* Seems to have less input lag than the sixad wrapper
* The Bluetooth stack is not disabled or crippled in any way
* The service uses a simple program to disconnect each controller after being idle for 10 minutes (needed because the sixaxis plugin doesn't respect the Bluetooth IdleTimeout setting).
* Unlike sixad, no perceivable lag when the controller auto-disconnects.

Major issues:
* ~~~It's necessary to import the udev environment for the helper script to operate correctly. This works well, except that udev mangles variables that include a hyphen, such as the USB $DEVPATH. This doesn't affect Bluetooth rules, but, at least on Pi, USB rules won't get processed correctly. Until a more generic fix is made, automatic Bluetooth pairing will only work on Pi.~~~ (A [better ](https://github.com/psyke83/sixaxis/commit/5e40ebe82e60df2ca615f4b88522868c424ce0e9)workaround has been committed, so this should work on all devices).
* ~~~BlueZ 5.48 needs to be installed in order for third-party controllers (Shanwan) to pair via Bluetooth. I'll try adapting the latest Debian source (5.47) to the newer release, but I don't have third-party controllers to test. In the meantime, the default BlueZ packages in stretch (and most likely jessie) are sufficient for the official controllers.~~~ BlueZ 4.58 will be installed to support third-party controllers, but I only have official controllers for testing.

Minor:
* I could use feedback on potential issues with the sixaxis-timeout program, but it seems to do the job well and doesn't hog CPU either when idle or buttons are continuously pressed.
* ~~~Kodi doesn't seem to have proper mappings for the controller when the new driver is installed, but it can be easily reconfigured. Will investigate.~~~ The new hid-sony driver uses new key mappings, so Kodi's default configuration no longer matches. Not really a bug, but an inconvenience.


psyke83:
Here's the ugly workaround that I'd appreciate input on how to solve properly:
https://github.com/psyke83/sixaxis/blob/master/sixaxis-helper.sh#L9-L16

To clarify, udev mangles hyphens into forward slashes in DEVPATH; for example:
```/devices/platform/soc/3f980000.usb/usb1/1-1/1-1.2/1-1.2:1.0/0003:054C:0268.0007/input/input3/event2```
Becomes this:
```/devices/platform/soc/3f980000.usb/usb1/1/1/1/1.2/1/1.2:1.0/0003:054C:0268.0007/input/input3/event2```

Without the ugly hack above, the USB udev rules won't be processed correctly (which are needed to set ```ds3_settrust``` when plugging in the controller). When that doesn't work, automatic BT pairing won't work until you manually set the controller's MAC address as trusted.

**Comment by psyke83** [comment]
@pjft would appreciate if you could give this a test (with your two official controllers). Assuming that it even works at all, I'd like to hear feedback on input latency, especially re: any dropped events when both controllers are in use, as you noticed with recent sixad versions.

Your Shanwan controllers won't work via BT until I sort out the new BlueZ packages. This PR uninstalls ps3controller as part of its install process, so be sure to limit testing to your backup card. Thanks.

**Comment by pjft** [comment]
-- thanks for tagging me here, I certainly will!

I am not a subscriber to the RetroPie-Setup repository, so I had missed it originally.

Will report back during the weekend, if it suits.

**Comment by psyke83** [comment]
Maybe hold off for a bit - I've just tested on jessie and noticed some issues.

**Comment by psyke83** [comment]
Updated PR & repo to working correctly on jessie - ready for testing.

The new hid-sony driver uses new key mappings, so you'll need to reconfigure your controller. It also explains why Kodi's mappings don't work out of the box.

**Comment by psyke83** [comment]
Just a small update, no new functional changes.
* clean up kernel patch to disable sensors
* handle case in which bluetoothd is not executable after uninstalling pre-systemd versions of sixad
* add version check conditional for bluez "backport" function (which is still disabled anyway).

**Comment by psyke83** [comment]
Big update - it will now install BlueZ 5.48, meaning that third-party controllers should be compatible. These packages are built by me, with no special patches etc.; I simply manually updated from the latest 5.47-1 version. My source is here (as well as the binaries): https://github.com/psyke83/bluez-retropie

@pjft : can you refresh the PR and try with your Shanwan controllers? Only try on your development image; I have tested the BlueZ upgrade/downgrade thoroughly, but I may have overlooked something.

If the controllers don't pair or there's any kind of problem, ```cat /var/log/syslog | grep sixad``` will be a good starting point. Also, the full build log from ```~/RetroPie-Setup/logs``` would be useful. Thanks!

**Comment by psyke83** [comment]
Updated the PR to make pairing work correctly. BlueZ's sixaxis requires authorization on cable connect only, so the procedure to pair is:

* In the RetroPie Bluetooth menu, choose ```Register and Connect to Bluetooth Device```
* When sixaxis is installed, the ```Searching ...``` text will be modified to ```Searching (Sixaxis USB pairing mode enabled) ...```
* While the above dialog is showing, connect your controller via USB, and the device will be authorized in the Bluetooth stack.

Scanning standard Bluetooth devices is not impaired by this change, so hopefully this approach is acceptable.

**Comment by joolswills** [comment]
Thanks for this. Note that when I tested native pairing on Jessie I couldn't get it to work. It worked on my Ubuntu machine using bluetoothctl though. I assumed it was an issue with the Jessie bluez version.

**Comment by pjft** [comment]
I'll look into this later today - thank you!

**Comment by psyke83** [comment]
@joolswills - I only pushed the changes to fix first-time pairing 20 minutes ago. Were you trying that version? Only in that latest PR would you see the different ```Searching (Sixaxis USB pairing mode enabled) ...``` text.

I've made modifications to the bluetooth script to support BlueZ 5.48, but been testing those changes against the old BlueZ 5.23 on jessie to ensure there's no incompatibility introduced. I've just tested again, and scan/pair still works as expected on a standard BT mouse with the 5.23 stack on jessie.

**Comment by psyke83** [comment]
@joolswills I may have misunderstood that you tested the PR, sorry. To The PR achieves pairing via:
```
bluez_cmd "default-agent" "10" "Authorize" "yes"
```
So it sends "default-agent", then waits for 10 seconds. During that time, if the "Authorize" string is matched, it sends the "yes" reply to enable the authorization. Same instructions as this: https://wiki.gentoo.org/wiki/Sony_DualShock

This can only work if "default-agent" is issued before cable connect, and the service request is triggered on the USB connection. So we can't automate this via udev, etc.

This PR more than likely could also allow pairing with DualShock4 controllers (without needing the userspace driver).

**Comment by joolswills** [comment]
Yeah sorry if I wasn't clear. I haven't tested the PR. I just tested manually with bluetoothctl. It was reliable on Ubuntu (newer bluez), but not on Jessie when I tried. If your method works reliably on Jessie, then maybe I did something wrong, or Jessie had a bluez update that resolved it.

**Comment by joolswills** [comment]
Reference - https://github.com/RetroPie/sixad/issues/9#issuecomment-255553149

Glad if it's working as would be nice to get rid of sixad

**Comment by psyke83** [comment]
Yeah, it's understandable to see different results. The behaviour varies between BlueZ releases due to changes in the cable connect policy for the sixaxis plugin.

For older BlueZ versions, you can simply "trust" a known MAC addresses even when it's not connected. But the new BlueZ is stricter, and until you authorize the device, the MAC address is not listed in "devices" output.

After testing manually, I've discovered that this is the minimum steps to authorize (the wiki has superflous steps);
```
[bluetooth]# default-agent
(connect via USB cable - can't be already connected)
[agent]Authorize service service_uuid (yes/no): yes
```

**Comment by joolswills** [comment]
Odd why it didn't work reliably for me on Jessie as I did the same steps. I'll give your PR a go when I have a chance. Thanks.

**Comment by psyke83** [comment]
@pjft: I just reinstalled a fresh jessie image - it turns out that packages were being left unconfigured due to an unresolved dependency (libdw1):

```
dpkg: dependency problems prevent configuration of bluez:
 bluez depends on libdw1 (>= 0.127); however:
  Package libdw1 is not installed.

dpkg: error processing package bluez (--install):
 dependency problems - leaving unconfigured
Setting up libbluetooth3:armhf (5.48-1~rpi1) ...
Setting up libbluetooth-dev:armhf (5.48-1~rpi1) ...
dpkg: dependency problems prevent configuration of bluetooth:
 bluetooth depends on bluez; however:
  Package bluez is not configured yet.

dpkg: error processing package bluetooth (--install):
 dependency problems - leaving unconfigured
```

I've updated the PR with that dependency, so it should now install all BlueZ packages from a default image. Please refresh the PR and retry.

Your pairing issue seemed related to the SDP record, which is expected to happen on an older BlueZ release. The new BlueZ is supposed to set a SDP record, which presumably would properly configure the Bluetooth HID type to be the correct "input-gaming" type, etc.

We can debug it, but let's wait until you've tried the new PR with all packages installed correctly, just in case that was the issue.

**Comment by psyke83** [comment]
Raspbian added some modifications to the debian source that I missed, including a change to the default dbus security policy. I've rebuilt the packages with the patches included, so Bluetooth should now work reliably.

If you're retesting the PR, you may need to uninstall and reinstall sixaxis to trigger re-download/install of the new BlueZ packages.

**Comment by psyke83** [comment]
@joolswills: this is ready for proper scrutiny now that Shanwan controllers are confirmed to work.

@pjft correctly pointed out that the difference in controller mappings between the old and new driver will cause users who have no other input method (or SSH) to get stuck in the RetroPie-Setup menu after installation.

I could patch the driver to use the old mappings, but I'm not so sure this is a good idea. The Raspberry Pi firmware is due to be bumped to 4.14 soon, and the mappings are going to change when this happens anyway. Thoughts?

**Comment by pjft** [comment]
Just to make it clear: even after a reboot those users will be left with an unusable installation unless they have another controller type, keyboard or SSH available and manage to trigger the input configuration menu in ES for this known but remapped controller. 

If it was RetroPie Setup only we could probably force a reboot after the installation. 

I was suggesting to alter the es_input file via the script to remap the buttons to the new ones or delete that previous mapping so that ES will launch the input configuration menu on start. 

They'd still be stick on RetroPie Setup so an automatic reboot after 10secs unless someone stops it is probably a good idea. 

Thank you.

**Comment by psyke83** [comment]
I noticed an issue with joy2key related to variance in the DEVPATH reported by udev in the sixad virtual uinput device vs. hid-sony.

With hid_sony, content of 'device' variable:
```
Device(u'/sys/devices/platform/soc/3f201000.serial/tty/ttyAMA0/hci0/hci0:11/0005:054C:0268.000D/input/input17/js0')
```
The current code, ```sysdev_path = '/sys' + os.path.dirname(device.get('DEVPATH')) + '/'```, resolves to the following:
```
/sys/devices/platform/soc/3f201000.serial/tty/ttyAMA0/hci0/hci0:11/0005:054C:0268.000D/input/
```

This is the incorrect path - it should include ```/input17``` - which leads to joy2key breaking with the new hid_sony driver.

The new function will properly normalize the path to DEVPATH, then check if "name" exists in the current path. If not, it sets the path to the parent. This fixes the hid_sony driver with joy2key and continues working with the old sixad uninput virtual device. I'm reasonably confident this won't break anything else, but would appreciate a fresh pair of eyes on it.

**Comment by psyke83** [comment]
I've updated the PR to use the old key mappings, as well as a fix for Kodi and an update hook to ensure the module stays up to date.

@pjft,

Can you give the PR another test when it's convenient? Make sure to remove the old sixaxis module before updating. Also make sure to remove the customhidsony module; if necessary, do it manually via ```sudo dkms remove hid-sony/0.1.1 --all```. (this is only needed for people testing older PRs. I didn't bump the customhidsony version between PR versions, but it won't be an issue if/when it's finished and merged)

I'd like you to focus on the following:
* Verify that the keymaps are the same compared to the old driver. They're identical for me; here's old mapping via sixad vs new:
```
--- "PLAYSTATION(R)3 Controller.cfg"    2018-02-09 17:14:01.467483330 +0000
+++ "Sony PLAYSTATION(R)3 Controller.cfg"       2018-02-10 02:29:15.898780150 +0000
@@ -1,4 +1,4 @@
-input_device = "PLAYSTATION(R)3 Controller"
+input_device = "Sony PLAYSTATION(R)3 Controller"
 input_driver = "udev"
 input_r_y_plus_axis = "+3"
 input_r_x_minus_axis = "-2"
```
N.B. You may still need to remap controls. The sixad virtual uinput device doesn't prefix the vendor ("Sony" in my case) in the device name, but the old *and* new hid-sony driver does. This means that if you only mapped via Bluetooth using sixad and not via USB, you will probably need to map your controller again. But you shouldn't find yourself stuck in the menu with incorrect bindings.

* Verify that the idle disconnect timeout works OK on both official and Shanwan controllers. To speed up the process, you can edit ```/usr/bin/sixaxis-helper.sh``` and change SIXAXIS_TIMEOUT from the default of 600 (10 mins) to a lower value. I know it works OK on official, but I just need to be sure that the udev rule is matching your Shanwan correctly.

* See if Kodi works out of the box. It will work with your official controller, but I'm not so sure about your Shanwan (with or without this PR): https://github.com/xbmc/peripheral.joystick/tree/master/peripheral.joystick/resources/buttonmaps/xml/linux

The Shanwan keymap listed there may not match your controller's product name. Whether it does or not, a modified version would need to be created in order to work with the new hid-sony driver (due to the axis/button count being different).

* If possible, send me an excerpt of your log, preferably after trying the idle disconnect:
```cat /var/log/syslog | grep sixaxis```

Thanks.

**Comment by psyke83** [comment]
I've just tested the transition process, using only the controller to navigate (apart from setting up the PR branch)

1. Fresh RetroPie install
2. Set up PR branch (only step via ssh)
3. Map controller via USB
4. Install ps3controller, pair and map controller again (because the hid-sony driver prefixes the vendor)
5. Enable the PR and install the sixaxis module

The result: 
* the controller turns off during sixaxis installation (expected)
* the controller refuses to pair when turned back on (expected)
* connecting via USB, and the controller works fine. It's possible to exit out of the script without any issue.
* to set up pairing, I enter the Bluetooth menu, choose ```Register and connect...```, then unplug and replug the controller.

Result: the controller pairs fine and has mapping available immediately.

Considering the results, I updated the PR to clarify the proper process when assuming that the menu is being driver by the controller:

Help text now says:
```It's highly recommended that you map your controller via USB *before* installation!```

Bluetooth scan says:
```
 Searching ... 

(To register a DualShock controller, please unplug and
replug your controller while this text is visible...)
```

It's important to map your controller via USB before installing sixaxis, or else you may get stuck in EmulationStation.

**Comment by psyke83** [comment]
I just refreshed the PR. Some updates:

* Given the fact that the next firmware will be kernel 4.14, which will use the new hid-sony driver button mappings, I've decided to get rid of the patch to use the old mappings. The change in button mappings is inevitable.
* I've split the BlueZ backport into a separate module - which is not selected for installation by ```sixaxis```, as it's only needed if you have a third-party controller. The module is now also capable of building these packages entirely from source similar to sdl/sdl2, as well as an install from binary function (which can be changed to fetch from retropie's servers later).
* I've gotten rid of the edited Kodi keymap file in favour of a function that will patch the original version of the file instead. I suppose the function could be extended to patch EmulationStation's ```es_input.cfg```, but that's probably overkill.

No hurry to review this - in fact, it might make more sense to defer until after the next stretch firmware is released; then, we won't need to worry about the changed button mappings. Cheers.

**Comment by joolswills** [comment]
thanks - will give it a test when I get a chance. Appreciate your efforts to improve this part of RetroPie.

File: scriptmodules/supplementary/bluetooth.sh [hunk 1]
```diff
@@ -46,33 +46,53 @@ function get_script_bluetooth() {
     echo "$name"
 }
 
+function bluez_cmd() {
```
joolswills: function naming is wrong - needs to include _bluetooth so it's part of the module.

scriptmodules/supplementary/bluetooth.sh
```diff
@@ -300,7 +320,7 @@ function boot_bluetooth() {
             while read mac_address; read device_name; do
                 macs+=($mac_address)
             done < <(list_registered_bluetooth)
-            local script="while true; do for mac in ${macs[@]}; do hcitool con | grep -q \"\$mac\" || echo \"connect \$mac\" | bluetoothctl >/dev/null 2>&1; sleep 10; done; done"
+            local script="while true; do for mac in ${macs[@]}; do hcitool con | grep -q \"\$mac\" || bluez_cmd \"connect \$mac\"; sleep 10; done; done"
```
joolswills: I don't think this will work as this script is backgrounded. the bluez_cmd won't be available.

psyke83: The issue with the new bluez is that its command-line parsing no longer works with a direct pipe unless you pad with an explicit sleep before and after the required command, i.e. something like:

```{ sleep 1; echo \"connect \$mac\"; sleep 1; } | bluetoothctl >/dev/null 2>&1```

Would something like the above be a satisfactory workaround to using the dedicated function? I'm not sure if I'm missing more escape characters for the curly braces etc., but I'll test.

Comment (psyke83) [reply scriptmodules/supplementary/bluetooth.sh]: Also, the original command is also missing the "quit" command, which may leave a rogue bluetoothctl process.

Comment (joolswills) [reply scriptmodules/supplementary/bluetooth.sh]: I never noticed any bluetoothctl processes being left when testing - it always quit after the pipe closed. So the bluetoothctl command line parsing got even worse :-) nice. If your workaround works, I'm happy to switch to it.

Comment (psyke83) [reply scriptmodules/supplementary/bluetooth.sh]: Well - testing bluez 5.43 on stretch:
```
{ echo version; sleep 1; } | bluetoothctl
```
gives: ```(bluetoothctl:16701): GLib-CRITICAL **: Source ID 23 was not found when attempting to remove it``` - but the process is gone, yes.

```
{ echo -e "version\nquit"; sleep 1; } | bluetoothctl
```
Will quit gracefully. Probably a good idea to do it.

My memory is foggy on whether sleep needs to happen first, but it's probably a good idea since the dedicated function properly waits for the correct "bluetooth#" prompt to appear, but a direct pipe wouldn't.

Comment (joolswills) [reply scriptmodules/supplementary/bluetooth.sh]: I agree. I don't remember this on Jessie, so maybe it's also new.

Comment (joolswills) [reply scriptmodules/supplementary/bluetooth.sh]: Didn't seem to need the sleep here at least for the version command.

Comment (psyke83) [reply scriptmodules/supplementary/bluetooth.sh]: Some commands are ok, including ```version``` and ``ìnfo ...``` but the "connect" command was failing for me on BlueZ 5.48. 

The Raspbian patches to BlueZ change the default dbus security policy to more permissive, so perhaps the behaviour was different in an older version of the PR where I hadn't yet included the patches. Will re-check it regardless.

File: scriptmodules/supplementary/custombluez.sh [hunk 3]
```diff
@@ -0,0 +1,81 @@
...
+
+function sources_custombluez() {
+    # git-buildpackage requires the master *and* pristine-tar branch to be available, so clone the entire repo
+    __persistent_repos=1 gitPullOrClone "$md_build/custombluez" https://github.com/psyke83/bluez-retropie.git
+}
+
+function build_custombluez() {
+    pushd custombluez
+    rm -rf "$md_build/build-area"
+    CFLAGS="${CFLAGS//-O2/-Os}" gbp buildpackage -us -uc
```
psyke83: Reverting to the lower optimization level appears to be necessary to avoid this build error:
```
/usr/bin/ld: client/main.o: relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
```

This happens on jessie but not stretch, as far as I can see. Running git-buildpackage from a shell manually on jessie will build the packages successfully (as ```-O2``` is not inserted via the CFLAGS environment variables).

scriptmodules/supplementary/bluetooth.sh
```diff
@@ -300,7 +314,7 @@ function boot_bluetooth() {
             while read mac_address; read device_name; do
                 macs+=($mac_address)
             done < <(list_registered_bluetooth)
-            local script="while true; do for mac in ${macs[@]}; do hcitool con | grep -q \"\$mac\" || echo \"connect \$mac\" | bluetoothctl >/dev/null 2>&1; sleep 10; done; done"
+            local script="while true; do for mac in ${macs[@]}; do hcitool con | grep -q \"\$mac\" || { echo \"connect \$mac\nquit\"; sleep 1; } | bluetoothctl >/dev/null 2>&1; sleep 10; done; done"
```
psyke83: Just to follow up, here's the behaviour with BlueZ 5.48:
```
pi@retropie:~/RetroPie-Setup $ echo version | bluetoothctl
[NEW] Controller B8:27:EB:1B:E6:3A retropie [default]
Agent registered
Agent unregistered
[DEL] Controller B8:27:EB:1B:E6:3A retropie [default]
vpi@retropie:~/RetroPie-Setup $
```
Notice that the version string begins to be printed before cutting off. Before I added the "sleep 1" to the end of ```bluez_cmd_bluetooth```, the response challenge was not completing correctly.

Compare to:
```
pi@retropie:~/RetroPie-Setup $ { echo version; sleep 1; } | bluetoothctl
[NEW] Controller B8:27:EB:1B:E6:3A retropie [default]
[NEW] Device C0:33:5E:05:DE:D2 Microsoft Sculpt Comfort Mouse
Agent registered
[Microsoft Sculpt Comfort Mouse]# version
Version 5.48
Agent unregistered
[DEL] Controller B8:27:EB:1B:E6:3A retropie [default]
pi@retropie:~/RetroPie-Setup $
```

Adding the sleep both here and in the (new) dedicated function should increase compatibility with newer BlueZ, without breaking legacy versions.

joolswills:
Buggy controllers yeah. It used to be needed for the 8bitdo ones. That was fixed in a firmware update. It's possible it may be useful for others.

The bluetoothctl tool is really not very good for automation. Needa a noninteractive mode really.

steeeb:
Always getting a PageTimeout using this pull request with a Shanwan controller, with sixad and custombluez installed.

scriptmodules/supplementary/custombluez.sh
psyke83: Will change that in the next refresh, thanks.

There's another issue that will need discussing before this PR is suitable to be included. Currently, the new driver exposes both digital and analog events for both of the trigger buttons (L2 & R2). The script currently sets a large axis fuzz on the analog axes to effectively disable them, so that only the digital inputs would be used, but disabling the analog inputs in this way will cause problems with Kodi, as they prefer to use the analog configuration (see: https://github.com/xbmc/peripheral.joystick/pull/143). Kodi has an "ignore input" feature to assist configuration of the controller: https://kodi.wiki/view/HOW-TO:PlayStation_4_controller

We might need to implement something similar for EmulationStation, otherwise it will be difficult to configure DS3 or DS4 pads correctly, as pressing L2 or R2 will register both digital and analog inputs simultaneously, causing the input configuration to assign the wrong key to the next button in the list erroneously.

scriptmodules/supplementary/sixaxis.sh
```diff
@@ -0,0 +1,53 @@
...
+rp_module_section="driver"
+
+function depends_sixaxis() {
+    local depends=(checkinstall libevdev-tools)
+
+    getDepends "${depends[@]}"
+
+    rp_callModule ps3controller remove
+    rp_callModule customhidsony
+    if isPlatform "rpi"; then
```
psyke83: Are we going to continue supporting jessie? This package or an equivalent that has the needed functionality isn't available in the older distro. If we're dropping support, I could simply remove this whole section and make things a bit cleaner.

joolswills: I would like to continue it at least for a bit - but if we drop support just for one package I don't mind perhaps. Will have a think.

urubu2007:
Hello guys, thanks for your effort on improve this part of Retropie! My only concern is that it is going to be incompatible with my 4 ps3 clone controllers, that are not of Shanwan type.. I tried to use your updates with them and really doesn't work with my controllers right now, so you guys could please keep the older version of sixad too (not using the custom hid-sony) for compatibility purposes? Thank you very much and best regards!

psyke83:
I've refreshed the patch set with a potential fix for Gasia controllers that identify as ```PLAYSTATION(R)3Conteroller-PANHAI``` via bluetooth.

I can't test, so I'd appreciate if @pjft and other owners can give it a shot. The patch is based on the comment from Marcelo (@mmbossoni) here: https://bugzilla.kernel.org/show_bug.cgi?id=200009

Hopefully I didn't misunderstand Marcelo's comment, but my genuine controller did fail to pair (blinking lights) if the function was entirely replaced instead of being isolated by name.

psyke83:
Updated the PR to include a potential fix for Gasia controllers connecting via Bluetooth. If anyone has a Gasia controller, feedback would be appreciated.

pjft:
Thanks for putting it together! I'll give it a whirl in the coming days, sure. What are we testing, just the pairing with Shanwan and official ones? Remind me again how should the pairing go - via the Bluetooth menu?

Thanks for the update!

psyke83:
@pjft I deleted a comment asking for you to test. I forgot that you had Shanwan pads instead of Gasia; only the latter needs testing, so you don't need to do anything. Sorry!

pjft:
Ah, apologies! Received it over email and came here to answer. Not a bother - glad to see you back in action :)

mmbossoni:
So,
I tested the patch for hid-sony and it worked on both of my controllers. I have a genuine ps3 controller and the gasia controller that causes problem.
Both of them are working just fine with the patch.

psyke83:
@mmbossoni,

Thanks for testing (and pointing out the potential solution on bugzilla)! Just to be clear, are you seeing the ```failed to send raw request, attempting fallback``` message in dmesg for your Gasia controller?

@joolswills 

The PR is updated with BlueZ 5.50 and various fixes, including compatibility issues with the bluetooth module. Testing with my official controller on stretch, association works seamlessly from the Retropie bluetooth menu (as long as you follow the on-screen prompt to plug in/unplug and replug the controller during  scanning). If we can confirm that all Gasia pads are working & that the trigger button patch in EmulationStation's master branch is safe, then this PR might be sufficient to completely replace the sixad daemon.

The PR has become a bit unwieldy again due to the amount of changes. I'm happy to split it up again if it will help make reviewing go smoother, but I'll spend some more time testing on my end regardless.

mmbossoni:
Yes, I got the message on dmesg

```
[  130.330936] input: Sony PLAYSTATION(R)3 Controller as /devices/pci0000:00/0000:00:01.3/0000:03:00.0/usb1/1-9/1-9:1.0/bluetooth/hci0/
hci0:71/0005:054C:0268.0006/input/input20
[  130.331030] sony 0005:054C:0268.0006: input,hidraw3: BLUETOOTH HID v80.00 Joystick [Sony PLAYSTATION(R)3 Controller] on 00:1a:7d:da:
71:03
[  140.411914] sony 0005:054C:0268.0006: failed to send raw request, attempting fallback
[  160.824399] sony 0005:054C:0268.0006: failed to send raw request, attempting fallback
```

V0rt3x667:
Q

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

tupc10:
Hello guys, I tested with my fake ps3 controllers and they just worked through usb connection, but when disconnecting from usb cable and trying to pair directly through bluetooth, retropie says that a new gamepad was found and their lights stay on paired status (no blinking), but I can't configure none of them through the retropie configure input, no buttons are recognized.. Just let me know if you need any log information to help, ok? Thanks for the great job and have a nice weekend! Cya

tupc10:
If it helps, I'm sending the output of the "info" command for my controller inside "bluetoothctl" prompt. Thanks again!

[Sony PLAYSTATION(R)3 Controller]# info
Device 00:26:5C:01:0D:A9 (public)
        Name: Sony PLAYSTATION(R)3 Controller
        Alias: Sony PLAYSTATION(R)3 Controller
        Class: 0x00000540
        Icon: input-keyboard
        Paired: no
        Trusted: yes
        Blocked: no
        Connected: yes
        LegacyPairing: no
        UUID: Human Interface Device... (00001124-0000-1000-8000-00805f9b34fb)
        Modalias: usb:v054Cp0268d0000

tupc10:
> @tupc10
> 
> Thanks for testing. Try installing the `evtest` package and run it from the terminal while the controller is connected. You should be able to check output from the USB connection so you have a frame of reference for how it should report button events when connected via Bluetooth.
> 
> Your bluetoothctl output indicates that it paired and trusted correctly, but it seems to have the wrong class/icon (my official controller reports class `0x00000508`, icon type `input-gaming`). I'm not certain if this is normal for a third-party controller, though.
> 
> Just in case the bluez stack is using a cached device enumeration from an older version, you may want to try removing the profile (via the bluetooth retropie script, or `remove 00:26:5C:01:0D:A9` via bluetoothctl) and then go through the pairing steps again.
> 
> Can you also let me know the exact device name enumerated when connecting from USB? You can find it in the dmesg log after plugging in. Thanks.

@psyke83 

Thanks for answering and sorry for the late answer! Here is the dmesg log when connecting my ps3 clone controller:

[12930.657739] usb 1-1.5: new full-speed USB device number 8 using dwc_otg
[12930.790750] usb 1-1.5: New USB device found, idVendor=054c, idProduct=0268
[12930.790771] usb 1-1.5: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[12930.790783] usb 1-1.5: Product: PLAYSTATION(R)3 Controller
[12930.790791] usb 1-1.5: Manufacturer: SZMY-POWER CO.,LTD.
[12930.805589] input: SZMY-POWER CO.,LTD. PLAYSTATION(R)3 Controller Motion Sensors as /devices/platform/soc/3f980000.usb/usb1/1-1/1-1.5/1-1.5:1.0/0003:054C:0268.0006/input/input9
[12930.868053] input: SZMY-POWER CO.,LTD. PLAYSTATION(R)3 Controller as /devices/platform/soc/3f980000.usb/usb1/1-1/1-1.5/1-1.5:1.0/0003:054C:0268.0006/input/input8
[12930.869046] sony 0003:054C:0268.0006: input,hiddev96,hidraw2: USB HID v81.11 Joystick [SZMY-POWER CO.,LTD. PLAYSTATION(R)3 Controller] on usb-3f980000.usb-1.5/input0

I checked another places on internet and most users reported that their ps3 clone controllers were classified as "input-gaming" on linux based systems, unlike happens with mine.

I repeated the bluetooth pairing process following your tips about remove the address using bluetoothctl and checking the evtest output, but unfortunately when connected through bluetooth evtest doesn't show any button that I have pressed.. Maybe because the wrong classification as "input-keyboard".

Anyway, thanks for your help and please let me know if there is anything more I could do for testing, ok?

Best regards!

mmbossoni:
@tupc10, @psyke83 

my GASIA controller shows up as a keyboard as well, but works as expected, so maybe this is not the problem.

tupc10:
> @tupc10, @psyke83
> 
> my GASIA controller shows up as a keyboard as well, but works as expected, so maybe this is not the problem.

@mmbossoni 

Hello, thanks for your answer! That is good to know, it seems that there is hope for my ps3 clone controllers then 😄 One thing that I noticed but don't know if it's relevant is that the name of my controllers are different when connecting through USB or bluetooth. They show up as ```SZMY-POWER CO.,LTD. PLAYSTATION(R)3 Controller``` over usb and ```Sony PLAYSTATION(R)3 Controller``` over bluetooth. I was able to use them over usb and bluetooth with the ```sixad``` driver compiled with the ```GASIA only``` option, but I did realized that connecting through USB after paired on bluetooth gave me double inputs, like there was 2 controllers connected simultaneously on Retropie, but in fact only 1 physical controller.

CookiePLMonster:
May I ask what is the state of this PR? How do the instructions presented here relate to `customhidsonly` driver, which seems to be the recommended one right now, eg. here: https://retropie.org.uk/forum/topic/20762/help-how-to-pair-use-four-dualshock-3-joyscticks-simultaneously-retropie-x86-and-retropie-rpi3/3

If the PR is still valid and relevant, I am willing to give it a try with my two legimate DS3 pads, even though sixad technically works well for me. Might just want to do a SD card backup prior to testing, in case I break something terribly.

psyke83:
I've split off some patches into different PRs to hopefully make things a bit easier to review.

Safe for review + merge at any time, independent of other changes:
* Bluez 5.50+ compatibility enhancement: https://github.com/RetroPie/RetroPie-Setup/pull/2594
* Joystick detection fix (new): https://github.com/RetroPie/RetroPie-Setup/pull/2592

Should be OK, but depends on #2594:
* BlueZ backport: https://github.com/RetroPie/RetroPie-Setup/pull/2593

Needs to be delayed until the above PRs - and new stable EmulationStation - is released:
* This PR (#2263)

Additionally, would you prefer that https://github.com/psyke83/sixaxis is forked into RetroPie? If so, I would of course need to update the scriptmodule to point to the new location.

jmbooth2000:
The good. This was easy to install and worked out of the box in Mame 2003, 2010, FBA, SNES, Genesis, N64, and PSX once you remapped the controls.

Even with the remapped controls it didn't work in adv-mame or reicast. I tried to manually reconfigure but I couldn't seem to get it to work. This is likely my fault vice the driver.

Strangely when I used evtest to look at the device it seemed supper sensitive and was continually reporting minor changes on all the analog controls. I don't remember this occurring with the old driver.

I'm going to continue to try and getting it working with adv-mame and riecast. I really want to get this working so I can use both a wii controller and ps3 controller at the same time.

scriptmodules/supplementary/sixaxis.sh
```diff
@@ -0,0 +1,81 @@
...
+
+rp_module_id="sixaxis"
+rp_module_desc="Helper service to install & configure the latest drivers for official and third-party DualShock controllers (ps3controller replacement)\n\nNote: For Shanwan/GASIA third-party controllers, enable third-party support in the configuration options.\n\nTo pair controllers, use the RetroPie Bluetooth menu, choose 'Register and Connect...', then follow the on-screen instructions."
+rp_module_licence="GPL2 https://raw.githubusercontent.com/RetroPie/sixaxis/master/COPYING"
+rp_module_section="driver"
+
+function depends_sixaxis() {
+    local depends=(checkinstall libevdev-tools)
+
+    getDepends "${depends[@]}"
```
joolswills: Is the array needed? Eg. Future platform dependency differences?

scriptmodules/supplementary/sixaxis.sh
```diff
@@ -0,0 +1,81 @@
...
+
+function sources_sixaxis() {
+    gitPullOrClone "$md_build/sixaxis" https://github.com/RetroPie/sixaxis.git
+}
+
+function build_sixaxis() {
+    cd sixaxis
+    make clean
+    make
+    md_ret_require+=("$md_build/sixaxis/bins/sixaxis-timeout")
```
joolswills: Not an issue but you don't have to use an array for a single file check. It's actually probably better to though - originally it only checked for single files and I think I extended it hence allowing either.

scriptmodules/supplementary/sixaxis.sh
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function configure_sixaxis() {
+    local sixaxis_config="$md_conf_root/all/sixaxis_timeout.cfg"
+
+    if [[ "$md_mode" == "install" ]] && [[ ! -f "$sixaxis_config" ]]; then
+        cat > "$sixaxis_config" << _EOF_
+# Set your preferred controller timeout in seconds (0 to disable)
+SIXAXIS_TIMEOUT=600
+_EOF_
```
joolswills: You could use the copyDefaultConfig if you wanted. Create the config file in tmp via mktemp and then use that function. Then if they changed the settings they get a rp-dist file they can reference. 

Not an issue either way.

lucasmr:
@psyke83 Any interest in making a kernel patch and submitting upstream? Bug is still open: https://bugzilla.kernel.org/show_bug.cgi?id=200009

PR #2265, opened 2018-01-21
amiberry: update for new SDL2 release

psyke83:
Preparation for @midwan's SDL2 version of amiberry. Not ready for merging.

### DIFF
--- a/scriptmodules/emulators/amiberry.sh
+++ b/scriptmodules/emulators/amiberry.sh
@@ -14,10 +14,15 @@ rp_module_desc="Amiga emulator with JIT support (forked from uae4arm)"
 rp_module_help="ROM Extension: .adf\n\nCopy your Amiga games to $romdir/amiga\n\nCopy the required BIOS files\nkick13.rom\nkick20.rom\nkick31.rom\nto $biosdir"
 rp_module_licence="GPL3 https://raw.githubusercontent.com/midwan/amiberry/master/COPYING"
 rp_module_section="opt"
-rp_module_flags="!x86 !mali !kms"
+rp_module_flags="!x86"
 
 function depends_amiberry() {
-    depends_uae4arm
+    local depends=(libpng12-dev libmpeg2-4-dev zlib1g-dev)
+    if ! isPlatform "rpi" || isPlatform "kms"; then
+        depends+=(libsdl2-dev libsdl2-image-dev libsdl2-ttf-dev)
+    fi
+
+    depends_uae4arm "${depends[@]}"
 }
 
 function sources_amiberry() {
@@ -25,18 +30,44 @@ function sources_amiberry() {
 }
 
 function build_amiberry() {
+    local amiberry_bin="$__platform-sdl2"
+    local amiberry_platform="$__platform-sdl2"
+    if isPlatform "rpi" && ! isPlatform "kms"; then
+        amiberry_bin="$__platform-sdl1"
+        amiberry_platform="$__platform"
+    elif isPlatform "odroid-xu"; then
+        amiberry_bin="xu4-sdl2"
+        amiberry_platform="xu4-sdl2"
+    elif isPlatform "tinker"; then
+        amiberry_bin="tinker"
+        amiberry_platform="tinker"
+    fi
+
     make clean
-    CXXFLAGS="" make PLATFORM="$__platform"
-    md_ret_require="$md_build/amiberry"
+    CXXFLAGS="" make PLATFORM="$amiberry_platform"
+    ln -sf "amiberry-$amiberry_bin" "amiberry"
+    md_ret_require="$md_build/amiberry-$amiberry_bin"
 }
 
 function install_amiberry() {
+    local amiberry_bin="$__platform-sdl2"
+    if isPlatform "rpi" && ! isPlatform "kms"; then
+        amiberry_bin="$__platform-sdl1"
+    elif isPlatform "odroid-xu"; then
+        amiberry_bin="xu4-sdl2"
+    elif isPlatform "tinker"; then
+        amiberry_bin="tinker"
+    fi
+
     md_ret_files=(
         'data'
+        "amiberry-$amiberry_bin"
         'amiberry'
     )
 }
 
 function configure_amiberry() {
     configure_uae4arm
+    moveConfigDir "$md_inst/controllers" "$configdir/all/retroarch/autoconfig"
+    moveConfigFile "$md_inst/conf/retroarch.cfg" "$configdir/all/retroarch.cfg"
 }
--- a/scriptmodules/emulators/uae4arm.sh
+++ b/scriptmodules/emulators/uae4arm.sh
@@ -17,7 +17,7 @@ rp_module_section="opt"
 rp_module_flags="!x86 !mali !kms"
 
 function depends_uae4arm() {
-    getDepends libsdl1.2-dev libsdl-gfx1.2-dev libsdl-ttf2.0-dev libguichan-dev libmpg123-dev libxml2-dev libflac-dev
+    getDepends libsdl1.2-dev libsdl-gfx1.2-dev libsdl-ttf2.0-dev libguichan-dev libmpg123-dev libxml2-dev libflac-dev "${@}"
 }
 
 function sources_uae4arm() {


psyke83:
@midwan,

So far I have the minimum changes needed to successfully build and install the rpiX-sdl2-dev version. To keep things simple, there will be a "amiberry" symlink created to refer to the appropriate binary. When you actually switch this code over to the master branch, I'll amend the PR to remove the "dev" branch (and presumably the "-dev" suffix for the target binary).

I've taken a look at the new menu items, and added the missing "controllers" configuration folder. I'm uncertain of the purpose of the retroarch configuration file - can this be compiled or run as a libretro core? Currently it will try to find the file at ```/opt/retropie/emulators/amiberry/conf/retroarch.cfg```, but I'm uncertain if extra configuration is needed (such as a symlink) to make this work as intended.

You also mentioned configuration changes, so if there are any other specific issues that need addressing, let me know.

I'm interested in testing the performance difference between dispmanx and plain SDL2, but I've hit an obstacle while attempting to set up my usual Amiga OS installation; when I click "Add Directory" in the Hard Drives menu, the program doesn't show a file selector; instead, the program refuses to respond to mouse inputs, necessitating the process to be terminated manually. Is this a known issue or is the binary not built correctly, perhaps? The 'Add Hardfile' dialog does open without issue.

psyke83:
I've tried compiling the different versions of amiberry:
sdl1-dev: stalls on black screen; mouse visible and moves, but GUI never renders
sdl2-dispmanx-dev: build error related to a missing header inclusion issue ([PR](https://github.com/midwan/amiberry/pull/214) sent). After fixing build, the program seems to launch correctly, but still hangs when clicking ```Add Directory```.

scriptmodules/emulators/amiberry.sh
```diff
@@ -17,22 +17,24 @@ rp_module_section="opt"
 rp_module_flags="!x86 !mali !kms"
 
 function depends_amiberry() {
-    depends_uae4arm
+    getDepends libxml2-dev libflac-dev libmpg123-dev libpng12-dev libmpeg2-4-dev
```
joolswills: missing sdl2 dependency.

scriptmodules/emulators/amiberry.sh
```diff
@@ -17,22 +17,24 @@ rp_module_section="opt"
...
 }
 
 function sources_amiberry() {
-    gitPullOrClone "$md_build" https://github.com/midwan/amiberry/
+    gitPullOrClone "$md_build" https://github.com/midwan/amiberry/ "dev"
```
joolswills: Just to check, but this is for testing now right (We will continue to track master).

joolswills:
It's worth testing with our custom sdl2 scaling options too. the sdl2 rpi driver is not ideal in some cases, as it doesn't support having a smaller window size and upscaling via videocore. I submitted our changes (actually @gizmo98 's changes) upstream, but they don't seem interested.

midwan:
Thanks for the feedback.
The file requester issue is certainly a bug, I'll get it fixed shortly.
The retroarch.cfg file is optional, and we use it (if found) for controller mapping under the "Custom" panel. We thought we'd pick a standard way of doing this, and the RetroArch standard looked good enough. If the file does not exist, the controller is used with the default mappings instead.

I'll check the issues you mentioned with the black screen, as the last time I tested this on RetroPie (yesterday) it seemed to work.

midwan:
Just fixed the final bug in the HDD Panel, which caused the Add Directory window to not draw the requester until a key was pressed. Thanks for bringing this up!

scriptmodules/emulators/amiberry.sh
psyke83: Yep. As I said, I'm also guessing the "-dev" suffix on the binary may also be gone when @midwan makes the switch to master. For now, the PR is just to help "kick the tires" and sort out potential issues with the new version.

scriptmodules/emulators/amiberry.sh
psyke83: Will add them when I next test - thanks.

scriptmodules/emulators/amiberry.sh
midwan: Yes, the "-dev" suffix will be gone once this is moved to master ;-)

psyke83:
@midwan,

Thanks - I should be able to continue testing tonight. Once I've got my setup configured, I'll try the different configurations (including some tests with the [dispmanx scaling](https://github.com/RetroPie/SDL-mirror/commit/5220245016a9aa31ac1ee5afb59f721aeecaceff) options, as @joolswills mentioned) to see the performance differences.

Regarding the retroarch.cfg - if it's being used to configure controller remapping, then perhaps it would be a good idea to create a symlink to, say, ```/opt/retropie/configs/all/retroarch.cfg```? My concern is that amiberry might makes changes to this file which would break retroarch itself.

midwan:
@psyke83 Amiberry treats this file as read-only, no changes are made to it.
It's expected that if the file is there, it was externally configured and we only parse it to set the controller options.

midwan:
Regarding dispmanx scaling, I've already implemented a scaling method inside Amiberry which keeps the correct aspect ratio, when using Dispmanx.
If using pure SDL2, then scaling is handled by the native SDL2 functions instead (SetLogicalSize).

midwan:
Also, might I suggest that the controller and retroarch paths are set to the relevant RetroPie paths (i.e. not amiberry/conf)? 
For example, I believe `/opt/retropie/configs/all/retroarch/autoconfig/` is the right location?

We wrote a relevant Wiki article here: https://github.com/midwan/amiberry/wiki/Setting-up-Input-Controllers

psyke83:
@HoraceAndTheSpider 

As far as I can see, the kickstart path is already accounted for. The ```uae4arm_configure``` function installs a symlink in ```/opt/retropie/emulators/amiberry/kickstarts``` pointing to ```/home/pi/RetroPie/BIOS``` (which is the officially designated kickstart path for all Amiga emulators, not ```BIOS/Amiga/``` as you said). It should work automatically without the need to provide a pre-configured adfdir.conf.

HoraceAndTheSpider:
Why not have the Amiga bios in a nice tidy folder though, like some others? There are 1000s if users who use that path with the pre made UAE configs for WHDLoad. It’s far from essential, but it might help some...

And how about the Configuration path? This is defaulting to Amiberry/conf atm I think. 

Again, I don’t see this as critical, i am just going from the feedback I receive from users, thinking how it could be more seem less for them.

psyke83:
What pre-made UAE configs are you talking about? The path we support is the one listed on the wiki: https://github.com/RetroPie/RetroPie-Setup/wiki/Amiga

We can discuss the prospect of moving the BIOS folder another time as there's already too much going on in this PR. Making an Amiga-specific BIOS folder would mean testing for regressions on *all* Amiga emulators that we support, including the libretro cores, which may necessitate adding unnecessary complexity to the system-specific retroarch.cfg.

If you're dealing with custom configurations that uses an incorrect path, edit the files to use the proper path, or make a symlink from ```BIOS``` to ```BIOS/Amiga```.

The ```conf``` folder is set up correctly as far as I can see.
```
$ ls -l /opt/retropie/emulators/amiberry/conf
lrwxrwxrwx 1 pi pi 41 Feb  5 00:54 /opt/retropie/emulators/amiberry/conf -> /opt/retropie/configs/amiga/amiberry/conf
```

As long as the kickstart is placed in the correct location, a fresh install of Amiberry will work correctly without the need to change any path configuration.

HoraceAndTheSpider:
Perhaps I am not explaining very well, sorry.

It is possible to load a .uae Config file direct from command line, and therefore from ES. 

Therefore many users are placing .uae Config files into `/home/pi/RetroPie/roms/amiga` because this way you can have game-specific setups which look normal on the rom list but take advantage of features such as HD support, JIT, or custom controls. 

It would seem a good idea (for Amiberry at least) to consider directing the Config path to there instead (yes the current one can work but this is suggested as a possible improvement change)

 The user has to option to change the path anyway, I just think this might be simpler for RetroPie users 
(Perhaps it will be irrelevant once LHA file support is added anyway, so as I said it’s not essential, just potential for simplifying the user experience) 


The commonly found .uae file are available online pre made or made using a Python script that is readily available (see the RetroPie forum or my GitHub for more information) ... 

you can select any folder you like for the bios but by default I use an Amiga specific sub folder, as some other emulators do.

I won’t be changing the path back to the main BIOS folder, but that is mostly because i think it is messy, (so there is no point creating a symlink) but maybe it is easier for me to add a change to amiberry to also scan a subfolder if a change won’t be considered here atm.

i appreciate there is maybe uae4all to consider for Pi 1 users but since uae4arm does not seem to be developed I hope it would at least be considered for the future.

psyke83:
I'll merge this for now since @midwan has already made the changes to the master branch.

@HoraceAndTheSpider - I still don't understand why you'd want the default config path to be in ```roms/amiga``` if you can already launch configs that are manually placed there? Or does it not load configurations properly unless you've already changed the path manually? I'll investigate this further when I have more time.

HoraceAndTheSpider:
The main benefit is the following use case:

- user places .uae file into rom/amiga as you have said
- user loads game from ES
- user makes minor change to setting (eg screen size)
- user enters the “paths” panel and saves the Config (by default the config file is saved to amiberry /conf) 

In this scenario the changes made by the user are not saved back to the loading location causing “confusion” as to why new setting is not being applied on future reloading. 

Unfortunately the nature of being able to change Config in-menu means it is not possible to change Amiberry so that the loaded file is saved back to the same location automatically, since a sister could wipe his originally loaded file with settings for one which is later loaded via the GUI. 


The problem is remedied by the setting of the path to the roms/amiga/ location , and will then display all Configs in the paths panel list. 


It’s really not that essential though, (and should be fixed through user knowledge) but is quite a common issue that does occur, and does hamper achieving a “play out of the box” scenario. 

I always recommend to users to set the Config path to the /roms/amiga folder to cover this off, but advice is sometimes missed sadly.

joolswills:
binaries for jessie and stretch have now been updated. thanks.

midwan:
Great work, thanks! :)

PR #2271, opened 2018-01-27
PPSSSPP: switch to master with fullscreen fix

psyke83:
PPSSPP defaults to fullscreen off, and since we have no way to initialize
a default config on first install, use the "--fullscreen" launch argument
for non-X11 platforms.

### DIFF
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -23,7 +23,7 @@ function depends_ppsspp() {
 }
 
 function sources_ppsspp() {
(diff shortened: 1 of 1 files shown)


psyke83:
@joolswills: we need either to switch to master or apply a patch against v1.5.4 to get PPSSPP building on stretch.

Master branch has the issue in which it currently doesn't use the full desktop resolution. There's a genuine bug there (as it ignores the fullscreen setting in the configuration), but even when it's fixed, we still need to do this, as the "fullscreen" option is set as false by default on config init.

joolswills:
As I mentioned on hangouts, I will be applying the patch to the v1.5.4 version. There has been too many breakages with master, and due to the number of platforms to test against, I think it's easiest to manage this way.

PR #2276, opened 2018-01-29
Stretch changes

joolswills:
This PR adds some additional stretch support -

* Enables running retropie_setup on Stretch
* Enables binaries (Almost all binaries are now built)
* Fixes a few packages for Stretch

lr-armsnes does not currently build on gcc 6.0 , due to some inline asm. I believe it should be a simple fix, but I am considering removing this module. It's no longer developed upstream, and had very few optimisations over pocketsnes really.

I'm currently rebuilding all packages again to make sure there are no issues with the CPU flags changes for the rpi1.


lilbud:
Would Stretch provide any benefits over Jessie? Or would it be pointless to update?

joolswills:
Please use the forum for discussion.

PR #2279, opened 2018-01-30
lr-beetle-psx: include hw core

psyke83:
Tagging @movisman - please test this PR and verify that the hw core builds & runs as expected.

As Jools has stated on the forum, it needs to be confirmed as working by someone in the project before being included. I currently don't have an x86 Linux installation available, and Jools is busy, so you'll need to be patient until I can sort out a testing environment.

Are you absolutely certain that Vulkan works? I would imagine that RetroArch itself needs to be built with Vulkan support enabled -  and we don't take any special steps to enable vulkan there.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -18,7 +18,7 @@ function depends_retroarch() {
     local depends=(libudev-dev libxkbcommon-dev libsdl2-dev libasound2-dev libusb-1.0-0-dev)
     isPlatform "rpi" && depends+=(libraspberrypi-dev)
     isPlatform "mali" && depends+=(mali-fbdev)
-    isPlatform "x11" && depends+=(libx11-xcb-dev libpulse-dev libavcodec-dev libavformat-dev libavdevice-dev)
+    isPlatform "x11" && depends+=(libx11-xcb-dev libpulse-dev libavcodec-dev libavformat-dev libavdevice-dev libvulkan-dev)
 
     # only install nvidia-cg-toolkit if it is available (as the non-free repo may not be enabled)
     if isPlatform "x86"; then
@@ -47,6 +47,7 @@ function build_retroarch() {
     isPlatform "kms" && params+=(--enable-kms)
     isPlatform "arm" && params+=(--enable-floathard)
     isPlatform "neon" && params+=(--enable-neon)
+    isPlatform "x11" && params+=(--enable-vulkan)
     ./configure --prefix="$md_inst" "${params[@]}"
     make clean
     make
--- a/scriptmodules/libretrocores/lr-beetle-psx.sh
+++ b/scriptmodules/libretrocores/lr-beetle-psx.sh
@@ -16,19 +16,26 @@ rp_module_licence="GPL2 https://raw.githubusercontent.com/libretro/beetle-psx-li
 rp_module_section="opt"
 rp_module_flags="!arm"
 
+function depends_lr-beetle-psx() {
+    local depends=(libvulkan-dev libgl1-mesa-dev)
+    getDepends "${depends[@]}"
+}
+
 function sources_lr-beetle-psx() {
     gitPullOrClone "$md_build" https://github.com/libretro/beetle-psx-libretro.git
 }
 
 function build_lr-beetle-psx() {
     make clean
-    make
-    md_ret_require="$md_build/mednafen_psx_libretro.so"
+    make HAVE_HW=1
+    md_ret_require=(
+        'mednafen_psx_hw_libretro.so'
+    )
 }
 
 function install_lr-beetle-psx() {
     md_ret_files=(
-        'mednafen_psx_libretro.so'
+        'mednafen_psx_hw_libretro.so'
     )
 }
 
@@ -36,6 +43,6 @@ function configure_lr-beetle-psx() {
     mkRomDir "psx"
     ensureSystemretroconfig "psx"
 
-    addEmulator 0 "$md_id" "psx" "$md_inst/mednafen_psx_libretro.so"
+    addEmulator 0 "$md_id" "psx" "$md_inst/mednafen_psx_hw_libretro.so"
     addSystem "psx"
 }
--- a/scriptmodules/supplementary/configedit.sh
+++ b/scriptmodules/supplementary/configedit.sh
@@ -231,7 +231,7 @@ function advanced_configedit() {
         'input_overlay_enable true false'
         "input_overlay _file_ *.cfg $rootdir/emulators/retroarch/overlays"
         "audio_driver $audio_opts"
-        'video_driver gl dispmanx sdl2 vg'
+        'video_driver gl dispmanx sdl2 vg vulkan'
         'menu_driver rgui xmb'
         'video_fullscreen_x _string_'
         'video_fullscreen_y _string_'


psyke83:
OK, I saw the thread in which you participated: https://retropie.org.uk/forum/topic/15914/x86-vulkan-driver-option-for-various-emulators-not-showing-up-but-dolphin-works/7

RetroArch clearly needs vulkan support as well, so I've added the changes to ensure it builds the vulkan driver on x11 platforms.

movisman:
Hi there,

Thanks a bunch for looking at this. I do appreciate the time!

So, from my side I already had the dependencies installed for Beetle HW Support (libvulkan-dev and libgl1-mesa-dev). I just found running "make HAVE_VULKAN=1 HAVE_OPENGL=1" compiled me a hardware version and this seems to work absolutely fine.

I must admit though, I did not know there would be a requirement to rebuild Retroarch to ensure Vulkan support.

However, Beetle was running in Vulkan, OpenGL and Software modes fine for me - I could tell Vulkan was enabled because the Vulkan specific options were taking effect.

You do have to make sure you either specify the Vulkan driver in the PSX retroarch.cfg file though (video_driver = "vulkan"), or apply it to the main retroarch.cfg file, otherwise the emulator falls back to software mode. It may be worth considering having 'vulkan' as one of the video driver options in the retroarch config editor (advanced options) for the setup script if possible, as at the moment I think you can only select from GL, dispmanx, SDL2, or vg. I just edited the file directly but maybe it's worth having in there as an option.

Would you still like me to test those commits? I am happy to back up my other two .sh files and overwrite with these, and the update/rebuild Beetle and Retroarch via the setup script and see if it all works. I am not running x86, my machine is x64, but it is running Ubuntu MATE 16.04 LTS so it should be ok/supported. I already have all the dependencies needed, but I haven't built Retroarch with the Vulkan flag.

This comment here:
https://www.reddit.com/r/emulation/comments/7hxke8/psa_psx_beetle_hw_is_absolutely_amazing/dquv1jp/

Confirms that the non-HW and HW are the same core. I think some people are confused by it because when Retroarch is installed both are available to choose from.

Would you suggest the best plan is to simply have a single build available via the optional/experimental installers, seeing as that covers all (software and hardware)? I don't think x86 users would have any need for the software only build, especially looking at that comment. If I am mistaken though, maybe both could be installed like Retroarch does by default, and you could choose which core to launch?

Thanks!

psyke83:
Thanks. I noticed upstream bugs for issues only with the HW core, but if it's the case that the hw and regular core are identical when you set software rendering in the hw core, it doesn't really make sense to distribute both. I'll update the PR and will also take a look at the retroarch config editor.

psyke83:
OK, I've added vulkan to the configedit options, but I won't remove the old core. I understand that it's possible to enable software mode in the HW core,  but it's possible that some users may be running on a system that doesn't support GL or Vulkan, which may result in the HW core not being able to successfully initialize to the point where they can actually modify the option. We try to support a broad range of configurations when possible. You're still free to use the -hw core and toggle between vulkan/gl/software in that single core.

Please test the PR as-is now. Make sure to undo all your local changes, then rebuild both retroarch and lr-beetle-psx with the PR applied to ensure it's working as intended.

movisman:
Hi there,

Thanks a lot for this.

Just a couple of other things to ask/note, when you add the vulkan flag for the retroarch build do you have any idea what this does? Because when I run retroarch --features it indicates Vulkan is already supported and I didn't rebuild it originally.

And regarding keeping the HW and SW core available, this does make sense - however in the core options for the HW core, there is this option: beetle_psx_hw_renderer_software_fb = "enabled"

Which I believe should mean, if Vulkan or OpenGL is not supported at all on a users system, it falls back to Software. So maybe both are not required after all?

However, I get for compatibility purposes it might be best to keep them both available. I guess the slight advantage to this too is that you can have a separate set of core options for software emulated core and the hardware one.

The third thing, which is slightly more embarrassing, is my Github knowledge is really very basic, so i've not tested anyone's PR before. As it is just a couple of files changed, I was going to backup my originals and simply overwrite the content. But what is the correct way to test a PR?

Thanks a lot.

psyke83:
First: You rebuilt retroarch after installing libvulkan-dev - correct? 

If you run ```./configure --help``` on the retroarch source, it'll show the available options along with the default setting when no argument is passed. As long as libvulkan-dev is installed, Vulkan will be compiled by default:
```
  --enable-vulkan             Enable   Vulkan support
```
So, adding the configure flag is not strictly necessary, but I added it just to make absolutely sure Vulkan is selected  (also useful if retroarch later changes the default build configuration).

Second: it may be the case that it can select the software fallback automatically in the case where OpenGL or Vulkan are not available, but that assumption needs to be tested to ensure that we're not  limiting options for users that are running on such a configuration. Even though the core is blacklisted for arm, I suppose that the Pi might be a good candidate to test this hypothesis, since neither GL nor Vulkan is available with the standard vendor graphics driver. I can test that and see if the hw core launches, but I'll also want to test on a native x86/x86_64 distro.

For now, it will be helpful if you can test the PR and ensure that the hw and regular cores work as intended.

Third: you can test this PR with this procedure:
```
cd RetroPie-Setup
git remote add psyke83 https://github.com/psyke83/RetroPie-Setup
git fetch psyke83
git stash
git checkout psyke83/beetle_hw
```
To go back to the regular branch:
```
cd RetroPie-Setup
git checkout master
git stash pop
```

The ```stash```/```stash pop``` are needed to hide/restore your local modifications. If you manually clean your source before testing, those steps are unnecessary.

movisman:
Hi there,

Thanks a lot again for the help.

Yep, so on the first point, I did indeed rebuild Retroarch from source after installing libvulkan-dev, so this totally explains why Vulkan works for me even without the flag. And makes total sense to have it in there anyway.

Regarding the second point, yep that all makes sense too. Does Beetle-PSX even work on the Pi? If it compiles and runs (but just runs poorly) I guess you can indeed test that the HW core launches and falls back to a software render in the absence of OpenGL and Vulkan.

But yep - I will test your PR and let you know if both HW and regular cores work as intended. Cheers for writing out the procedure to fetch your PR, I appreciate that.

Thanks!

psyke83:
It does work on the Pi - albeit very slowly. I haven't yet tested if the HW core initializes properly, but I can try it later tonight.

The scriptmodule blocks it on Pi and other arm targets, but I can manually remove the block to test. If the hw core also functions on Pi in software mode, it's likely that the software fallback will also work OK on regular x86* systems.

Perhaps if you tested with a custom xorg.conf that uses the vesa driver instead of your vendor driver, or change the configuration to disable opengl/vulkan. I admit that I haven't messed with an xorg.conf for a long time, as I've been using Linux in VMs for the past several years.

movisman:
I'm doing all my config on my live physical machine which is hopefully going to become a little box which pretty much only runs RetroPie, so i'll probably want to take some config backups before I start messing with xorg.conf, etc.

This is also a project for me to learn a little bit of Linux, my first foray into it has been over Christmas when I got this OS installed, and i've been working on getting everything working nicely in RetroPie since and plan to end up with a machine that boots into ES similar to how the Pi does. I've documented everything i've done so far, and will probably end up sharing it in case any other users find it useful.

Regarding the PR, I performed the steps as you mentioned, and the extra config option appears for Vulkan, which is good, and there are two cores now and emulators.cfg all looks correct, all good - but the HW emulator just loads the regular core for some reason.

When comparing the files, indeed the HW one is larger than the regular one (3.8mb instead of 3mb), but it is not as big as the one I was using before (the one I compiled myself), which is 4.2mb. When I load this one I get the hardware options.

Just to clarify, I am a bit novice with github but I found to get it working I had to clone the beetle repository, and then I ran:
make HAVE_VULKAN=1 HAVE_OPENGL=1

This gave me the 4.2mb file which then works.

I see the .so script you adjusted uses HAVE_HW but that should work according to the code in the makefile on their repository as if you have it - it sets those values to 1 anyway. Like I said, I am not too up on github and code stuff but not sure if you can look at this?

Cheers!

psyke83:
It seems that the hw core doesn't build correctly when the build tree is not clean: https://bugs.archlinux.org/task/56827

So, I've changed the PR to switch to the hw core instead of providing both. I tested on Pi, and can confirm it still initializes properly. The hw core is now being built correctly, as I can see the vulkan/opengl/software core options.

Please retest. You'll need to reset your local branch due my having done a force-push on my remote branch. This should be enough:
```
git fetch psyke83
git reset
```

movisman:
Hi again,

Yep, it looks good to me - it launches just as before and the file is 4.2mb, the same as the one I generated myself before.

You still have to specify Vulkan as a core override in retroarch.cfg, or at a higher level, even if you select Vulkan as the renderer in the core. The two have to match otherwise it looks like it falls back to software.

The only thing I noticed, and i'm not sure if this is a side effect from the testing i've done, but when installing the Beetle core, looking at the script I was expecting it to remove the non-HW version from the list, but it hasn't done so.

See the attached emulators.cfg and you'll see what I mean. If you are only using one core going forward, can this get cleaned up so there is just one to choose from? It is possible it is something my side left over from the PR from earlier.

Thanks!

[emulators.cfg.zip](https://github.com/RetroPie/RetroPie-Setup/files/1679927/emulators.cfg.zip)

psyke83:
Don't worry about the orphaned ```lr-beetle-psx-hw``` entry - only you and I will have that issue from testing an earlier version of the PR. Feel free to manually delete it from the file.

If/when the PR is merged, users' existing ```lr-beetle-psx``` entry will be modified to point to the new hw core without creating a new entry, so it'll be fine.

psyke83:
I'll go ahead and merge this. You verified that the hw core is working as expected on a Vulkan-enabled system, and I was able to verify that the software fallback works on the Pi. Thanks for the assistance with testing.

movisman:
Hi there,

Yes, this is what I figured about the orphaned entry in emulators.cfg, I just thought i'd mention it just in case.

Just to double check all is well on that front, I cleaned up my emulators.cfg, and removed the Beetle package. Reinstalling it added the lr-beetle-psx entry in correctly, and removing the package removes it, all as intended. So all good there. The config addition is there, and the core launches as expected with no issues.

And indeed, GL is the default for RetroArch, so it'd be down to the user to change their video_driver either globally or within PSX specific retroarch.cfg to Vulkan, if they want to make any use of that renderer. Otherwise it will not use it even if selected within the core. Me personally, I have my PSX retroarch.cfg set to use the Vulkan video driver, and I also set the shader path to a .SLANGP shader that works with Vulkan (the default shader I use is a GLSLP one, which does not load when Vulkan is being utilised because those types are incompatible). All other cores then use GL and my standard shader. Only Vulkan enabled cores I will set overrides for.

Thanks a bunch for the help, input and time to incorporate and merge these changes. I now need to figure out a little more Git stuff as it is all fairly new to me, just in case I need to test any other PR's further down the road. Do I have to run any clean up or anything like that for the PR/remote repository I was testing from to keep everything tidy?

Thanks!

psyke83:
Return to the master branch via ```git checkout master```, and ```git stash pop``` if you had changes stashed - but you should undo any similar modification you've made related to this PR before updating the source, otherwise there will be a conflict when you update the script.

If you want to fully remove the ```psyke83``` remote and delete the PR branch:

```
git remote -D psyke83
git branch -D beetle_hw
```

movisman:
Hi,

Yep, this morning I already returned to the master branch with git checkout master, and I looked up on the web about removing the remote repository, so I did that too. When running the git branch command above, beetle_hw was not found so I think all is already cleaned up? And running a git branch -v shows only the master.

Regarding the git stash and pop commands, when I ran these initially before testing the PR, i just got 'no local changes to save' so those didn't do anything for me. I hadn't changed any element of the setup script itself so i'm guessing that's why. I still need to learn a bit more about all of this so I might not be correct.

I updated the script all installed cores and there were no errors. So I think all is ok.

Cheers

PR #2281, opened 2018-02-01
Update ir-mame2003-plus.sh

arcadez:
As i've now split off mame2003-plus to be a separate standalone core, so therefore i need to make
some changes here also.

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003-plus.sh
+++ b/scriptmodules/libretrocores/lr-mame2003-plus.sh
@@ -12,21 +12,61 @@
 rp_module_id="lr-mame2003-plus"
 rp_module_desc="Arcade emu - updated MAME 0.78 port for libretro with added game support"
 rp_module_help="ROM Extension: .zip\n\nCopy your MAME roms to either $romdir/mame-libretro or\n$romdir/arcade"
-rp_module_licence="NONCOM https://raw.githubusercontent.com/libretro-mirrors/mame2003-plus-libretro/master/docs/mame.txt"
-rp_module_section="exp"
+rp_module_licence="NONCOM https://github.com/arcadez/-mame2003-plus-libretro/blob/master/docs/mame.txt"
+rp_module_section="main"
 
 function sources_lr-mame2003-plus() {
-    gitPullOrClone "$md_build" https://github.com/libretro-mirrors/mame2003-plus-libretro.git
+    gitPullOrClone "$md_build" https://github.com/arcadez/-mame2003-plus-libretro
 }
 
 function build_lr-mame2003-plus() {
-    build_lr-mame2003
+    rpSwap on 750
+    make clean
+    local params=()
+    isPlatform "arm" && params+=("ARM=1")
+    make ARCH="$CFLAGS" "${params[@]}"
+    rpSwap off
+    md_ret_require="$md_build/mame2003-plus_libretro.so"
 }
 
 function install_lr-mame2003-plus() {
-    install_lr-mame2003
+    md_ret_files=(
+        'mame2003-plus_libretro.so'
+        'README.md'
+        'changed.txt'
+        'whatsnew.txt'
+        'whatsold.txt'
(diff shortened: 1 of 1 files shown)


scriptmodules/libretrocores/lr-mame2003-plus.sh
```diff
@@ -12,21 +12,61 @@
 rp_module_id="lr-mame2003-plus"
 rp_module_desc="Arcade emu - updated MAME 0.78 port for libretro with added game support"
 rp_module_help="ROM Extension: .zip\n\nCopy your MAME roms to either $romdir/mame-libretro or\n$romdir/arcade"
-rp_module_licence="NONCOM https://raw.githubusercontent.com/libretro-mirrors/mame2003-plus-libretro/master/docs/mame.txt"
-rp_module_section="exp"
+rp_module_licence="NONCOM https://github.com/arcadez/-mame2003-plus-libretro/blob/master/docs/mame.txt"
+rp_module_section="main"
```
joolswills: this will need to stay in experimental for now.

scriptmodules/libretrocores/lr-mame2003-plus.sh
```diff
@@ -12,21 +12,61 @@
...
 
 function sources_lr-mame2003-plus() {
-    gitPullOrClone "$md_build" https://github.com/libretro-mirrors/mame2003-plus-libretro.git
+    gitPullOrClone "$md_build" https://github.com/arcadez/-mame2003-plus-libretro
```
joolswills: why the hyphen repo prefix in the repo name btw?

scriptmodules/libretrocores/lr-mame2003-plus.sh
```diff
@@ -12,21 +12,61 @@
...
 
 function build_lr-mame2003-plus() {
-    build_lr-mame2003
+    rpSwap on 750
+    make clean
+    local params=()
+    isPlatform "arm" && params+=("ARM=1")
+    make ARCH="$CFLAGS" "${params[@]}"
+    rpSwap off
+    md_ret_require="$md_build/mame2003-plus_libretro.so"
```
joolswills: this should still use the parent module function with some changes to avoid duplication (as with the configure).

I will do the code changes to the module as it will be quicker than going to and fro with the PR.

scriptmodules/libretrocores/lr-mame2003-plus.sh
arcadez: Ok then

scriptmodules/libretrocores/lr-mame2003-plus.sh
arcadez: When i forked the plus mirror the git did the naming

scriptmodules/libretrocores/lr-mame2003-plus.sh
arcadez: Ok and thanks

scriptmodules/libretrocores/lr-mame2003-plus.sh
joolswills: That's strange. Never seen that before when forking.

Are you planning on keeping the libretro github up to date or is your repo meant to replace it ?

arcadez: Hmm come to think of it i might have renamed it after forking it was a while ago and well
,my memory ain't what it was :)

With regards to your question i have maintained both cores simultaneously they lineup codewise 
the only difference being just incase i made a mistake with the folder renaming i've not added those changes to the mame2003-plus mirror as yet.

Anyway look as i said with regards to the retropie setup files feel free to hook up with whatever
core you would prefer it doesn't have to be mine just lemmie know if you choose the mirror
and i'll update it for that purpose.

Regards!!

joolswills:
Created a new PR https://github.com/RetroPie/RetroPie-Setup/pull/2307

PR #2282, opened 2018-02-04
wifi: fixes for stretch

psyke83:
* Use ifconfig invocations instead of ifup/ifdown
* Ensure wlan0 is up before initiating scan
* Fix display of IP address & only show wlan0 interface address

Also fix IP address display in retropiemenu.

### DIFF
--- a/scriptmodules/supplementary/wifi.sh
+++ b/scriptmodules/supplementary/wifi.sh
@@ -14,9 +14,23 @@ rp_module_desc="Configure Wifi"
 rp_module_section="config"
 rp_module_flags="!x11"
 
+function _set_interface_wifi() {
+    local state="$1"
+
+    if [[ "$state" == "up" ]]; then
+        if ! ifup wlan0; then
+            ip link set wlan0 up
+        fi
+    elif [[ "$state" == "down" ]]; then
+        if ! ifdown wlan0; then
+            ip link set wlan0 down
+        fi
+    fi
+}
+
 function remove_wifi() {
     sed -i '/RETROPIE CONFIG START/,/RETROPIE CONFIG END/d' "/etc/wpa_supplicant/wpa_supplicant.conf"
-    ifdown wlan0 &>/dev/null
+    _set_interface_wifi down 2>/dev/null
 }
 
 function list_wifi() {
@@ -44,6 +58,8 @@ function connect_wifi() {
     local type
     local options=()
     i=0
+    _set_interface_wifi up 2>/dev/null
+    sleep 1
     while read essid; read type; do
         essids+=("$essid")
         types+=("$type")
@@ -132,8 +148,11 @@ _EOF_
 }
 
 function gui_connect_wifi() {
-    ifdown wlan0 &>/dev/null
-    ifup wlan0 &>/dev/null
+    _set_interface_wifi down 2>/dev/null
+    _set_interface_wifi up 2>/dev/null
+    # BEGIN workaround for dhcpcd trigger failure on Raspbian stretch
+    systemctl restart dhcpcd &>/dev/null
+    # END workaround
     dialog --backtitle "$__backtitle" --infobox "\nConnecting ..." 5 40 >/dev/tty
     local id=""
     i=0
@@ -142,14 +161,18 @@ function gui_connect_wifi() {
         id=$(iwgetid -r)
         ((i++))
     done
-    [[ -z "$id" ]] && printMsgs "dialog" "Unable to connect to network $essid"
+    if [[ -z "$id" ]]; then
+        printMsgs "dialog" "Unable to connect to network $essid"
+        _set_interface_wifi down 2>/dev/null
+    fi
 }
 
 function gui_wifi() {
     local default
     while true; do
-        local ip_int=$(ip route get 8.8.8.8 2>/dev/null | head -1 | cut -d' ' -f8)
-        local cmd=(dialog --backtitle "$__backtitle" --cancel-label "Exit" --item-help --help-button --default-item "$default" --menu "Configure WiFi\nCurrent IP: $ip_int\nWireless ESSID: $(iwgetid -r)" 22 76 16)
+        local ip_current=$(ip route get 8.8.8.8 2>/dev/null | awk '{print $NF; exit}')
+        local ip_wlan=$(ip route ls dev wlan0 2>/dev/null | awk 'END {print $7}')
+        local cmd=(dialog --backtitle "$__backtitle" --cancel-label "Exit" --item-help --help-button --default-item "$default" --menu "Configure WiFi\nCurrent IP: $ip_current\nWireless IP: $ip_wlan\nWireless ESSID: $(iwgetid -r)" 22 76 16)
         local options=(
             1 "Connect to WiFi network"
             "1 Connect to your WiFi network"
--- a/scriptmodules/supplementary/retropiemenu.sh
+++ b/scriptmodules/supplementary/retropiemenu.sh
@@ -150,7 +150,7 @@ function launch_retropiemenu() {
             mc
             ;;
         showip.rp)
-            local ip="$(ip route get 8.8.8.8 2>/dev/null | head -1 | cut -d' ' -f8)"
+            local ip="$(ip route get 8.8.8.8 2>/dev/null | awk '{print $NF; exit}')"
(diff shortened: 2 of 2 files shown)


HerbFargus:
Thanks I'll test this out. Also noticed raspi-config has an equivalent wifi config now though you have to manually put in the ssid since it doesn't scan for networks. Also has an option for predictible network interface names. Not sure if/how that will affect the wifi module if someone enables it from raspi-config

HerbFargus:
Can confirm it worked on both my Jessie and stretch builds. Didn't test with the wifi import txt file though

psyke83:
Thanks. Enabling predictable interface names has the effect of renaming ```eth0``` to ```enxb827ebb14c90```, but my ```wlan0``` remains unchanged. Regardless, it seems to break wifi connectivity, at least when using our scriptmodule. I'll see if this can be easily fixed soon.

Is the wifi import still useful? It seems to be duplicating existing functionality, as recent firmware versions automatically copy /boot/wpa_supplicant.conf to the appropriate location during boot as a means to allow automatic wireless configuration after flashing (similar to /boot/ssh).

HerbFargus:
>Is the wifi import still useful? 

Imo yes in cases like my handheld where I can enable the wifi without the use of a keyboard.

Or are you saying users should just add /boot/wpa_supplicant.conf instead as it's a more automated process without needing to go through the setup script to import the file?

psyke83:
Actually, wifi *does* work when predictive names are enabled. I must have entered the wrong PSK by accident or something the last time, as it seems to work fine now.

If you create a wpa_supplicant.conf and put it in /boot, it gets automatically applied. This is a feature of Raspbian built into the raspberrypi-net-mods package (which is installed by default in the lite and full images): https://github.com/RPi-Distro/raspberrypi-net-mods/blob/master/debian/raspberrypi-net-mods.service

The only disadvantage is that you need to ensure the configuration is valid, i.e., it contains the proper configuration preceding the network array, i.e.:
```
ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev
update_config=1
country=GB

network={
        ssid="..."
        psk="..."
}
```

Not a big deal since all you need to do is have a valid copy of your wpa_supplicant.conf config prepared before flashing (which you can copy verbatim from a running image).

HerbFargus:
>The only disadvantage is that you need to ensure the configuration is valid, i.e., it contains the proper configuration preceding the network array, i.e.:

This isn't a concern to me so long as it is documented properly. I don't mind if the option is removed from the wifi module for a stretch release but for the time being it might be worth keeping for people still on jessie. Would be good to get jools feedback, but the new wpa_supplicant method is favourable to me.

psyke83:
I was just pointing out that it's (at least with new firmwares) a duplicate feature. I won't remove it in this PR, as our version might be a bit more resilient. 

The disadvantage of the official method is that the section of the wpa_supplicant.conf preceding the actual network configuration varies between jessie and stretch, so copying a wpa_supplicant.conf from a jessie installation for use on stretch would probably break wireless.

psyke83:
Unfortunately I had to update the PR with a minor kludge. If you boot stretch with no active wifi configuration, the dhcpcd service is in a strange state:
```
Warning: dhcpcd.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

When attempting to bring up wlan0 via the wifi menu, the proper dhcpcd hooks for the wireless interface are never triggered, so an IP address doesn't get assigned. Invoking a systemctl daemon-reload and restarting dhcpcd seems to kick it into proper operation.

Fortunately, if the system boots with a configured wpa_supplicant.conf, wlan0 will be brought up correctly, so this little kludge can be isolated to the wifi menu as opposed to being invoked during every boot.

psyke83:
Closing for a bit until I sort some stuff out.

psyke83:
OK, now tested and confirmed to be working fine on both jessie and stretch.

scriptmodules/supplementary/wifi.sh
```diff
@@ -16,7 +16,7 @@ rp_module_flags="!x11"
 
 function remove_wifi() {
     sed -i '/RETROPIE CONFIG START/,/RETROPIE CONFIG END/d' "/etc/wpa_supplicant/wpa_supplicant.conf"
-    ifdown wlan0 &>/dev/null
+    ifdown wlan0 &>/dev/null || ip link set wlan0 down &>/dev/null
```
joolswills: if the second case works on jessie too, we can just use that rather than trying both ?

scriptmodules/supplementary/wifi.sh
```diff
@@ -142,13 +147,16 @@ function gui_connect_wifi() {
...
+        printMsgs "dialog" "Unable to connect to network $essid"
+        ifdown wlan0 &>/dev/null || ip link set wlan0 down &>/dev/null
+    fi
 }
 
 function gui_wifi() {
     local default
     while true; do
-        local ip_int=$(ip route get 8.8.8.8 2>/dev/null | head -1 | cut -d' ' -f8)
+        local ip_int=$(ip route ls dev wlan0 2>/dev/null | awk 'END {print $7}')
```
joolswills: This should show the current IP, not the current wlan0 interface ip. eg an ethernet cable could be connected. I assume the above change would not show the LAN ip if wifi was off ?

scriptmodules/supplementary/wifi.sh
joolswills: also is &>/dev/null actually needed for the ip link set command ? might just be ok to redirect stderr.

scriptmodules/supplementary/wifi.sh
joolswills: if we are going to switch to showing just the wlan0 ip address, the wording should change. Or maybe we can show both. Not sure I want this change for now though.

HerbFargus: >Or maybe we can show both

I like the idea of showing both. granted showing ethernet stuff in the wifi module is a bit misleading, or at the very least have it show them with the showip module in the retropie menu. perhaps the module should have been called networking instead of wifi ;)

scriptmodules/supplementary/wifi.sh
psyke83: No, the second case doesn't work on jessie. 

On stretch, the wlan0 device is not enumerated in /etc/network/interfaces, and the dhcpcd service is supposed to automatically trigger wpa_supplicant as needed. On jessie, wlan0 is configured via /etc/network/interfaces and wpa_supplicant gets triggered via that, and simply bringing up the interface (or also restarting dhcpcd) is not enough to establish a connection, as wpa_supplicant doesn't trigger via dhcpcd.

joolswills: ok thanks.

scriptmodules/supplementary/wifi.sh
psyke83: I'm happy to either clarify the text or revert.

Maybe we could have:
```
local ip_int $(ip route ls dev wlan0 2>/dev/null | awk 'END {print $7}')
local ip_route $(ip route get 8.8.8.8 2>/dev/null | awk '{print $NF; exit}')"
```
Then print:
```
Current wireless IP: $ip_int (default route: $ip_route)
```

Better or just leave it to the previous behaviour?

joolswills: I'll merge as is for now.

scriptmodules/supplementary/wifi.sh
joolswills: actually, it would make sense then do have a function which does the jessie or stretch method and call that - would de-duplicate the code a bit.

psyke83: I was considering that, but figured that it's potentially not a problem isolated to Raspbian jessie -> stretch. Other boards may benefit from trying both the regular ifup and then trying the fallback.

scriptmodules/supplementary/wifi.sh
joolswills: Happy for your change above, although ip_wlan would be better var name no (int = internal ? the ethernet ip address is internal also).

scriptmodules/supplementary/wifi.sh
joolswills: it can still be in a function to de-duplicate.

joolswills:
Thanks for this.

PR #2283, opened 2018-02-04
Updated Commander Genius

gerstrong:


gerstrong:
This updates Commander Genius to a much better version (2.0.5)

psyke83:
Hi @gerstrong. I'd like to review this, but it seems that you've pulled in an unnecessary merge commit. Can you try updating the RetroPie upstream origin and rebasing your PR branch against it to eliminate the merge?

HerbFargus:
I thought we were pulling from master which should already be the latest?

gerstrong:
It is the latest stabile 2.0.5There is 2.1 but this ist very instabile still...Am 04.02.2018 8:55 nachm. schrieb HerbFargus <notifications@github.com>:I thought we were pulling from master which should already be the latest?

psyke83:
@gerstrong,

If you look at the contents of the PR that's being proposed, most of your commit is in conflict with what's already present in our repository. For example: https://github.com/RetroPie/RetroPie-Setup/commit/61ca248f14d2fbaeea52f140dbe5fb22d934894c#diff-54c0e665cf7039cdbb2ecf9167aa7a08

The current scriptmodule, without this PR, is going to build from the master branch of https://github.com/gerstrong/Commander-Genius.git. I can see you added a ```v205Release``` tag to that repo, so unless we add that tag to the scriptmodule, it will always try to build from the latest master.

I suggest that you checkout a fresh copy of the RetroPie remote's master branch and apply any necessary changes, because the current PR is not suitable for merging.

joolswills:
I will add the missing dependency and switch to the v206Release tag

joolswills:
I'm having some trouble with the 2.0.6 version cmake - will open a ticket at your github.

joolswills:
https://github.com/gerstrong/Commander-Genius/issues/322

gerstrong:
got it fixed! Thanks for reporting

joolswills:
Thanks - will you be doing another tagged release we can point to ?

BTW if you had a stable branch you maintained (eg called 2.0.x) we could track that, then people could update to the latest stable easily without us updating the RetroPie module.

joolswills:
Have a linking issue with the last release - for reference - https://github.com/gerstrong/Commander-Genius/issues/324

PR #2286, opened 2018-02-08
add Digger Remastered to experimental

proyvind:
This is my digger branch which adds cmake & joystick support. :)

### DIFF
--- a/retropie_packages.sh
+++ b/retropie_packages.sh
@@ -9,7 +9,7 @@
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
 #
 
-__version="4.3.10"
+__version="4.3.11"
 
 [[ "$__debug" -eq 1 ]] && set -x
 
--- a/scriptmodules/ports/cgenius.sh
+++ b/scriptmodules/ports/cgenius.sh
@@ -15,11 +15,11 @@ rp_module_licence="GPL2 https://raw.githubusercontent.com/gerstrong/Commander-Ge
 rp_module_section="exp"
 
 function depends_cgenius() {
-    getDepends build-essential cmake libcurl4-openssl-dev libvorbis-dev libogg-dev libsdl2-dev libsdl2-image-dev libboost-dev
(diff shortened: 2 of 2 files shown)


scriptmodules/ports/digger.sh
```diff
@@ -0,0 +1,41 @@
...
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="digger"
+rp_module_desc="Digger Remastered"
+rp_module_licence="GPL https://raw.githubusercontent.com/sobomax/digger/master/README.md"
+rp_module_section="exp"
+
+function depends_digger() {
+    getDepends cmake libsdl2-dev
```
psyke83: It looks like you need to add zlib1g-dev:
```
pi@retropie:~/RetroPie-Setup $ readelf -d /opt/retropie/ports/digger/digger  | grep lib
 0x00000001 (NEEDED)                     Shared library: [libSDL2-2.0.so.0]
 0x00000001 (NEEDED)                     Shared library: [libz.so.1]
 0x00000001 (NEEDED)                     Shared library: [libm.so.6]
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
```

scriptmodules/ports/digger.sh
```diff
@@ -0,0 +1,41 @@
...
+}
+
+function sources_digger() {
+    gitPullOrClone "$md_build" https://github.com/proyvind/digger.git joystick
+}
+
+function build_digger() {
+    cmake . -DCMAKE_INSTALL_PREFIX="$md_inst"
+    make
+    md_ret_require="$md_build"
```
psyke83: This should point to the digger binary so it can send the appropriate return code when the file is missing (i.e. build fails), otherwise it will try to proceed to the install/configure phases erroneously.

joolswills: I noticed a module with this the other day - I think one contributed one had it, then there was some c&p. There are 4 currently like this I am going to fix.

PR #2290, opened 2018-02-10
retroarch - enable ffmpeg on Raspbian Stretch or above (and recent Ub…

joolswills:
…untu)


joolswills:
binaries have been updated

PR #2294, opened 2018-02-14
pegasus-fe: patch vendor graphics libraries

psyke83:
Needed until binary packages are updated to link against the correct
libbrcm*.so libraries from new firmware revisions.

(Hopefully doesn't need merging if the author can fix the binaries soon)


mmatyas:
I'm on it!

joolswills:
Closing as this is fixed in the upstream repo. Thanks @mmatyas

PR #2308, opened 2018-02-22
resolve error + one other update to mame2010 build

markwkidd:
I updated the mame2010 readme and switched to markdown to enjoy some formatting. This creates an error as reported by @mediamogul here: https://retropie.org.uk/forum/post/133826

Pending the approval of my next PR ( https://github.com/libretro/mame2010-libretro/pull/94 ) mame2010 will autogenerate a mame.ini file if there isn't one -- therefore I'm suggesting that `mame.ini` from this script be removed as well.

In this PR I have also omitted the original mame 0.139 "whatsnew" file which seems out of place. No offense if you reject this change.

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2010.sh
+++ b/scriptmodules/libretrocores/lr-mame2010.sh
@@ -34,7 +34,6 @@ function install_lr-mame2010() {
     md_ret_files=(
         'mame2010_libretro.so'
         'README.md'
-        'mame.ini'
     )
 }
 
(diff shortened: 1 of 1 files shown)


estefan3112:
lr-MAME2010 now installs in Retropie/Ubuntu 16.04 from source, but every game crashes (they all work with lr-MAME2003). Ubuntu tells me that "This package does not seem to be installed correctly." Can I help further?

joolswills:
Please use the forum for support.

PR #2310, opened 2018-02-22
Full game description in .rvm for scraper to hash

adelpha:
This change makes the script write the full game descrption (e.g. "Grim Fandango (Windows/English)") into the .rvm files that ES uses to generate the gamelist. The intent is to give .rvm files unique hashes that can be used by scrapers. Exactly the same as recent PR to scummvm.sh.


adelpha:
I wrote .svm on the commit name like an idiot (too much ScummVM recently) but the code is correct.

PR #2312, opened 2018-02-24
Fix OpenBlok build warning

mmatyas:
The portable installation flag was slightly incorrect, which produced a warning during the build and set up a different directory structure (the game otherwise worked fine). This patch makes the program truly portable.


joolswills:
Thanks

PR #2313, opened 2018-02-24
mame2003 now already includes hiscore.dat

markwkidd:
`hiscore.dat` is compiled into the mame2003 core itself starting with this commit: https://github.com/libretro/mame2003-libretro/commit/9e397494e0b55a1c11494959b0e23946508ab38b

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -57,8 +57,8 @@ function configure_lr-mame2003() {
     mkUserDir "$biosdir/mame2003/samples"
 
     # copy hiscore.dat
-    cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/mame2003/"
-    chown $user:$user "$biosdir/mame2003/"{hiscore.dat,cheat.dat}
+    cp "$md_inst/metadata/"{cheat.dat} "$biosdir/mame2003/"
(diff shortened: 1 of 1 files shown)


joolswills:
Why was it embedded? technically before the hiscore.dat was customisable. Not my call anyway, but not sure embedding things like this is a good idea.

Not going to merge as is, as there is no need for the `{}` if copying a single file. I will make the change.

joolswills:
Also, the install is shared with the lr-mame-2003-plus so I will need to make sure that has the change also, or the scriptmodule will need further changes.

markwkidd:
mame2003 is used on a lot of systems where users have trouble accessing the
filesystem, which I know isn't an issue with retropie.

You're right in that the user would need to compile mame2003 to see new
hiscore adds. Maybe this can become a core setting that either uses
precompiled or loads from the dat depending on the setting.

markwkidd:
In fact @joolswills -- if it's ok with you -- I'm going to close this issue and and I'll plan to add a core option to mame2003 to instruct mame2003 to optionally load the `hiscore.dat` live versus precompiled.

I've started an issue to track that plan:

markwkidd:
Follow-up. I submitted a PR for a new core option to toggle between the built-in `hiscore.dat` and an external `hiscore.dat`: https://github.com/libretro/mame2003-libretro/pull/324

joolswills:
thanks.

PR #2318, opened 2018-02-27
zdoom: add support for Hexen 1 Series & Heretic

psyke83:
Requires the following wads:
* Hexen: ```hexen.wad```
* Hexen Deathkings: ```hexen.wad``` & ```hexdd.wad```
* Heretic: ```heretic.wad```

Additionally, the scriptmodule will automatically rename supported wad
files to lowercase if necessary.

Tested and confirmed as working after ripping the necessary .wad files from my Steam installation.

Just a note on the added invocation of pushd/popd: the Hexen Deathkings file is actually a pwad that depends on ```hexen.wad``` to be present in order to launch correctly. If zdoom doesn't find hexen.wad in the current directory, the expansion won't be able to launch unless its location is specified via the additional argument ```-file /path/to/hexdd.wad``` -  but that must be invoked with ```-iwad /path/to/hexen.wad```. If you prefer to make a wrapper script that handles this complexity instead, let me know and I can update the PR.

### DIFF
--- a/scriptmodules/libretrocores/lr-prboom.sh
+++ b/scriptmodules/libretrocores/lr-prboom.sh
@@ -46,11 +46,19 @@ function _add_games_lr-prboom() {
         ['doom2']="Doom 2"
         ['tnt']="TNT - Evilution"
         ['plutonia']="The Plutonia Experiment"
+        ['heretic']="Heretic - Shadow of the Serpent Riders"
+        ['hexen']="Hexen - Beyond Heretic"
+        ['hexdd']="Hexen - Deathkings of the Dark Citadel"
     )
     local game
+    local doswad
     local wad
     for game in "${!games[@]}"; do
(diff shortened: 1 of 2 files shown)


psyke83:
Forgot to specify doswad as local scope - will amend that later. This is not related to stretch support so it's low priority.

joolswills:
pushd / popd no longer required ?

psyke83:
Correct. From: https://zdoom.org/wiki/command_line_parameters

> ZDoom will look for the IWAD in the current directory, in the same directory as zdoom.exe, in the directory set in the DOOMWADDIR environment variable, and in the directory set in the HOME environment variable. (Verification needed)

Despite the disclaimer, the env var seems to work as intended and the expansion loads correctly.

joolswills:
thanks.

Issue #2296, opened 2018-02-15
Wolf4SDL Not Launching 'Spear of Destiny' (Mission 1)


mediamoshpit:
This is in reference to a potential issue [first mentioned here](https://retropie.org.uk/forum/topic/16260/wolf4sdl-launching-spear-of-destiny-mission-1). 

I believe the current mission one launch command of

`/opt/retropie/ports/wolf4sdl/bin/wolf4sdl-spear --mission 1`

might need to be

`/opt/retropie/ports/wolf4sdl/bin/wolf4sdl-spear`

psyke83:
I can't reproduce. Please post the output of /dev/shm/runcommand.log after trying to launch mission 1.

Posting the contents of your wolf4sdl folder may also help troubleshoot. Mission 1, 2 and 3 work properly for me after copying the ```*.SOD``` and ```*.SD*``` files from my Steam installation and renaming to lowercase.

```
pi@retropie:~/RetroPie/roms/ports/wolf3d $ ls -l
total 10260
-rw-r--r-- 1 pi pi    1072 Jan 28 02:29 audiohed.sod
-rw-rw-r-- 1 pi pi    1156 Dec  3  1992 audiohed.wl1
-rw-r--r-- 1 pi pi  328620 Jan 28 02:29 audiot.sod
-rw-rw-r-- 1 pi pi  132613 Dec  3  1992 audiot.wl1
-rw-rw-r-- 1 pi pi   34101 Dec 22  1992 catalog.exe
-rw-r--r-- 1 pi pi     522 Jan 28 02:51 config.sod
-rw-r--r-- 1 pi pi   68290 Jan 28 02:29 gamemaps.sd1
-rw-r--r-- 1 pi pi  135024 Jan 28 02:29 gamemaps.sd2
-rw-r--r-- 1 pi pi  148534 Jan 28 02:29 gamemaps.sd3
-rw-r--r-- 1 pi pi  148534 Jan 28 02:50 gamemaps.sod
-rw-rw-r-- 1 pi pi   27425 Dec  3  1992 gamemaps.wl1
-rw-r--r-- 1 pi pi     402 Jan 28 02:29 maphead.sd1
-rw-r--r-- 1 pi pi      86 Jan 28 02:29 maphead.sd2
-rw-r--r-- 1 pi pi      86 Jan 28 02:29 maphead.sd3
-rw-r--r-- 1 pi pi      86 Jan 28 02:50 maphead.sod
[...]

PR #2319, opened 2018-02-27
wolf4sdl: ensure SOD ep 2/3 follow dispmanx setting

psyke83:
Needed to fix episode 2/3 not stretching to entire fullscreen on Pi as discussed in #2296.

This feels dirty, but the introduction of "wolf4sdl-spear{2,3}" breaks the assumption that a single binary corresponds to a single port target. Is there a nicer way to do it?


joolswills:
I'm not sure about this - I will need to check.

joolswills:
Although I have no problem with this being done for dispmanx (btw it should be possible to have fullscreen without sdl1 dispmanx driver, so long as there is a framebuffer mode for the output res), I think it will need another change due to:

```
[[ "$bin" != "wolf4sdl-spear" ]] && addPort "$bin" "wolf3d" "Wolfenstein 3D" "$md_inst/bin/$bin"
```

?

joolswills:
The other thing, is this should probably be changed, as we do with doom/quake, so that multiple port entries are added rather than having a single entry. Can have one for wolfenstein, and perhaps 1 for spear, or maybe one for each mission.

psyke83:
I don't think so? If you look at the next line, it sets up the three missions:
```
    for bin in "wolf4sdl-spear" "wolf4sdl-spear2" "wolf4sdl-spear3"; do
        ((mission++))
        addPort "$bin" "wolf3d" "Wolfenstein 3D" "$md_inst/bin/wolf4sdl-spear --mission $mission"
    done`
```

The issue is that get_bins_wolf4sdl is deriving its list from the build targets, which worked well when each port target was a build target. The three spear ports share the one binary, and only ```wolf4sdl-spear``` was being processed for dispmanx.

Also, to clarify, mission 2/3 is using fullscreen, but it's not utilizing the full screen, so it looks more like 75% fullscreen. Come to think of it, DOSbox has the same behaviour (~75% fullscreen scaled by default, scales closer to 100% when dispmanx enabled).

joolswills:
(It's not that important I think, but just to mention it).

psyke83:
Setting it up to have separate entries in ES would be nice - agreed.

joolswills:
```
[[ "$bin" != "wolf4sdl-spear" ]]
```
won't match against "wolf4sdl-spear2" "wolf4sdl-spear3" though which the bin function now returns. It should be `[[ "$bin" != wolf4sdl-spear* ]]` no ?

joolswills:
since you are now adding those names onto the bins function.

psyke83:
Ah yes, you're right, sorry.

joolswills:
your changes look good, but they user will get left over entries in the emulators.cfg if upgrading and the default entry will be wrong.

I wonder if it might be worth cleaning this up in this case and removing the old entries on upgrade.

psyke83:
Jools,

Should be sorted now.

Note: I was going to use ```addPort``` within a ```[[ "$md_mode" == "remove" ]]``` condition. There's two problems, however:
1. ```addPort``` never calls ```addEmulator``` for the remove case, and would need this:
```
diff --git a/scriptmodules/helpers.sh b/scriptmodules/helpers.sh
index dd3c500c..e7138d82 100644
@@ -1186,9 +1186,9 @@ _EOF_
             delSystem "$id" "ports"
         fi
     else
-        [[ -n "$cmd" ]] && addEmulator 1 "$id" "$port" "$cmd"
         addSystem "ports"
     fi
+    [[ -n "$cmd" ]] && addEmulator 1 "$id" "$port" "$cmd"
 }

 ## @fn addEmulator()
```
2. Even if helpers.sh is changed (and it probably might be a good idea in general), the return command in ```addEmulator```'s remove case appears to break the subshell in wolf4sdl.sh - it will break out of the loop after the first invocation of the ```addEmulator``` function. Not sure if that's a shell bug, but calling ```delEmulator``` directly avoids the issue.

Should I add the helpers change in a separate commit? Other ports aren't cleaning up the emulators.cfg on uninstall correctly, and the invocation of ```delEmulator``` doesn't encompass the remaining "wolf4sdl" entry on uninstall here either.

psyke83:
By subshell, I mean this:
```
    # remove obsolete emulator entries
    while read -r game; do
        delEmulator "${game%% *}" "wolf3d"
    done < <(get_opts_wolf4sdl; echo -e "wolf4sdl-spear2\nwolf4sdl-spear3")
```
When ```addPort``` is used (instead of ```delEmulator```), only the first result of ```get_opts_wolf4sdl...``` is processed due to interference from the return command here: https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/helpers.sh#L1228

joolswills:
Ah yes - addPort should call removeEmulator  - I can do the change though.

odd regarding the subshell. I will check that.

psyke83:
Assuming helpers.sh is changed, this would work:
```
diff --git a/scriptmodules/ports/wolf4sdl.sh b/scriptmodules/ports/wolf4sdl.sh
index 889bd28d..416c6ff8 100644
--- a/scriptmodules/ports/wolf4sdl.sh
+++ b/scriptmodules/ports/wolf4sdl.sh
@@ -42,10 +42,11 @@ function _add_games_wolf4sdl() {
     local game
     local wad

-    # remove obsolete emulator entries
-    while read -r game; do
-        delEmulator "${game%% *}" "wolf3d"
-    done < <(get_opts_wolf4sdl; echo -e "wolf4sdl-spear2\nwolf4sdl-spear3")
+    if [[ "$md_mode" == "remove" ]]; then
+        for game in $(get_opts_wolf4sdl; echo -e "wolf4sdl-spear2\nwolf4sdl-spear3"); do
+            addPort "${game%% *}" "wolf3d" "Wolfenstein 3D" "$cmd" "$wad"
+        done
+    fi

     for game in "${!games[@]}"; do
         wad="$romdir/ports/wolf3d/$game"
```
Will I change to this?

joolswills:
yeah. although I would like to look into why it didn't work with the process substitution loop (which has the advantage in general that it works with things with spaces etc and parent variables)

joolswills:
I made the helpers adjustment

psyke83:
PR refreshed with the above change. I rebased my local branch to include your helpers.sh change; wolf4sdl removal seems to get rid of the legacy entries as well as wolf4sdl, so it looks OK now. Thanks.

scriptmodules/ports/wolf4sdl.sh
```diff
@@ -30,11 +30,31 @@ function get_opts_wolf4sdl() {
...
+        ['vswap.sd1']="Wolfenstein 3D - Spear of Destiny Ep 1"
+        ['vswap.sd2']="Wolfenstein 3D - Spear of Destiny Ep 2"
+        ['vswap.sd3']="Wolfenstein 3D - Spear of Destiny Ep 3"
+    )
+    local game
+    local wad
+
+    # remove obsolete emulator entries
+    if [[ "$md_mode" == "remove" ]]; then
+        for game in $(get_opts_wolf4sdl; echo -e "wolf4sdl-spear2\nwolf4sdl-spear3"); do
```
joolswills: this is wrong - as it will loop through `wolf4sdl-sw-v14 -DCARMACIZED -DUPLOAD` etc

psyke83: It's what I intended. The next line filters ```$game``` to omit anything after space, like the old ```get_bins_wolf4sdl```. I just did it that way to avoid assigning a new var once.

psyke83: Ah do you mean it loops through each phrase on a newline? I'll check.

joolswills: Yep. The other loop didn't but that one will.

joolswills: It will call addPort with 

```
wolf4sdl-sw-v14
-DCARMACIZED
-DUPLOAD
wolf4sdl-3dr-v14
-DCARMACIZED
wolf4sdl-gt-v14
-DCARMACIZED
-DGOODTIMES
wolf4sdl-spear
-DCARMACIZED
-DGOODTIMES
-DSPEAR
wolf4sdl-spear2
wolf4sdl-spear3
```

psyke83: Yes, you're right. If you can hang on a bit, I'll do some research to see if we can do anything to circumvent the problem with the return interfering with process substitution loop.

joolswills:
The problem with the other loop is not the return but the addPort function reading from stdin and consuming the extra parameters from the parent loop.

in addPort

```bash
    if [[ -t 0 ]]; then
        cat >"$file" << _EOF_
#!/bin/bash
"$rootdir/supplementary/runcommand/runcommand.sh" 0 _PORT_ "$port" "$game"
_EOF_
    else
        cat >"$file"
    fi
```

so the fix is probably something like

```
    # remove obsolete emulator entries
    if [[ "$md_mode" == "remove" ]]; then
        while read game; do
            addPort "${game%% *}" "wolf3d" "Wolfenstein 3D" "$cmd" "$wad" </dev/null
        done < <(get_opts_wolf4sdl)
    fi
```

joolswills:
I should move round the logic in addPort actually, so that the launch script is not created on removal.

joolswills:
eg

```
diff --git a/scriptmodules/helpers.sh b/scriptmodules/helpers.sh
index 74d5c0c4..c1328303 100644
@@ -1167,18 +1167,6 @@ function addPort() {
         mv "$configdir/$port" "$md_conf_root/"
     fi
 
-    if [[ -t 0 ]]; then
-        cat >"$file" << _EOF_
-#!/bin/bash
-"$rootdir/supplementary/runcommand/runcommand.sh" 0 _PORT_ "$port" "$game"
-_EOF_
-    else
-        cat >"$file"
-    fi
-
-    chown $user:$user "$file"
-    chmod +x "$file"
-
     # remove the ports launch script if in remove mode
     if [[ "$md_mode" == "remove" ]]; then
         rm -f "$file"
@@ -1188,6 +1176,18 @@ _EOF_
             delSystem "$id" "ports"
         fi
     else
+        if [[ -t 0 ]]; then
+            cat >"$file" << _EOF_
+#!/bin/bash
+"$rootdir/supplementary/runcommand/runcommand.sh" 0 _PORT_ "$port" "$game"
+_EOF_
+        else
+            cat >"$file"
+        fi
+
+        chown $user:$user "$file"
+        chmod +x "$file"
+
         [[ -n "$cmd" ]] && addEmulator 1 "$id" "$port" "$cmd"
         addSystem "ports"
     fi

```

joolswills:
I have changed the helpers addPort function - the `</dev/null`  is not needed in my example code, due to the order change.

psyke83:
I've rebased again and tested - it indeed works now. The PR is updated to go back to the process substitution loop (without the now-unnecessary redirect).

psyke83:
Also, would it not be a good idea to hide some of these functions (adding underscore prefix)?
```328/wolf4sdl            : Wolf4SDL - port of Wolfenstein 3D / Spear of Destiny engine : build configure_dispmanx_off configure_dispmanx_on configure depends game_data get_opts install sources remove help
```

I'm thinking of hiding ```get_opts_wolf4sdl``` at the very least, and perhaps ```game_data_wolf4dl```, unless you want to have the option of manually installing the demo assets? Worth it or will I just leave alone?

joolswills:
I like game data to be visible so it shows up from retropie_setup (and can be used to re-install game data manually without knowing the name of the function).

get_opts can be hidden.

psyke83:
OK, PR updated with function renamed to ```_get_opts_wolf4sdl```.

joolswills:
Thanks!

PR #2323, opened 2018-02-27
add Digger Remastered to experimental

proyvind:
screwed up previous pull request, recreating again with issues fixed

### DIFF
--- a/scriptmodules/ports/digger.sh
+++ b/scriptmodules/ports/digger.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
You can re-use existing PRs by fixing / rebasing and force pushing btw.

psyke83:
Adding to what @joolswills said - make sure to re-open the PR before force-pushing a change or else you'll need to jump through hoops to make the PR track your branch again.

joolswills:
Thanks for the module.

joolswills:
Probably should have done more testing before accepting this, but assumed the folders were correct. 
There is nothing saved to `$home/.config/digger"` so it makes an unneeded symlink. digger uses a `$home/.digger.rc` and a `$home/.digger.sco` though. I will fix this up.

joolswills:
the pushd and popd are also not needed.

PR #2324, opened 2018-02-28
zdoom: extend support to Strife, Chex 3 & isolate supported ports

psyke83:
* lr-pboom does not support Hexen/Heretic/Strife/Chex, so isolate these
  to zdoom's configuration instance.
* Omit Chex 1/2, as Chex 2 requires special arguments to load chex.wad, and
  Chex 3 (freeware) includes remastered versions of the previous two episodes.

--

This should cover zdoom's list of supported commercial ports (keeping in mind that Chex 3 is a remaster, but includes the originals).


joolswills:
Thanks

PR #2331, opened 2018-03-14
joy2key: normalize device path & check parent directory

psyke83:
The location of the name, id/vendor & id/product descriptors are not
consistent with udev's DEVPATH for every kernel driver.

Fix this by properly normalizing the path correctly, then shift to the parent
directory if the "name" file does not exist.

Fixes the hid_sony driver.


joolswills:
thanks.

PR #2333, opened 2018-03-16
mupen64plus: workaround GLideN64 shader cache crash

psyke83:
Crash seems reproducible only on gcc 6.3.0 via Rasbian stretch, but the
workaround should not negatively impact other systems.

See: https://github.com/gonetz/GLideN64/issues/1665


joolswills:
thanks.

dankcushions:
Great!

It might be worth doing the same change as a PR for GlideN64. I fear your comment on https://github.com/gonetz/GLideN64/issues/1665 will/has become buried.

PR #2335, opened 2018-03-18
Yquake2

psyke83:
Supports Quake II and both official mission packs. Can use OpenGL 1.x, 3.x or software
rendering via SDL2 (of which the latter is used for GLES targets such as Raspberry Pi).

The .pak files and video folder need to be installed from each of the original
folders (maintaining the same folder structure):

* baseq2: Quake II (full or shareware)
* rogue:  Quake II - Ground Zero
* xatrix: Quake II - The Reckoning

If no data is present, the function will automatically fetch and install the shareware data.

Finally, optional CD audio is supported via ripped .ogg files. See the README.md for details.

### DIFF
--- a/scriptmodules/ports/yquake2.sh
+++ b/scriptmodules/ports/yquake2.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="yquake2"
+rp_module_desc="yquake2 - The Yamagi Quake II client"
+rp_module_licence="GPL2 https://raw.githubusercontent.com/yquake2/yquake2/master/LICENSE"
+rp_module_section="exp"
+rp_module_flags=""
+
+function depends_yquake2() {
+    local depends=(libgl1-mesa-dev libglu1-mesa-dev libogg-dev libopenal-dev libsdl2-dev libvorbis-dev zlib1g-dev)
+
+    getDepends "${depends[@]}"
+}
+
+function sources_yquake2() {
+    gitPullOrClone "$md_build" https://github.com/yquake2/yquake2.git
+    # workaround for hang on startup
+    sed -i "$md_build/src/backends/unix/system.c" -e '/setegid/d' -e '/setreuid/d'
+}
+
+function build_yquake2() {
+    rm -rf release
+    cmake .
+    make clean
+    make
+    md_ret_require="$md_build/release/quake2"
+}
+
+function install_yquake2() {
+    md_ret_files=(
+        'release/baseq2'
+        'release/q2ded'
+        'release/quake2'
+        'release/ref_gl1.so'
+        'release/ref_gl3.so'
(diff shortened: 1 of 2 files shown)


scriptmodules/ports/yquake2.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+
+rp_module_id="yquake2"
+rp_module_desc="yquake2 - The Yamagi Quake II client"
+rp_module_licence="GPL2 https://raw.githubusercontent.com/yquake2/yquake2/master/LICENSE"
+rp_module_section="exp"
+rp_module_flags=""
+
+function depends_yquake2() {
+    local depends=(libgl1-mesa-dev libglu1-mesa-dev libogg-dev libsdl2-dev libvorbis-dev zlib1g-dev)
```
joolswills: libraspberrypi-dev dependency instead of mesa for rpi ?

scriptmodules/ports/yquake2.sh
```diff
@@ -0,0 +1,99 @@
...
+            addPort "$md_id" "quake2" "${games[$game]}" "$cmd" "${game%%/*}"
+        fi
+    done
+}
+
+function game_data_yquake2() {
+    local unwanted
+
+    if [[ ! -f "$romdir/ports/quake2/baseq2/pak1.pak" && ! -f "$romdir/ports/quake2/baseq2/pak0.pak" ]]; then
+        cd "$__tmpdir"
```
joolswills: I don't think this is needed ?

scriptmodules/ports/yquake2.sh
```diff
@@ -0,0 +1,99 @@
...
+        # Get shareware game data
+        downloadAndExtract "https://deponie.yamagi.org/quake2/idstuff/q2-314-demo-x86.exe" "$romdir/ports/quake2/baseq2" "-j -LL"
+    fi
+
+    # rename existing files that are likely to cause conflicts or unwanted default settings
+    for unwanted in $(find "$romdir/ports/quake2" -maxdepth 2 -name "*.so" -o -name "*.cfg" -o -name "*.dll"); do
+        mv "$unwanted" "$unwanted.orig"
```
joolswills: perhaps we should just remove?

scriptmodules/ports/yquake2.sh
joolswills: Sorry - I misread the PR. software rendering on gles. So we don't need this dependency for the RPI.

joolswills:
Thanks for the PR. a few people have asked for quake2 and I never got around to looking at it. Just a couple of minor points I mentioned. Cheers

psyke83:
Updated to reflect your comments.

You're right, libraspberrypi-dev is not needed even for Pi - but I did overlook the libopenal-dev dependency, which is now included.

Cheers.

actraiser30:
Hi,i installed yquake2 put the original game files in the correct folders but i only have the shortcut for quake 2 in the ports and not the 2 mission packs Ground Zero and the Reckoning that i also put in the quake2 folder eg: xatrix and rogue folders,i reinstalled yquake 2 after but still no mission packs in ports, is it possible to add shortcuts to the ports for the mission packs?

blargelJim:
Hi psyke83, joolswills. After some discussion and work with the yamagi quake 2 guys, there is now support for a convenient way of switching weapons using a controller (https://github.com/yquake2/yquake2/pull/392). I've been digging through the retropie repo without finding any config.cfg file for quake 2 but Im pretty sure there is one made especially for the retropie (since it has bindings for the controller per default).
I would like to update it so that there are default bindings for switching weapons in quake 2. Do you know where that config.cfg file sits and if there is any chance I can update it?

PR #2348, opened 2018-03-27
Add some mupen64plus optimizations

gizmo98:
-use default "Video-Rice" for Conker and Zelda games. These game have problems if GLideN64 is used (translucent and missing textures).
-remove Gauntlet Legends from blacklist
-use better Audio-SDL resampler (less crackle)
-fix Perfect Dark highres autosetting


joolswills:
Thanks!

Issue #2327, opened 2018-03-12
Amiga LHA File Support


HoraceAndTheSpider:
Apologies in advnace that I am having a bit of a blonde moment in trying to add this code myself, and failing miserably at finding the right places to make the changes, so i will have to revent to making a request...

Amiberry’s development branch has just had a new file support added, LHA. This is commonly used as a single file solution to packaged WHDLoad games, as this format is able to accommodate certain constraints of the Amiga filesystem within a contained archive. (Eg. Issues with certain filenames etc)

The support is fairly simple. Just by adding `-autowhdload=gamefile.lha` on the command line.

For RetroPie to fully take advantage of this (and it was designed with RetroPie in mind) I’d like to suggest that the `es_systems.cfg` has .lha as a supported Amiga filetype and that the following be added to the Amiberry launch script on line 13 (right after the .uae file support)

```
[elif [[ "$rom" == *.lha ]]; then
"$emulator" -autowhdload="$rom" -G]) 
```

Details on how the system is able to have game-specific settings, without using individual Config files, is provided on the Amiberry Wiki Page , and you can expect a large number of game-specifics to be added very soon.

I appreciate that the devs here might want to wait until this change reaches the master branch, but I anticipate that will be quite soon anyway.

Thanks!

psyke83:
Does #2328 look OK? We can delay merging until amiberry's master branch catches up. This feature is working on the dev branch? I'd like to compile a build from the dev branch (via a local edit to my script) to make sure everything's ok.

HoraceAndTheSpider:
It does look good, thank you @psyke83

I guess that explains why I could not find the script when I searched - I did not realise it was still shared with Uae4arm!

Yes, the feature has been added to the dev version, but you will need to download the full package and not just update the binary as the whdbooter files need including. I have been doing extensive testing to but I made the script changes by hand. 

I also have not created an extensive XML file for the improved game compatibility, but so far I have been pleasantly surprised with how many games work on the default options. I hope I can do that before the next merge with the master branch by it actually isn’t essential. 

There is extra info here if you need it:

https://github.com/midwan/amiberry/wiki/WHDLoad-Auto-booting

Thanks again

PR #2352, opened 2018-04-05
amiberry - add support for whdload auto-load - #2327

joolswills:
* add lha extension
 * adjust launch script to use new -autowhdload parameter
 * rework whdboot folder set-up

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -4,7 +4,7 @@
 ags_exts=".exe"
 ags_fullname="Adventure Game Studio"
 
-amiga_exts=".adf .adz .dms .ipf .uae .zip .sh"
+amiga_exts=".adf .adz .dms .ipf .lha .sh .uae .zip"
 amiga_fullname="Commodore Amiga"
 
 amstradcpc_exts=".cdt .cpc .dsk"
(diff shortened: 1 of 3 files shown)


HoraceAndTheSpider:
Visual review, this is looking good - i will try and get a chance for a proper install test later.

HoraceAndTheSpider:
Worked for me from source, but not from binary (the latter only created the whdboot folder?)

HoraceAndTheSpider:
One thing i have noticed is the defaults in `hostprefs.conf` are wrong ... I can fix default on the Dev branch of Amiberry, but it might be a little while till pushed to the Master

basically it tries to use joystick number 2 as a the mouse controller by default. 

```
CONTROLLER_MOUSE_1=joy1
CONTROLLER_MOUSE_2=joy2
```

joolswills:
It won't work from binary until the bins are updated. Can't do much about the hostprefs until it's merged to master. Perhaps upstream can cherry pick that fix for master.

HoraceAndTheSpider:
Great stuff.

The hostconfig fix will be in the next master update, which I am discussing with Dimitris so hopefully we can push that out soon as well.

PR #2356, opened 2018-04-06
Update to add 5 themes to esthemes.sh

RetroHursty69:
I have created 5 new themes to be added to EmulationStation. Crips, Crisp_Light, Primo, Primo_Light and Back2Basics. The light versions only cover about one 3rd of the systems and collections compared to the full versions of the themes. The full versions of Primo and Crisp over over 230 systems and collections, so i made the light versions for people mindful of microsd space. Also it is hoped that light versions may perform better with less media to load. Thank you in advance for adding these themes.


joolswills:
thanks!

PR #2359, opened 2018-04-08
Create a fully seperated mame2003-plus setup file

arcadez:
Possible fix for samples not working in the mame2003-plus core


arcadez:
Sure, i just assumed to fix the samples and any other path problems it would require me
as you seen to create a fully seperate mame2003-plus setup file, i may just leave this to 
others more versed in all things Rpi to come up with a more viable alternative.

Regards.

joolswills:
the current code should set up the paths based on the name of the core so it should be different. If there is a problem with a path / location I will need to details on that and can look into a fix.

markwkidd:
@arcadez I think we no longer need this due to improvements in all three repositories. do you think it's ok to close?

PR #2363, opened 2018-04-10
Sixaxis groundwork

psyke83:
Due to the new kernel 4.14, we now have different button maps for the DS3's USB connection (via hid-sony) compared to both the 4.9 kernel version's driver as well as the current ps3controller driver. In order to at least maintain consistency with the native driver, I've split the sixaxis PR into two. 

This "groundwork" PR does two things: 
* bring the ```customhidsony``` driver into sync with the default 4.14 kernel driver so it has the same new mappings; the only difference between rpi-4.14.y and rpi-4.15.y version of hid_sony will be the Shanwan vibration fix (as well as my change to disable motion sensors).
* add basic bluetooth pairing support.

Neither of these changes will break ps3controller, but will allow people the option to use the native driver via Bluetooth (as long as they understand that it will not yet have an inactivity timeout) - and helps lay the groundwork for future inclusion of the sixaxis helper.

The original PR now only has only the major changes - the helper and custom Bluez module. Those require more time and proper review, and thus can wait until much later.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -80,9 +80,15 @@ function bluez_cmd_bluetooth() {
 function list_available_bluetooth() {
     local mac_address
     local device_name
+    local info_text="\n\nSearching ..."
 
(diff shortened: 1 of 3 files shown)


joolswills:
thanks.

Issue #2295, opened 2018-02-14
lr-mgba isn't just a Game Boy Advance (gba) emulator, it also emulates Game Boy (gb) and Game Boy Color (gbc).


vaguerant:
I've added it to the `/opt/retropie/configs/gb/emulators.cfg` and `/opt/retropie/configs/gbc/emulators.cfg` files on my end. I tried to find where these are on the repo but had no luck or else I would have made a pull request, sorry.

While native mGBA supports the Game Boy, Super Game Boy, Game Boy Color and Game Boy Advance, the libretro version is admittedly more GBA-focussed. It doesn't seem to have support for the Super Game Boy, or BIOSes for the other Game Boys besides Advance. Still, it's a very solid option for Game Boy/Color games and already included in RetroPie, so why not include the option to run GB/C games with it?

Darknior:
Yes you are real ... it can be cool to add it for the two other gameBoy.

https://github.com/libretro/mgba

joolswills:
What advantages does it have over the other emulators ? Have you tested it?

Darknior:
Like you write it, i have add it to my RETROPIE easily :)
I don't try playing games with it for the moment, but an other emu is always cool to prevent a bug of one emulator on a game and trying with an other one.
And like you write mGBA have an active development :)
For me it's a good idea to add it to GB and GBC.

vaguerant:
Just wanted to note here that upstream at libretro/mgba, mGBA has been updated to be in-line with current standalone mGBA, meaning it now supports (Super) Game Boy (Color) BIOSes, full SGB border and palette support, Game Boy Color enhanced palettes for original Game Boy games, etc. I'd say this will make mGBA easily the best Game Boy/Color/Advance emulator available in RetroPie in terms of its capabilities. These updates are not in a stable release yet; RetroPie users can update their lr-mgba from source to get these updates right now.

[...]

PR #2366, opened 2018-04-12
lr-mgba.sh: enable lr-mgba for GB and GBC (SGB also supported)

hhromic:
As mentioned in #2295, `lr-mgba` is mature enough now to support GB/GBC/SGB as well as GBA.
I've been using it for several weeks now and all works fine. The emulator detects the system type automatically and can use all BIOSes available, even Super Gameboy.
The emulation speed in my standard-configured RPI3 is very satisfactory too.
This is for the `master` upstream repository of `lr-mgba`, used when you install it from source.
I strongly suggest the binary version in RetroPie is updated too to make all work out-of-the-box.

The following BIOS files were used for testing each system type:

    a860e8c0b6d573d191e4ec7db1b1e4f6  gba_bios.bin
    32fbbd84168d3482956eb3c5051637f5  gb_bios.bin
    dbfce9db9deaa2567f6a84fde55f9680  gbc_bios.bin
    d574d4f9c12f305074798f54c091a8b4  sgb_bios.bin

I kept the original logic used to set the emulator as default, e.g. set it as default in non `armv6` platforms.
This should keep users of slow hardware unaffected if they are using `lr-gambatte` already.

### DIFF
--- a/scriptmodules/libretrocores/lr-mgba.sh
+++ b/scriptmodules/libretrocores/lr-mgba.sh
@@ -10,8 +10,8 @@
 #
 
 rp_module_id="lr-mgba"
-rp_module_desc="GBA emulator - MGBA (optimised) port for libretro"
-rp_module_help="ROM Extensions: .gba .zip\n\nCopy your Game Boy Advance roms to $romdir/gba\n\nCopy the required BIOS file gba_bios.bin to $biosdir"
+rp_module_desc="(Super) Game Boy Color/GBA emulator - MGBA (optimised) port for libretro"
+rp_module_help="ROM Extensions: .gb .gbc .gba .zip\n\nCopy your Game Boy roms to $romdir/gb\nGame Boy Color roms to $romdir/gbc\nGame Boy Advance roms to $romdir/gba\n\nCopy the recommended BIOS files gb_bios.bin, gbc_bios.bin, sgb_bios.bin and gba_bios.bin to $biosdir"
 rp_module_licence="MPL2 https://raw.githubusercontent.com/libretro/mgba/master/LICENSE"
 rp_module_section="main"
 rp_module_flags=""
@@ -40,11 +40,14 @@ function install_lr-mgba() {
 }
 
 function configure_lr-mgba() {
-    mkRomDir "gba"
-    ensureSystemretroconfig "gba"
-
-    local def=1
-    isPlatform "armv6" && def=0
-    addEmulator $def "$md_id" "gba" "$md_inst/mgba_libretro.so"
-    addSystem "gba"
+    local system
+    local def
+    for system in gb gbc gba; do
+        def=0
+        [[ "$system" == "gba" ]] && ! isPlatform "armv6" && def=1
+        mkRomDir "$system"
+        ensureSystemretroconfig "$system"
+        addEmulator "$def" "$md_id" "$system" "$md_inst/mgba_libretro.so"
+        addSystem "$system"
+    done
 }


hhromic:
If this change gets merged and the binary updated, I will update the Wiki to include BIOS file information too.

scriptmodules/libretrocores/lr-mgba.sh
```diff
@@ -40,11 +40,12 @@ function install_lr-mgba() {
...
 
 function configure_lr-mgba() {
-    mkRomDir "gba"
-    ensureSystemretroconfig "gba"
-
     local def=1
     isPlatform "armv6" && def=0
-    addEmulator $def "$md_id" "gba" "$md_inst/mgba_libretro.so"
-    addSystem "gba"
+    for system in gb gbc gba; do
```
psyke83: You should ensure the ```system``` variable is local scope

scriptmodules/libretrocores/lr-mgba.sh
```diff
@@ -40,11 +40,12 @@ function install_lr-mgba() {
...
+        mkRomDir "$system"
+        ensureSystemretroconfig "$system"
+        addEmulator "$def" "$md_id" "$system" "$md_inst/mgba_libretro.so"
```
psyke83: I'm not certain if @joolswills wants this patch included at all, but if so, perhaps he can chime in on whether the previous defaults should be preserved, or if all three systems should use ```lr-mgba``` as set by this patch (I think not).

scriptmodules/libretrocores/lr-mgba.sh
hhromic: oh forgot that, thanks for spotting.

hhromic:
Perhaps the correct thing to do is to follow the same approach/logic than for `lr-genesis-plus-gx` and `lr-picodrive`, two competing emulators for various Sega 8/16 bits consoles:

`lr-picodrive` is used as default for: **megadrive mastersystem segacd sega32x**.
`lr-genesis-plus-gx` is used as default for: **gamegear sg-1000**.
`lr-genesis-plus-gx` is used as alternative for:  **megadrive mastersystem segacd**.

Then I can rework the patch to achieve this style of behaviour for the Game Boy consoles:

`lr-gambatte` is used as default for: **gb gbc**
`lr-mgba` is used as default (in non `armv6`) for: **gba**
`lr-mgba` is used as alternative for: **gb gbc**
`lr-gpsp` is used as default (in `armv6`) for: **gba**
`lr-vba-next` is used as alternative for: **gba**

How does this sound?

hhromic:
Hi @psyke83 , @joolswills , did you have any chance to review this PR?
Let me know of your thoughts to keep working on it and get it merged.
If you are not happy with the defaults handling I propose (check the edited comment), we can discard that, but at least I think we should allow lr-mgba to be an alternative for GB/GBC even if it is not their default emulator.
Cheers and thanks for all your work!

joolswills:
I will add this as an alternative but I am not going to change any defaults currently.

hhromic:
Good stuff. Do you want me to refactor this PR accordingly?

joolswills:
If you would like - please squash the commits also. alternatively I am happy to implement it when I next have some time.

hhromic:
No problem from my side, I was also planning to squash the commits. Will update shortly. Thanks for the feedback.

hhromic:
Update the PR now:
* Updated the module description and help.
* Added GB/GBC as alternative but not default.
* Kept (as originally) GBA as default only in non armv6 platforms.
Thanks!

joolswills:
Thanks

hhromic:
@joolswills , before I forget , it is possible to also update the binary for lr-mgba to be up-to-date to its current master? Thanks!

joolswills:
Can do. It was updated a few weeks ago btw.

hhromic:
Great, should be fine then. When you have time, it won't hurt if you update again just in case.
Thanks a million.

hhromic:
Hi @joolswills , sorry to bring this up again. Looks like users in the forum are needing to update `lr-mgba` from source to get the latest features enabled recently: https://retropie.org.uk/forum/post/146141
Maybe a refresh of the binary is a good idea? Cheers!

Issue #2365, opened 2018-04-12
Add .7z and .zip extensions to Atari 5200 in platforms.cfg.


Eckaji:
lr-atari800 has support for both the 7z and zip archive formats.

Please add these extensions to Atari 5200 at your earliest convenience so that all future users are able to use these archive formats without additional configuration.

PR #2367, opened 2018-04-14
Added .7z and .zip to atari5200 in platforms.cfg

zerojay:
Fix for issue #2365.


joolswills:
we try to add in alphabetical order (not all are correct, but prefer to add new extensions correctly0. Will add myself.

PR #2369, opened 2018-04-15
overload license and changelog filenames for mame2003-plus

markwkidd:
If I understand the function of this script, the PR will point `lr-mame2003-plus.sh` to the location of the changelog without having to re-implement any more of the underling `lr-mame2003.sh` script.

Because twinaphex merged the PR which reformated and relocated these files in mame2003-plus it's a fair bet that this change will eventually work it's way back to mame2003 as well but who knows when.


markwkidd:
I'm actually going to close this because it seems like these changes are about to be merged into mame2003-libretro so the filenames can simply be changed in `lr-mame2003.sh` instead of here.

PR #2375, opened 2018-04-18
fix lr-mame2003.sh tested by actual retropie user

markwkidd:
Yes there was a botch in my last PR. This PR has been tweaked and tested by @grant2258 and shown to work in real life.

For at least the immediate future there are some folks with RetroPie that will hopefully be able to test things if, for some unknown reason, future updates are ever necessary again to this script.

Thanks for being cool while we try to make something good happen with mame2003 and mame2003-plus.

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -39,7 +39,6 @@ function install_lr-mame2003() {
         'README.md'
         'CHANGELOG.md'
         'metadata'
-        'metadata/nvram'
     )
 }
 
@@ -54,25 +53,23 @@ function configure_lr-mame2003() {
         mkRomDir "$mame_dir"
         mkRomDir "$mame_dir/$name"
         ensureSystemretroconfig "$mame_dir"
-        
-        # currently nvram is the only folder to prepopulate
+
+        # currently nvram is the only folder to repopulate
         for mame_sub_dir in nvram; do
             mkRomDir "$romdir/$mame_dir/$name/$mame_sub_dir"
         done
-        
+
         # prepopulate nvram patches
-        cp -r "$md_inst/nvram" "$romdir/$mame_dir/$name/$nvram"
+        cp -r "$md_inst/metadata/nvram" "$romdir/$mame_dir/$name/$nvram"
         chown -R $user:$user "$romdir/$mame_dir/$name/$nvram"
     done
 
     mkUserDir "$biosdir/$name"
     mkUserDir "$biosdir/$name/samples"
-    chown -R $user:$user "$biosdir"
-
 
     # prepopulate hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$name/"
-    chown -R $user:$user "$biosdir/$name/"{hiscore.dat,cheat.dat}
+    chown -R $user:$user "$biosdir"
     # Set core options
     setRetroArchCoreOption "${name}-skip_disclaimer" "enabled"
     setRetroArchCoreOption "${name}-dcs-speedhack" "enabled"
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -39,6 +39,7 @@ function install_lr-mame2003() {
         'README.md'
         'CHANGELOG.md'
         'metadata'
+        'metadata/nvram'
     )
 }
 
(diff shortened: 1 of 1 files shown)


grant2258:
I can comfirm ive compiled and installed both mame2003 and mame2003+ and they are working

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -47,27 +47,29 @@ function configure_lr-mame2003() {
...
     for mame_dir in arcade mame-libretro; do
         mkRomDir "$mame_dir"
         mkRomDir "$mame_dir/$name"
         ensureSystemretroconfig "$mame_dir"
 
-        for mame_sub_dir in cfg ctrlr diff hi inp memcard nvram snap; do
-            mkRomDir "$mame_dir/$name/$mame_sub_dir"
+        # currently nvram is the only folder to repopulate
+        for mame_sub_dir in nvram; do
+            mkRomDir "$romdir/$mame_dir/$name/$mame_sub_dir"
```
joolswills: Why have you added $romdir ? Why are the other folders removed ? Looping through a single folder doesn't make sense.

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -47,27 +47,29 @@ function configure_lr-mame2003() {
...
 
     mkUserDir "$biosdir/$name"
     mkUserDir "$biosdir/$name/samples"
 
-    # copy hiscore.dat
+    # prepopulate hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$name/"
-    chown $user:$user "$biosdir/$name/"{hiscore.dat,cheat.dat}
-
+    chown -R $user:$user "$biosdir"
```
joolswills: What is this change for ? We shouldn't need to chown the whole biosdir

joolswills:
Made a couple of comments - the changes don't make sense currently so cannot accept as is.

joolswills:
The last PR does indeed have a mistake - I will revert it for now - there is no $nvram variable.

markwkidd:
If you're happy with the restored-nvram version and it's working then it seems I can close this PR. Many thanks!

joolswills:
not sure what you mean by restored nvram version sorry.

markwkidd:
The current code, after your reversion, does not prepopulate the .nv files
from metadata/nvram. That was one of the intentions of the recent PRs to
this script that I tried to accomplish.

That absence is not a crisis as only two games currently use nvram patches,
but worth mentioning.

I'll just be glad to not be responsible for any more RetroPie setup issues
to be honest

joolswills:
I will re-implement the nvram copying when I have a moment. thanks.

joolswills:
Im happy for you to just open an issue for heads up on upstream changes if you prefer.

markwkidd:
Thanks, I'll go with an Issue in the future :)

scriptmodules/libretrocores/lr-mame2003.sh
grant2258: mame_dir is mame-libretro
 romdir is /home/pi/RetroPie/roms

i was just using the full path the other dirs where  removed because mame makes them itself

scriptmodules/libretrocores/lr-mame2003.sh
grant2258: wasnt sure if the sudo to run the script would create the files with uid 0 so i just done the whole dir it should be owned by the pi users anyway if the script doesnt creat with uid 0 it can just be removed

scriptmodules/libretrocores/lr-mame2003.sh
joolswills: MkRomDir adds $romdir already so I don't think the above will work correctly.

grant2258: @joolswills ah i see what you mean i should have used mkdir in that statement i see what was wrong with that. Thanks for the clarification.

grant2258: all that remains to be added is 

cp -r "$md_inst/metadata/nvram" "$romdir/$mame_dir/$name/$nvram"

unless you have a tidier way of doing it

grant2258: just for some clarification 

/libretro savefile dir/mame2003-plus/diff/
/libretro savefile dir/mame2003-plus/nvram/
/libretro savefile dir/mame2003-plus/hi/
/libretro savefile dir/mame2003-plus/cfg/
/libretro savefile dir/mame2003-plus/memcard/

Im not sure where retropie sets the libretro savefile dir the contents should be going into this dir im assuing its been set to the content dir

PR #2377, opened 2018-04-20
Updated Commander Genius version

gerstrong:
Updated Commander Genius version


joolswills:
Please can you squash PRs like this when submitting, so we don't have multiple commits / merge commtis etc.

joolswills:
I'll squash and merge this time. thanks.

PR #2383, opened 2018-04-24
Added bombermaaan.sh script.

bjaraujo:
I would like if you could add this port to retropie experimental for others to install and test.

### DIFF
--- a/scriptmodules/ports/bombermaaan.sh
+++ b/scriptmodules/ports/bombermaaan.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="bombermaaan"
+rp_module_desc="Bombermaaan - Classic bomberman game"
+rp_module_licence="GPL3 https://github.com/bjaraujo/Bombermaaan/blob/master/LICENSE.txt"
+rp_module_section="exp"
+rp_module_flags="!mali !kms"
+
+function depends_bombermaaan() {
+    getDepends cmake libsdl1.2-dev libsdl-mixer1.2-dev build-essential
+}
+
+function sources_bombermaaan() {
+    gitPullOrClone "$md_build" https://github.com/bjaraujo/Bombermaaan.git
+}
+
+function build_bombermaaan() {
+    cd trunk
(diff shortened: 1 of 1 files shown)


scriptmodules/ports/bombermaaan.sh
```diff
@@ -0,0 +1,47 @@
...
+
+function sources_bombermaaan() {
+    gitPullOrClone "$md_build" https://github.com/bjaraujo/Bombermaaan.git
+}
+
+function build_bombermaaan() {
+    cd trunk
+    cmake . -DCMAKE_INSTALL_PREFIX="$md_inst" -DLOAD_RESOURCES_FROM_FILES:BOOL=ON
+    make
+    mv src/Bombermaaan bombermaaan
```
joolswills: I don't think the mv is needed. Can just check for the file in trunk/src

scriptmodules/ports/bombermaaan.sh
```diff
@@ -0,0 +1,47 @@
...
+        'trunk/res/images'
+        'trunk/res/sounds'
+    )
+}
+
+function configure_bombermaaan() {
+    mkUserDir "$home/.config"
+    moveConfigDir "$home/.config/bombermaaan" "$md_conf_root/bombermaaan"
+    addPort "$md_id" "bombermaaan" "Bombermaaan" "pushd $md_inst; $md_inst/bombermaaan; popd"
+}
```
joolswills: missing EOL

scriptmodules/ports/bombermaaan.sh
```diff
@@ -0,0 +1,47 @@
...
+        'trunk/bombermaaan'
+        'trunk/levels'
+        'trunk/res/images'
+        'trunk/res/sounds'
+    )
+}
+
+function configure_bombermaaan() {
+    mkUserDir "$home/.config"
+    moveConfigDir "$home/.config/bombermaaan" "$md_conf_root/bombermaaan"
```
joolswills: Just want to confirm it actually uses this location for config storage ?

joolswills:
Made a few more comments. I haven't had a chance to test yet (which I usually do before merging)

scriptmodules/ports/bombermaaan.sh
bjaraujo: mv is just to move and rename to lower case.

scriptmodules/ports/bombermaaan.sh
bjaraujo: It probably shouldn't need that. Should I delete that?

scriptmodules/ports/bombermaaan.sh
bjaraujo: I will add.

scriptmodules/ports/bombermaaan.sh
joolswills: Yes please.

scriptmodules/ports/bombermaaan.sh
joolswills: ok. I'm happy with that.

bjaraujo:
Ok. I have made the changes. Please test.

joolswills:
Thanks.

PR #2387, opened 2018-04-25
Prevent Bash parameter expansion in filenames passed to retroarch

jarettmillard:
When attempting to launch a ROM with a "$" in the filename, runcommand.sh passes the filename to retroarch in a way that allows for Bash parameter expansion. I encountered this in the wild when trying to run "WarioWare, Inc. - Mega Microgame$!": the $! is a valid Bash parameter and was expanded.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -878,8 +878,8 @@ function get_sys_command() {
     COMMAND="$(default_emulator get emu_cmd)"
 
     # replace tokens
-    COMMAND="${COMMAND//\%ROM\%/\"$ROM\"}"
-    COMMAND="${COMMAND//\%BASENAME\%/\"$ROM_BN\"}"
+    COMMAND="${COMMAND//\%ROM\%/\'$ROM\'}"
+    COMMAND="${COMMAND//\%BASENAME\%/\'$ROM_BN\'}"
 
     # special case to get the last 2 folders for quake games for the -game parameter
     # remove everything up to /quake/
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. I will need to test this and make sure it doesn't affect anything else due to the way the filename is escaped.

jarettmillard:
I found a good way to escape all the special characters with `printf` without using quotes.

```
Executing: /opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-mgba/mgba_libretro.so --config /opt/retropie/configs/gba/retroarch.cfg /home/pi/RetroPie/roms/gba/WarioWare\,\ Inc.\ -\ Mega\ Microgame\$\!\ \(USA\).gba --appendconfig /dev/shm/retroarch.cfg
```

```
Executing: /opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-snes9x2010/snes9x2010_libretro.so --config /opt/retropie/configs/snes/retroarch.cfg /home/pi/RetroPie/roms/snes/Ganpuru\ -\ Gunman\'s\ Proof\ \(Japan\).sfc --appendconfig /dev/shm/retroarch.cfg
```

hhromic:
I also believe that escaping the `$` is the correct thing to do.

hhromic:
Ah indeed an array might be a better idea for this! Will keep an eye and test when implemented.
Good catch from @jarettmillard nevertheless.

joolswills:
fixed by escaping command before using eval.

joolswills:
(Switching to array instead of eval needs a fair amount of work, and may break something so I went with a quick fix).

PR #2388, opened 2018-04-26
Remove 'eudora-updated', Sort list alphabetically

ChoccyHobNob:
Now that all my updates are merged into AmadhiX's version of Eudora,
mine is no longer required.

Took the opportunity to sort the list alphabetically too.


joolswills:
Any sorting should be done in a separate commit but I don't want to change the order right now - I may do some changes to the order later.

ChoccyHobNob:
OK I'll close this and just submit a remove of eudora-updated

joolswills:
No need to do a new PR for this. Will be quicker for me to remove it.

PR #2391, opened 2018-04-29
Version update

gerstrong:
I hope it is okay this time. I hard synched my fork against your repo and now just one local commit. If that's not okay, let me know how you want to have it squash...

### DIFF
--- a/scriptmodules/ports/cgenius.sh
+++ b/scriptmodules/ports/cgenius.sh
@@ -19,7 +19,7 @@ function depends_cgenius() {
 }
 
 function sources_cgenius() {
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. The commit message is not really descriptive enough, but I will squash and merge and change it. Better would be something like

```
cgenius - bump version to v2.2.2
```

btw - rather than us just buming the version when you tag it, how about a stable branch we can point to - then users will always get the latest version ?

gerstrong:
Sure, I will be more descriptive in future. actually these releases are stable enough to be tagged, still I want to fixate them for now in case I mess something up on the branch I am working on right now. I have been reorganizing a bit, there will be a stable branch shortly where we can point to. But first let me get some things cleaned up first...

PR #2394, opened 2018-05-02
Commander Genius now uses the stable branch instead of a fixed tag

gerstrong:
Use stable branch instead of a version


joolswills:
thanks.

PR #2398, opened 2018-05-05
lr-mame2003-plus fixes - #2397

joolswills:


markwkidd:
@grant2258 tagging you so you can see the patch that, pending testing, would go into RetroPie setup

grant2258:
@joolswills there is one problem

in /opt/retropie/configs/mame-libretro/emulators.cfg

lr-mame2003 = "/opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-mame2003/mame2003_libretro.so --config /opt/retropie/configs/mame-libretro/retroarch.cfg %ROM%"

should be    -L /opt/retropie/libretrocores/lr-mame2003/mame2003_plus_libretro.so

not  -L /opt/retropie/libretrocores/lr-mame2003/mame2003_libretro.so  thats what its set to now

PR #2400, opened 2018-05-07
retroarch - updated to v1.7.3

joolswills:
* add --disable-qt configure flag
 * disable udev sort as it the changed behaviour may cause more problems than it solves - https://retropie.org.uk/forum/topic/17276/usb-ports-no-longer-determine-player-number-retroarch-1-7-1

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -37,13 +37,14 @@ function depends_retroarch() {
 }
 
 function sources_retroarch() {
-    gitPullOrClone "$md_build" https://github.com/libretro/RetroArch.git v1.7.1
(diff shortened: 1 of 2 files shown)


markwkidd:
Hi joolswills. I know there is an issue related to the changed udev sort behavior in RetroArch but it hasn't been going anywhere. I would like to take your udev patch and make it into a PR for a configurable RetroArch option in retroarch.cfg.

I think that my "role" in the libretro team being such as it is I may be able to help this get some new traction.

I'll also say that I've done a tiny bit of coding for RetroArch here and there, but I haven't yet implemented a new retroarch.cfg option. If you know how to do that I would be grateful of your help constructing the PR.  Otherwise I'll see if I can figure it out myself.

PR #2403, opened 2018-05-12
Add support for tinker for few emulator

JyuHo:
Hello,

The changes made make it possible to have a support on the emulators concerned. We are working on an image based on Armbian Stretch with the supplied 4.4.120-rockchip kernel. The changes also work on the ISO TinkerOS provided by ASUS.
If you want to do the tests on an armbian image we have a script that automates all the part of the dependencies etc.

https://github.com/Tinkerpie/Armbian-Setup

### DIFF
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -23,10 +23,18 @@ function depends_ppsspp() {
 }
 
 function sources_ppsspp() {
-    gitPullOrClone "$md_build/ppsspp" https://github.com/hrydgard/ppsspp.git v1.5.4
+    if isPlatform "tinker"; then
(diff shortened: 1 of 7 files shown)


joolswills:
Thanks. Looks good. I might change the gcc options later (just -mcpu= rather than the arch/tune).

joolswills:
ppsspp didn't work on the ASUS linaro-alip tinker image - I believe it worked before when I tested but that was some time ago. I've not had a chance to look into it further - it looks like a GL header compatibility issue. It might be better for us to target Armbian anyway as it's likely better supported - I will give it a try.

joolswills:
Note there is a PR to switch to upstream master of ppsspp. I have already made some changes for tinker compatibility, but will also test that on Armbian when it's complete. https://github.com/RetroPie/RetroPie-Setup/pull/2423

PR #2411, opened 2018-05-26
Steven Selph's scraper: prefer screenshots

UncleRus:
This PR adds option to download screenshots instead of box art for
Steven Selph's scraper.


joolswills:
thanks.

PR #2423, opened 2018-06-09
ppsspp & lr-ppsspp updates

psyke83:
Changes for ppsspp:
* Use the master branch (v1.6.3 is missing important EGL fixes)
* Disable Wayland; libgles2-mesa-dev pulls in libwayland-dev, but our SDL2
  build doesn't include Wayland support, which causes a build failure.
* Vulkan is compiled in (due to NO_VULKAN builds not completing successfully).
* The latest master build defaults to fullscreen on all ARM builds, so we
  can remove the explicit command-line argument.

lr-ppsspp:
* Switch to same master branch.
* minimize duplicate code by using ppsspp scriptmodule functions when possible.


psyke83:
Compiles and launches fine on RPi3 & booted into fullscreen by default on a fresh install with no prior config. The only problem I had was that my DS3 controller doesn't seem to be mapped correctly.

As I mentioned in desc, v1.6.3 tag is useless to us due to vital EGL fixes being merged after that tag. I think they're moving to v1.7 soon, so if we keep up to date with the changes and make sure to send any needed fixes for breakage, hopefully we can lock to the v1.7 tag quite soon.

psyke83:
FYI, I've updated the PR to also update lr-ppsspp. Builds & runs fine on my Pi3 as well. My DS3 controller works fine (as expected, since it uses retroarch mappings), but it still seems a bit slower than the native binary, sadly.

gerble1000:
psyke83 
i have tried your commits and seem to get a gray screen when saving game. any ideas, we are getting no errors on output except a standard glerror :)
i am using a rpi3 jesse latest release

psyke83:
I don't know about a grey screen, but it does seem that lr-ppsspp can randomly segfault when attempting to load or save a save state via the hotkey+left/right shoulder combo.

Save states seem to work correctly in the standalone version, however. Considering that the build configuration for both modules in this PR is identical apart from the LIBRETRO cmake option, it's likely a bug in the upstream libretro, not here. Will look more into it when I have more time.

gerble1000:
thanx for the quick reply :) but just to clarify that we are getting the gray screen when using the inbuild save function in the game and not the savestate

psyke83:
Small update: I tried a fresh rebuild yesterday (11th) and the segfault on loading save states issue is still present with lr-ppsspp. It may not be isolated to Pi, however: https://github.com/libretro/ppsspp/issues/13

Additionally, issues with game crashes related to *in-game* saving also seems to have an open ticket: https://github.com/libretro/ppsspp/issues/16

It's likely that the PR itself is fine (apart from potential compile issues for another platform that Jools mentioned to me in private chat), but the question is whether we want to switch to the new release that has these known bugs, or wait until they're squashed. TBH I don't use ppsspp often, and it's a bit prohibitive to constantly re-compile it constantly on a Pi.

joolswills:
I didn't have a chance to finish off looking into the tinker board issues. I had problems compiling ppsspp before this PR (but it definitely worked a while back when I did the tinker support), but I wanted to spend a bit more time on that before this PR (and changes) go ahead - sorry for the delays though. Thanks for the update.

psyke83:
Jools,

Updated the patchset to fix conflicts with recent commits. I compile-tested ppsspp and lr-ppsspp for Pi, and can confirm that they works fine, but still crash randomly if using the state load/save as discussed earlier. Should make things cleaner for all targets including the tinker and vero4k targets, but of course I can't really test them myself.

I see that v1.7.1 was tagged 12 days ago, so we can use that if you prefer not to track master (but I'd have to test again to make absolutely sure this point release doesn't have Pi-related issues vs latest master).

joolswills:
Thanks. Will check (Sorry I've not had time to finish looking into this PR - I still haven't got around to completing going over our existing tinker support as earlier code was made for the linaro image they provide, whereas I believe some later changes provided by users may be been tested on Armbian).

joolswills:
I've taken too long on this sorry. I'm going to merge and will fix up any issues with non RPI/x11 targets later. If you're happy with that Psyke83?

psyke83:
It'll negatively impact RPi users who use lr-ppsspp due to the save-state issue, but the standalone version works OK in that regard. Go ahead and merge if you're happy.

joolswills:
I'm going to merge. Can revisit if people moan and I will deny all knowledge :-)

Thanks for your work on this.

i30817:
Wayland being missing (probably) makes the core crash on a self-built retroarch without X11 support (only wayland). I did this because enabling X11 caused unacceptable delay on RA menus. Oh well.

PR #2425, opened 2018-06-12
love - allow installation of v0.10.2 in addition to latest version

rsn8887:
Many games for Love are incompatible with the latest version (v11). For example, the bundled game [Mari0](http://stabyourself.net/mari0/) requires v0.10.2, and the amazing game [Not Tetris](http://stabyourself.net/nottetris2/) requires 0.7.2.

With this PR, I am trying to add the option to install versions 0.10.2 and 0.7.2 as additional emulators for the system "love". I looked at advmame-x.xx as examples how this might work.

The goal is to allow users to use the runcommand menu to choose the version of Love on a per-game basis, similar to how it works with Advmame. The choices should be love (latest version, currently v11.1, the default), love-0.10.2, or love-0.7.2.

I am pretty sure I probably forgot something in this PR. I just copy-pasted from love.sh and made the obvious adjustments.

Can someone who is more familiar with RetroPie-Setup look over my first attempt and see if I forgot something before merging this?

Thank you!

### DIFF
--- a/scriptmodules/ports/love-0.10.2.sh
+++ b/scriptmodules/ports/love-0.10.2.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="love-0.10.2"
+rp_module_desc="Love - 2d Game Engine v0.10.2"
+rp_module_help="Copy your Love games to $romdir/love"
+rp_module_licence="GPL3 https://bitbucket.org/rude/love/raw/7b520c437317626da2102de1aafdad0e67b54bf5/license.txt"
+rp_module_section="opt"
+rp_module_flags="!aarch64"
+
+function depends_love-0.10.2() {
+    depends_love
+}
+
+function sources_love-0.10.2() {
+    hg clone https://bitbucket.org/rude/love/#0.10.2 "$md_build"
+}
+
+function build_love-0.10.2() {
+    build_love
+}
+
+function install_love-0.10.2() {
+    install_love
+}
+
+function game_data_love-0.10.2() {
+    game_data_love
+}
+
+function configure_love-0.10.2() {
+    configure_love
+}


joolswills:
If we do implement more than one version, it will need to be done in a way to avoid code duplication - eg, there should be a single module that does the work, and other modules that share it. I cannot include this as it is currently, sorry.

rsn8887:
Thanks for looking over this.

At the moment, I really don't know enough about bash scripting and RetroPie-Setup to improve this PR much.

I just used AdvanceMame for reference, since it has two versions. But that also has code-duplication.

Is there another module that does something similar, that you could recommend as a reference?

rsn8887:
I force pushed a change. The code duplication is removed. Is this what you had in mind?

joolswills:
thanks. We don't use () when calling functions elsewhere, so please remove those, and I will test / merge.

rsn8887:
Thanks for looking into this. I removed () on function calls.

joolswills:
Did you build and test the 0.7.2 version on the RPI ? It uses SDL1 by the looks of it, so it can't share the main love.sh dependencies.

joolswills:
If 0.7.2 builds and works fine on the RPI with SDL1 and the other dependencies are the same with the main love.sh, the main love.sh depends function can be adjusted to include sdl1/sdl2 depending on the module id.

rsn8887:
I pushed a change that hopefully fixes it, but it is untested at the moment. I should be able to test it later today.

rsn8887:
At the moment, everything works apart from launching games with lua v.0.7.2. To get v0.7.2 to compile, I had to change the dependencies for 0.7.2 from libsdl2 to libsdl1.2, and from libluajit to liblua. It looks like it compiles ok now, but when I try to run a game (Not Tetris) using 0.7.2, it just goes back to Emulationstation. I suspect it has something to do with the dependencies...

joolswills:
I don't think it's dependencies since it built. It's probably down to the SDL support being incompatible with the RPI or something.

I had assumed when you posted the PR that the binaries produced had been tested etc.

rsn8887:
I will do some further testing and hopefully fix this PR with another commit today or tomorrow. If Love 0.7.2 really does not run for some reason, it will still be worthwhile to add 0.10.2 at least.

rsn8887:
It looks like SDL 1.2 is built without OpenGL support, but Love 0.7.2 requires it. Since it is a very old version, I removed it.

So everything works now, and this PR will allow users to run all the games that require Love 0.10 but do not work with Love 0.11.

joolswills:
There is no opengl on the RPI on the default gpu driver. Will check and merge.

joolswills:
(gles can work with sdl, but likely requires some custom code)

joolswills:
Thanks.

PR #2426, opened 2018-06-14
Update lr-fbalpha.sh

barbudreadmon:
`profile=performance` argument doesn't do anything (it disappeared ~2 years ago), also, considering you keep dat files in pifba, perhaps you should do the same here.
Also, you don't plan to include the other subsystems ? Considering fbalpha run properly some games that won't on picodrive/genplus_gx, it's kinda sad.


joolswills:
Thanks.

I added neogeo cd subsystem the other day - I haven't got around to implementing the others yet.

PR #2429, opened 2018-06-15
Add lr-81: libretro core for Sinclair ZX81

SuperCatFooz:
I noticed this core sitting in the RetroArch repository, and decided to try compiling it on the Pi and seeing how well it works.  Since it ended up working more-or-less perfectly from the look of things, and because the ZX81 isn't available at all in RetroPie yet, I decided to write a script module for it.

Due to some control-related issues (specifically having to set player 2's controller type to "Sinclair Keyboard" in order to get inputs to work), I put the core in the Experimental section.  The RetroArch page also didn't mention whether or not .zip or .7z files work, so I excluded them for the time being.

I should also note that some colorized ROM files are present within the repository's `colorized` folder for use with the core's Chroma feature.  However, the author of the ROMs appear to have given permission to redistribute the files on the repository (plus they're redistributed on his website), so I'm not sure if they should be excluded or not.

### DIFF
--- a/scriptmodules/libretrocores/lr-81.sh
+++ b/scriptmodules/libretrocores/lr-81.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

joolswills:
Note the file had windows line endings which caused errors to be thrown. I have corrected it to LF now.

PR #2430, opened 2018-06-16
lr-fbalpha - added support for additional systems

joolswills:
* added pcengine (+supergrafx & turbografx-16), gamegear, mastersystem, megadrive, and sg-1000

This is currently untested. I believe the emulator needs specific romsets as it does with arcade for these systems.


joolswills:
Does anyone have an opportunity to test this with a fbalpha megadrive romset etc ?

PR #2432, opened 2018-06-22
Added new rp_module_sites variable to scripts.

lashower:
rp_module_sites is a list of sites related to the programs.

While this is not hooked into anything in retropie_packages.sh, it could be useful in the future.

My plan is to pull the list with my RetroPieWebConsole application.


joolswills:
Sorry but I don't want to add more module variables yet. I have some plans for some additional ones in the future that actually may contain this info, but don't want to add anything yet.

PR #2434, opened 2018-06-23
Add lr-pokemini: experimental libretro core for Pokemon Mini

SuperCatFooz:
Title pretty much says it all.  I decided to take this opportunity to edit platforms.cfg to add Pokemon Mini platform definitions, since that seems to be how es_systems determines what file types to use.  (I might also submit a PR for the ZX81, as that script appears to not place file extensions otherwise, thus requiring the user to edit es_systems manually to add the extensions.)

I could just be misunderstanding how scriptmodules work, so feel free to change as you see fit.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -166,7 +166,7 @@ pcengine_theme="pcengine"
 pcfx_exts=".pce .ccd .cue .iso .chd"
 pcfx_fullname="PC-FX"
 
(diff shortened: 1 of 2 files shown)


joolswills:
I have fixed the missing zx81 platform.cfg entries.

I will check this PR when I get a chance and squash/merge (Please squash commits in the future for small changesets).

PR #2438, opened 2018-07-01
Add support for the Vero4k

hissingshark:
The Vero4k is the most recent box running OSMC.  Similar to an ODROID-C2 I think, but with it's own quirks and requirements.

Note - this is in parallel with a PR for emulationstation.
Usual fixes and detection type changes for the platform.

My PR relies on a 2nd commit to upgrade `gitPullOrClone`.  This is to address platforms that don't have binaries available for download and are forced to build from master.  I've been working on this fork in my spare time for over a year, and always have to hold off on my PR because of upstream breakage in master.  If there's no stable branch or version tag available upstream then you can't install the package.  Now you can specify a last working commit to build from instead.

I haven't used an `rp_module_flags=!vero4k` type flag at this stage.  I've not tried to build every single emulator yet.  Should I veto those with such a flag?  Or wait for feedback from a user who actually cares to use it and then investigate?

There's a lot to look over here I know.  I've tried to follow the style guide but I realise as a 1st PR here I may have tripped.
Many thanks in advance.


joolswills:
Looks mostly fine - thanks for the contribution. I will make a few comments for some minor changes when I get a moment.

scriptmodules/emulators/retroarch.sh
```diff
@@ -56,6 +57,7 @@ function build_retroarch() {
     isPlatform "arm" && params+=(--enable-floathard)
     isPlatform "neon" && params+=(--enable-neon)
     isPlatform "x11" && params+=(--enable-vulkan)
+    isPlatform "vero4k" && params+=(--enable-dbus --enable-libusb --enable-threads --enable-mali_fbdev --enable-zlib --enable-alsa --enable-freetype --disable-x11 --disable-vulkan --with-opengles_libs='-L/opt/vero3/lib')
```
joolswills: some of these flags we currently "autodetect" - but if we are to force them, we may want to do it for all the platforms - eg enabling zlib is not vero4k specific.

scriptmodules/helpers.sh
```diff
@@ -345,13 +347,19 @@ function gitPullOrClone() {
...
             git+=" --depth 1"
         fi
         [[ "$branch" != "master" ]] && git+=" --branch $branch"
         echo "$git \"$repo\" \"$dir\""
         runCmd $git "$repo" "$dir"
     fi
+    if [[ "$commit" ]]; then
+        pushd "$dir" > /dev/null
+        echo "Winding back $repo: $branch to commit #$commit"
+        runCmd git checkout $commit
```
joolswills: is it possible to use `git -C` rather than pushd / popd ?

scriptmodules/libretrocores/lr-pcsx-rearmed.sh
```diff
@@ -16,7 +16,11 @@ rp_module_licence="GPL2 https://raw.githubusercontent.com/libretro/pcsx_rearmed/
 rp_module_section="main"
 
 function depends_lr-pcsx-rearmed() {
-    getDepends libpng12-dev libx11-dev
+    if isPlatform "vero4k"; then
+        getDepends libpng-dev
```
joolswills: We handle libpng12-dev -> libpng-dev mapping based on OS version - if you set _os_debian_ver=9 it will do this afair. Once we drop jessie support we can lost all the libpng12-dev for libpng-dev also.

scriptmodules/supplementary/sdl2.sh
```diff
@@ -38,6 +39,8 @@ function depends_sdl2() {
     isPlatform "mali" && depends+=(mali-fbdev)
     isPlatform "kms" && depends+=(libdrm-dev libgbm-dev)
     isPlatform "x11" && depends+=(libpulse-dev libegl1-mesa-dev libgles2-mesa-dev libglu1-mesa-dev)
+    # for vero4k start depends again due to unwanted X packages
```
joolswills: possible to avoid putting the depends again by excluding them for vero4k earlier ?

scriptmodules/system.sh
```diff
@@ -349,3 +352,11 @@ function platform_imx6() {
     __default_makeflags="-j2"
     __platform_flags="arm armv7 neon"
 }
+
+function platform_vero4k() {
+    __default_cflags="-I/opt/vero3/include -L/opt/vero3/lib -O3 -march=armv8-a+crc -mtune=cortex-a53 -mfpu=neon-fp-armv8 -mfloat-abi=hard -ftree-vectorize -funsafe-math-optimizations"
```
joolswills: Would prefer -O2 for default optimisation level (many of the emulators override it anyway)

joolswills:
Thanks again - made a few comments. Overall very nice though.

scriptmodules/supplementary/sdl2.sh
hissingshark: As it's to avoid X11 packages, do you mean they could go into `isPlatform "x11"` perhaps?  Certainly works for my box.
I was basically trying not to break other peoples stuff. But if a platform doesn't bear the `x11` flag I don't see why they'd need anything X related.

scriptmodules/system.sh
hissingshark: Done.

scriptmodules/helpers.sh
hissingshark: `-C`.  Every day's a schools day.  Should work fine.  Building now.

scriptmodules/libretrocores/lr-pcsx-rearmed.sh
hissingshark: Ah yes, I see what you're doing in `helpers.sh`.  It's a bit pedantic here I know, but I was there anyway trying to exclude the unneeded libx11-dev.  In retrospect - based on the x11 issue discussed below - I'll filter that for `x11` platforms then?  And keep the "generic" libpng12-dev for now.  Again it works for me either way, just don't want to break anyone else's builds.

scriptmodules/emulators/retroarch.sh
hissingshark: This was one of the first things I built, before I'd even considered a full blown RetroPie fork.  My build params are belt and braces from the day.
I've just reviewed the configure script, done some test builds and `./retroarch --features`.  Nearly everything is auto-detected by the build system, based on installed packages.
My params are now leaner to say the least!  How embarrassing.

joolswills:
Is this ready to go? - please let me know and I will merge.

hissingshark:
I've responded to each comment in turn and submitted commit #479dbe9 to address the feedback.

As long as you're happy with the revisions i.e. I've not broken anything, then it's good to go.  In particular that would be the X11 and Mesa package related changes in `sdl.sh` and `lr-pcsx-rearmed.sh` which I can't test here.

joolswills:
are you ok to squash the PR feedback changes into the first commit ? I have a couple more comments also.

scriptmodules/emulators/scummvm.sh
```diff
@@ -17,7 +17,12 @@ rp_module_section="opt"
 rp_module_flags=""
 
 function depends_scummvm() {
-    local depends=(libmpeg2-4-dev libogg-dev libvorbis-dev libflac-dev libmad0-dev libpng12-dev libtheora-dev libfaad-dev libfluidsynth-dev libfreetype6-dev zlib1g-dev libjpeg-dev)
+    local depends=(libmpeg2-4-dev libogg-dev libvorbis-dev libflac-dev libmad0-dev libtheora-dev libfaad-dev libfluidsynth-dev libfreetype6-dev zlib1g-dev libjpeg-dev)
+    if isPlatform "vero4k"; then
+        depends+=(vero3-userland-dev-osmc libpng-dev)
+    else
+        depends+=(libpng12-dev)
+    fi
```
joolswills: would prefer to handle the png logic as mentioned before - using the automatic system (set the raspbian version to 9 and it should work automatically). I will be changing all libpng12-dev to libpng-dev in the future (and put a workaround in for older systems to use libpng12-dev)

hissingshark:
No problem.  All commits squashed.
Ah yes I spotted another `libpng` I'd overlooked too.

scriptmodules/emulators/ppsspp.sh
```diff
@@ -157,9 +164,9 @@ function configure_ppsspp() {
     ln -snf "$romdir/psp" "$md_conf_root/psp/PSP/GAME"
 
     if isPlatform "tinker"; then
-        addEmulator 0 "$md_id" "psp" "$md_inst/PPSSPPSDL --fullscreen %ROM%"
+        addEmulator 1 "$md_id" "psp" "$md_inst/PPSSPPSDL --fullscreen %ROM%"
     else
-        addEmulator 0 "$md_id" "psp" "$md_inst/PPSSPPSDL %ROM%"
+        addEmulator 1 "$md_id" "psp" "$md_inst/PPSSPPSDL %ROM%"
```
joolswills: Sorry just noticed this as I was about to merge - Not sure this is relevant to the PR ? Changing the default emulator for ppsspp.

hissingshark: Yes good spot, thank you.
I couldn't access the runcommand popup to select it back when I was runtime testing (starting emulationstation via a mobile SSH app) so it wouldn't start.  We can't use the libretro plugin at the moment, but that's no need to make the standalone default here.
Fixed -> commited -> squashed.

hissingshark:
Happy to be here!  And thank you for the constructive feedback throughout the process.

joolswills:
The SDL2 changes broke SDL2 building from source - this may actually be related to some tickets about sdl2 not working for source based platforms other than X11 or RPI targets (Odroid / Tinker).

Have made a change here to only exclude the deps from vero4k but it's not ideal - 0b2e2378cc183d1a6c85c097f05e5f2fe870b38d

hissingshark:
Thanks for letting me know.  I am sorry about that.  I was [concerned ](https://github.com/RetroPie/RetroPie-Setup/pull/2438#discussion_r199364633 ) about the bigger picture that I couldn't test.  I guess that has finally come home to roost.

[Your fix](https://github.com/RetroPie/RetroPie-Setup/commit/0b2e2378cc183d1a6c85c097f05e5f2fe870b38d) does seem a more sensible flow.  What don't you like about it?

joolswills:
well it's ok, but ideally I guess we would build sdl with X in also for vero4k (as we do on the rpi). I guess I don't like code for platforms I can't test :-)

PR #2442, opened 2018-07-16
Fix threaded ALSA audio driver selection

cmitu:
Fix the mis-spelling of the **alsathread** audio driver in RetroArch Config. Reported in https://retropie.org.uk/forum/topic/18430/lr-armsnes-and-retropie-4-4/30.


joolswills:
thanks.

PR #2445, opened 2018-07-22
Update for the advmame scriptmodule.

cmitu:
* get the package source from the github.com repository so it's not taken from the SF download page and uploaded on the RetroPie files server.
* run `autoreconf.sh` before configuring the build (needed to produce the `configure` script).
* update the build function to check for success using `$md_ret_files`

Started by https://retropie.org.uk/forum/topic/18473/update-of-advancemame-did-not-work

### DIFF
--- a/scriptmodules/emulators/advmame.sh
+++ b/scriptmodules/emulators/advmame.sh
@@ -26,20 +26,22 @@ function _update_hook_advmame() {
 }
 
 function depends_advmame() {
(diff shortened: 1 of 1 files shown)


joolswills:
Pretty sure we get the source archive from github (the one linked from the advancemame website). However happy to switch to using github directly for checking out the source. Good spot on the missing md_ret_files. Thanks.

PR #2455, opened 2018-08-12
dependency fixes to helpers, runcommand and mupen64plus

hhromic:
While working on a bare-minimum version of Raspbian Stretch with RetroPie, I discovered dependencies missing in runcommand and mupen64plus that are assumed to be installed: `fbset` and `libraspberrypi-bin`. I also slightly enhanced ensureFBMode() in helpers to account for a potentially non-existing `/etc/fb.modes`, e.g. for the case `fbset` is not installed beforehand. All details are in the commits.

* mupen64plus - add libraspberrypi-bin as dependency
* runcommand - add fbset and libraspberrypi-bin as dependencies
* helpers - ensureFBMode(): no-op if /etc/fb.modes doesn't exist


joolswills:
thanks.

PR #2456, opened 2018-08-12
add .zip extension to n64 platform for lr-mupen64plus

hhromic:
The `lr-mupen64plus` retroarch core does support zip files, add them as supported files for Nintendo 64.
Tested and working.


hhromic:
thanks for accepting, I now updated the Wiki aswell to indicate that lr-mupen64plus can load zip files too.

PR #2459, opened 2018-08-14
helpers - document and improve the ask() function

hhromic:
Someone forgot to document the `ask()` function in `helpers.sh`.  Fix that :)

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -56,6 +56,11 @@ function fnExists() {
     return $?
 }
 
+# @fn ask()
+# @param message string or array of messages for the prompt
+# @brief Asks the user for a yes/no decision.
+# @retval 0 if the user selected the "yes" option
+# @retval 1 if the user gave any other input, including "no"
 function ask() {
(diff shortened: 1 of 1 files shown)


hhromic:
fine-tuned the function a bit too on the spot:
* set `ans` as local variable
* do not allow backslashes to escape any characters in `read`

joolswills:
Actually it was intentional as the function will be removed - it's a leftover from very early versions of RetroPie-Setup and isn't used anywhere. I will remove it at some point.

hhromic:
Ah, a vestigial function. oh well.. you can discard this PR then if that is the case.

hhromic:
I can confirm (via grep) that no script in the repository is using ask() currently.
What stops its removal?  I can modify this PR to remove the function instead and keep RetroPie tidy.
Let me know.

joolswills:
I just haven't got around to it. Appreciate the offer, but I don't think it's needed to do a PR for the removal.

hhromic:
No problem. I will close this PR then. Cheers and great job in general with RetroPie, very impressive bash programming.

joolswills:
Thanks for your support.

PR #2460, opened 2018-08-15
runcommand - use existing VIDEO_CONF variable

hhromic:
Another small fix:
* The `VIDEO_CONF` variable is initialised at the beginning of the script with the appropiate file path. However, the `default_mode()` function does not use it and instead hard-codes the file path.


joolswills:
Thanks

PR #2461, opened 2018-08-16
Revert "dependency fixes to helpers, runcommand and mupen64plus"

joolswills:
Reverts RetroPie/RetroPie-Setup#2455

Reverting due to OSMC dependency issue - no raspberrypi-bin on OSMC.


hhromic:
Ah I just realised that indeed the conditional is there, but OSMC also matches `rpi` for isPlatform.
This is an unfortunate problem. Perhaps the conditional can be done against running on Raspbian?
I will study it better. In the meantime I'm going to split the reverted PR into individual changes so we can re-add the other unrelated fixes I submitted. Sorry again.

hhromic:
I just checked and can confirm that the following packages supply `tvservice` for OSMC and Xbian:
* `rbp-userland-osmc` for OSMC
* `xbian-package-firmware` for Xbian
So the workaround should be trivial.

PR #2463, opened 2018-08-16
runcommand - add fbset as dependency

hhromic:
The runcommand script uses the `fbset` binary and in some
systems this package might not be installed by default.


joolswills:
Thanks

PR #2465, opened 2018-08-17
helpers: getDepends() improve raspberrypi userland packages workaround

hhromic:
Here is an improvement for getDepends() on handling the raspberrypi userland packages. In particular, I found out we were also missing the `libraspberrypi0` package. Now the workaround should be complete.

The raspberrypi userland packages have different names on osmc and xbian:

    libraspberrypi0 -> rbp-userland-osmc, xbian-package-firmware
    libraspberrypi-bin -> rbp-userland-osmc, xbian-package-firmware
    libraspberrypi-dev -> rbp-userland-dev-osmc, xbian-package-firmware

This commit:

* Makes existing workaround code simpler and handle missing `libraspberrypi0` case.
* Improves existing workaround that ensures up-to-date userland packages.
* Bumps the minimum userland version from `1.20170703-1` to `1.20180328-1`.

Validation:

* Confirmed package names mapping using official repositories:
   Raspbian: <http://mirrordirector.raspbian.org/raspbian/dists/stretch/firmware/binary-armhf/Packages>
   OSMC: <http://ftp.fau.de/osmc/osmc/apt/dists/stretch/main/binary-armhf/Packages>
   Xbian: <http://apt.xbian.org/dists/stable/main/binary-armhf/Packages>
* Confirmed that the new userland version exists for both `jessie` and `stretch` distributions.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -203,15 +203,20 @@ function getDepends() {
 
     for required in $@; do
 
-        # workaround for different package names on osmc / xbian
-        if [[ "$required" == "libraspberrypi-bin" ]]; then
-            isPlatform "osmc" && required="rbp-userland-osmc"
-            isPlatform "xbian" && required="xbian-package-firmware"
-        fi
-        if [[ "$required" == "libraspberrypi-dev" ]]; then
-            isPlatform "osmc" && required="rbp-userland-dev-osmc"
-            isPlatform "xbian" && required="xbian-package-firmware"
-        fi
+        # workaround for raspberrypi userland packages having different names on osmc / xbian
+        # libraspberrypi0 -> rbp-userland-osmc, xbian-package-firmware
+        # libraspberrypi-bin -> rbp-userland-osmc, xbian-package-firmware
+        # libraspberrypi-dev -> rbp-userland-dev-osmc, xbian-package-firmware
+        case "$required" in
+            libraspberrypi0|libraspberrypi-bin)
+                isPlatform "osmc" && required="rbp-userland-osmc"
+                isPlatform "xbian" && required="xbian-package-firmware"
+                ;;
+            libraspberrypi-dev)
+                isPlatform "osmc" && required="rbp-userland-dev-osmc"
+                isPlatform "xbian" && required="xbian-package-firmware"
+                ;;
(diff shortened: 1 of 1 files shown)


joolswills:
libraspberrypi0 should never be included as a dependency though in a module as it's referenced by other packages (libraspberrypi-bin / libraspberrypi-dev) so that's not needed.

The code to force a specific library version doesn't need to be updated - that could break jessie - it's to force a version that had a library rename change.

joolswills:
best to leave for now - I may change to case / esac if we have any more in the future.

joolswills:
I see you said the userlane version bump is ok on Jessie, but I don't see any reason to bump it currently for the reasons above (it was for a specific purpose).

hhromic:
No problem! I just wanted to make it more complete/robust.
I realise now that because everything, one way or another, comes from compilation and not pure binaries. e.g. closed source, a module should never ask for `libraspberrypi0` alone. Fair point, but shall `getDepends()` rely on user-modules to behave properly?

Regarding the version upgrade, if there was a very specific reason then I also understand to leave it as-is. I just tried to keep it more current. Firmware upgrades on RPI are particularly beneficial, specially for the newest RPI3B+.

Finally, regarding the case/esac, I was thinking the rest of the workarounds in `getDepends()` can be integrated into a unified case/esac block for readability/clearness of the code. If you wish I can develop this improvement and send another PR for you to review/merge.

Cheers!

joolswills:
Not sure what you mean by "User Modules". getDepends works with the RetroPie modules we have - we check code we accept into RetroPie. No modules should include binary libraries - same for things like libpng etc. 

I would be reluctant to do a code refactor on getDepends at this time. I don't think the rest of the logic would be any better in a case/esac.

hhromic:
Oh sorry.  I normally see users sending PRs etc. for these modules, hence I informally tend to call them "user modules". I agree that getDepends() and others in helpers are fundamental functions that the potential for breaking them without a good reason is not worth it. No worries.
Anyway, I will still keep an eye on my spare time to review/test code in RetroPie and send more PRs to fix things if I find them. Looks like I'm enjoying more improving RetroPie than actually playing games :)
Again, amazing job you guys have done with this project.

PR #2468, opened 2018-08-20
Script to install, update and remove Skyscraper using RetroPie-Setup

muldjord:
Per user request I've created a script for inclusion in the RetroPie-Setup which allows the Skyscraper game scraper to be installed, updated and removed. I've also included a small help text to clarify the usage after installation.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -15,22 +15,30 @@ rp_module_help="IMPORTANT! In order for Skyscraper to work properly you need to 
 rp_module_licence="GPL3 https://raw.githubusercontent.com/muldjord/skyscraper/master/LICENSE"
 rp_module_section="opt"
 
-skysource="/home/$user/skysource"
-skyconfig="/home/$user/.skyscraper"
+function get_skysource() {
+    local skysource="/home/$user/skysource"
+    echo "$skysource"
+}
+
+function get_skyconfig() {
+    local skyconfig="/home/$user/.skyscraper"
+    echo "$skyconfig"
+}
 
 function depends_skyscraper() {
     local depends=(qt5-default)
 }
 
 function sources_skyscraper() {
+    local skysource=$(get_skysource)
+    
     if [[ -f "$skysource/Skyscraper" ]]; then
-        cd $skysource
+        cd "$skysource"
 	./update_skyscraper.sh
-	sleep 5
     else
-        mkdir $skysource
-        cd $skysource
-        curl https://raw.githubusercontent.com/muldjord/skyscraper/master/update_skyscraper.sh | bash
+        mkdir "$skysource"
+        cd "$skysource"
+	wget -q -O - https://raw.githubusercontent.com/muldjord/skyscraper/master/update_skyscraper.sh | bash
     fi
 
     cp Skyscraper $md_build
@@ -44,21 +52,18 @@ function install_skyscraper() {
 }
 
 function remove_skyscraper() {
+    local skysource=$(get_skysource)
+    local skyconfig=$(get_skyconfig)
+    
     if [[ -f "$skysource/Skyscraper" ]]; then
-	cd $skysource
+	cd "$skysource"
 	make uninstall
-	rm -Rf $skysource
+	rm -Rf "$skysource"
     fi
 
     clear
 
     if [[ -d "$skyconfig" ]]; then
-	read -n1 -p "Do you wish to delete the $skyconfig cache and config folder (y/n)? " purgeall
-	echo \n
-	case $purgeall in  
-	    y|Y) echo "Deleting $skyconfig folder..."
-		 rm -Rf $skyconfig
-		 ;; 
-	esac
+	rm -Rf "$skyconfig"
     fi
 }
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,64 @@
...
+#
+
+rp_module_id="skyscraper"
+rp_module_desc="Scraper for multiple frontends by Lars Muldjord"
+rp_module_help="IMPORTANT! In order for Skyscraper to work properly you need to exit your frontend. For EmulationStation you can do so by pressing F4.\nTo run Skyscraper in Simple Mode, exit your frontend and simply type 'Skyscraper' and press enter. Then answer the questions you are given and let Skyscraper do the work.\nIf you want better control check the available options with 'Skyscraper --help'.\n\nBe sure to check the elaborate documentation at github: https://github.com/muldjord/skyscraper"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/muldjord/skyscraper/master/LICENSE"
+rp_module_section="opt"
+
+skysource="/home/$user/skysource"
+skyconfig="/home/$user/.skyscraper"
```
joolswills: we don't allow globals in the modules. can use a function to return the location in the module.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,64 @@
...
+
+function depends_skyscraper() {
+    local depends=(qt5-default)
+}
+
+function sources_skyscraper() {
+    if [[ -f "$skysource/Skyscraper" ]]; then
+        cd $skysource
```
joolswills: should be quoted.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,64 @@
...
+	./update_skyscraper.sh
```
joolswills: wrong indent

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,64 @@
...
+	sleep 5
```
joolswills: why the sleep ?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,64 @@
...
+    else
+        mkdir $skysource
+        cd $skysource
+        curl https://raw.githubusercontent.com/muldjord/skyscraper/master/update_skyscraper.sh | bash
+    fi
```
joolswills: quotes - also please use wget instead of curl as we ensure wget is installed.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,64 @@
...
+
+    if [[ -d "$skyconfig" ]]; then
+	read -n1 -p "Do you wish to delete the $skyconfig cache and config folder (y/n)? " purgeall
+	echo \n
+	case $purgeall in  
+	    y|Y) echo "Deleting $skyconfig folder..."
+		 rm -Rf $skyconfig
+		 ;; 
+	esac
+    fi
```
joolswills: no user interaction can be included in the remove function.

joolswills:
Thanks - Unfortunately there are many issues with the module in its current form. I have pointed out some, but there are other issues too. It may just be better to include this as a feature request and I will look into it at some point.

scriptmodules/supplementary/skyscraper.sh
joolswills: I would prefer to not use a 3rd party script for the building but include the steps in our module.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,64 @@
...
+        'Skyscraper'
+    )
+}
+
+function remove_skyscraper() {
+    if [[ -f "$skysource/Skyscraper" ]]; then
+	cd $skysource
+	make uninstall
+	rm -Rf $skysource
+    fi
```
joolswills: wrong indent.

muldjord:
Thank you very much for your comments. I've tried to incorporate your changes. The only thing I am a bit in doubt about is the faulty indent you mentioned.
Concerning the inclusion of the script from my repository, I understand why you wouldn't want that as you don't control it. I can easily just paste my script in there if I can make it conform to your format. Would that be acceptable to you? The only reason I use it directly from my repository, is in relation to updates I might make myself. But I can easily update both places if I ever change it, no problem.

Thank you for taking the time to look through this.

joolswills:
You have used tab indentation in places instead of 4 spaces

I still think it may be best for me to just do a module for this - it may take less time than the back and forth for the review. This module has a lot of things I would change before inclusion.

muldjord:
I understand. I will let this rest then. Thanks again. :)

cmitu:
Can I take a stab at this ? I've done some preliminary work based on Steven Selph's `scraper` module.

joolswills:
sure. thanks.

cmitu:
Proposed a new version #2497.

joolswills:
Closing this in favour of the newer PR.

PR #2470, opened 2018-08-21
system - add missing dependencies for get_retropie_depends()

hhromic:
Just catched these missing deps too:

* `ca-certificates` is needed by `git` for https repositories
* `procps` is needed by `system.sh`:

      scriptmodules/system.sh: line 20: free: command not found
      scriptmodules/system.sh: line 21: free: command not found

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -166,7 +166,7 @@ function get_os_version() {
 }
 
 function get_retropie_depends() {
-    local depends=(git dialog wget gcc g++ build-essential unzip xmlstarlet python-pyudev)
+    local depends=(git ca-certificates dialog wget gcc g++ build-essential unzip xmlstarlet python-pyudev procps)
 
     if ! getDepends "${depends[@]}"; then
         fatalError "Unable to install packages required by $0 - ${md_ret_errors[@]}"


joolswills:
procps is an essential package - I don't think we need to include it in retropie. If you have a minimal install without it you will have to just make sure you have the basic system set-up.

hhromic:
The way I see it, is that regardless of packages being essential or not, RetroPie *needs them* and the job of `get_retropie_depends()` is to *ensure that all of its dependencies are met*, without any pre-assumption.
I understand "essential" packages are supposed to be always present but this is not guaranteed, e.g. not strictly "essential" like the kernel :). Adding procps as dependency of RetroPie would just ensure it is always installed. In many systems this will be a no-op install because as you say such packages are always installed by many distros. However for these special cases of minimal systems it will make sure RetroPie has everything it needs available and for the rest of the users is 100% harmless.

I really hope I can change your mind because RetroPie is already very impressively programmed/maintained to manage all of its module dependencies, working really well on my minimal system. I have had to manually install almost nothing by myself (only exceptions are the few PRs you already merged). In the same vein, this PR is just to keep polishing these packages that, as you said, were not expected to be missing and it would be a shame that there are some small things left behind.

Btw, I'm currently testing retropie-setup on a Docker standard debian image, which is even slimmer than my Raspbian nano image, and RetroPie handles dependencies very well even there. This is a remarkable achievement.

joolswills:
I don't agree - Debian / Ubuntu doesn't even let you remove procps without warning you will remove dependencies that are essential so I don't see it as an issue. What next ? Adding dependencies for fileutils or systemd? This has never been an issue before - if you are building some custom tiny OS, it's up to you to make sure you have important base packages installed.

hhromic:
I agree there is a limit on adding dependencies, true that. However a big difference is that the system doesn't even boot-up without an init system (well, you can argue that you can always do `init=/bin/sh` and bring-up all manually) but in contrast a system does work fine without procps.
Anyway, you are the boss and I respect your decision :+1: 

To wrap-up, you also are discarding the `ca-certificates` dependency? Shall I update the PR to only include that one and not procps?

joolswills:
I will add the dependency for that as it is possible to remove that without issues.

hhromic:
thanks!

PR #2471, opened 2018-08-22
Deprecate stdin usage (fixes #2467)

### LINKED ISSUE #2467: applyPatch/addPort called without stdin fails when stdin is not a terminal
Hello @joolswills, I believe I found a bug in `applyPatch` and `addPort` from `helpers.sh`.
**Warning:** this issue report turned a bit long :P.

When you run a script that uses `applyPatch` with a stdin redirection, i.e. not connected to a terminal, the `applyPatch` function always tries to apply a patch from stdin regardless if a real patch file was provided. And in the latter case, the real patch file is overwritten with whatever content was passed to stdin. The same problem affects `addPort` which also relies on detecting stdin being a terminal.
Note: the reason I'm experiencing this corner case is because I'm writing my own script that calls `retropie_packages.sh` and sometimes I give my script content via stdin for automation.

You can reproduce this with e.g. `lr-mupen64plus` that applies a patch from a file, not stdin:
https://github.com/RetroPie/RetroPie-Setup/blob/8cf6a70440d7f633ad9f4782961f7db664a8f0c7/scriptmodules/libretrocores/lr-mupen64plus.sh#L43

For example (on an RPI machine):

    sudo $HOME/Retropie-Setup/retropie_packages.sh lr-mupen64plus <<<"something"

fails because stdin is not a terminal and therefore `applyPatch` blindly consumes the given stdin overwritting `$md_data/01_new_lib_names.diff` with `something`. After the above command you can verify that indeed the real patch file was overwritten and hence the patching fails.

The problem comes from the following snippet in `applyPatch`:
https://github.com/RetroPie/RetroPie-Setup/blob/8cf6a70440d7f633ad9f4782961f7db664a8f0c7/scriptmodules/helpers.sh#L911-L914

I see a couple of options to solve this:

1. Make sure the patch file doesn't exist before consuming stdin. Nothing else needs changing. Example:
   ```sh
   if [[ ! -f "$patch" && ! -t 0 ]]; then
       cat >"$patch"
   fi
   ```

2. Do not try to detect stdin and instead always use patch files. This is my preferred solution, see below.

All modules that use in-line patching should have instead a proper file with the corresponding patch and remove the in-lines from the module itself. Currently there is a mixture of modules using patches from files and in-line in the module. I think this is a bad practice as it is ugly/error prone and should be discouraged.

The following are the affected modules that use `applyPatch` with in-line patches:

    $ grep -r applyPatch | grep '<<'
    scriptmodules/supplementary/xpad.sh:    applyPatch "retropie.diff" <<\_EOF_
    scriptmodules/emulators/atari800.sh:        applyPatch rpi_fixes.diff <<\_EOF_
    scriptmodules/emulators/fuse.sh:        applyPatch cursor.diff <<\_EOF_
    scriptmodules/emulators/scummvm-sdl1.sh:        applyPatch rpi-sdl1.diff <<\_EOF_
    scriptmodules/emulators/scummvm.sh:        applyPatch rpi_enable_scalers.diff <<\_EOF_
    scriptmodules/emulators/uae4all.sh:    applyPatch guichan.diff <<\_EOF_

Also, the current patches on disk are somewhat very inconsistently named:

    $ find scriptmodules -name *.diff -or -name *.patch
    scriptmodules/ports/darkplaces-quake/rpi.diff
    scriptmodules/supplementary/customhidsony/0001-hidsony-nomotionsensors.diff
    scriptmodules/emulators/retroarch/03_disable_udev_sort.diff
    scriptmodules/emulators/retroarch/02_disable_search.diff
    scriptmodules/emulators/retroarch/01_hotkey_hack.diff
    scriptmodules/emulators/ppsspp/02_tinker_options.diff
    scriptmodules/emulators/ppsspp/01_egl_name.diff
    scriptmodules/emulators/xm7/01_fix_build.diff
    scriptmodules/emulators/mupen64plus/0001-GLideN64-use-emplace.patch
    scriptmodules/emulators/quasi88/01_fixes.diff
    scriptmodules/emulators/advmame/01_rpi_framebuffer.diff
    scriptmodules/libretrocores/lr-hatari/01_libcapsimage.diff
    scriptmodules/libretrocores/lr-mupen64plus/01_new_lib_names.diff

I think these should be always `.patch` and the name descriptive of what the patch does. A good example is `0001-GLideN64-use-emplace.patch`. This is a standard naming convention. Doing this migration and fixing `applyPatch` to not use stdin to overwrite patches is trivial.

If you agree with any of these solutions, let me know and I can test and prepare a PR.

hhromic:
As discussed in #2467, deprecate stdin usage in `applyPatch` / `addPort`.
This is a somewhat big PR, so I will summarise the commits here:

* a367202 and 9534c17 remove stdin functionality from `applyPatch` and `addPort`.
* 67cdb4f to a89c3f5 moves and rebases the in-line patches into files for affected modules.
* 4c3146c updates the only module using stdin with `addPort` to use new approach.

In the case of `applyPatch`, I did the following:

* Tested patching in every affected module and in every supported platform (see below).
* Rebased the patches so they apply cleanly into current source code. Some patches didn't even apply anymore due to being outdated, this is now fixed too.

In the case of `addPort`, I did the following:

* Changed the signature of the function to accept a custom script as a parameter instead of stdin.
* Checked and updated the only module using addPort with stdin and updated to new signature.
* Tested all modules using `addPort` and verify they are still creating launch scripts properly.

Because the change in `applyPatch` is sensitive, I wrote the following test and verified that there were no errors and all patches are applied cleanly:
```sh
#!/usr/bin/env bash

PLATFORMS=(
  rpi1 rpi2 rpi3
  odroid-c1 odroid-c2 odroid-xu
  tinker
  x86 generic-x11
  armv7-mali
  imx6
  vero4k
)

PACKAGES=(
  atari800 darkplaces-quake fuse
  mupen64plus scummvm scummvm-sdl1
  uae4all xpad
)

for platform in "${PLATFORMS[@]}"; do
  for package in "${PACKAGES[@]}"; do
    printf "Package '$package' in platform '$platform' ...\n\n"
    __platform=$platform ./retropie_packages.sh "$package" clean
    __platform=$platform ./retropie_packages.sh "$package" sources
    printf "\n"
  done
done
```

Take your time to review this PR and let me know if I'm missing something. I'm happy to help.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1147,6 +1147,7 @@ function delSystem() {
 ## @param name display name for the launch script
 ## @param cmd commandline to launch
 ## @param game rom/game parameter (optional)
+## @param script launch script to use instead of the default (optional - requires game to be set)
 ## @brief Adds a port to the emulationstation ports menu.
 ## @details Adds an emulators.cfg entry as with addSystem but also creates a launch script in `$datadir/ports/$name.sh`.
 ##
@@ -1166,6 +1167,7 @@ function addPort() {
     local file="$romdir/ports/$3.sh"
     local cmd="$4"
     local game="$5"
+    local script="$6"
 
     # move configurations from old ports location
     if [[ -d "$configdir/$port" ]]; then
@@ -1185,13 +1187,13 @@ function addPort() {
 
     mkUserDir "$romdir/ports"
 
-    if [[ -t 0 ]]; then
+    if [[ -z "$script" ]]; then
         cat >"$file" << _EOF_
 #!/bin/bash
 "$rootdir/supplementary/runcommand/runcommand.sh" 0 _PORT_ "$port" "$game"
 _EOF_
     else
-        cat >"$file"
+        cat >"$file" <<< "$script"
     fi
 
     chown $user:$user "$file"
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -908,14 +908,10 @@ function applyPatch() {
     local patch="$1"
     local patch_applied="${patch##*/}.applied"
 
-    # patch is in stdin
-    if [[ ! -t 0 ]]; then
-        cat >"$patch"
-    fi
-
     if [[ ! -f "$patch_applied" ]]; then
         if patch -f -p1 <"$patch"; then
             touch "$patch_applied"
+            printMsgs "console" "Successfully applied patch: $patch"
         else
             md_ret_errors+=("$md_id patch $patch failed to apply")
             return 1
--- a/scriptmodules/emulators/atari800.sh
+++ b/scriptmodules/emulators/atari800.sh
@@ -25,32 +25,7 @@ function depends_atari800() {
 function sources_atari800() {
     downloadAndExtract "$__archive_url/atari800-3.1.0.tar.gz" "$md_build" 1
     if isPlatform "rpi"; then
-        applyPatch rpi_fixes.diff <<\_EOF_
---- a/src/configure.ac
-+++ b/src/configure.ac
-@@ -136,7 +136,8 @@ if [[ "$a8_target" = "ps2" ]]; then
-     LDFLAGS="$LDFLAGS -L${PS2SDK}/ports/lib"
- fi
- if [[ "$a8_target" = "rpi" ]]; then
--    CC="${RPI_SDK}/bin/arm-linux-gnueabihf-gcc"
-+    [[ -z "$RPI_SDK" ]] && RPI_SDK="/opt/vc"
-+    CC="gcc"
-     CFLAGS="$CFLAGS -I${RPI_SDK}/include -I${RPI_SDK}/include/SDL -I${RPI_SDK}/include/interface/vmcs_host/linux -I${RPI_SDK}/include/interface/vcos/pthreads"
-     LDFLAGS="$LDFLAGS -Wl,--unresolved-symbols=ignore-in-shared-libs -L${RPI_SDK}/lib"
- fi
-@@ -309,8 +310,9 @@ dnl BeOS has a real issue with redundant-decls
-         AC_DEFINE(SUPPORTS_PLATFORM_CONFIGURE,1,[Additional config file options.])
-         AC_DEFINE(SUPPORTS_PLATFORM_CONFIGSAVE,1,[Save additional config file options.])
-         AC_DEFINE(SUPPORTS_PLATFORM_PALETTEUPDATE,1,[Update the Palette if it changed.])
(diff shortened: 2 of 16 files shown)


hhromic:
For completeness, here is the output of the testing script: [test-results.log](https://github.com/RetroPie/RetroPie-Setup/files/2311349/test-results.log)

hhromic:
Oh I forgot to clarify something. When I say I tested all supported platforms, I meant that I tested the conditionals for applying the patches (by explicitly setting `__platform`), not that I used real devices. This is because I'm only testing patching after the `sources` action, not building.

scriptmodules/helpers.sh
```diff
@@ -1189,13 +1187,13 @@ function addPort() {
...
 
-    if [[ -t 0 ]]; then
+    if [[ -z "$script" ]]; then
         cat >"$file" << _EOF_
 #!/bin/bash
 "$rootdir/supplementary/runcommand/runcommand.sh" 0 _PORT_ "$port" "$game"
 _EOF_
     else
-        cat >"$file"
+        cat >"$file" <<< "$script"
```
joolswills: Perhaps just using cp ?

joolswills: ah - sorry I misunderstood - you are loading the custom scripts into a var and using that as a parameter.  ok.

joolswills: not 100% sure I like this workaround - it might be better for the scriptmodule itself to just copy the custom script in place. However I will keep for now, but may change later.

joolswills:
Thanks. Looks good to me.

scriptmodules/helpers.sh
hhromic: Yes, indeed yesterday I also kept thinking about a very similar idea.
Like patches, these custom launch scripts should be also kept in the `$md_data` directory of the module needing it and the filename passed to `addPort` instead of the content, to be copied accordingly. I agree is better like that.
This would be consistent with how `applyPatch` works too. I will give it a thought and propose a PR. Thanks for the feedback !

hhromic:
No problem. I will keep the branch for some days in case it breaks for someone so we can fix it quickly.

scriptmodules/helpers.sh
joolswills: In the case of the one example we have, it requires variables from the script though, so a separate file would not work like that unless we replaced tokens in it. I was thinking more that the module itself just creates the file - keep it outside of addPort. Eg we have a parameter for addPort like "noscript" which makes it not create a launch script so we can then just do it from the module. But the current way is ok I guess.

joolswills: or we just overwrite the script after calling addPort with the custom one

PR #2474, opened 2018-08-23
addPort - remove script parameter and handle in scriptmodule

joolswills:
@hhromic I was thinking to just to it like this - simplifies it I think.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1147,11 +1147,8 @@ function delSystem() {
 ## @param name display name for the launch script
 ## @param cmd commandline to launch
 ## @param game rom/game parameter (optional)
-## @param script launch script to use instead of the default (optional - requires game to be set)
 ## @brief Adds a port to the emulationstation ports menu.
 ## @details Adds an emulators.cfg entry as with addSystem but also creates a launch script in `$datadir/ports/$name.sh`.
-##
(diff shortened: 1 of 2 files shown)


hhromic:
Yep, definitively this looks cleaner/simpler.
Perhaps I would only add a small comment in the scriptmodule to explain why this "override" is done, future devs might appreciate. Something like this:
```sh
addPort "$md_id" "cavestory" "Cave Story" "$md_inst/nxengine_libretro.so"
# custom launch script: if the main executable is not found, warn the user
local file="$romdir/ports/Cave Story.sh"
```

ps: ah sorry forgot to remove the stdin snippet in the help before :)

hhromic:
looks good to me

actraiser30:
Hi, where is the addPort_custom_script in the retropie setup menu?

joolswills:
Please use the forum for help.

actraiser30:
i cannot find any info in the forum

joolswills:
You can ask - this is not the right place for help.

actraiser30:
anyway,it will be great to add such a feature to retropie so people can add a custom port by them self.

joolswills:
you can do this already by writing a scriptmodule.

actraiser30:
Anyway thanks it will be great to include such a script in the retropie setup menu for manual port installing
sorry for mess with your post here.

PR #2485, opened 2018-09-16
Fix lr-parallel-n64 build from source on tinker platform

EriHoss:
Hi, I wanted to test the lr-parallel-n64 core on my TinkerBoard (with Armbian 5.59, kern 4.14) but was unable to get it to build.  I was able to build it with these changes.

When building with platform="kms" the makefile exits immediately. 

I'm guessing eventually the lr-parallel-n64 Makefile will understand "kms" ?

The arguments inside the params array also need their quoting fixed, so that the variables in the makefile get set correctly when the array is expanded.

### DIFF
--- a/scriptmodules/libretrocores/lr-parallel-n64.sh
+++ b/scriptmodules/libretrocores/lr-parallel-n64.sh
@@ -35,10 +35,9 @@ function build_lr-parallel-n64() {
     if isPlatform "rpi" || isPlatform "odroid-c1"; then
         params+=(platform="$__platform")
     elif isPlatform "tinker"; then
-        params+=(platform="kms")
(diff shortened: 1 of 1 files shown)


joolswills:
The tinker support was added originally with the ASUS linaro linux image. I intend to look into Armbian and perhaps make that the main supported OS, but for now I will leave this as is until I have had a chance to do some testing.

joolswills:
The quoting is wrong though as you point out - was a user contributed patch in this case. Not sure how they managed to contribute it in a broken state (Would have assumed it was tested). Will sort that also.

joolswills:
Actually due to the broken quoting, I think I will accept this for now, and test later. Thanks for the PR.

joolswills:
Actually due to the broken quoting, I think I will accept this for now, and test later. Thanks for the PR.

PR #2490, opened 2018-09-24
Add lr-x1 - Sharp X1 libretro core

SuperCatFooz:
Yet another obscure Japanese home computer!  This one requires two BIOS files in `/home/pi/RetroPie/BIOS/xmil`, named IPLROM.X1 [CRC32 7B28D9DE] and IPLROM.X1T [CRC32 2E8B767C].  Tested and works full-speed on a Raspberry Pi 3B without overclocking.

Since the makefile is within the `libretro` folder, rather than at the root of the source files, I added `/libretro/` wherever needed.  Little bit hacky, but it works well enough and doesn't seem to cause any issues [I'm not sure if this is the preferred way to handle this however].

This commit also modifies `platforms.cfg` to add a name and extensions for the X1.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -227,6 +227,9 @@ wonderswan_fullname="Wonderswan"
 wonderswancolor_exts=".7z .wsc .zip"
 wonderswancolor_fullname="Wonderswan Color"
 
+x1_exts=".dx1 .zip .2d .2hd .tfd .d88 .88d .hdm .xdf .dup .cmd"
+x1_fullname="Sharp X1"
+
 x68000_exts=".dim .hdf .2hd .d88 .m3u"
 x68000_fullname="X68000"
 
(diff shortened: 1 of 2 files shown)


tomaz82:
A word of advise, if you commit something that's supposed to be just 3-4 added lines, and it says every single line in the file is changed, you probably did something wrong, my guess in this case would be that you changed the line endings.

SuperCatFooz:
Ouch! Nice catch.

SuperCatFooz:
From what I've tested so far, there doesn't _seem_ to be anything awry about the core despite being not officially part of RetroArch.  (The only exception being a known bug where opening a disk using the built-in emulator GUI causes a softlock, but I don't think that would be a huge issue since games are loaded via runcommand anyway).  

Despite that, I've still placed the core in experimental for now since it's unofficial, and given that the last commit to the emulator was in 2017, I don't see it moving out of experimental any time soon.

joolswills:
thanks.

PR #2495, opened 2018-09-28
Dolphin libretro core is missing sources

cmitu:
Small fix to actually pull the Dolphin core sources before building them.


joolswills:
Not sure how I managed to include this module in the state it's in - oops. there is another fix needed as the function is wrong. Will sort.

cmitu:
OK, then I'll close this one.

joolswills:
Should be ok now. Thanks for the heads up / fix. The sources function should have been the depends_ function so I renamed it, and added a sources function with the pull.

PR #2496, opened 2018-09-28
Fix line endings on lr-x1.sh

SuperCatFooz:
I didn't notice that the line endings for lr-x1.sh were CRLF, which causes a few syntax errors in RetroPie-Setup and hides the script from the experimental menu.  One run of `dos2unix` later, the file works as intended.


joolswills:
ouch. Thanks. I should have checked.

joolswills:
How did you test it btw on your RPI when it had the wrong line endings ? Or did you re-edit the file after testing ?

SuperCatFooz:
I re-edited the file right before submitting; just making sure everything in the file was consistent.  I guess Notepad++ auto-converts LFs to CRLFs.

PR #2497, opened 2018-10-01
Add Skyscraper as an alternate scraper

cmitu:
This PR is adding a re-worked version of the #2468 PR by @muldjord.

The scriptmodule is derived from the `scraper.sh` module for the GUI part, for users to get a similar interface for scraping outside Emulationstation.
Skyscraper GUI

I'd like to get some review for the choices I've added by default when configuring Skyscraper on the 1st install and on what should be present/not present in the Gui (especially from @muldjord).

* When installing through the scriptmodule, the existing configuration folder is preserved; it's also linked to `/opt/retropie/configs/all/skyscraper` as for any other module. The `config.ini` is generated only if one is not already present, to preserve any user customizations.
* Skyscraper has the concept of caching the resources downloaded online, then using the `artwork.xml` file to compose the image in the gamelist. Therefore, scraping will save some resources (configured in `~/.skyscraper/dbs`) on disk and the Gui has an option to purge these cached resources and the current DB size is shown after each scrape.
* I added an option to not cache any resources on-disk (the `--nolocaldb` option) and user can choose to scrape like this.
* Gamelist generation
   - this is configurable by user, default is the normal ES location, but can be toggled through the GUI to the ROM folder.
   - depending on the choice of gamelist location, the media location is either `~/.emulationstation/downloaded_medi` or (relative) in `$ROMFOLDER/media`. 
* The Scraping operation - I've monitored the Skyscraper topic in the RetroPie forums and I think @muldjord recommends scraping online and then a scrap from the local cached resources. I've implemented it like this (2 scrape invocations per system scraped), **except** when the user disabled caching. @muldjor - I would like your feedback on this.

The GUI configured the following parameters
* **-g/-o** : location of the gamelist and generated artwork (*Use rom folder for ...*)
* **--forcefilename** : either use ROM name retured by scraping or force file name as game name.
* **--nolocaldb**: cache scraped resources or not
* **--nocovers/--nomarquees/--videos**: by default marquees, screenshots and covers are downloaded (set in the .ini file), but videos are not. These options enable/disable downloading of those files.
* **-s <scraping_source>**: allow the user to select the scraping source, from the ones supported by Skyscraper. All of them are present, **except** localdb. I don't know if it makes sens to add it here or not.
Scrape source selection

Feedback welcome. Once this is added, I think we'll need to modify the Wiki to explain the usage and options available through the GUI.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -1,0 +1,687 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="skyscraper"
+rp_module_desc="Scraper for EmulationStation by Lars Muldjord"
+rp_module_licence="GPLv3.0 https://raw.githubusercontent.com/muldjord/skyscraper/master/LICENSE"
+rp_module_section="opt"
+
+function depends_skyscraper() {
+    getDepends qt5-default p7zip-full
+}
+
+function sources_skyscraper() {
+    gitPullOrClone "$md_build" "https://github.com/muldjord/skyscraper" "$(_latest_ver_skyscraper)"
+}
+
+function build_skyscraper() {
+    qmake
+    make
+    md_ret_require="$md_build/Skyscraper"
+}
+
+function install_skyscraper() {
+    md_ret_files=(
+        'Skyscraper'
+        'LICENSE'
+        'README.md'
+    )
+}
+
+# Purge all Skyscraper caches
+function _purge_skyscraper() {
+    local platform
+
+    while read platform; do
+        # Find any folders underneath the Local DB cache and clear them
+        _clear_platform_skyscraper "$platform"
+    done < <(find "$configdir/all/skyscraper/dbs" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
+}
+
+function _clear_platform_skyscraper() {
+    local platform="$1"
+
+    [[ ! -d "$configdir/all/skyscraper/dbs/$platform" ]] && return
+
+    # Remove any folder underneath the platform, try to keep any customized files
+    pushd "$configdir/all/skyscraper/dbs/$platform"
+    find . -maxdepth 1 -mindepth 1 -type d -print0 -exec rm -fr {} \;
+    rm -f db.xml
+    popd
+}
+
+# Purge by cached platform
+function _purge_platform_skyscraper() {
+    local options=()
+    local system
+
+    while read system; do
+        # If there is no 'db.xml' file underneath the folder, skip it, it means folder is empty
+        [[ ! -f "$configdir/all/skyscraper/dbs/$system/db.xml" ]] && continue
+
+        # Get the size on disk of the system and show it in the select list
+        local size=$(du -sh  "$configdir/all/skyscraper/dbs/$system" | cut -f1)
+        options+=("$system" "$size" OFF)
+    done < <(find "$configdir/all/skyscraper/dbs" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
+
+    # If not folders are found, show an info message instead of the selection list
+    if [[ ${#options[@]} -eq 0 ]] ; then
+        printMsgs "dialog" "Nothing to delete ! No cached systems found in \n$configdir/all/skyscraper/dbs "
+        return
+    fi
+
+    local cmd=(dialog --backtitle "$__backtitle" --radiolist "Select platform to purge" 20 60 12)
+    local platform=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
+
+    # Exit if no platform chosen
+    [[ -z "$platform" ]] && return
+
+    _clear_platform_skyscraper "$platform"
+}
+
+function _get_ver_skyscraper() {
+    if [[ -f "$md_inst/Skyscraper" ]]; then
+        echo $("$md_inst/Skyscraper" -h | grep 'Running Skyscraper'  | cut -d' '  -f 3 2>/dev/null)
+    fi
+}
+
+function _latest_ver_skyscraper() {
+    wget -qO- https://api.github.com/repos/muldjord/skyscraper/releases/latest | grep -m 1 tag_name | cut -d\" -f4
+}
+
+# List any non-empty system found in the ROM parent folder
+function _list_systems_skyscraper() {
+    find "$romdir"  -mindepth 2 -maxdepth 2 -type f -not -name '+*' | sed 's%/[^/]*$%%' | sort -u
+}
+
+function remove_skyscraper() {
+    # On removal of package, purge the Local DB
+    _purge_skyscraper
+}
+
+function configure_skyscraper() {
+    if [[ "$md_mode" == "remove" ]]; then
+        return
+    fi
+
+    # Check if this a first time install
+    local local_config
+    local_config=$(readlink -qn "$home/.skyscraper")
+
+    # Handle the cases where the user has an existing Skyscraper installation.
+    if [[ -d "$home/.skyscraper" && "$local_config" != "$configdir/all/skyscraper" ]]; then
+        # We have an existing Skyscraper installation, but not handled by this scriptmodule.
+        # Since the $HOME/.skyscraper folder will be moved, make sure the Local DB and Import folders are moved separately
+        local f_size
+        f_size=$(du --total -sm "$home/.skyscraper/dbs" "$home/.skyscraper/import" 2>/dev/null | grep total | cut -f 1 )
+        printMsgs "console" "INFO: Moving the Local DB and Import folders to new configuration folder (total: $f_size Mb)"
+
+        local folder
+        for folder in dbs import; do
+            mv "$home/.skyscraper/$folder" "$home/.skyscraper-$folder" && \
+                printMsgs "console" "INFO: Moved "$home/.skyscraper/$folder" to "$home/.skyscraper-$folder""
+        done
+
+        # When having an existing installation, changes are the gamelist is in the ROMs folder
+        # Create a GUI config file with this setting pre-set.
+        iniConfig " = " '"' "$configdir/all/skyscraper.cfg"
+        iniSet "use_rom_folder" 1
+    fi
+
+    moveConfigDir "$home/.skyscraper" "$configdir/all/skyscraper"
+
+    # When migrating an existing installation, make sure we move the Local DB and Import folder back the new conf folder
+    for folder in dbs import; do
+        if [[ -d "$home/.skyscraper-$folder" ]]; then
+            printMsgs "console" "INFO: Moving "$home/.skyscraper-$folder" back to configuration folder"
+            mv  "$home/.skyscraper-$folder" "$configdir/all/skyscraper/$folder"
+        fi
+    done
+
+    _init_config_skyscraper
+    chown -R $user:$user "$configdir/all/skyscraper"
+}
+
+function _init_config_skyscraper() {
+    local md_conf_dir="$configdir/all/skyscraper"
+
+    cat > "$md_conf_dir/config.ini.rp" << _EOF_
+[main]
+frontend="emulationstation"
+dbFolder="$md_conf_dir/dbs"
+importFolder="$md_conf_dir/import"
+inputFolder="$romdir"
+unattend="true"
+unattendSkip="true"
+noHints="false"
+threads="2"
+maxFails="30"
+lang="en"
+region="wor"
+videos="false"
+marquees="true"
+wheels="false"
+covers="true"
+screenshots="true"
+artworkXml="$md_conf_dir/artwork.xml"
+
+_EOF_
+
+    # Make sure the `artwork.xml` and other conf file(s) are present, but don't overwrite them on upgrades.
+    # For now we just have one file, but more might come up in the future.
+    local f_conf
+    for f_conf in artwork.xml; do
+        if [[ -f "$md_conf_dir/$f_conf" ]]; then
+            cp "$md_build/$f_conf" "$md_conf_dir/$f_conf.default"
+        else
+            cp "$md_build/$f_conf" "$md_conf_dir"
+        fi
+    done
+
+    # If we don't have a previous config.ini file, copy the dist one
+    [[ ! -f "$md_conf_dir/config.ini" ]] && cp "$md_conf_dir/config.ini.rp" "$md_conf_dir/config.ini"
+
+    # Try to find the rest of the necesary files from the qmake build file
+    # They should be listed in the `unix:examples.file` configuration line
+    if [[ $(grep unix:examples.files "$md_build/skyscraper.pro" 2>/dev/null | cut -d= -f2-) ]]; then
+        local files=$(grep unix:examples.files "$md_build/skyscraper.pro" | cut -d= -f2-)
+        local file
+
+        for file in $files; do
+            # Copy the files to the configuration folder. Skip config.ini and artwork.xml, they're already handled
+            if [[ $file != "artwork.xml" && $file != "config.ini" ]]; then
+                cp -f "$md_build/$file" "$md_conf_dir"
+            fi
+        done
+    else
+        # Fallback to the known resource files list
+        cp -f "$md_build/artwork.xml.example"* "$md_conf_dir"
+
+        # Copy resources
+        local resource_file
+        for resource_file in ARTWORK.md README.md mameMap.csv tgdb_developers.json tgdb_publishers.json hints.txt; do
+            cp -f "$md_build/$resource_file" "$md_conf_dir"
+        done
+    fi
+
+    # Copy the rest of the folders
+    cp -rf "$md_build/resources" "$md_conf_dir"
+
+    # Create the import folders
+    local folder
+    for folder in boxart snaps covers marquees screenshots textual videos wheels; do
+        mkUserDir "$md_conf_dir/import/$folder"
+    done
+
+    # Create the LocalDB cache folder
+    mkUserDir "$md_conf_dir/dbs"
+}
+
+# Scrape one system, passed as parameter
+function _scrape_skyscraper() {
+    local system="$1"
+
+    [[ -z "$system" ]] && return
+
+    iniConfig " = " '"' "$configdir/all/skyscraper.cfg"
+    eval $(_load_config_skyscraper)
+
+    local gamelist
+    local media_dir
+    local relative
+
+    if [[ "$use_rom_folder" -eq 1 ]]; then
+        gamelist="$romdir/$system"
+        media_dir="$romdir/$system/media"
+        # If we're saving to the ROM folder, then use relative paths in the gamelist
+        relative="--relative"
+    else
+        gamelist="$home/.emulationstation/gamelists/$system"
+        media_dir="$home/.emulationstation/downloaded_media/$system"
+    fi
+
+    local -a params=("--unattend" "--skipped")
+
+    params+=(-o "$media_dir")
+    params+=(-g "$gamelist" "$relative")
+    params+=(-p "$system")
+    params+=(-s "$scrape_source")
+
+    if [[ "$download_videos" -eq 1 ]]; then
+        params+=(--videos)
+    fi
+
+    if [[ "$cache_marquees" -eq 0 ]]; then
+        params+=(--nomarquees)
+    fi
+
+    if [[ "$cache_covers" -eq 0 ]]; then
+        params+=(--nocovers)
+    fi
+
+    if [[ "$cache_screenshots" -eq 0 ]]; then
+        params+=(--noscreenshots)
+    fi
+
+    if [[ "$cache_wheels" -eq 0 ]]; then
+        params+=(--nowheels)
+    fi
+
+    if [[ "$rom_name" -eq 1 ]]; then
+        params+=(--forcefilename)
+    fi
+
+    if [[ "$remove_brackets" -eq 1 ]]; then
+        params+=(--nobrackets)
+    fi
+
+    if [[ "$force_refresh" -eq 1 ]]; then
+        params+=(--refresh)
+    fi
+
+    # trap ctrl+c and return if pressed (rather than exiting retropie-setup etc)
+    trap 'trap 2; return 1' INT
+        sudo -u "$user" "$md_inst/Skyscraper" "${params[@]}"
+        echo -e "\nCOMMAND LINE USED:\n $md_inst/Skyscraper" "${params[@]}"
+        sleep 2
+    trap 2
+}
+
+# Scrape all systems
+function _scrape_all_skyscraper() {
+    local system
+
+    while read system; do
+        system=${system/$romdir\//}
+        _scrape_skyscraper "$system" "$@" || return 1
+    done < <(_list_systems_skyscraper)
+}
+
+# Scrape a list of systems, chosen by the user
+function _scrape_chosen_skyscraper() {
+    local options=()
+    local system
+    local i=1
+
+    while read system; do
+        system=${system/$romdir\//}
+        options+=($i "$system" OFF)
+        ((i++))
+    done < <(_list_systems_skyscraper)
+
+    if [[ ${#options[@]} -eq 0 ]] ; then
+        printMsgs "dialog" "No populated ROM folders were found in $romdir."
+        return
+    fi
+
+    local choices
+    local cmd=(dialog --backtitle "$__backtitle" --checklist "Select ROM Folders" 22 76 16)
+
+    choices=($("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty))
+
+    # Exit if nothing was chosen or Cancel was used
+    [[ ${#choices[@]} -eq 0 || $? -eq 1 ]] && return 1
+
+    local choice
+
+    for choice in "${choices[@]}"; do
+        choice=${options[choices*3-2]}
+        _scrape_skyscraper "$choice" "$@"
+    done
+}
+
+function _load_config_skyscraper() {
+    echo "$(loadModuleConfig \
+        'rom_name=0' \
+        'use_rom_folder=0' \
+        'download_videos=0' \
+        'cache_marquees=1' \
+        'cache_covers=1' \
+        'cache_wheels=1' \
+        'cache_screenshots=1' \
+        'scrape_source=screenscraper' \
+        'remove_brackets=0' \
+        'force_refresh=0'
+    )"
+}
+
+# Try to guess the most appropriate editor. On Debian derivatives, we have `sensible-editor` for that.
+function _open_editor_skyscraper() {
+    local editor
+
+    if [[ -n $(command -v sensible-editor) ]]; then
+        sensible-editor "$1"
+    else
+        editor="${EDITOR:-nano}"
+        $editor "$1"
+    fi
+}
+
+function _gui_advanced_skyscraper() {
+    local db_size
+    local -A help_strings_adv
+
+    iniConfig " = " '"' "$configdir/all/skyscraper.cfg"
+    eval $(_load_config_skyscraper)
+
+    help_strings_adv=(
+        [1]="Toggle whether screenshots are cached locally when scraping.\n\nSkyscraper option: \Zb--noscreenshots\Zn"
+        [2]="Toggle whether covers are cached locally when scraping.\n\nSkyscraper option: \Zb--nocovers\Zn"
+        [3]="Toggle whether wheels are cached locally when scraping.\n\nSkyscraper option: \Zb--nowheels\Zn"
+        [4]="Toggle whether marquees are cached locally when scraping.\n\nSkyscraper option: \Zb--nomarquees\Zn"
+        [E]="Opens the configuration file \Zbconfig.ini\Zn in an editor."
+        [F]="Opens the artwork definition file \Zbartwork.xml\Zn in an editor."
+        [P]="Purge \ZbALL\Zn all cached resources from the LocalDB."
+        [S]="Purge all cached resources from the LocalDB for a chosen system."
+    )
+
+    while true; do
+        db_size=$(du -sh "$configdir/all/skyscraper/dbs" 2>/dev/null | cut -f 1 || echo 0m)
+        [[ -z "$db_size" ]] && db_size="0Mb"
+
+        local cmd=(dialog --backtitle "$__backtitle" --help-button --colors --no-collapse --default-item "$default" --ok-label "Ok" --cancel-label "Back" --title "Advanced options and commands" --menu "\n  Local cache is currently at $db_size\n\n" 21 65 12)
+
+        local options=("-" "LOCAL Cache")
+
+        if [[ "$cache_screenshots" -eq 1 ]]; then
+            options+=(1 "Cache screenshots (Enabled)")
+        else
+            options+=(1 "Cache screenshots (Disabled)")
+        fi
+
+        if [[ "$cache_covers" -eq 1 ]]; then
+            options+=(2 "Cache covers (Enabled)")
+        else
+            options+=(2 "Cache covers (Disabled)")
+        fi
+
+        if [[ "$cache_wheels" -eq 1 ]]; then
+            options+=(3 "Cache wheels (Enabled)")
+        else
+            options+=(3 "Cache wheels (Disabled)")
+        fi
+
+        if [[ "$cache_marquees" -eq 1 ]]; then
+            options+=(4 "Cache marquees (Enabled)")
+        else
+            options+=(4 "Cache marquees (Disabled)")
+        fi
+
+        options+=("-" "PURGE commands")
+        options+=(S "Purge system")
+        options+=(P "Purge all (!)")
+
+        options+=("-" "EXPERT - edit configurations")
+        options+=(E "Edit 'config.ini'")
+        options+=(F "Edit 'artwork.xml'")
+
+        local choice=$("${cmd[@]}" "${options[@]}" 2>&1 > /dev/tty)
+
+        if [[ -n "$choice" ]]; then
+            local default=$choice
+
+            case "$choice" in
+
+                1)
+                    cache_screenshots="$((cache_screenshots ^ 1))"
+                    iniSet "cache_screenshots" "$cache_screenshots"
+                    ;;
+
+                2)
+                    cache_covers="$((cache_covers ^ 1))"
+                    iniSet "cache_covers" "$cache_covers"
+                    ;;
+
+                3)
+                    cache_wheels="$((cache_wheels ^ 1))"
+                    iniSet "cache_wheels" "$cache_wheels"
+                    ;;
+
+                4)
+                    cache_marquees="$((cache_marquees ^ 1))"
+                    iniSet "cache_marquees" "$cache_marquees"
+                    ;;
+
+                S)
+                    _purge_platform_skyscraper
+                    ;;
+
+                P)
+                    local confirm=$(dialog --clear --defaultno --colors --yesno  "\Z1\ZbAre you sure ?\Zn\nThis will \Zb\ZuERASE\Zn all locally cached scraped resources" 8 60 2>&1 >/dev/tty)
+                    if [[ ! -n "$confirm" ]]; then
+                        _purge_skyscraper
+                    fi
+                    ;;
+
+                E)
+                    _open_editor_skyscraper "$configdir/all/skyscraper/config.ini"
+                    ;;
+
+                F)
+                    _open_editor_skyscraper "$configdir/all/skyscraper/artwork.xml"
+                    ;;
+
+                HELP*)
+                    # Retain choice
+                    default="${choice/HELP /}"
+                    if [[ ! -z "${help_strings_adv[${default}]}" ]]; then
+                    dialog --colors --no-collapse --ok-label "Close" --msgbox "${help_strings_adv[${default}]}" 22 65 >&1
+                    fi
+            esac
+        else
+            break
+        fi
+    done
+}
+
+function gui_skyscraper() {
+    if pgrep "emulationstatio" >/dev/null; then
+        printMsgs "dialog" "This scraper must not be run while Emulation Station is running or the scraped data will be overwritten. \n\nPlease quit from Emulation Station, and run RetroPie-Setup from the terminal"
+        return
+    fi
+
+    iniConfig " = " '"' "$configdir/all/skyscraper.cfg"
+    eval $(_load_config_skyscraper)
+    chown "$user":"$user" "$configdir/all/skyscraper.cfg"
+
+    local -a s_source
+    local -a s_source_names
+    local -A help_strings
+
+    s_source=([1]=screenscraper [2]=arcadedb [3]=thegamesdb [4]=igdb [5]=mobygames [6]=openretro [7]=worldofspectrum)
+    s_source+=([10]=localdb [11]=import)
+
+    s_source_names=([1]=ScreenScraper [2]=Arcade Database [3]=TheGamesDB [4]=IGDB [5]="Moby Games" [6]="OpenRetro" [7]="World of Spectrum")
+    s_source_names+=([10]="LocalDB Cache" [11]="Import Folder")
+
+    local default
+    local ver
+
+    # Help strings for this GUI
+    help_strings=(
+        [1]="Scrape all systems found in \Zb$romdir\Zn."
+        [2]="Choose which systems to scrape from the ones found in \Zb$romdir\Zn."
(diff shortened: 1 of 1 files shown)


muldjord:
I'd be happy to provide some feedback / suggestions when I get back home
from work today. :)

muldjord:
First of all, this looks really good! Great work so far. Here are my thoughts:
* The design of the GUI should probably be designed around retaining the cache as much as possible. The cache is such a central part of Skyscraper in order to provide efficient rescrapes when adding new roms, and to save bandwidth for the sources (this is more important than people realize). I would hate for the users to disable it because they don't understand it. It would be better to put any "Purge" or "Cache disable" options under an "Advanced" menu.

Initially I feel like the following options would make sense:

* Scrape from [source]
  * Set user key or id:passwd [key / id:passwd] ('-u key' or '-u id:passwd')
* Gamelist / ROM names [SOURCE NAME / filename] (--forcefilename)
* Bracket info [ENABLE / disable] (--nobrackets)
* Force cache refresh [enable / DISABLE] (--refresh)
* Scrape videos [enable / DISABLE] (--videos)
* Artwork
  * Configure artwork (Start "nano -w ~/.skyscraper/artwork.xml" or something like that)
  * Reset 'artwork.xml' (copy in the default artwork.xml in case user has broken it)
* Advanced
  * Edit 'config.ini' (Start "nano -w ~/.skyscraper/config.ini" or something like that)
  * Configure non-default paths
    * Set input rom folder [folder] (-i or configure in config.ini)
    * Set media export folder [folder] (-o or configure in config.ini)
    * Set gamelist export folder [folder] (-g or configure in config.ini)
  * Local cache
    * Disable screenshot caching [true / FALSE] (--noscreenshots)
    * Disable cover caching [true / FALSE] (--nocovers)
    * Disable wheel caching [true / FALSE] (--nowheels)
    * Disable marquee caching [true / FALSE] (--nomarquees)
    * Purge type [cover/screenshots/wheel/marquee/video]
    * Purge module [all/thegamesdb, etc...]
    * Purge selected type and module ("--purgedb t:type,m:module", if module is 'all' it should just use "--purgedb t:type")
    * Purge chosen platform, x MB (delete "~/.skyscraper/dbs/[platform]")
    * Purge all resources, x MB (delete "~/.skyscraper/dbs", should probably give a warning before doing this)
* Scrape single system
* Scrape all systems

Anything in CAPITALS (such as DISABLE) should probably be the default.

I realize that what I've put in is pretty elaborate. The purging could probably be fine with just the "Purge chosen platform" to begin with. I added the rest because it is an option in Skyscraper.

I will gladly help out with any help texts for each of these menu options.

I'm unsure how to implement the key / id:passwd part of the sources / modules. Some of the requires a key or id:passwd to be set to work at all. And some of them will benefit greatly from it. For instance "screenscraper" is restricted to 1 thread if no id:passwd is provided. But users can gain more threads by helping out with the database or buying more.

Let me know what you think. Of course everything I've put in is just suggestions based on how I feel like Skyscraper should be used. Thank you for giving this a shot!

cmitu:
> > ```
> > Scrape from ONLINE: <online_scraper>
> > ...
> > Scrape from CACHE: Import
> > Scraper from CACHE: LocalCache
> > ```
> > What do you think ?
> 
> Yes, but it should probably be
> 
> * Online source: [source]
> * Local source: [localdb/import]
>   That's what I use in the Skyscraper documentation aswell.

 OK, I'll try to find a way make it exclusive. Right now the source is just one selection and the user scrapes by using one of the
  - scrape all systems
  - scrape selected systems.
  
 
> 
> > ...but what I think you mean is that this list of files could change, so that the installation performed by the script will not copy all the necesary files.
> > Am I correct ?
> 
> It is imperative that Skyscraper is installed with the "sudo make install" command. The files are copied to the /usr/local folders and I check those upon running Skyscraper, to make sure the example files and other files are up-to-date inside ~/.skyscraper. If it's done just by copying the files into ~/.skyscraper, my entire file checking routine breaks.

 The script does the installation and the update and makes sure these files are copied there - it's similar to a `make install`. I noticed this check (from `/usr/local`) - but how would this break the file checking routine ? Remember the scraper is run as an user, there's no write access outside the user's homepage and the 
  
> 
> > I'm not sure about this one - right now online scraping is done with --refresh. You think we should remove this options from the scraping operation and have a separate refresh option ?
> 
> I would prefer not to have '--refresh' added by default. The idea is that the user can rescrape their entire library whenever they want, and it will only contact the online sources for the new roms they have added. I feel a responsibility to make sure Skyscraper doesn't strain the sources. And I feel this GUI needs to reflect that. :) If the user don't use the advanced features anyways, they will get the same data from the cache and from the online source anyway. And with a helpful text on the entry, I think it will be useful as a bool.

 You're right, I'll modify this accordingly.
 
> 
> Concerning the purging, I don't know man. It's up to you I guess. I kind of agree that the advanced usage can be left out for a first version. :)
> 
> I'm sorry if I at any point seem a bit on the fence about any of this. It bugs me a bit to "boil it down" to the bare minimum like this, as Skyscraper is "just another scraper" if the user doesn't take advantage of the artwork compositor, the config.ini possibilities, the user/id for the modules and of course the cache options. But I understand the need to, I just don't like it. :D

 Skyscraper is not 'just another scraper' and users will still benefit from it, event without having access to advanced features from the GUI. You are already an advanced user and having all these features not available seems to you that we're hiding or obscuring these things for the user. Think of it this way - we can add more features later in the GUI as more users will get in touch with it and they discover the possibilities.

muldjord:
> The script does the installation and the update and makes sure these files are copied there

Oh, so it copies it to the same folders in /usr/local? Then it should work as expected, and then yes, the only problem would be if I change the installation script.

> Think of it this way - we can add more features later in the GUI as more users will get in touch with it and they discover the possibilities.

Yes, I agree.

cmitu:
> > The script does the installation and the update and makes sure these files are copied there
> 
> Oh, so it copies it to the same folders in /usr/local? Then it should work as expected (I thought it just copied the files directly to "~/.skyscraper" which wouldn't work well). Then yes, the only problem would be if I change the installation script.

 Not exactly. What I mean is that your existing routine is:
  * **make install** copies the files in `/usr/local/etc`
  * when running, `Skyscraper` checks the files in `/usr/local/etc` and then copies them into `~/.skyscraper`.

but the scriptmodule will copy those exact files into '~/.skyscraper` directly, so the result is the same - the most recent files from your distribution will end up in the user's configuration folder. The only exceptions to this rule are `config.ini` and `artwork.xml` (those are copied/generated only on the 1st installation).

muldjord:
It might be ok. I can't remember if there are issues. I spend a lot of time on those checking routines, but it's been a long since I did them, so I can't remember the details. Only way to test it properly would be to just install it from scratch using this script (without having any of the files in /usr/local) and see what happens.

cmitu:
> It might be ok. I can't remember if there are issues. I spend a lot of time on those checking routines, but it's been a long since I did them, so I can't remember the details. Only way to test it properly would be to just install it from scratch using this script (without having any of the files in /usr/local) and see what happens.

This is how I tested first, before starting any development of the module, and this is how I now test the GUI actions and their options. In both my test installations (Raspbian - PI3 and Ubuntu - PC), I didn't install anything in `/usr/local`.

cmitu:
Based on the feedback from 2 days ago, I've made some changes to the GUI. I've only pushed the changes in my fork repository, but not yet in this PR.

## Main screen

 Main GUI
 I've added the options you suggested:
 * _Force cache refresh_ for `--refresh`
 * _Bracket info_ for `--nobrackets`
 * changed a bit the Scraper selection screen and merged the LocalDB and the Import module
 Scrape Selection

* added an **Advanced** sub-dialog

## The Advanced section
Advanced screen

The advanced screen is split in 3 sections. I moved here the stats for the Local Cache (from the main GUI)
1. The **Local Cache** options you suggested to be moved here - caching of Wheels, Covers, Screenshots, Marquees
2. The _Purging_ options - either by platform or for all resources. When selecting the purging of _All_ resources, a warning will be shown.

Erase all warning

3. The - really - advanced options to edit the configuration files. 

Let me know what you think. 


Things that I know can be improved in this iteration (but I haven't got to it yet):
- (**Purging**) when purging an individual system, show how much space should be freed in the system selection list.
- (**UI**) Replace the _Enabled_/_Disabled_ states with something like **Enabled**/_Disabled_ vs _Enabled_/**Disabled** so it's clear it's a 2 state option.
- (**Install**) instead of putting the list of necessary files in the script, parse the `.pro` file (used by `qmake`) and extract the list of files from there. Important files like `config.ini` and `artwork.xml` will be handled individually (i.e. making sure we're not overwriting them on upgrade).
- (**UI**) add a Help button to document the option.

cmitu:
@joolswills I'll buzz you for a review when I'm thinking it's good enough.

muldjord:
Great work, looks good. One comment: For the "Select scraping source" it should probably say "Local: LocalDb cache" and "Local: Import folder". So it's "Online" and "Local".

cmitu:
OK, I got the previous points implemented - except the UX part, because it doesn't look so good.

* I added a size indicator when purging a system. The total size of the local DB is shown when entering the Advanced sub-menu
* Installation now looks in the `skyscraper.pro` and gets all the files declared as **unix:examples.files** in order to copy them in the config folder.
* I added a `<Help>` button that explains the options, with the corresponding CLI parameter that's associated with that option (where it applies).

@muldjord you can test the GUI and see the help messages by copying the `skyscraper.sh` file into `$HOME/RetroPie-Setup/scriptmodules/supplementary`, then running

``` sh
cd $HOME/RetroPie-Setup/
sudo ./retropie_packages skyscraper gui
```

This will just show the GUI and you'll be able to change options, see how the Help looks like. When trying to scrape, it will probably give you an error unless you install the package first.

The package is in the **optional** section (we can move it to experimental at first); if installed, the GUI can be started from the main `Configuration/Tools` menu in RetroPie-Setup.

Let me know if you think we could add other options from the ones available (`--unpack` ?, `--lang` ?) in Skyscraper.

muldjord:
Currently testing the ui:
  * When choosing "Cancel" in "Scrape chosen systems" it says "ROMS have been scraped" instead of just going back to the main Skyscraper menu.
  * In "Scrape from..." menu "Open Retro" should be "OpenRetro".
  * For "Remove bracket info", when it's "Disabled", the B in bracket suddenly becomes uppercase.

For help texts:
  * For "Scrape from..." help text it says "Open Retro", which should be "OpenRetro".
  * For "Scrape from..." help text it would probably make sense to add a small help text saying something like "Some sources need a <key> or <user:password> to work, which can be set per source in config.ini" or similar. Otherwise I foresee a lot of support requests complaining about certain sources not working.
  * For "Purge system" is says "Skyscraper parameter: --purgedb <system>". This is not correct. The '--purgedb' option can't purge a system. It can only purge certain types from certain modules. The "Skyscraper parameter" line could probably just be removed. I assume it currently just removes the corresponding folders without actually instantiating Skyscraper right now.
  * For "Download videos" it should probably say "Toggle the download and caching of videos when scraping. This also toggles whether the videos will be included when generating the final game list." or similar.
  * For all of the "Cache <resource type>" options the help should be "Toggle whether <resource type> are cached locally when scraping".

I think that is all I have for now.

On whether more options should be supported:
  * I think we should leave '--unpack' out for now, as it is a bit of an experimental feature and requires "7z" to be installed.
  * '--region' would be useful, although it is only supported by the "ScreenScraper" source right now (and should probably say so in the help text). But I think it could make sense to add it with "eu" (default), "us" and "jp" regions. It will always try all regions regardless, but the one chose is the one it will prioritize highest. If '--region' is left out, "eu" is the region Skyscraper prioritizes highest by default, then the rest.
  * '--lang' could also be implemented, but I don't have a good grasp on how much people use this. Personally I always leave it out to just use the default "en", but some users might want "es", "de" or others. This relates mostly, if not only, to the texts. Maybe also screenshots, I can't remember.

cmitu:
> 
> 
> Currently testing the ui:
> 
>     * When choosing "Cancel" in "Scrape chosen systems" it says "ROMS have been scraped" instead of just going back to the main Skyscraper menu.

 Yeah, that's a reminder from the original (`scraper`) script. I think I can fix it.
> 
>     * In "Scrape from..." menu "Open Retro" should be "OpenRetro".

 OK, will fix.
 
> 
>     * For "Remove bracket info", when it's "Disabled", the B in bracket suddenly becomes uppercase.

 That's a leftover from the UI experiment, I'll fix it.
 
> 
> 
> For help texts:
> 
>     * For "Scrape from..." help text it says "Open Retro", which should be "OpenRetro".

OK, I'll change it.

> 
>     * For "Scrape from..." help text it would probably make sense to add a small help text saying something like "Some sources need a  or user:password to work, which can be set per source in config.ini" or similar. Otherwise I foresee a lot of support requests complaining about certain sources not working.

Yes, that's a good idea, I'll add this.

> 
>     * For "Purge system" is says "Skyscraper parameter: --purgedb ". This is not correct. The '--purgedb' option can't purge a system. It can only purge certain types from certain modules. The "Skyscraper parameter" line could probably just be removed. I assume it currently just removes the corresponding folders without actually instantiating Skyscraper right now.

OK, I'll fix it.

> 
>     * For "Download videos" it should probably say "Toggle the download and caching of videos when scraping. This also toggles whether the videos will be included when generating the final game list." or similar.
> 
>     * For all of the "Cache " options the help should be "Toggle whether  are cached locally when scraping".
> 

OK, I'll fix it. 
 
> 
> I think that is all I have for now.
> 
> On whether more options should be supported:
> 
>     * I think we should leave '--unpack' out for now, as it is a bit of an experimental feature and requires "7z" to be installed.
> 
>     * '--region' would be useful, although it is only supported by the "ScreenScraper" source right now (and should probably say so in the help text). But I think it could make sense to add it with "eu" (default), "us" and "jp" regions. It will always try all regions regardless, but the one chose is the one it will prioritize highest. If '--region' is left out, "eu" is the region Skyscraper prioritizes highest by default, then the rest.
> 
>     * '--lang' could also be implemented, but I don't have a good grasp on how much people use this. Personally I always leave it out to just use the default "en", but some users might want "es", "de" or others. This relates mostly, if not only, to the texts. Maybe also screenshots, I can't remember.

 OK, I'll leave those aside for now. The scriptmodule also installs `p7zip`, so users can use `--unpack` from the CLI.
 
 Once the module is fine, the hard part will begin - writing the wiki documentation :). I'll get back after adding the corrections, thanks for the feedback!

muldjord:
I just had a look at the command that's executed when it scrapes. It add both '--unattend' and '--unattendskip'. It should by default only use "/opt/retropie/supplementary/skyscraper/Skyscraper --unattend" and then the rest. '--unattendskip' and '--unattend' are mutually exclusive and '--unattend' should be default. I'll let you know if I stumble upon other stuff.

muldjord:
Just checked it out, looks good!

cmitu:
OK, then.
@joolswills - can you also take a look to spot any problems ?

cmitu:
[.]..]
> Instead of cloning master, you can do:
> 
> ```
> LATEST=`wget -q -O - "https://api.github.com/repos/muldjord/skyscraper/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'`
> wget -N https://github.com/muldjord/skyscraper/archive/${LATEST}.tar.gz
> [create directory to unpack to]
> ...
> That will grab the latest release and unpack it to where you need it to be.

 I see. We can do that. The script already has a `latest_ver_skyscraper` function that gets the latest release tag (using `wget` and doing something similar to what you described) and RetroPie's git functions can download a tagged release instead of using the master branch.
 
> 
> UPDATE:
> The install script stalled on my Pi 3b+ at:
> 
> ```
> = = = = = = = = = = = = = = = = = = = = =
> Configuring 'skyscraper' : Scraper for EmulationStation by Lars Muldjord
> = = = = = = = = = = = = = = = = = = = = =
> ```
> Nothing happens after that. I've let it sit for 5 minutes now.

I'll take a look at it, it could be because of the latest changes I did to inspect the `skyscraper.pro` file.

cmitu:
I've modified the install to get the latest tagged release (which now it seems to be 2.7.5 ?), but I cannot reproduce the problem you're seeing during configuration. Are you installing from scratch or over an existing installation ?

muldjord:
I know why it stalls now. It stalls at:
```
+ cp -a /home/pi/.skyscraper/. /opt/retropie/configs/all/skyscraper/
```
Which is a problem. It tries to copy everything from .skyscraper to the configs dir. If users already have a substantial cache in there, it will copy it entirely. In my case, it was copying 14 GB of cached data. What is that step supposed to do? Can it be made to only copy needed files? Also, if any files needs to be in /opt/retropie/configs/all/skyscraper/ they should probably be symlinked instead of copied, so the authoritative ones are in .skyscraper, which is where Skyscraper looks for them. Thoughts on this?

cmitu:
Yeah, I was afraid of that. As I said at the beginning

> When installing through the scriptmodule, the existing configuration folder is preserved; it's also linked to /opt/retropie/configs/all/skyscraper as for any other module. [..]

The scriptmodule symlinks the `~/.skyscraper` to `/opt/retropie/configs/all/skyscraper` and preserves the contents. This happens also for ES (`~/.emulationstation` is a symlink also) or any other module that stores data under `~/` dot-folders / or `~/.config`. 
It's perfectly fine, it's just that for Skyscraper existing installations which have a lot of data in the Local DB this takes some time.

muldjord:
But it doesn't seem to be symlinking, it's actually copying the data with "cp -a" which is a physical copy. A symlinking would be near-instant and doesn't take up more space.

cmitu:
> 
> 
> But it doesn't seem to be symlinking, it's actually copying the data with "cp -a" which is a physical copy. A symlinking would be near-instant and doesn't take up more space.

No, actually the symlink is the other way around. It could probably be faster if using `mv`, but it probably tries to be safe and does the copying first.

muldjord:
Oh... Then that will be an issue. I foresee plenty of current users running into trouble here. Not only will they potentially run out of disk space before the copy is done, but they will also be wondering why it got stuck without a message about what it's doing.

muldjord:
Can we perhaps check for a current ./skyscraper/dbs folder when installing it, and make it move the folder during the installation instead? Then it would go on to configuring later, and then the dbs folder is already moved. That would probably work.

cmitu:
> Oh... Then that will be an issue.

Could be, depending on the size of the Local DB. I'm just using a standard RetroPie setup helper here.

> Can we perhaps check for a current ./skyscraper/dbs folder when installing it, and make it move the folder during the installation instead?

Yes, that's what I was suggesting above in https://github.com/RetroPie/RetroPie-Setup/pull/2497#issuecomment-427668386. I'll think of a way to safely do that - space wouldn't be an issue on a standard RetroPie install, only on a PC, where the user would have `/opt` mounted separately than `/home`.

cmitu:
OK, I modified the configuration step with an explicit test for an existing Skyscraper installation. During the fist install, if **`$HOME/.skyscraper`** is a folder (and not a symlink) then
* the `dbs` and `import` folders are moved to `$HOME/.skyscraper-dbs` and `$HOME/.skyscraper-import`
* the files in `$HOME/.skyscraper` are copied to the new conf folder and a symlink is created as `$HOME/.skyscraper`. This is handled by the RetroPie's function `moveConfigDir`.
* if the folders "$HOME/.skyscraper-dbs` and `$HOME/.skyscraper-import` are found, then they're moved to the new conf folder.

This should handle the case of - existing install + first configuration through the scriptmodule.

I also modified the configuration file to handle the purging of the cache during the uninstall and eliminated the `remove` function (where the purge was happening until now).

@muldjord - you can test the new module for installation when a large Local DB is present - but make a backup first if you can (maybe move a partial DB into place instead of copying the leaving the whole Local DB into place).
I've tested if a few times (with about 3 Gbs created artificially in the `dbs` and `import` folders) and it seems ok.

muldjord:
Just tested installation and removal, and it seems to work as expected, great work! No stalling on configuring anymore, and my dbs and import folders are moved to opt as expected. :)

cmitu:
Thanks for testing. 
Now I'm thinking if we also need to switch the default `gamelist.xml` file location when on existing installs or just to leave a note for users to switch to using the ROM folders using the GUI. By default, it uses the Emulationstation default folder (`~/.emulationstation/gamelist/<system>`).

muldjord:
Good point. I think we could just add a note about it in the Help text for the option "Use rom folder for gamelist and images". Something along the lines of "Note: When you change this option, it won't remove the gamelist and media from the old location. You need to do that manually.".

cmitu:
>  I think we could just add a note about it in the Help text for the option "Use rom folder for gamelist and images". Something along the lines of "Note: When you change this option, it won't remove the gamelist and media from the old location. You need to do that manually.". That will pretty much also tell people that if they have data in the location from previous scrapes they did outside of this ui, that won't be moved/deleted either.

 Yes, I can add it in the Help and also in the Wiki. 
> 
> EDIT: Btw, could you rename the option to "Use rom folder for gamelist and MEDIA" instead? Since it also contains video that makes more sense, and is also the name I use internally in Skyscraper.

Sure, no problem.

cmitu:
OK, done. 
Now the 'Use ROM folder for gamelist and media' defaults on **on** when migrating existing  installations. Added also the Help text when toggling the gamelist location options.

cmitu:
@joolswills This seems to - now - ready for review before merge.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+    md_ret_files=(
+    'Skyscraper'
+    'LICENSE'
+    'README.md'
+    )
+}
+
+# Purge all Skyscraper caches
+function _purge_skyscraper() {
+    find "$configdir/all/skyscraper/dbs" -maxdepth 1 -mindepth 1 -type d -print0 -exec rm -fr {} \;
```
joolswills: could just remove the dbs folder ? it's recreated I assume ?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+# Purge by cached platform
+function _purge_platform_skyscraper() {
+    local options=()
+    local system
+
+    while read system; do
+        # Get the size on disk of the system and show it in the select list
+        local size=$(du -sh  "$configdir/all/skyscraper/dbs/$system" | cut -f1)
+        options+=("$system" "$size" OFF)
+    done< <(find "$configdir/all/skyscraper/dbs" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
```
joolswills: space after done

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+
+
+function get_ver_skyscraper() {
+    [[ -f "$md_inst/Skyscraper" ]] && "$md_inst/Skyscraper" -h | grep 'Running Skyscraper'  | cut -d' '  -f 3 2>/dev/null
+}
+
+function latest_ver_skyscraper() {
+    # Prefer `wget` to `curl`, sinc it's already installed and used by the RetroPie script
+    wget -qO- https://api.github.com/repos/muldjord/skyscraper/releases/latest | grep -m 1 tag_name | cut -d\" -f4
+}
```
joolswills: is this function needed to be called manually by users ? if not then it should be hidden with _ prefix from retropie_packages.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+
+# List any non-empty system found in the ROM parent folder
+function list_systems_skyscraper() {
+    find "$romdir"  -mindepth 2 -maxdepth 2 -type f -not -name '+*' | sed 's%/[^/]*$%%' | sort -u
+}
```
joolswills: as above.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+
+    local cmd=(dialog --backtitle "$__backtitle" --checklist "Select ROM Folders" 22 76 16)
+    local choices
+
+    choices=($("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty))
+
+    # Exit if nothing was chosen or Cancel was used
+    [[ ${#choices[@]} -eq 0 || $? -eq 1 ]] && return 1
+
+    for c in "${choices[@]}"; do
```
joolswills: not local

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+        params+=(--nobrackets)
+    fi
+
+    if [[ "$force_refresh" -eq 1 ]]; then
+        params+=(--refresh)
+    fi
+
+    # trap ctrl+c and return if pressed (rather than exiting retropie-setup etc)
+    trap 'trap 2; return 1' INT
+        sudo -u $user "$md_inst/Skyscraper" "${params[@]}"
```
joolswills: why sudo ?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+    eval $(_load_config_skyscraper)
+
+    local gamelist
+    local media_dir
+
+    if [[ "$use_rom_folder" -eq 1 ]]; then
+        gamelist="$romdir/$system"
+        media_dir="$romdir/$system/media"
+        # If we're saving to the ROM folder, then use relative paths in the gamelist
+        relative="--relative"
```
joolswills: local scope ?

joolswills:
found a bunch of things like variables that are not local scope. I haven't looked through everything but its worth going through and checking for others.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+        local cmd=(dialog --backtitle "$__backtitle"  --colors --cancel-label "Exit" --help-button --no-collapse --cr-wrap --default-item "$default" --menu "   Skyscraper: game scraper by Lars Muldjord ($ver)\\n \\n" 21 65 10)
+        local -a options=(
+            1 "Scrape all systems"
+            2 "Scrape chosen systems"
+        )
+
+        local source_found=0
+        local online="Online"
+
+        for i in "${!s_source[@]}"; do
```
joolswills: local scope.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+
+    if [[ $(command -v sensible-editor) ]]; then
+        sensible-editor "$1"
+    else
+        editor=${EDITOR:-nano}
+        $editor "$1"
+    fi
+
+}
+
```
joolswills: we don't generally have spaces at the start and end of functions like this.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+    )"
+}
+
+# Try to guess the most appropriate editor. On Debian derivatives, we have `sensible-editor` for that.
+function _open_editor_skyscraper() {
+
+    if [[ $(command -v sensible-editor) ]]; then
+        sensible-editor "$1"
+    else
+        editor=${EDITOR:-nano}
```
joolswills: prefer quotes for things like this.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+        'scrape_source=screenscraper' \
+        'remove_brackets=0' \
+        'force_refresh=0'
+    )"
+}
+
+# Try to guess the most appropriate editor. On Debian derivatives, we have `sensible-editor` for that.
+function _open_editor_skyscraper() {
+
+    if [[ $(command -v sensible-editor) ]]; then
```
joolswills: might be clearer to use -n

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,680 @@
...
+                    cache_marquees="$((cache_marquees ^ 1))"
+                    iniSet "cache_marquees" "$cache_marquees"
+                   ;;
+
+                S)
+                    _purge_platform_skyscraper
+                    ;;
+
+                P)
+                    confirm=$(dialog --clear --beep  --defaultno --colors --yesno  "\Z1\ZbAre you sure ?\Zn\nThis will \Zb\ZuERASE\Zn all locally cached scraped resources" 8 60 2>&1 >/dev/tty)
```
joolswills: we don't use beep anywhere else. Also we don't use --colors elsewhere, so might be best to stick with the look of other dialogs ?

scriptmodules/supplementary/skyscraper.sh
cmitu: I think so. I'll test and see if there's any error reported.

scriptmodules/supplementary/skyscraper.sh
cmitu: No, I'll hide it.

joolswills: thanks. might be worth checking other functions also - eg get_ver

scriptmodules/supplementary/skyscraper.sh
cmitu: `--beep` is a leftover experiment, but colors are used mainly in the Help section for formatting and only for Bold/Underline. I'll eliminate any formatting on the main screen, but I'd like to keep them in the Help sections.

cmitu:
@joolswills thank you for the review. I'll address the changes and fix them. Is there any linter for shell sources - I've not found anything else besides ShellCheck, which is a bit limited.

joolswills:
Think i've just used shellcheck in the past occasionally, but we don't follow all it's recommendations (and it's limited as you have noted).

scriptmodules/supplementary/skyscraper.sh
muldjord: The folder is recreated, but there are example files in the folder that should not be deleted. It might even be worth considering not deleting the platform subfolders themselves, since the user might have created custom dbs/[platform]/priorities.xml files for each platform. So basically a purge for all systems should remove only dbs/[platform]/db.xml files and all folders in dbs/[platform]/[folders]. Same goes for purging a single platform.

cmitu: OK, I'll modify the `purge_..` functions to take this into account. I knew something about the `priorities.xml`, but though it's a global file.

scriptmodules/supplementary/skyscraper.sh
cmitu: That was the original code in the `scraper` module (https://github.com/RetroPie/RetroPie-Setup/blob/13b99f039c2dafa801593c5afd712ed6a32600a1/scriptmodules/supplementary/scraper.sh#L136). I assume because the RetroPie-Setup runs as `root`, but the scraping is done by the user.

joolswills: Ah yes. forgot about that.

cmitu:
@joolswills I've modified the script after the review, mainly
* added `local` to the vars that had it missing. 
* renamed internal functions and prefixed them with `_`.
* modified the function style to remove extra lines at the beginning/end
* minimized the use of `--colors`  on the GUI's front menus and left any formatting on the Help pages.

I've also changed the purge functions to take into account what @muldjord mentioned about the keeping the conf file when deleting a system's cache.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+rp_module_section="opt"
+
+function depends_skyscraper() {
+    getDepends qt5-default p7zip-full
+}
+
+function sources_skyscraper() {
+    local latest_ver
+    latest_ver=$(_latest_ver_skyscraper)
+
```
joolswills: not sure there is any need for a variable here at all.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+    qmake
+    make
+    md_ret_require="$md_build/Skyscraper"
+}
+
+function install_skyscraper() {
+    md_ret_files=(
+    'Skyscraper'
+    'LICENSE'
+    'README.md'
```
joolswills: wrong indent (see other modules).

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+    [[ ! -d "$configdir/all/skyscraper/dbs/$platform" ]] && return
+
+    # Remove any folder underneath the platform, try to keep any customized files
+    pushd "$configdir/all/skyscraper/dbs/$platform"
+    find . -maxdepth 1 -mindepth 1 -type d -print0 -exec rm -fr {} \;
+    rm -f db.xml
+    popd
+
+}
+# Purge by cached platform
```
joolswills: white space above in function.

no gap between functions.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+
+    # Exit if no platform chosen
+    [[ -z "$platform" ]] && return
+
+    _clear_platform_skyscraper "$platform"
+}
+
+
+function _get_ver_skyscraper() {
+    [[ -f "$md_inst/Skyscraper" ]] && "$md_inst/Skyscraper" -h | grep 'Running Skyscraper'  | cut -d' '  -f 3 2>/dev/null
```
joolswills: maybe use `if` here so it's not all on one line as it's rather long.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+}
+
+function _latest_ver_skyscraper() {
+    # Prefer `wget` to `curl`, sinc it's already installed and used by the RetroPie script
```
joolswills: remove this comment please.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+function _list_systems_skyscraper() {
+    find "$romdir"  -mindepth 2 -maxdepth 2 -type f -not -name '+*' | sed 's%/[^/]*$%%' | sort -u
+}
+
+function configure_skyscraper() {
+    # When removing the package, clean up the Local DB
+    if [[ "$md_mode" == "remove" ]]; then
+        _purge_skyscraper
+        return
+    fi
```
joolswills: the purge function should probably be a remove_ function

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+
+    # Handle the cases where the user has an existing Skyscraper installation.
+    if [[ -d "$home/.skyscraper" && "$local_config" != "$configdir/all/skyscraper" ]]; then
+        # We have an existing Skyscraper installation, but not handled by this scriptmodule.
+        # Since the $HOME/.skyscraper folder will be moved, make sure the Local DB and Import folders are moved separately
+        local f_size
+        f_size=$(du --total -sm "$home/.skyscraper/dbs" "$home/.skyscraper/import" 2>/dev/null | grep total | cut -f 1 )
+        printMsgs "console" "INFO: Moving the Local DB and Import folders to new configuration folder (total: $f_size Mb)"
+
+        for folder in dbs import; do
```
joolswills: no local.

joolswills:
made a few more comments - please go over the whitespace in the file. eg I found a function with no gay from the one above. In some functions you have more lines than needed in places with 2 line gaps etc - looks inconsistent.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+            if [[ "$scrape_source" == "${s_source[$i]}" ]]; then
+                [[ $i -ge 10 ]] && online="Local"
+                options+=(3 "Scrape from ${s_source_names[$i]} ($online)")
+                source_found=1
+            fi
+        done
+
+        if [[ ! $source_found ]]; then
+             options+=(3 "Scrape from Screenscraper (Default)")
+             scrape_source="screenscraper" # default scraping source if none found
```
joolswills: wrong indent.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
+#!/usr/bin/env bash -x
```
joolswills: please remove the +x as pointed out by @hhromic

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,696 @@
...
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="skyscraper"
+rp_module_desc="Scraper for EmulationStation by Lars Muldjord"
+rp_module_licence="GPLv3.0 https://raw.githubusercontent.com/muldjord/skyscraper/master/LICENSE"
+rp_module_section="opt"
+
+function depends_skyscraper() {
+    getDepends qt5-default p7zip-full
```
joolswills: What is p7zip-full needed for ?

cmitu: It's used by the `--unpack` parameter to unzip/unpack any 7z packed ROMs and hash them. See 
https://github.com/muldjord/skyscraper/blob/24ed2abb1639864dc5343ca6065865f492210694/src/skyscraper.cpp#L133

joolswills: ok thanks.

scriptmodules/supplementary/skyscraper.sh
cmitu: Fixed.

scriptmodules/supplementary/skyscraper.sh
cmitu: It was supposed to be a one liner to get the version, but I think it's clearer to use an `if`. Modified accordingly.

scriptmodules/supplementary/skyscraper.sh
cmitu: OK, fixed.

scriptmodules/supplementary/skyscraper.sh
cmitu: OK, fixed.

scriptmodules/supplementary/skyscraper.sh
cmitu: OK, I removed it and put the call inline

scriptmodules/supplementary/skyscraper.sh
cmitu: Initially it was, but I thought to spare an extra function definition and add it to the `configure`'s remove call. But it's clearer if moved to a separate function - I added the `remove_` back.

scriptmodules/supplementary/skyscraper.sh
cmitu: OK, fixed.

scriptmodules/supplementary/skyscraper.sh
cmitu: Done, thanks for noticing @hhromic.

cmitu:
>  made a few more comments - please go over the whitespace in the file. eg I found a function with no gap from the one above. In some functions you have more lines than needed in places with 2 line gaps etc - looks inconsistent.

I'll check over the whitespace and then commit the latest changes.

cmitu:
OK, did a whitespace (spurious newlines and a few 3 space indents) cleanup and seems ok.

joolswills:
Thanks for the changes. If you are happy, please can you squash the commits, and I will do a final quick check and merge. Appreciate the contribution - it's good to have another scraper implemented (especially with the detail you have given the interface).

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,687 @@
...
+        if [[ -f "$md_conf_dir/$f_conf" ]]; then
+            cp "$md_build/$f_conf" "$md_conf_dir/$f_conf.rp-dist"
+        else
+            cp "$md_build/$f_conf" "$md_conf_dir"
+        fi
+    done
+
+    # If we don't have a previous config.ini file, copy the dist one
+    [[ ! -f "$md_conf_dir/config.ini" ]] && cp "$md_conf_dir/config.ini.rp-dist" "$md_conf_dir/config.ini"
+
```
joolswills: Might it be worth using copyDefaultConfig. It could be that this isn't exactly what you need, but with current retropie modules we only create a rp-dist if the standard config is changes from the default. This may be useful or not to you.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -0,0 +1,687 @@
...
+    # Try to find the rest of the necesary files from the qmake build file
+    # They should be listed in the `unix:examples.file` configuration line
+    if [[ $(grep unix:examples.files "$md_build/skyscraper.pro" 2>/dev/null | cut -d= -f2-) ]]; then
+        local files=$(grep unix:examples.files "$md_build/skyscraper.pro" | cut -d= -f2-)
+        local file
+
+        for file in $files; do
+            # Copy the files to the configuration folder. Skip config.ini and artwork.xml, they're already handled
+            if [[ $file != "artwork.xml"  && $file != "config.ini" ]]; then
```
joolswills: very minor but I noticed you have 2 spaces before &&

muldjord:
I second that. Thank you very much for your work on this @cmitu, I look forward to seeing where this goes when people start using it.

scriptmodules/supplementary/skyscraper.sh
cmitu: OK, will fix it.

scriptmodules/supplementary/skyscraper.sh
cmitu: The idea here was to generate a basic default for a new installation and to save it in case editing it will bungle the config. By default skyscraper has a `config.ini.example` but it can run without.  I can change the suffix to keep the convention for `.rp-dist`.

joolswills: Happy to leave the decision to you. Just wanted to mention the function. Whatever you think is best in this case. Thanks

cmitu:
@joolswills added the modifications and squashed the commits. Module is in 'optional', should appear alongside the `scraper` module.

cmitu:
> I second that. Thank you very much for your work on this @cmitu, I look forward to seeing where this goes when people start using it.

Yep, will be interesting. Unfortunately I won't have time until Sunday to do any additions to the wiki describing the Gui and usage, but once merged you should probably make a new topic in the forums for people to report problems for the module specifically - instead of cluttering the existing topic (which is already all time 4th most popular topic).

muldjord:
Yes, you are right, it's probably fitting with a new topic. I will do that once it's merged.

cmitu:
> Proper credit will be given both places of course. :)

Eh [with Canadian accent], don't feel you need to do that -

PR #2498, opened 2018-10-03
Update esthemes.sh

waweedman:
i have updated my waweedman pii-wii' and add 3 newthemes 'waweedman Blade-360 'waweedman Venom''waweedman Spider-Man'


scriptmodules/supplementary/esthemes.sh
```diff
@@ -171,6 +171,9 @@ function gui_esthemes() {
         'lipebello spaceoddity'
         'lipebello swineapple'
         'waweedman pii-wii'
```
waweedman: update

waweedman: update

PR #2500, opened 2018-10-05
Allow `lr-mess` to finish building on the Raspberry Pi

cmitu:
Based on the discussion in https://retropie.org.uk/forum/topic/18126/lr-mess-and-lr-mess2016-not-compiling/19.

* increase the total memory allocated to 2Gb
* limit the compilation to 2 processes, in the memory critical sections the system is just swapping anyway.

### DIFF
--- a/scriptmodules/libretrocores/lr-mess.sh
+++ b/scriptmodules/libretrocores/lr-mess.sh
@@ -24,7 +24,7 @@ function sources_lr-mess() {
 }
 
 function build_lr-mess() {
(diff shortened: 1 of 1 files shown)


joolswills:
2 processes is default for the rpi. 2gb seems a lot and it compiled for me with less before. I will need to do some testing before changing the swap size.

joolswills:
I increased to 2000MB for lr-mess and lr-mame but left out the makeflags changes. Thanks for the heads up.

PR #2502, opened 2018-10-07
KDE Neon is now based on Ubuntu 18.04 LTS.

benpro:
We need to set `__os_debian_ver="9"` as KDE Neon is now based on Ubuntu 18.04
LTS and we need the little libpng-dev trick.
Old KDE Neon use Ubuntu 16.04 LTS.


joolswills:
thanks.

PR #2504, opened 2018-10-09
Fix lr-fbalpha build on unrecognized neon platforms

EriHoss:
Fix for neon platforms not recognized by lr-fbalpha makefile.

See Issue: https://github.com/libretro/fbalpha/issues/249


joolswills:
thanks.

PR #2505, opened 2018-10-11
retroarch - refresh patches

hhromic:
The following patches for `retroarch` needed a rebase after the recent update to `v1.7.5`:

* `01_hotkey_hack.diff`
* `02_disable_search.diff`

This PR refreshes them so they apply cleanly when compiling from source.

### DIFF
--- a/scriptmodules/emulators/retroarch/01_hotkey_hack.diff
+++ b/scriptmodules/emulators/retroarch/01_hotkey_hack.diff
@@ -1,8 +1,8 @@
 diff --git a/input/input_driver.c b/input/input_driver.c
-index 3d5ee6ca73..c9b58a6f1f 100644
+index f49fd82..d13ba8e 100644
 --- a/input/input_driver.c
 +++ b/input/input_driver.c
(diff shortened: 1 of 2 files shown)


joolswills:
I already commented on this here - https://github.com/RetroPie/RetroPie-Setup/commit/41fe1fa1bc54d3032ba5c3fd0b237c30b2226e76

They don't "need" a rebase. They still apply fine (with fuzz). This PR isn't needed imho but I will merge.

joolswills:
I will need to check these before merge - it's easier if I just rebase them myself and commit rather than testing this PR.

hhromic:
Oh sorry, didn't notice that conversation. GitHub's UI is a bit obscure on these commit-specific comments.
In relation, I do believe that keeping patches clean is a good/healthy software maintenance practice :)

I'm sorry, I forgot to mention that I did test the patches before submitting the PR, the idea is to save you some manual testing time too. But good stuff you merged the rebases yourself now. Thanks!

PR #2512, opened 2018-10-19
escape $ inside heredoc

win-t:
escape `$` inside heredoc


joolswills:
good spot - thanks for that.

PR #2520, opened 2018-10-25
Fix the install from a binary package and a scraping error

cmitu:
* the configure phase needed the build folder, breaking binary installs.
  Fix this by adding the necessary files to the binary install folder and making configure look there.
* added a few more samples for the 'import' and 'dbs' folders. 
* fixed a scraping problem because of a stray space in the sudo scraping command
* default config enables 'wheels' download and scraping and removes the maxFails
~~* added 'unpack' by default. This should alleviate the need to add --unpack from cli when
  scraping zipped ROMs from screenscraper.~~

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -33,6 +33,20 @@ function install_skyscraper() {
         'Skyscraper'
         'LICENSE'
         'README.md'
+        'config.ini.example'
+        'artwork.xml'
+        'artwork.xml.example1'
+        'artwork.xml.example2'
+        'artwork.xml.example3'
+        'artwork.xml.example4'
+        'ARTWORK.md'
+        'tgdb_developers.json'
+        'tgdb_publishers.json'
+        'mameMap.csv'
+        'hints.txt'
+        'dbs'
+        'import'
+        'resources'
     )
 }
 
@@ -163,16 +177,13 @@ unattend="true"
 unattendSkip="true"
 noHints="false"
 threads="2"
-maxFails="30"
 lang="en"
 region="wor"
 videos="false"
 marquees="true"
-wheels="false"
+wheels="true"
 covers="true"
 screenshots="true"
-artworkXml="$md_conf_dir/artwork.xml"
-
 _EOF_
 
     # Make sure the `artwork.xml` and other conf file(s) are present, but don't overwrite them on upgrades.
@@ -180,9 +191,9 @@ _EOF_
     local f_conf
     for f_conf in artwork.xml; do
         if [[ -f "$md_conf_dir/$f_conf" ]]; then
-            cp "$md_build/$f_conf" "$md_conf_dir/$f_conf.default"
+            cp "$md_inst/$f_conf" "$md_conf_dir/$f_conf.default"
         else
-            cp "$md_build/$f_conf" "$md_conf_dir"
+            cp "$md_inst/$f_conf" "$md_conf_dir"
         fi
     done
 
@@ -203,26 +214,28 @@ _EOF_
         done
     else
         # Fallback to the known resource files list
-        cp -f "$md_build/artwork.xml.example"* "$md_conf_dir"
+        cp -f "$md_inst/artwork.xml.example"* "$md_conf_dir"
 
         # Copy resources
         local resource_file
         for resource_file in ARTWORK.md README.md mameMap.csv tgdb_developers.json tgdb_publishers.json hints.txt; do
-            cp -f "$md_build/$resource_file" "$md_conf_dir"
+            cp -f "$md_inst/$resource_file" "$md_conf_dir"
         done
     fi
 
     # Copy the rest of the folders
-    cp -rf "$md_build/resources" "$md_conf_dir"
+    cp -rf "$md_inst/resources" "$md_conf_dir"
 
-    # Create the import folders
+    # Create the import folders and add the sample files.
     local folder
-    for folder in boxart snaps covers marquees screenshots textual videos wheels; do
+    for folder in covers marquees screenshots textual videos wheels; do
         mkUserDir "$md_conf_dir/import/$folder"
     done
+    cp -rf "$md_inst/import" "$md_conf_dir"
 
-    # Create the LocalDB cache folder
+    # Create the LocalDB cache folder and add the sample files
     mkUserDir "$md_conf_dir/dbs"
+    cp -fr "$md_inst/dbs" "$md_conf_dir"
 }
 
 # Scrape one system, passed as parameter
@@ -236,7 +249,7 @@ function _scrape_skyscraper() {
 
     local gamelist
     local media_dir
-    local relative
+    local relative=""
 
     if [[ "$use_rom_folder" -eq 1 ]]; then
         gamelist="$romdir/$system"
@@ -251,7 +264,8 @@ function _scrape_skyscraper() {
     local -a params=("--unattend" "--skipped")
 
     params+=(-o "$media_dir")
-    params+=(-g "$gamelist" "$relative")
+    params+=(-g "$gamelist")
+    [[ -n "$relative" ]] && params+=("$relative")
     params+=(-p "$system")
     params+=(-s "$scrape_source")
 
@@ -333,7 +347,7 @@ function _scrape_chosen_skyscraper() {
     local choice
 
     for choice in "${choices[@]}"; do
-        choice=${options[choices*3-2]}
+        choice="${options[choice*3-2]}"
         _scrape_skyscraper "$choice" "$@"
     done
 }


scriptmodules/supplementary/skyscraper.sh
```diff
@@ -287,10 +300,12 @@ function _scrape_skyscraper() {
         params+=(--refresh)
     fi
 
+    local param_list="${params[@]}"
```
joolswills: Why the additional variable rather than using the array directly?

joolswills: I saw on the forum it's due to variable expansion, but shouldn't need a new var. Did you try `${params[*]}`

cmitu: I didn't, but I'll give it a shot and see if I can improve it.

cmitu: It turns out it was the quoting was passing the whole params array as a parameter. Scraper does this correctly [here](https://github.com/RetroPie/RetroPie-Setup/blob/9549a01a44308af922daa7c8acd8a4269dfdd0b2/scriptmodules/supplementary/scraper.sh#L136), I've must have added it myself and broke it during the review.
I've added another fix found during testing for this modification.

hhromic: Interesting. To my understanding the quotes should actually be there to prevent word splitting. With quotes the paramters are expanded, quoted and passed one by one. See this quick test:
```sh
#!/usr/bin/env bash
function print() { local arg; for arg in "$@"; do echo "argument: $arg"; done}
myarray=(this is a "test with" five words)
print "${myarray[@]}"
```
Which works as expected:
```
$ bash test.sh
argument: this
argument: is
argument: a
argument: test with
argument: five
argument: words
```
Maybe there is something else giving you weird results. The `sudo` call perhaps?

hhromic: And if you remove the quotes:
```sh
print ${myarray[@]}
```
You get the undesired effect of word splitting:
```
$ bash test.sh
argument: this
argument: is
argument: a
argument: test
argument: with
argument: five
argument: words
```

cmitu: Interesting, I'll do a test to see how this runs through `sudo`.

cmitu: It doesn't make a difference, using your test, if using `sudo` or not. Quoted expansion happens every time.

hhromic: What is the exact error you receive when quoting the expanded array? I'm very curious now.

I just tested through `sudo` like this:
```sh
#!/usr/bin/env bash
myarray=(/boot /mnt /proc "no exist")
sudo ls -1lad "${myarray[@]}"
```
And it works as expected (notice how `ls` can't find the single dir `no exist` and does find each element of the array correctly):
```
$ bash test.sh
ls: cannot access no exist: No such file or directory
drwxr-xr-x   4 root root 2048 May  2  2015 /boot
drwxr-xr-x   4 root root 4096 Jan 10  2014 /mnt
dr-xr-xr-x 255 root root    0 May 20 14:04 /proc
```
I'm skeptic of the program mishandling the quotes because the shell expands all arguments before even calling the program.

cmitu: @hhromic there's no error quoting the array in the script, but I found what happens. The `$relative` variable can be empty and it's passed as a quoted empty string by `"${myarray[@]}"` , which does not happen for `${myarray[@]}` . `Skyscraper` gets the empty parameter and assumes it's a filename, thus the error message.

@joolswills I'll do a proper fix for this.

hhromic: brilliant! well done :) that was tricky to spot

cmitu: @joolswills fixed by selectively adding `$relative`.

hhromic: I got thinking, why not initialise `params` further up (before the conditionals) and then just add parameters according to the conditions? you could get rid of the temporary `gamelist`, `media_dir` and `relative` variables all-together like that. For example:
```sh
local -a params=("--unattend" "--skipped")
if [[ "$use_rom_folder" -eq 1 ]]; then
    params+=(-g "$romdir/$system")
    params+=(-o "$romdir/$system/media")
    # If we're saving to the ROM folder, then use relative paths in the gamelist
    params+=(--relative)
else
    params+=(-g "$home/.emulationstation/gamelists/$system")
    params+=(-o "$home/.emulationstation/downloaded_media/$system")
fi
```
I think it makes cleaner and more robust code :)

hhromic: You could also compact code like this for less clutter:
```sh
if [[ "$cache_screenshots" -eq 0 ]]; then
    params+=(--noscreenshots)
fi
# is equivalent to this
[[ "$cache_screenshots" -eq 0 ]] && params+=(--noscreenshots)
```

joolswills:
Thanks for this.

joolswills:
Let's leave it a week, and then if all good, ill do a binary build and move it to optional ?

cmitu:
Yes, I think that's ok - but we can leave it more than 1 week. 
I've only tested the 'binary install' by simulating a 'configure' after 'install' and removing the build folder in between. I think I can fiddle around with `archive_url` and simulate a binary download from a local web server just to make sure it works fine.

scriptmodules/supplementary/skyscraper.sh
cmitu: @hhromic thanks for the tips! I'll probably peruse the script for possible optimizations and add your observations to it, in a subsequent revision.

muldjord:
Just update setup script and installed Skyscraper from source through the Experimental package. Choosing "edit 'config.ini'" and "edit 'artwork.xml'" just returns to the menu. Currently running the script locally on the Pi with keyboard connected directly to the Pi.

I also noticed issues with this in the old package, where it would first ask me which editor I preferred, and after choosing "nano" it opened the nano editor, but keyboard input didn't work. This was through a remote ssh session from Ubuntu to the Pi.

muldjord:
"Purge all!" option exits to command line.

muldjord:
When running either "Scrape all systems" or "Scrape system" directly on the Pi, the output is b0rked:
skyscraper_output

cmitu:
@muldjord The edit part is a bug, I think the editor is not started as the user and you might be right about the the editing being eaten by the setup script. Since I don't have a keyboard for the Pi, I only tested this via `ssh`.
I'll take a look at the scraping output - why it looks so displaced.

muldjord:
You had the editing working over ssh? I can't get it to work either way. The result is the same. When I use the cursor keys to move around in the text in nano, it just prints out garbage on the screen. I can exit nano with ctrl+x, but it prints garbage while doing so, so it's only because I know what I am supposed to type that I can exit nano.

cmitu:
@muldjord I see the problem over ssh on the PI also. The directional arrows and control chars are not shown correctly . It's solved by running as `$user`, instead of the `root`. Replacing [here](https://github.com/RetroPie/RetroPie-Setup/blob/792ebb8a0e59b8871bba578a3252866e0526d9de/scriptmodules/supplementary/skyscraper.sh#L378) with 
```
sudo -u $user $editor "1"
```
fixes this.

But I cannot reproduce the output during scraping or the purge all errors. If I scrape directly from the Pi screen (i.e. no ssh connection) using a gamepad, the output looks fine - the only things that's awkward is the output buffering (which you've been working on lately by adding `fflush`).

'Purge all' also works, I don't get an error and the folders are purged - if I go to purge a system I get the 'there's no system to purge' message. Note that 'Purge all' has the 'cancel' button selected by default, you need to press left (for <kbd>Ok</kbd>) and then press ok/enter to initiate the purge.

muldjord:
I was able to reproduce the exiting when choosing "Purge all!" both on the Pi and over ssh. I initiated a scraping of "Scrape all platforms", waited for it to scrape a few games, and then ctrl+c'd it. It then says "Scraping aborted". I then went into the advanced menu and chose "Purge all!" and chose "Yes". Then it exited to command line.

Concerning nano problems. I changed line 375 to:
```
sudo -u $user sensible-editor "$1"
```
and line 378 to:
```
sudo -u $user $editor "$1"
```
Same thing as before both for ssh and local. When I press the keyboard arrow keys in nano I get this:
skyscraper_garbled_nano

The faulty output from Skyscraper seemingly not doing carriage returned as reported in my previous post, I have been unable to reproduce that again. So it just seems unstable as it clearly happened.

In general it seems that terminal output is handled quite oddly when running anything from within the setup script. It seems to buffer a lot before actually printing it to screen, and it seems to not flush very often, and definitely not on every "\n" as I would normally expect.

cmitu:
> I was able to reproduce the exiting when choosing "Purge all!" both on the Pi and over ssh. I initiated a scraping of "Scrape all platforms", waited for it to scrape a few games, and then ctrl+c'd it. It then says "Scraping aborted". I then went into the advanced menu and chose "Purge all!" and chose "Yes". Then it exited to command line.
> 

 Aha. Yes, I can reproduce it now also. 

> Concerning nano problems. I changed line 375 to:
> 
> ```
> sudo -u $user sensible-editor "$1"
> ```
> and line 378 to:
> 
> ```
> sudo -u $user $editor "$1"
> ```
> Same thing as before both for ssh and local. When I press the keyboard arrow keys in nano I get this:
> skyscraper_garbled_nano

Hm, I can't reproduce this. Did you restart the script after making the modifications ?

> 
> I nave not been able to reproduce the missing carriage returns in the output as reported in my previous post. So it just seems unstable as it clearly happened.
> 
> In general it seems that terminal output is handled quite oddly when running anything from within the setup script. It seems to buffer a lot before actually printing it to screen, and it seems to not flush very often, and definitely not on every "\n" as I would normally expect.

I think I have an idea how this can be improved, let me do some tests I might have a fix for this also.

muldjord:
> Hm, I can't reproduce this. Did you restart the script after making the modifications ?
Yes, several times. I also did a modification to verify that it was using the correct script, and the minor change I made came through, so it's definitely updated and using the modification. It's a clean RetroPie 4.4 on a RPi3b+ with just Skyscraper installed, so not sure what could be causing it. But it happens both locally and through ssh.

cmitu:
I don't think that's something in the installation, or the ES settings (ES is not even running when you try to edit).  I think I found a fix for the output buffering and I've added the editing fix from above. Can you try [this version](https://raw.githubusercontent.com/cmitu/RetroPie-Setup/skyscraper-pr2/scriptmodules/supplementary/skyscraper.sh) and see if you can still see the editing problems and the output buffering ?

muldjord:
I still get garbage when pressing any key on the keyboard while in nano with the version you linked. When I chose "Edit 'artwork.xml'" I get an odd error aswell:
> File ...work.xml is being edited (by root with nano 2.7.4, PID 2093); continue?

General output works a lot better now! It flushes the way I would expect now.

cmitu:
> I still get garbage when pressing the cursor keys on the keyboard while in nano with the version you linked. When I chose "Edit 'artwork.xml'" I get an odd error aswell:
> 
> > File ...work.xml is being edited (by root with nano 2.7.4, PID 2093); continue?

 There is probably a lock file for the .xml file. Use `ls -l ~/.skyscraper` to spot it and delete it. 
 I'll see if I can figure out a way make `nano` work, though it's going to be tricky since I can't reproduce the problem anymore. I wonder if it's because of the terminal type used. If you run 
```
echo $TERM
```
from the session/terminal where you start the script, what's the output ?

muldjord:
> from the session/terminal where you start the script, what's the output ?

xterm-256color

cmitu:
@muldjord I think I found the problem to the editing part. 
Using `dialog` for the gui 'steals' the stdin and stdout and the editor doesn't get the input correctly. It happens when running the main setup, if I run the Skyscraper GUI with `sudo ~/RetroPie-Setup/retropie_packages skyscraper gui` then editing is fine (regardless of using `sudo` or not). I don't have a solution yet, but we can use a `dialog` editing box as a last resort.

cmitu:
@muldjord OK, found the solution for the file edit - you can test it again from [here](https://raw.githubusercontent.com/cmitu/RetroPie-Setup/skyscraper-pr2/scriptmodules/supplementary/skyscraper.sh).
I've been looking over the other error reported - the `Purge All` crashing. It's because of the interrupt <kbd><Ctrl+C</kbd> during the scraping, but haven't figured out the solution yet. It happens also for the `scraper` module, where the interrupt handling routine originated.

muldjord:
Confirmed, editing now works as expected, awesome! :)

muldjord:
Just removed the config.ini and reinstalled Skyscraper to check which default flags are in there. 'unattendSkip="true"' should be removed, it can't coexist with "unattend".

I was thinking, couldn't we just copy in the config.ini.example to start off with? I've made it so it has all of the available options and there's some explanations in there aswell. Would help the user get started properly.

cmitu:
@muldjord The switches on the GUI act as CLI switches, they don't modify the `config.ini` - for now. 

My idea was to set some reasonably defaults on the configuration file and then use the CLI switches to modify them. I know some of them are already defaults - so they look like setting them looks redundant. As a rule, defaults can change in applications, so I have the habit of settings them explicitly in the config file - that's just a habit of mine. 
Of course, we can mimic or copy the `config.ini.example` you provide in the installation and use it as a default, that's not a problem. One thing I'd like to keep is to explicitly disable video download, because you don't have a `--novideos` switch (or am I mistaken ?) and have the user enabling video download themselves (thus using the CLI switch to download them).

cmitu:
@muldjord The problem about the `config.ini` editing through the GUI is that you can have multiple sections with the same config option (i.e. _main_  might have `X=true`, _[nes]_ could have `X=false`), so there's no really an easy way to know what the end result would be. 

I'm fine copying the default `example` config.

For the _Purge All_ problem, it turns out the issue is a bit more complicated and it's not localized to this particular script - the same happens for `scraper`. I should open a separate issue for this, because solving could involve the modification of the external `setup.sh` scriptmodule. 
Basically, the problem is that when started from the `setup` screen, the Ctrl+C triggers a SIGPIPE in the main script (`retropie_setup.sh`), which terminates it completely.

muldjord:
Hey, what is the status of the script as of today? :) Is it now copying the config.ini.example unedited as mentioned? And I think it's fine that we don't let the script itself edit the config.ini for now. Just keep it working with command line options. I have an idea for this btw. I could add a command line option to Skyscraper that let's you manipulate the config.ini that is used. So something like '--configedit [section],[variable]=[value]' and all it would do was just edit that and quit. Would that be useful in regards to the script?

The userediting of config.ini from the copied config.ini.example is still relevant though, so just leave that in. Many options are still necessary to set, for instance the userid and password for certain scraping modules. So we need to have the useredit of config.ini for that to be "easy" to add.

Other than those things, I don't think I have more comments as of now.

cmitu:
@muldjord I'm preparing another PR for the modifications you requested, I'm just giving it a few more tests.

muldjord:
Great, sounds like you guys got your hands dirty with the propagating signal. :) Looking forward to having that fixed as well when you find an acceptable solution.

PR #2522, opened 2018-10-26
added lr-fbalpha2018 module

joolswills:
upstream fbalpha moved to github, so libretro deprecated the old fbalpha repository and renamed it to fbalpha2018.
lr-fbalpha is now a fork of the upstream development and will be more up to date, but will require more frequent
romset changes - see https://retropie.org.uk/forum/topic/19741/new-fb-alpha-libretro-pre-v0-2-97-44

This PR is still undergoing testing.


barbudreadmon:
I don't recommend using the fba2018 though, it'll just add more confusion.

joolswills:
Going to leave it for now.

PR #2528, opened 2018-11-08
openbor-3400.sh

crcerror:
Added old 3400 branch with PAK and CLI support
This version is an outdated version of OpenBOR
It supports full load/save function
It supports loading PAKs (so extracting is not needed anymore)


crcerror:
Added openbor-6xxx branch
There are lots of more things to do :(
libGL should be build from source code
Maybe binaries should be provided as download link directly :(
Would appreciate some help ;)

crcerror:
@hhromic Can you please help me to get those two scriptmodules integrated into RetroPie?

hhromic:
@crcerror I would suggest you to go one scriptmodule at a time.
Given that the title of this PR is for openbor-3400, let's start with that one first.

Can you refresh this PR to only include the openbor-3400 scriptmodule and any recent updates you have done? Also please make a single commit so it can be reviewed easier.

crcerror:
Added new PR

PR #2529, opened 2018-11-12
Don't remove analogues on EmulationStation post config

raelgc:
I've in a different on PR https://github.com/RetroPie/EmulationStation/pull/499 to make EmulationStation work with both digital and analogue directionals, but the `/opt/retropie/supplementary/emulationstation/configscripts/emulationstation.sh` script actually remove analogues from final configuration.

So this PR is about to keep analogues directionals.


hhromic:
this PR seems senseful, the only improvement I would do is the commit log message :)

joolswills:
Thanks. Please can you make your commit message a bit more verbose and I'll merge. See my commit history for examples :-)

raelgc:
@joolswills Sorry, just made an amend to the commit with a bit more explanation.

joolswills:
Thanks.

PR #2532, opened 2018-11-16
Add Splitwolf: Splitscreen Wolfenstein 3D

lazd:
# Splitwolf

Based on the same setup script as Wolf4SDL, this PR adds [_Splitwolf_ from Team Raycast](https://bitbucket.org/linuxwolf6/splitwolf/wiki/Home), 2-4 player splitscreen Wolfenstein 3D with new game modes and full gamepad support!

1540921030208-screen-shot-2018-10-30-at-10 36 31-am

That's right, you can frag Nazis with up to 4 of your friends on the big screen!
# Notable Features

* Modern FPS gamepad controls (strafe/move with left stick, rotate with right stick)
* Smooth gamepad controls let you strafe, turn, and creep up slowly behind your enemies
* Classic co-op and 9 new game modes (Instagib, Nazi Zombies, Carry the Flag, Melee Rampage, Defuse the Bomb, and more)
* Graphic, zoomable minimap with all player locations
* Deployable MG42 machine gun
* Armored vehicle equipped with a mounted gatling gun spawns in areas with lots of enemies
* Gib animations when enemies are killed with gatling gun at close range or punched with fists in Rampage mode
* Support for [SDL_GameControllerDB](https://github.com/gabomdq/SDL_GameControllerDB) with controller mappings for all known game controllers
* Ability to remap all 13 supported gamepad buttons
* Support for SDL_GameControllerDB with controller mappings for all known game controllers
* Ability to set per-gamepad sensitivity settings (so you can turn faster without pressing run)
* Support enabling/disabling moving with mouse (player 1 only)
* Support mapping strafe left/right and weapon prev/next to keyboard keys

# Installation

Installation works identically to Wolf4SDL, with the same logic to check which game files are available. Binaries were compiled on Raspbian 8 (Jessie), so they work properly on old and new RetroPie setups alike.

Game files should be added to `roms/ports/wolf3d/` before installation. If not present, the install script will download the shareware versions of Wolf3D and Spear of Destiny.

# Support

For RetroPie-specific Splitwolf support and questions (RetroPie installation, game files, etc), post [on the RetroPie forums](https://retropie.org.uk/forum/category/8/help-and-support). 

For general Splitwolf support and questions (how to build, support for other platforms, map packs, etc), post on the [Team Raycast forums](http://raycast.teamforum.ru/viewforum.php?f=7).

If you already have Splitwolf up and running and you've found a bug (glitch with the game, crash while playing, etc), please report it using [the Splitwolf issue tracker](https://bitbucket.org/linuxwolf6/splitwolf/issues?status=new&status=open).

If you've found a bug with the RetroPie Setup installer, please report it using [the RetroPie-Setup issue tracker](https://github.com/RetroPie/RetroPie-Setup/issues).

**DO NOT USE ISSUE TRACKERS FOR GENERAL SUPPORT OR QUESTIONS, YOU WILL NOT BE HELPED!**

See the [Splitwolf wiki](https://bitbucket.org/linuxwolf6/splitwolf/wiki/Home) for additional details, compilation instructions, game mode instructions, etc.

## Credits

None of this would be possible without the hard work of the following people:

**Multiplayer framework, new game modes:** LinuxWolf

**Gamepad support, configuration UI, and RetroPie implementation:** lazd

**Additional art:** DoomJedi, Untrustable, Atina, PSTrooper, ArmanAhmadi

**Title Screen:** Atina

**Title Screen Font:** Tormentor667

**SDL Port**: Moritz "Ripper" Kroll (http://www.chaos-software.de.vu)

**Original Wolfenstein 3D:**  id Software (http://www.idsoftware.com)

### DIFF
--- a/scriptmodules/ports/splitwolf.sh
+++ b/scriptmodules/ports/splitwolf.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="splitwolf"
+rp_module_desc="SplitWolf - 2-4 player split-screen Wolfenstein 3D / Spear of Destiny"
+rp_module_help="Game File Extension: .wl6, .sod, .sd2, .sd3\n\nCopy your game files to $romdir/ports/wolf3d/\n\nIf you add new game files, run: sudo ~/RetroPie-Setup/retropie_packages.sh splitwolf configure"
+rp_module_licence="NONCOM https://bitbucket.org/linuxwolf6/splitwolf/src/master/license-mame.txt"
+rp_module_section="exp"
+
+function depends_splitwolf() {
+    getDepends libsdl2-dev libsdl2-mixer-dev
+}
+
+function sources_splitwolf() {
+    gitPullOrClone "$md_build" https://bitbucket.org/linuxwolf6/splitwolf.git
+}
+
+function _get_opts_splitwolf() {
+    echo 'splitwolf-wolf3d VERSION_WOLF3D_SHAREWARE=y' # shareware v1.4
+    echo 'splitwolf-wolf3d_apogee VERSION_WOLF3D_APOGEE=y' # 3d realms / apogee v1.4 full
+    echo 'splitwolf-wolf3d_full VERSION_WOLF3D=y' # gt / id / activision v1.4 full
+    echo 'splitwolf-sod VERSION_SPEAR=y' # spear of destiny
+    echo 'splitwolf-sodmp VERSION_SPEAR_MP=y' # spear of destiny mission packs
+    echo 'splitwolf-spear_demo VERSION_SPEAR_DEMO=y' # spear of destiny
+}
+
+function game_data_splitwolf() {
+    if [[ ! -d "$md_inst/bin/lwmp" ]]; then
+        # Get game assets
+        downloadAndExtract "https://bitbucket.org/linuxwolf6/splitwolf/downloads/lwmp.zip" "$md_inst/bin/"
+    fi
+}
+
+function add_games_splitwolf() {
+    declare -A games_wolf4sdl=(
+        ['vswap.wl1']="Splitwolf - Wolf 3D Demo"
+        ['vswap.wl6']="Splitwolf - Wolf 3D"
+        ['vswap.sod']="Splitwolf - Spear of Destiny Ep 1"
+        ['vswap.sd2']="Splitwolf - Spear of Destiny Ep 2"
+        ['vswap.sd3']="Splitwolf - Spear of Destiny Ep 3"
+        ['vswap.sdm']="Splitwolf - Spear of Destiny Demo"
+    )
+
+    add_ports_wolf4sdl "$md_inst/bin/splitwolf.sh %ROM%" "splitwolf"
+}
+
+
+function build_splitwolf() {
+    mkdir -p "bin"
+    local opt
+    while read -r opt; do
+        local bin="${opt%% *}"
+        local defs="${opt#* }"
+        make clean
+        make "$defs" DATADIR="$romdir/ports/wolf3d/"
+        mv "$bin" "bin/$bin"
+        md_ret_require+=("bin/$bin")
+    done < <(_get_opts_splitwolf)
+}
+
+function install_splitwolf() {
+    cp "$md_build/gamecontrollerdb.txt" "$md_build/bin/"
+    md_ret_files=('bin')
+}
+
+function configure_splitwolf() {
+    local game
+
+    mkRomDir "ports/wolf3d"
+
+    if [[ "$md_mode" == "install" ]]; then
+        game_data_splitwolf
+        game_data_wolf4sdl
+        cat > "$md_inst/bin/splitwolf.sh" << _EOF_
+#!/bin/bash
+
+function get_md5sum() {
+    local file="\$1"
+
+    [[ -n "\$file" ]] && md5sum "\$file" 2>/dev/null | cut -d" " -f1
+}
+
+function launch_splitwolf() {
+    local wad_file="\$1"
+    declare -A game_checksums=(
+        ['6efa079414b817c97db779cecfb081c9']="splitwolf-wolf3d"
+        ['a6d901dfb455dfac96db5e4705837cdb']="splitwolf-wolf3d_apogee"
+        ['b8ff4997461bafa5ef2a94c11f9de001']="splitwolf-wolf3d_full"
+        ['b1dac0a8786c7cdbb09331a4eba00652']="splitwolf-sod"
+        ['25d92ac0ba012a1e9335c747eb4ab177']="splitwolf-sodmp --mission 2"
+        ['94aeef7980ef640c448087f92be16d83']="splitwolf-sodmp --mission 3"
+        ['35afda760bea840b547d686a930322dc']="splitwolf-spear_demo"
+    )
+        if [[ "\${game_checksums[\$(get_md5sum \$wad_file)]}" ]] 2>/dev/null; then
+            $md_inst/bin/\${game_checksums[\$(get_md5sum \$wad_file)]} --splitdatadir $md_inst/bin/lwmp/ --split 2 --splitlayout 2x1
+        else
+            echo "Error: \$wad_file (md5: \$(get_md5sum \$wad_file)) is not a supported version"
+        fi
+}
+
+launch_splitwolf "\$1"
+_EOF_
+        chmod +x "$md_inst/bin/splitwolf.sh"
+    fi
+
+    add_games_splitwolf
+
+    moveConfigDir "$home/.splitwolf" "$md_conf_root/splitwolf"
+}
--- a/scriptmodules/ports/wolf4sdl.sh
+++ b/scriptmodules/ports/wolf4sdl.sh
@@ -28,24 +28,32 @@ function _get_opts_wolf4sdl() {
     echo 'wolf4sdl-3dr-v14 -DCARMACIZED' # 3d realms / apogee v1.4 full
     echo 'wolf4sdl-gt-v14 -DCARMACIZED -DGOODTIMES' # gt / id / activision v1.4 full
     echo 'wolf4sdl-spear -DCARMACIZED -DGOODTIMES -DSPEAR' # spear of destiny
+    echo 'wolf4sdl-spear-sw -DCARMACIZED -DSPEARDEMO -DSPEAR' # spear of destiny demo
 }
 
 function add_games_wolf4sdl() {
-    local cmd="$1"
-    declare -A games=(
+    declare -A -g games_wolf4sdl=(
         ['vswap.wl1']="Wolfenstein 3D demo"
         ['vswap.wl6']="Wolfenstein 3D"
         ['vswap.sd1']="Wolfenstein 3D - Spear of Destiny Ep 1"
         ['vswap.sd2']="Wolfenstein 3D - Spear of Destiny Ep 2"
         ['vswap.sd3']="Wolfenstein 3D - Spear of Destiny Ep 3"
+        ['vswap.sdm']="Wolfenstein 3D - Spear of Destiny Demo"
     )
+
+    add_ports_wolf4sdl "$md_inst/bin/wolf4sdl.sh %ROM%" "wolf3d"
+}
+
+function add_ports_wolf4sdl() {
+    local port="$2"
+    local cmd="$1"
     local game
     local wad
 
-    for game in "${!games[@]}"; do
+    for game in "${!games_wolf4sdl[@]}"; do
         wad="$romdir/ports/wolf3d/$game"
         if [[ -f "$wad" ]]; then
-            addPort "$md_id" "wolf3d" "${games[$game]}" "$cmd" "$wad"
+            addPort "$md_id" "$port" "${games_wolf4sdl[$game]}" "$cmd" "$wad"
         fi
     done
 }
@@ -78,6 +86,12 @@ function game_data_wolf4sdl() {
         # Get shareware game data
         downloadAndExtract "http://maniacsvault.net/ecwolf/files/shareware/wolf3d14.zip" "$romdir/ports/wolf3d" "-j -LL"
     fi
+    if [[ ! -f "$romdir/ports/wolf3d/vswap.sdm" && ! -f "$romdir/ports/wolf3d/vswap.sod" ]]; then
+        cd "$__tmpdir"
+        # Get shareware game data
+        downloadAndExtract "http://maniacsvault.net/ecwolf/files/shareware/soddemo.zip" "$romdir/ports/wolf3d" "-j -LL"
+    fi
+
     chown -R $user:$user "$romdir/ports/wolf3d"
 }
 
@@ -111,6 +125,7 @@ function launch_wolf4sdl() {
         ['b1dac0a8786c7cdbb09331a4eba00652']="wolf4sdl-spear --mission 1"
         ['25d92ac0ba012a1e9335c747eb4ab177']="wolf4sdl-spear --mission 2"
         ['94aeef7980ef640c448087f92be16d83']="wolf4sdl-spear --mission 3"
+        ['35afda760bea840b547d686a930322dc']="wolf4sdl-spear-sw"
     )
         if [[ "\${game_checksums[\$(get_md5sum \$wad_file)]}" ]] 2>/dev/null; then
             $md_inst/bin/\${game_checksums[\$(get_md5sum \$wad_file)]}
@@ -124,7 +139,7 @@ _EOF_
         chmod +x "$md_inst/bin/wolf4sdl.sh"
     fi
 
-    add_games_wolf4sdl "$md_inst/bin/wolf4sdl.sh %ROM%"
+    add_games_wolf4sdl
 
     moveConfigDir "$home/.wolf4sdl" "$md_conf_root/wolf3d"
 


joolswills:
Thanks. I think rather than duplicating a lot of the functionality this could be implemented reusing some of the code from the other module which is what we do elsewhere when modules share similar code.

lazd:
@joolswills glad to make that change if possible, can you point me to an example? I'll see what can be shared.

joolswills:
eg. amiberry uses some of uae4arm module code.

In regards to this module it should be able to reuse gamedata from the wolfsdl module as well as the code there.

DoomJedi:
Can I help this conversation? Member of SplitWolf team.

lazd:
@joolswills good call on the game data, I was thinking about that. I will make that change.

I will investigate what I can share between wolf4sdl's install script and this one and get back to you shortly.

lazd:
@joolswills ok, I've pushed a commit that uses the same ROM directory as wolf4sdl (`roms/wolf3d`) and am compiling binaries to test it now.

After investigating to see if there is anything to share between the wolf4sdl install script and this one, I found that the only function we could share directly is downloading game files (even dependencies are different as we're using SDL2 here). However, the existing wolf4sdl setup script does not compile a Spear of Destiny Demo binary, so it doesn't make sense for wolf4sdl to download Spear of Destiny Demo files.

Instead of modifying wolf4sdl in this pull request to support Spear of Destiny Demo, or modifying wolf4sdl's setup script to download game files it doesn't use, I propose that I send a subsequent pull request that does the following:

1. Uses [my SDL2 fork of wolf4sdl](https://github.com/lazd/wolf4sdl/tree/sdl2remap) that adds full game controller support
2. Builds a Spear of Destiny Demo binary
3. Shares `depends_wolf4sdl` with Splitwolf
4. Shares `game_data_wolf4sdl` with Splitwolf 

This subsequent pull request will have to come after the holidays, but I will definitely send it. What do you say?

lazd:
@joolswills, think we can get this in as-is and update Wolf4SDL later?

joolswills:
no sorry. I think the module needs work to less duplicate the functionality of the wolf4sdl module but I don't have time to help you do this currently. I may look into this at some point and re-implement the module myself so it shares functions.

lazd:
Ok, I have consolidated:

* `game_data_wolf4sdl` -  This means that wolf4sdl will download Spear of Destiny Demo game files, but does not use them
* `add_games_wolf4sdl` - associative array of game files and their names is passed

I cannot consolidate:

* `depends_splitwolf` - Splitwolf uses `libsdl2-dev libsdl2-mixer-dev`, wolf4sdl uses `libsdl1.2-dev libsdl-mixer1.2-dev`
* `install_bin_splitwolf` - wolf4sdl does not provide binary installation
* `sources_splitwolf` - Splitwolf uses different sources
* `build_wolf4sdl` and `_get_opts_splitwolf` - Splitwolf uses different make options and creates 2 additional binaries
* `install_splitwolf` - This performs operations that are Splitwolf-specific
* `configure_splitwolf` - There is enough different stuff going on here (different binaries and source files supported, different emulator entries to delete) that it is non-trivial to consolidate

@joolswills Is this enough? I can't think of anything else that is worth consolidating.

joolswills:
The game data could have parameters so it downloads only required files.

There should be no install_bin - there should be sources_ and install_ only. install_bin is for binary only packages which we try and avoid.

scriptmodules/ports/wolf4sdl.sh
```diff
@@ -31,14 +31,8 @@ function _get_opts_wolf4sdl() {
 }
 
 function add_games_wolf4sdl() {
+    eval "declare -A games="${2#*=}
```
joolswills: there is no need to do this - you can access games from this function if games is set in the parent with local scope.

scriptmodules/ports/wolf4sdl.sh
```diff
@@ -73,11 +67,21 @@ function game_data_wolf4sdl() {
     pushd "$romdir/ports/wolf3d"
     rename 'y/A-Z/a-z/' *
     popd
-    if [[ ! -f "$romdir/ports/wolf3d/vswap.wl6" && ! -f "$romdir/ports/wolf3d/vswap.wl1" ]]; then
+    if [[ ! -f "$romdir/ports/wolf3d/vswap.wl1" &&
+          ! -f "$romdir/ports/wolf3d/vswap.wl6"
+    ]]; then
```
joolswills: no need to reformat existing code as part of the addition.

lazd:
> There should be no install_bin - there should be sources_ and install_ only. install_bin is for binary only packages which we try and avoid.

We provide binaries because the Splitwolf build takes about 5 minutes on a 3B+ and 45 minutes on a Zero! You want us to remove the option to install from binary?

lazd:
> The game data could have parameters so it downloads only required files.

Instead of doing this, I'll just modify wolf4sdl to build the Spear of Destiny Demo so people can play it. Otherwise I'll just be adding code I'll remove when I fix wolf4sdl in my next PR. Agreed?

lazd:
@joolswills just pushed a commit that fixes formatting and updates wolf4sdl to create a Spear of Destiny Demo binary and link.

scriptmodules/ports/wolf4sdl.sh
lazd: I originally moved the `declare -A wolf3d_games` for the array up to the top of the `splitwolf.sh` and tried to use it as a global variable, but `add_games_wolf4sdl` inside of `wolf4sdl.sh` could not access it, so instead I passed it in and used this hack to define it because you cannot pass an associative array in bash.

If there is a better way, I'll gladly do it, but I am not aware.

scriptmodules/ports/splitwolf.sh
```diff
@@ -0,0 +1,124 @@
...
+    declare -A games=(
+        ['vswap.wl1']="Splitwolf - Wolf 3D Demo"
+        ['vswap.wl6']="Splitwolf - Wolf 3D"
+        ['vswap.sod']="Splitwolf - Spear of Destiny Ep 1"
+        ['vswap.sd2']="Splitwolf - Spear of Destiny Ep 2"
+        ['vswap.sd3']="Splitwolf - Spear of Destiny Ep 3"
+        ['vswap.sdm']="Splitwolf - Spear of Destiny Demo"
+    )
+
+    add_games_wolf4sdl "$md_inst/bin/splitwolf.sh %ROM%" "$(declare -p games)"
```
lazd: @joolswills this is where I pass `games` in `splitwolf.sh` -- if I defined it as `local` above, `add_games_wolf4sdl` would not be able to access it.

joolswills: it would be able to access it

bash local scope variables are usable in child functions - we use that for the $md_ variables which are passed to the modules - they are local in the parent.

joolswills:
>We provide binaries because the Splitwolf build takes about 5 minutes on a 3B+ and 45 minutes on a Zero! You want us to remove the option to install from binary?

Yes please.

joolswills:
>Instead of doing this, I'll just modify wolf4sdl to build the Spear of Destiny Demo so people can play it. Otherwise I'll just be adding code I'll remove when I fix wolf4sdl in my next PR. Agreed?

I think I understand - go ahead for now.

scriptmodules/ports/splitwolf.sh
lazd: No, that doesn't work. If i use `declare -A` at the top of `splitwolf.sh`, I cannot access it inside of `splitwolf.sh`'s functions functions nor can I access it inside of `wolf4sdl.sh`'s functions, even if I do `declare -g -A` (global).

Honestly, we are way outside of my knowledge when it comes to bash scripting. What you're describing does not seem possible, and I do not see another way to accomplish this. If you can tell me precisely the code change you want to see, I'll gladly do it, but I've been reading about this for an hour and I cannot come up with another way.

lazd:
>>We provide binaries because the Splitwolf build takes about 5 minutes on a 3B+ and 45 minutes on a Zero! You want us to remove the option to install from binary?

> Yes please.

@joolswills I don't understand. You can choose to install Splitwolf from source, it is not a binary-only release.

image

Many other packages provide binaries, including:

* lincity
* micropolis
* minecraft
* supertux

Why shouldn't Splitwolf provide a binary install option in addition to source? It needs to build 6 separate binaries, and 45 minutes is a long time to compile for people on less powerful Raspberry Pis.

psyke83:
@lazd,

Just to clarify re: binary installations, see: https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/packages.sh#L233

If you remove the custom install_bin_splitwolf function, binary installation can still occur via the generic installBin function (also in that script) which sources binaries that Jools builds from the RetroPie server. Binaries are provided for main and opt packages, so your scriptmodule would be covered.

lazd:
@psyke83, thanks for clarifying, I had no idea that was how it worked; I thought @joolswills was saying that binary installation would not be supported! I have removed the function.

@joolswills How are the binaries generated? I assume they will include the contents of `lwmp.zip` and the `gamecontrollerdb.txt` that need to be in the `bin/` directory for splitwolf to work.

psyke83:
The binary will include the entire contents of ```$md_inst``` that is produced after running the depends/sources/build/install (in a chroot environment). I can see from the current code that you move those files to the ```bin``` folder which are marked for installation to the module's installation directory, so as long as your script is building from source correctly, yes, they would be included in the generated binaries.

DoomJedi:
So...where this stands now?

lazd:
Hey @joolswills, do let me know if there is anything else I can do to get this ready to merge. Thanks!

joolswills:
Just the variable scope stuff I mentioned above.

scriptmodules/ports/splitwolf.sh
```diff
@@ -0,0 +1,119 @@
...
+
+function build_splitwolf() {
+    mkdir "bin"
+    local opt
+    while read -r opt; do
+        local bin="${opt%% *}"
+        local defs="${opt#* }"
+        make clean
+        make $defs DATADIR="$romdir/ports/wolf3d/"
+        mv $bin "bin/$bin"
```
joolswills: vars should be quoted.

scriptmodules/ports/splitwolf.sh
```diff
@@ -0,0 +1,119 @@
...
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="splitwolf"
+rp_module_desc="SplitWolf - 2-4 player split-screen Wolfenstein 3D / Spear of Destiny"
+rp_module_help="Game File Extension: .wl6, .sod, .sd2, .sd3\n\nCopy your game files to $romdir/ports/wolf3d/\n\nIf you add new game files, run: sudo ~/RetroPie-Setup/retropie_packages.sh splitwolf configure"
+rp_module_licence="NONCOM https://bitbucket.org/linuxwolf6/splitwolf/src/master/license-mame.txt"
+rp_module_section="opt"
+rp_module_flags="dispmanx !mali !kms"
```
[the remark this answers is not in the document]
joolswills: as it's sdl2 dispmanx should be removed also.

scriptmodules/ports/splitwolf.sh
```diff
@@ -0,0 +1,119 @@
...
+        ['vswap.sd2']="Splitwolf - Spear of Destiny Ep 2"
+        ['vswap.sd3']="Splitwolf - Spear of Destiny Ep 3"
+        ['vswap.sdm']="Splitwolf - Spear of Destiny Demo"
+    )
+
+    add_games_wolf4sdl "$md_inst/bin/splitwolf.sh %ROM%" "splitwolf" "$(declare -p games)"
+
+    moveConfigDir "$home/.splitwolf" "$md_conf_root/splitwolf"
+
+    setDispmanx "$md_id" 1
```
joolswills: this is an sdl1 only feature.

joolswills:
Went over the code again and noticed a few things. Thanks.

scriptmodules/ports/splitwolf.sh
```diff
@@ -0,0 +1,119 @@
...
+
+function depends_splitwolf() {
+    getDepends libsdl2-dev libsdl2-mixer-dev
+}
+
+function sources_splitwolf() {
+    gitPullOrClone "$md_build" https://bitbucket.org/linuxwolf6/splitwolf.git
+
+    if [[ ! -d "$md_build/lwmp" ]]; then
+        cd "$__tmpdir"
```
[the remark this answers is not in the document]
lazd: Good call. Also, I realize this entire routine should be moved to the configure script, next to `game_data_wolf4sdl`, as this is downloading the Splitwolf-specific assets (8-way sprites, weapon sprites, etc).

lazd:
@joolswills thanks for the thorough review, I have implemented all of your feedback and tested locally.

There was one issue I could not address, the issue of variable scoping for the `$games` variable.

> Regarding a reply I saw in email but I think you removed regarding variable scope. 

The reply you're referencing is here: https://github.com/RetroPie/RetroPie-Setup/pull/2532#discussion_r235552716

The issue here is that we need the associative array `$games` to be different between `splitwolf.sh` and `wolf4sdl.sh`, and we need to use it inside of the `add_games_wolf4sdl` function, which is declared in `wolf4sdl.sh` and also used by `splitwolf.sh`.

Now, from your replies, I gather that you do not want me to pass the variable around, and instead want the variable to be declared at the top of the file an accessed within a function.

So, if I declare it with a unique name,  `$wolf_games`, at the top of `wolf4sdl.sh` as follows:

```sh
declare -A -g wolf_games=(
    ['vswap.wl1']="Wolfenstein 3D demo"
    ['vswap.wl6']="Wolfenstein 3D"
    ['vswap.sd1']="Wolfenstein 3D - Spear of Destiny Ep 1"
    ['vswap.sd2']="Wolfenstein 3D - Spear of Destiny Ep 2"
    ['vswap.sd3']="Wolfenstein 3D - Spear of Destiny Ep 3"
    ['vswap.sdm']="Wolfenstein 3D - Spear of Destiny Demo"
)
```

Then, I can use it within `wolf4sdl.sh`'s `add_games_wolf4sdl`, and it works fine, as expected.

However, if I then declare `$wolf_games` inside of `splitwolf.sh` as follows:

```sh
declare -A -g wolf_games=(
    ['vswap.wl1']="Splitwolf - Wolf 3D Demo"
    ['vswap.wl6']="Splitwolf - Wolf 3D"
    ['vswap.sod']="Splitwolf - Spear of Destiny Ep 1"
    ['vswap.sd2']="Splitwolf - Spear of Destiny Ep 2"
    ['vswap.sd3']="Splitwolf - Spear of Destiny Ep 3"
    ['vswap.sdm']="Splitwolf - Spear of Destiny Demo"
)
```

It does not override the `$wolf_games` variable declared in `wolf4sdl.sh`, so when I run `add_games_wolf4sdl` from within `splitwolf.sh`, I incorrectly get the `$wolf_games` variable  for wolf4sdl.

The solution to this problem is to serialize `$games` and pass it as an argument to `add_games_wolf4sdl`, which then evaluates it, declaring it within the scope of the function. That way, `splitwolf.sh` can define its own, unique `$games` associative array and pass it to `wolf4sdl.sh`'s `add_games_wolf4sdl` function. This is what I have implemented, hence the `eval "declare -A games="${2#*=}` within `add_games_wolf4sdl`.

This is the only way I found to pass associative arrays between functions in Bash (source https://stackoverflow.com/a/8879444). If there is a better way to do this, I'll gladly implement it, but declaring the variable globally will not work.

joolswills:
I think you misunderstood. You declare it inside the parent function, as a local array. Then use it in the child function.

joolswills:
```diff
diff --git a/scriptmodules/ports/splitwolf.sh b/scriptmodules/ports/splitwolf.sh
index e3eac2d3..4b985482 100644
--- a/scriptmodules/ports/splitwolf.sh
+++ b/scriptmodules/ports/splitwolf.sh
@@ -111,7 +111,7 @@ _EOF_
         ['vswap.sdm']="Splitwolf - Spear of Destiny Demo"
     )
 
-    add_games_wolf4sdl "$md_inst/bin/splitwolf.sh %ROM%" "splitwolf" "$(declare -p games)"
+    add_games_wolf4sdl "$md_inst/bin/splitwolf.sh %ROM%" "splitwolf"
 
     moveConfigDir "$home/.splitwolf" "$md_conf_root/splitwolf"
 }
diff --git a/scriptmodules/ports/wolf4sdl.sh b/scriptmodules/ports/wolf4sdl.sh
index 58991531..49b8b122 100644
--- a/scriptmodules/ports/wolf4sdl.sh
+++ b/scriptmodules/ports/wolf4sdl.sh
@@ -32,7 +32,6 @@ function _get_opts_wolf4sdl() {
 }
 
 function add_games_wolf4sdl() {
-    eval "declare -A games="${3#*=}
     local port="$2"
     local cmd="$1"
     local game
@@ -136,7 +135,7 @@ _EOF_
         ['vswap.sdm']="Wolfenstein 3D - Spear of Destiny Demo"
     )
 
-    add_games_wolf4sdl "$md_inst/bin/wolf4sdl.sh %ROM%" "wolf3d" "$(declare -p games)"
+    add_games_wolf4sdl "$md_inst/bin/wolf4sdl.sh %ROM%" "wolf3d"
 
     moveConfigDir "$home/.wolf4sdl" "$md_conf_root/wolf3d"
 
```

joolswills:
In addition, the add_games functions need to be standalone, so both modules need that function that should contain the array declaration - another _add_games can be created that goes through the array.

lazd:
Ok, @joolswills, I've made the change. Each script now has its own `add_games_*` function that declares `games_wolf4sdl` globally, and there is a shared `add_ports_wolf4sdl` that iterates over the global variable `games_wolf4sdl` and adds the ports.

Is there anything else here for me to correct?

scriptmodules/ports/splitwolf.sh
```diff
@@ -0,0 +1,122 @@
...
+
+function configure_splitwolf() {
+    local game
+
+    mkRomDir "ports/wolf3d"
+
+    # remove obsolete emulator entries
+    while read game; do
+        delEmulator "${game%% *}" "splitwolf"
+    done < <(_get_opts_splitwolf)
```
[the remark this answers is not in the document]
lazd: If this code is unnecessary I can remove it from both wolf4sdl and splitwolf install scripts, just say the word.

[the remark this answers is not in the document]
joolswills: You don't need it then. It was added to wolf4sdl to remove entries from older versions of the module. Since this is a new  module it's not needed.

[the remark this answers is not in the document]
lazd: @joolswills done deal, I removed it from Splitwolf only.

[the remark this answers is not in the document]
joolswills: It's needed for wolf4sdl.

joolswills:
Please can you squash this PR and I will merge. Thanks.

lazd:
@joolswills squashed and ready to merge. Thanks for your thorough reviews and help getting this into shape.

A couple of questions:

1. How should users update RetroPie-Setup to get the new script?
2. When will a binary install be available?

joolswills:
They can choose "update retropie-setup" to just have it appear. Binaries won't be available for exp modules, but I am happy to move it later and build binaries etc. Could open a thread on the forum for user testing and get some feedback.

joolswills:
Merged - thanks for your hard work on this.

DoomJedi:
Couldn't find SplitWolf in latest retropie...did I miss it? If not, why was it removed?

joolswills:
@DoomJedi It's in the experimental section - but please use the forum for support.

DoomJedi:
Thank you, glad to hear, will check there. Sorry for bumping this thread.

PR #2533, opened 2018-11-16
Create lr-kronos.sh

barbudreadmon:
see https://retropie.org.uk/forum/topic/19930/new-sega-saturn-libretro-core-kronos


joolswills:
thanks.

PR #2534, opened 2018-11-18
added lr-freechaf module

recompileorg:
This is for the FreeChaF ChannelF libretro core

### DIFF
--- a/scriptmodules/libretrocores/lr-freechaf.sh
+++ b/scriptmodules/libretrocores/lr-freechaf.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. On phone so haven't tested. Looks generally ok but we will need a platforms.cfg entry. I will test and review asap.

recompileorg:
>  Looks generally ok but we will need a platforms.cfg entry.

Something like this?
```
channelf_exts=".bin .rom"
channelf_fullname="Fairchild ChannelF"
channelf_platform="channelf"
```

recompileorg:
Thanks.  I've added the entry to platforms.cfg with the changes you suggested.

joolswills:
Sorry - you have included unrelated changes in platforms.cfg now.

I will add the module manually.

PR #2537, opened 2018-11-20
runcommand - add support for rom directories with extension in name

hhromic:
This is a working and tested draft of the idea proposed in the forum for supporting direct launching of game directories in lr-scummvm from EmulationStation.
The changes are minimal and unobtrusive (I hope) so should be easy to review and shouldn't affect any existing behaviour.

**Relevant forum posts:**
https://retropie.org.uk/forum/post/168019
https://retropie.org.uk/forum/post/168309
https://retropie.org.uk/forum/post/168365

**Summary:**
If the given rom is a directory and has an extension in the name, collect files inside the rom directory with that same extension, then if there is exactly one file inside the rom directory with the same extension as the rom directory, use it for the %ROM% token to be replaced.
This behaviour is useful for example in the lr-scummvm emulator to support launching games directly when selecting the game directory in EmulationStation if it has a supported system extension.

**Tests:**
* runcommand correctly launches all existing file roms for arcade, nes, snes, megadrive, etc [existing behaviour].
* runcommand correctly launches scummvm games without an extension in the directory name, giving the svm file name inside [existing behaviour].
* runcommand correctly launches scummvm games with an extension in the directory, giving just the directory name [new added behaviour].
* game art is correctly displayed for all the above cases [$ROM_BN is still correctly assigned].

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -149,6 +149,8 @@ function get_params() {
             ROM="$4"
             ROM_BN_EXT="${ROM##*/}"
             ROM_BN="${ROM_BN_EXT%.*}"
+            ROM_EXT="${ROM_BN_EXT##*.}"
+            [[ "$ROM_BN" == "$ROM_EXT" ]] && ROM_EXT=""  # no extension case
             if [[ "$COMMAND" == "_PORT_" ]]; then
                 CONF_ROOT="$CONFIGDIR/ports/$SYSTEM"
                 EMU_SYS_CONF="$CONF_ROOT/emulators.cfg"
@@ -876,8 +878,21 @@ function get_sys_command() {
 
     COMMAND="$(default_emulator get emu_cmd)"
 
+    # if rom is a directory and has an extension in the name, collect
+    # files inside the rom directory with that same extension
+    local -a romdir_files
+    if [[ -d "$ROM" && -n "$ROM_EXT" ]]; then
+        mapfile -t romdir_files < <(compgen -G "$ROM/*.$ROM_EXT")
+    fi
+
     # replace tokens
-    COMMAND="${COMMAND//\%ROM\%/\"$ROM\"}"
+    if [[ "${#romdir_files[@]}" -eq 1 ]]; then
+        # if there is exactly one file inside the rom directory with
+        # the same extension as the rom directory, use it for %ROM%
+        COMMAND="${COMMAND//\%ROM\%/\"${romdir_files[0]}\"}"
+    else
+        COMMAND="${COMMAND//\%ROM\%/\"$ROM\"}"
+    fi
     COMMAND="${COMMAND//\%BASENAME\%/\"$ROM_BN\"}"
 
     # special case to get the last 2 folders for quake games for the -game parameter


joolswills:
was hoping to discuss before creating a PR as I am not sure I want to implement this (and not sure I want to implement it in this way).

I am not clear on the 3rd case

>runcommand correctly launches scummvm games with an extension in the directory, giving just the directory name [new added behaviour].

Please can you detail how this works via ES. Since with ES you only select files not directories ?

joolswills:
(Giving an example of the commandline before and after so I am clear what this is resolving - I may prefer to resolve this with a launch script and not via runcommand, since I am not sure what else benefits from it currently).

joolswills:
I think I understand now - ES will pass the folder name if the folder has an extension ?

Looking through the code changes, I am not sure I agree with them being minimal and unobtrusive - looks overly complicated to me. My current feeling would be to do this outside of ES via a launch script.

hhromic:
Yes, exactly! ES considers the folder as a ROM by itself if the folder name has a supported system extension, like any regular ROM file, and will pass this to runcommand on click instead of entering the folder. This is the feature I'm looking to exploit to make launching lr-scummvm games more user-friendly.

The current approach I have in the lr-scummvm experimental scriptmodule is to create a custom cmd for `addEmulator` that appends `/*.svm` to the `%ROM%` token (as I described in the forum post https://retropie.org.uk/forum/post/168309). This doesn't need to change runcommand but forces the user to always follow the convention of naming their game folders with the `.svm` extension (otherwise games won't work) and also depends on shell globbing (not 100% of my liking). An alternative to globbing is to force a particular name for the inner file, such as `%ROM%/gameid.svm`, however this might not be of the liking of users, specially those also using the standalone scummvm.

In this PR, instead the idea is to make runcommand aware of rom directories with a file inside to set `%ROM%` accordingly and transparently. I do agree that this is an arbitrary convention that currently adjusts to lr-scummvm. The benefit is that `addEmulator` doesn't need any special cmd hence less "hackish". Also games would still work for users that do not want to rename their game folders.

The usage of a launch shell script in my opinion is a bit cumbersome because the shell script will need to wrap the call to retroarch and might loose the runcommand features for libretro cores such as verbose logging, render resolution and config append. Also is an extra thing to add to the scriptmodule. I definitively don't prefer this route.

Nonetheless, after re-thinking again about all this, maybe the current approach I have (the custom cmd in the scriptmodule) is sufficient if you are okay with using `%ROM%/*.svm` (the globbing) and if we are okay with forcing users to always use the `.svm` extension in their folder names. My only big concern are users that do not read the documentation and will be clueless why the games don't work as expected without the folder extension.

What do you think?

hhromic:
Forgot to add examples of commandline with this PR applied:

Current behaviour with a megadrive rom:
```
megadrive
lr-genesis-plus-gx
/home/pi/RetroPie/roms/megadrive/Sonic the Hedgehog 2 (World).zip
/opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-genesis-plus-gx/genesis_plus_gx_libretro.so --config /opt/retropie/configs/megadrive/retroarch.cfg "/home/pi/RetroPie/roms/megadrive/Sonic the Hedgehog 2 (World).zip"
```

Current behaviour with a scummvm game folder without an extension, in this case in ES you have to go inside the folder and select the `sky.svm` file inside (note runcommand again doesn't change %ROM%):
```
scummvm
lr-scummvm
/home/pi/RetroPie/roms/scummvm/Beneath a Steel Sky/sky.svm
/opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-scummvm/scummvm_libretro.so --config /opt/retropie/configs/scummvm/retroarch.cfg "/home/pi/RetroPie/roms/scummvm/Beneath a Steel Sky/sky.svm"
```

New behaviour with a scummvm game folder having supported extension (note how runcommand automatically added `atlantis.svm` to %ROM% as needed by lr-scummvm):
```
scummvm
lr-scummvm
/home/pi/RetroPie/roms/scummvm/Indiana Jones and the Fate of Atlantis.svm
/opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-scummvm/scummvm_libretro.so --config /opt/retropie/configs/scummvm/retroarch.cfg "/home/pi/RetroPie/roms/scummvm/Indiana Jones and the Fate of Atlantis.svm/atlantis.svm"
```

Also I saw there is already a special case consideration for quake games inside runcommand that is of the same type of handling for a corner case, hence I didn't think my proposal is so out of place:
```sh
    # special case to get the last 2 folders for quake games for the -game parameter
    # remove everything up to /quake/
    local quake_dir="${ROM##*/quake/}"
    # remove filename
    quake_dir="${quake_dir%/*}"
    COMMAND="${COMMAND//\%QUAKEDIR\%/\"$quake_dir\"}"
```

joolswills:
although the quakedir case is used by more than one module, and is less of a workaround imho (Just another runcommand variable). I am not yet convinced as I think it would be simpler with a launch script.

I don't think the launch script would lose features. It would just adjust the commandline passed to runcommand so if a directory was chosen, it would look for the file inside and if it exists, it would pass it to runcommand.

This also ensures no interaction with anything else that we have overlooked.

hhromic:
Thanks for the comments, let me study more closely the idea of a launch script and how it can be done to keep the features for LR cores. I think runcommand relies on string matching over the command to run to enable/disable things. Will come back with a conclusion after testing in the coming days.

hhromic:
Hi @joolswills , sorry for the long delay coming back to this. I have been experimenting on the idea of using a launch script as suggested, however I think that the `addEmulator` function has a limitation that prevents on achieving what we want to do. Let me explain. So far, the idea I'm working on is this:

1. Create a launch script wrapper to be used as the "emulator" for the scummvm system. This launch script's job is to check if the passed %ROM% is a directory and find the required file inside. If so, then launch retroarch with the appropiate rom path. Let's name this script `scummvm-retroarch.sh`. I chose this name with `retroarch-` to make sure runcommand's `retroarch_append_config` function works properly.
2. Then in the scriptmodule, during the configure phase, create an emulator entry for the scummvm system using `addEmulator` with the following parameters: `addEmulator 0 "$md_id" "scummvm" "/path/to/scummvm-retroarch.sh %ROM%"`.

However, this doesn't work as expected because `addEmulator` has the following functionality:
```sh
    # automatically add parameters for libretro modules
    if [[ "$id" == lr-* && "$cmd" != "$emudir/retroarch/bin/retroarch"* ]]; then
        cmd="$emudir/retroarch/bin/retroarch -L $cmd --config $md_conf_root/$system/retroarch.cfg %ROM%"
    fi
```
And therefore ends up writing this to the `emulators.cfg` file:
```
lr-scummvm = "/opt/retropie/emulators/retroarch/bin/retroarch -L /path/to/scummvm-retroarch.sh %ROM% --config /opt/retropie/configs/scummvm/retroarch.cfg %ROM%"
```
Which obviously is not correct. The problem is that `lr-*` modules are expected to be libretro core libraries and nothing else like launch scripts. I'm not sure if this is intentional or an overlook.

On the other hand, I understand the process of launching a ROM from EmulationStation is this:
1. From `es_systems.cfg` get the command for the given system. In this case: `/opt/retropie/supplementary/runcommand/runcommand.sh 0 _SYS_ scummvm %ROM%`
2. From there call runcommand with the selected ROM (file or directory)
3. Runcommand retrieves the emulator cmd from `emulators.cfg` and executes it accordingly. Here is where the launch script would come into play.

However, you said:
> (...) It would just adjust the commandline passed to runcommand so if a directory was chosen, it would look for the file inside and if it exists, it would pass it to runcommand.

And therefore I got confused, in which stage the launch script should adjust the commandline to pass to runcommand?  We could use `addSystem` to change how EmulationStation launches runcommand but this would break any other emulator for the same system, e.g. standalone scummvm.

I believe the approach I'm working on (setting the launch script in `emulators.cfg` through `addEmulator`) is the right way to proceed as the launcher is just for this particular emulator (lr-scummvm), however `addEmulator` is preventing me to configure the emulator line correctly. Perhaps, I'm missing something?

Apologies for the long post and looking forward for your feedback :)

hhromic:
Just to add a bit of more information, I went to check what scriptmodules benefit/use the `addEmulator` snippet above (id ~= lr-* && cmd != retroarch), and couldn't find any (no output below):

    $ grep addEmulator -r scriptmodules/libretrocores | grep retroarch
    $ grep addEmulator -r scriptmodules | grep retroarch

Perhaps the check `"$cmd" != "$emudir/retroarch/bin/retroarch"*` needs to be improved to detect libretro cores instead of the retroarch binary?

hhromic:
Ok I think I have a better idea. How about using something like `romdir-lr-scummvm` as the emulator id for `addEmulator`? That would allow me to use any cmdline, e.g. the wrapper launch script. In this case, it would create *an emulator that launches lr-scummvm using romdirs or romfiles* and that would even be selectable by the end-user.

What do you think ? Sorry for all the noise :)

hhromic:
So, yes, the above idea worked as expected :).
I now successfully created a launcher script for lr-scummvm that tries to find svm files inside a passed rom directory and, if found, launches retroarch with the right rom file, outside runcommand. It looks like this (the paths are automatically set from $emudir, $md_inst, etc. in the scriptmodule during creation):
```sh
#!/usr/bin/env bash
ROM=$1; shift
SVM_FILES=()
[[ -d $ROM ]] && mapfile -t SVM_FILES < <(compgen -G "$ROM/*.svm")
[[ ${#SVM_FILES[@]} -eq 1 ]] && ROM=${SVM_FILES[0]}
/opt/retropie/emulators/retroarch/bin/retroarch \
    -L "/opt/retropie/libretrocores/lr-scummvm/scummvm_libretro.so" \
    --config "/opt/retropie/configs/scummvm/retroarch.cfg" \
    "$ROM" "$@"
```
This launch script must contain `retroarch` in the name for runcommand to support all RetroArch features such as appendConfig and verbose logging. Launch art also works as expected, as runcommand only sees the rom directory name and not any internal files.

Then, finally all is glued together with this invocation of `addEmulator` (the emulator id doesn't start with `lr-` so the function doesn't rewrite the cmd):
```sh
addEmulator 0 "romdir-$md_id" "scummvm" "$md_inst/retroarch-romdir.sh %ROM%"
```

In my tests all seems to be working fine without modifying runcommand, so yes, this PR is no longer necessary (close it as you need). I will update the pending lr-scummvm scriptmodule now with this and submit the PR soon for reviewing.

joolswills:
Sorry I have been busy, but your solution is along the lines of what I had in mind (in fact your solution is arguably better as mine would have required some changes to the helpers as I had overlooked some details (I was thinking to do like you have, but call runcommand from the script, but then runcommand would have been launched twice (and it would have required not using runcommand in the es config). A bit like how the ports work (which doesn't use runcommand via es).

hhromic:
Exactly, indeed I wanted to avoid calling runcommand multiple times and instead follow a straight-flat chain. I also think that runcommand is capable enough to be used for almost everything, and should be used as the entry point from ES whenever possible.
I will close this PR now, thanks for the discussion and sorry again for the noise.

PR #2539, opened 2018-11-22
Adding .chd to Saturn supported extensions

raelgc:
`lr-beetle-saturn` supports `.chd` images, so adding it to the list of supported files.


joolswills:
Thanks.

PR #2540, opened 2018-11-22
Fixes and optimizations for the Skyscraper scriptmodule

cmitu:
Based on the conversation on the previous `skyscraper` PR by @muldjord and @hhromic.

* Fix the editing in advanced configuration by providing STDOUT/STDIN for console editors (e.g nano/vim) and running the editor as `$user`.
* Disable output buffering during the scraping session (
* Use the Skyscraper shipped configuration file (config.ini.example) as a started config file.
* Add some optimizations suggested in the previous PR #2520 by @hhromic.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -167,25 +167,6 @@ function configure_skyscraper() {
 function _init_config_skyscraper() {
     local md_conf_dir="$configdir/all/skyscraper"
 
-    cat > "$md_conf_dir/config.ini.rp" << _EOF_
-[main]
-frontend="emulationstation"
-dbFolder="$md_conf_dir/dbs"
-importFolder="$md_conf_dir/import"
-inputFolder="$romdir"
-unattend="true"
-unattendSkip="true"
-noHints="false"
-threads="2"
-lang="en"
-region="wor"
-videos="false"
-marquees="true"
-wheels="true"
-covers="true"
-screenshots="true"
-_EOF_
-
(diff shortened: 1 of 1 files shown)


hhromic:
LGTM, just wondering why the indent between the `trap`s ?  Trap is just a statement in bash, not a control block like `if`, `while` or `for`.
```sh
    # trap ctrl+c and return if pressed (rather than exiting retropie-setup etc)
    trap 'trap 2; return 1' INT
        sudo -u "$user" "$md_inst/Skyscraper" "${params[@]}"
        sudo -u "$user" stdbuf -o0  "$md_inst/Skyscraper" "${params[@]}"
        echo -e "\nCOMMAND LINE USED:\n $md_inst/Skyscraper" "${params[@]}"
        sleep 2
    trap 2
```

hhromic:
Btw I like your coding style, the associative arrays for the help strings are very elegant :)

hhromic:
@cmitu makes sense. My humble suggestion would be to align the indent to the traps, like it is in the `scraper.sh` module too. Sorry to be a nuisance :)

joolswills:
thanks.

muldjord:
Will test this out in the coming week. Thank you guys.

muldjord:
Been testing it a bit today. Found one error. The script needs to also copy the "tgdb_*.json" files to the ~/.skyscraper (actually the /opt/retropie... folder, you know what I mean) folder as they are used by the thegamesdb module for identifying Developer and Publisher from their ID's. Otherwise Developer and Publisher will be blank when using the 'thegamesdb' module.

Other than that, it seems to be working well now.

muldjord:
Tested, works well now.

PR #2541, opened 2018-11-26
Couple of minor fixes

movisman:
Fix Debian link for FS-UAE script
Fix missing aspect ratio on advanced config edit


joolswills:
Thanks but the commit messages are not suitable (questions to me should not be in commit messages). I will sort these issues.

movisman:
Hi,

No worries, apologies and thanks in advance.

Cheers

Issue #2524, opened 2018-10-30
CTRL+C during scraping may trigger the exit of the main script later on


cmitu:
(Started from https://github.com/RetroPie/RetroPie-Setup/pull/2520#issuecomment-433611505). 

I've noticed that using Ctrl+C while scraping, either using `scraper` or `skyscraper`, has the side effect of exiting the RetroPie-Setup script after choosing any subsequent option that produces any output. The script exits with exit code 141 (SIGPIPE), seemingly because of the `tee` command that pipes the output to the RetroPie-Setup log.

Steps to reproduce (using `scraper`, since it's fewer clicks):
1. Open the RetroPie-Setup script, then open the `scraper` module (Configuration/Tools).
2. Start a scraping session via `Scrape all system` or `Scrape chosen system`. 
3. Interrupt the scraping with Ctrl+C. The script correctly shows that _Scraping was aborted_ and returns to the `scraper` selection GUI.
4. Choose `Upgrade scraper to the last version` (last option in the GUI). The RetroPie-Setup script exits. (**$?=141**)

The problem doesn't appear when running the `scraper` gui with ` sudo ./retropie_packages.sh scraper gui`.

The problem seems to be in 
https://github.com/RetroPie/RetroPie-Setup/blob/2b7e724dbe8d2a24a34a7dcf10870d2dcf4dc669/scriptmodules/admin/setup.sh#L373-L384
`tee` catches somehow the `SIGINT` and stops, then any output from the `rp_callModule "$choice" gui` call will trigger a SIGPIPE in the `gzip` call and ends the script.

A possible fix would be to replace the `tee` call with 
``` sh
{
# call anything here
} | (tee -i >(gzip --stdout >"$logfilename"))
```
which doesn't trigger the script to exit, but still mangles the next gui command that has any output (after which GUI options are working fine)

Do you think this warrants more investigation/time or should be something that belongs to the scriptmodule to handle ?

joolswills:
I will look into it.

hhromic:
Hi guys, I got entertained by this interesting puzzle, and I think I found out the problem.
[...]

PR #2545, opened 2018-11-27
setup - stop ctrl+c signal from reaching tee / gzip and exiting retropie-setup - #2524

joolswills:
* use setsid for tee / gzip so they always complete
 * https://github.com/RetroPie/RetroPie-Setup/issues/2524

My commit message is a bit crap but here's a proposed fix based on the discussion. Only difference is I put it in a function to avoid duplication.


joolswills:
ok done. Thanks for the feedback.

PR #2546, opened 2018-11-27
lr-mupen64plus - add patch from @psyke83 to fix broken build and refresh existing patch

### LINKED ISSUE #2544: lr-mupen64plus is not building in current master
Hi @joolswills , since a while now `lr-mupen64plus` is not building for RPI in current master.

The reason is that the Makefile broke again in a recent PR. This is the same problem that you addressed before in PR https://github.com/libretro/mupen64plus-libretro/pull/73.

As you know, there is a long-standing PR from @psyke83 that addresses old libnames being used and he recently updated it to also fix the build again: https://github.com/libretro/mupen64plus-libretro/pull/39
However, the upstream maintainers seem to be away for the moment and the PR hasn't been merged.

I think there are three courses of action for this:
1) Locally patch the Makefile to fix the build, i.e. applying @psyke83 's PR from the scriptmodule.
2) Set the build to a known-to-be-good commit (similar to what you did in `lr-parallel-n64` recently). In this case the commit would be the one before the PR that broke the build: `bb64262862`.
3) Try to poke again the upstream maintainers to merge the PR from @psyke83 (ideal solution). Perhaps you have more influence over this than me :)

hhromic:
Tested and building fine now. Fixes #2544

### DIFF
--- a/scriptmodules/libretrocores/lr-mupen64plus.sh
+++ b/scriptmodules/libretrocores/lr-mupen64plus.sh
@@ -39,8 +39,9 @@ function depends_lr-mupen64plus() {
 
 function sources_lr-mupen64plus() {
     gitPullOrClone "$md_build" https://github.com/libretro/mupen64plus-libretro.git
-    # needed until https://github.com/libretro/mupen64plus-libretro/pull/39 is accepted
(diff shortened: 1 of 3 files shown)


joolswills:
thanks.

PR #2551, opened 2018-12-04
add ScummVM port for libretro (lr-scummvm) to experimental

hhromic:
Finally I think I'm happy for this scriptmodule to be merged in RetroPie.
It has been tested for a while by various users in the forum and everyone seems happy with it.

**Useful Links**

* Forum Thread: <https://retropie.org.uk/forum/topic/19693/lr-scummvm-request-for-comments-and-testing>
* Libretro Docs for Scummvm: <https://docs.libretro.com/library/scummvm/>
* Default Joypad Mapping (I will add this info to the RetroPie wiki after merging): <https://docs.libretro.com/library/scummvm/#joypad>
* An evolving RetroArch-specific compatibility list (a lot of games are working fine) is being maintained in: <https://docs.google.com/spreadsheets/d/1a87BnQJc7_fTD5zF55JXcm79d5_l_d1ghDxtiubTnDE>

**Module Features**

* The scriptmodule enables MT-32 emulation during compilation.
* The scriptmodule automatically downloads and installs necessary extra data for the core from the official libretro repository, e.g. soundfont, game data, default theme, etc.
* The scriptmodule automatically configures the core to be usable out-of-the-box with standard settings, e.g. enable fluidsynth and the SC-55 soundfont, enable subtitles, etc.
* A patch is included to fix some cases where trying to exit directly, e.g. using HOTKEY+START, freezes RetroArch (not the system). This patch will be submitted upstream as well.
* Support for directly launching scummvm games from EmulationStation using the game directory instead of entering the directory itself (just suffix the game directory name with `.svm`).
* Games do not need to be added using the ScummVM GUI, just uncompressed into their own directories in `roms/scummvm` and work out of the box.

### DIFF
--- a/scriptmodules/libretrocores/lr-scummvm.sh
+++ b/scriptmodules/libretrocores/lr-scummvm.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-scummvm"
+rp_module_desc="ScummVM port for libretro"
+rp_module_help="Copy your ScummVM games to $romdir/scummvm\n\nThe name of your game directories must be suffixed with '.svm' for direct launch in EmulationStation."
+rp_module_licence="GPL2 https://raw.githubusercontent.com/libretro/scummvm/master/COPYING"
+rp_module_section="exp"
+
+function sources_lr-scummvm() {
+    gitPullOrClone "$md_build" https://github.com/libretro/scummvm.git
+    # fix random freeze-on-quit (remove when patched upstream)
+    applyPatch "$md_data/01_fix_quit.diff"
+}
+
+function build_lr-scummvm() {
+    make clean -C backends/platform/libretro/build
+
+    local params=(HAVE_MT32EMU=1)
+    local platform
+    isPlatform "arm" && platform+="armv"
+    isPlatform "neon" && platform+="neon"
+    [[ -n "$platform" ]] && params+=(platform="$platform")
+
+    CXXFLAGS="$CXXFLAGS -DHAVE_POSIX_MEMALIGN=1" LDFLAGS="$LDFLAGS -shared -Wl,--no-undefined" \
+        make "${params[@]}" -C backends/platform/libretro/build
+    md_ret_require="$md_build/backends/platform/libretro/build/scummvm_libretro.so"
+}
+
+function install_lr-scummvm() {
+    md_ret_files=(
+        "backends/platform/libretro/build/scummvm_libretro.so"
+        "COPYING"
+    )
+}
+
+function configure_lr-scummvm() {
+    addEmulator 0 "$md_id" "scummvm" "$md_inst/romdir-launcher.sh %ROM%"
+    addSystem "scummvm"
+    [[ "$md_mode" == "remove" ]] && return
+
+    # ensure rom dir and system retroconfig
+    mkRomDir "scummvm"
+    ensureSystemretroconfig "scummvm"
+
+    # download and extract auxiliary data (theme, extra)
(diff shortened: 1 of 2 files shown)


scriptmodules/libretrocores/lr-scummvm.sh
```diff
@@ -0,0 +1,89 @@
...
+
+function install_lr-scummvm() {
+    md_ret_files=(
+        "backends/platform/libretro/build/scummvm_libretro.so"
+        "COPYING"
+    )
+}
+
+function configure_lr-scummvm() {
+    addEmulator 0 "romdir-$md_id" "scummvm" "$md_inst/retroarch-romdir.sh %ROM%"
```
joolswills: why romdir-$md_id ? $md_id should suffice.

scriptmodules/libretrocores/lr-scummvm.sh
```diff
@@ -0,0 +1,89 @@
...
+        iniSet "soundfont" "$biosdir/scummvm/extra/Roland_SC-55.sf2"
+        iniSet "gui_theme" "scummmodern"
+        iniSet "subtitles" "true"
+        iniSet "multi_midi" "true"
+        iniSet "gm_device" "fluidsynth"
+        chown $user:$user "$biosdir/scummvm.ini"
+    fi
+
+    # enable speed hack core option if running in arm platform
+    isPlatform "arm" && setRetroArchCoreOption "scummvm_speed_hack" "enabled"
```
joolswills: is this needed for all arm platforms ? Maybe better for just armv6 or so ? I assume there are negative effects or it would be a default in the libretro core ?

scriptmodules/libretrocores/lr-scummvm.sh
hhromic: No, as noted in https://github.com/RetroPie/RetroPie-Setup/pull/2537#issuecomment-443906376, the `addEmulator` function has the shortcoming that, when $md_id starts with `lr-`, then it overwrites the cmd passed with a static retroarch invocation. The affecting conditional is this:
```sh
    # automatically add parameters for libretro modules
    if [[ "$id" == lr-* && "$cmd" != "$emudir/retroarch/bin/retroarch"* ]]; then
        cmd="$emudir/retroarch/bin/retroarch -L $cmd --config $md_conf_root/$system/retroarch.cfg %ROM%"
    fi
```

I noticed that no scriptmodule benefits from the second part of the conditional, so maybe the conditional can be improved to check for a passed libretro core (`*.so`) instead of `retroarch`. If you agree with that, I actually would prefer that solution, I don't like the id name :)

Note that for `runcommand` to recognise a retroarch core launching, the name fo the cmd needs to have `retroarch` in its name.

scriptmodules/libretrocores/lr-scummvm.sh
hhromic: This is recommended by the lr-scummvm upstream docs (<https://docs.libretro.com/library/scummvm/>), hence I just followed it.

Their reasoning is this:

> This hack is considered 'safe' - games should run correctly when it is enabled, and most inaccuracies are imperceptible other than in a few edge cases. It remains a hack, though, and it is strongly recommended that users of desktop class machines keep it disabled.
> 
> However: For users of low power hardware (Android devices, single board computers), this hack is essentially mandatory for full speed operation of the core. For Android users in particular, the guides in the 'Usage' section of this document assume that the speed hack is enabled.

I myself and the testers in the forum has not noticed any problem with it.

scriptmodules/libretrocores/lr-scummvm.sh
joolswills: I'll come up with something.

scriptmodules/libretrocores/lr-scummvm.sh
joolswills: but does the core run at full speed on a rpi3 without it ? if so for example, we may want to leave it disabled.

hhromic: I will try with my set of games and come up with a conclusion. I didn't try further as they emphasize: *However: For users of low power hardware (Android devices, single board computers), this hack is essentially mandatory for full speed operation of the core.*

hhromic:
I updated the PR for the new changes in PR #2555 .
It looks way better now with a proper emulator id and also the launcher script name.
Thanks for the improvements!

joolswills:
Thanks.

PR #2552, opened 2018-12-04
steamlink: add overscan override

psyke83:
The steamlink client will refuse to launch when overscan is enabled
(which is enabled by default on our RetroPie image), but an override
is supported.

(they added this overscan detection to their scripts after I submitted the initial PR)

### DIFF
--- a/scriptmodules/ports/steamlink.sh
+++ b/scriptmodules/ports/steamlink.sh
@@ -14,14 +14,14 @@ rp_module_desc="Steam Link for Raspberry Pi 3"
 rp_module_licence="PROP https://steamcommunity.com/app/353380/discussions/0/1743353164093954254/"
 rp_module_section="exp"
 rp_module_flags="!mali !x86 !kms !rpi1 !rpi2"
-rp_module_help="Streaming games to your computer with Steam."
+rp_module_help="Streaming games to your computer with Steam.\n\nIf you experience a black screen during startup, you may need to disable the overscan_scale setting in your Pi's /boot/config.txt."
 
 function depends_steamlink() {
     getDepends python3-dev curl xz-utils
 }
 
 function install_bin_steamlink() {
-    local ver="1.0.4"
+    local ver="1.0.5"
     local url="http://media.steampowered.com/steamlink/rpi"
     wget -qO "$__tmpdir/steamlink_"$ver"_armhf.deb" ""$url"/steamlink_"$ver"_armhf.deb"
     dpkg -i "$__tmpdir/steamlink_"$ver"_armhf.deb"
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
.deb file has updated to 1.0.5 too, btw. Maybe it'd be worth asking them nicely for a way to retrieve the latest .deb from a constant link, so setup script doesn't fall behind? They seem to be very responsive on that matter thus far.

scriptmodules/ports/steamlink.sh
```diff
@@ -33,4 +33,12 @@ function remove_steamlink() {
 
 function configure_steamlink() {
     addPort "$md_id" "steamlink" "Steam Link" "/usr/bin/steamlink"
+
+    # RetroPie sets overscan by default, which requires an override
+    if egrep '^overscan_scale=1' /boot/config.txt >/dev/null 2>&1; then
```
joolswills: I don't think grep -E is needed.

also `&>/dev/null` would be more concise in bash.

scriptmodules/ports/steamlink.sh
```diff
@@ -33,4 +33,12 @@ function remove_steamlink() {
...
+        sudo -u $user bash -c "\
+            mkdir -p "~/.local/share/SteamLink/"
+            touch "~/.local/share/SteamLink/.ignore_overscan"
+        "
```
joolswills: mkUserDir and touch/chown to match rest of code please.

joolswills: actually mkUserDir isn't ideal, but we don't use sudo like this in other code, so it would be best to stick with the existing style (mkUserDir .local / .share etc - it's longwinded but perhaps we could improve that function later).

joolswills: probably best to make this folder by default anyway - and then just set the .ignore_overscan as part of this logic.

scriptmodules/ports/steamlink.sh
hhromic: even better: `grep -q '^overscan_scale=1' /boot/config.txt` doesn't produce output at all ;)

joolswills: it will if config.txt doesn't exist.

hhromic: right, sorry, I keep forgetting that RetroPie doesn't run only on RPIs :)

joolswills: This module is RPI only, but actually on a RPI config.txt may not be present. Boot partition may not be mounted or like with berryboot.

scriptmodules/ports/steamlink.sh
psyke83: I'm not sure if moveConfigDir is appropriate in this case, as `~/.local/share/SteamLink` is where the actual binaries are downloaded and updated too. There is also a `~/.local/share/Valve Software/Streaming Client/` dir which contains autogenerated configuration based on the active session, but is also the location to put an optional user configuration (streaming_args.txt). Maybe that file can be linked via moveConfigFile at least?

joolswills: Ah I see. Hmm. I'll leave it up to you :-)

psyke83: All your comments should be addressed.

I've only added streaming_args.txt via moveConfigFile, and intentionally left it blank, as adding unrecognized parameters - including commented lines - will break execution of the streaming_client binary.

Users will be able to dig into the supported arguments via a Google search, or perhaps someone will create a wiki entry on our site.

psyke83:
*shakes fist*

;)

I was following the example seen in steamcontroller.sh by creating a user spawned sudo shell, but I'm too lazy to check git blame on that part of the code to know if that code was you/sanctioned by you, haha.

scriptmodules/ports/steamlink.sh
```diff
@@ -32,5 +32,22 @@ function remove_steamlink() {
 }
 
 function configure_steamlink() {
+    local sl_dir="$home/.local/share/SteamLink"
+    local slconfig_dir="$home/.local/share/Valve Software/Streaming Client"
+
+    mkUserDir "$sl_dir"
+    mkUserDir "$slconfig_dir"
```
joolswills: if home/.local/share/Valve Software doesn't exist it will be made with root permissions so a chown is needed. (as the mkUserDir does a mkdir -p as root, but only chowns the directory passed, not the parent directories.

psyke83: Should be sorted now.

joolswills:
thanks.

PR #2553, opened 2018-12-05
image - move default gpu_mem and overscan_scale settings out of the E…

joolswills:
* users may want to change these and not have them reset to defaults when updating ES.

Currently testing this.

### DIFF
--- a/scriptmodules/admin/image.sh
+++ b/scriptmodules/admin/image.sh
@@ -95,6 +95,13 @@ function install_rp_image() {
         sed -i "s/quiet/quiet loglevel=3 consoleblank=0 plymouth.enable=0 quiet/" "$chroot/boot/cmdline.txt"
     fi
 
+    # set default GPU mem, and overscan_scale so ES scales to overscan settings.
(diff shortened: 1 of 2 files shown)


scriptmodules/admin/image.sh
```diff
@@ -95,6 +95,13 @@ function install_rp_image() {
         sed -i "s/quiet/quiet loglevel=3 consoleblank=0 plymouth.enable=0 quiet/" "$chroot/boot/cmdline.txt"
     fi
 
+    # set default GPU mem, and overscan_scale so ES scales to overscan settings.
+    iniConfig "=" "" "$chroot/boot/config.txt"
+    iniSet "gpu_mem_256" 128
+    iniSet "gpu_mem_512" 256
+    iniSet "gpu_mem_1024" 256
+    iniSet "overscan_scale" 1
```
joolswills: Instead of using iniSet I may append these settings now, with some comments on usage. Might be more user friendly then (eg to explain that the gpu_mem_XXX configs overrides the default gpu_mem and that users may want to disable overscan_scale if settings disable_overscan to 1)

PR #2554, opened 2018-12-05
Home folders

joolswills:
create ~/.config ~/.local ~/.local/share folder used by some modules @psyke83

### DIFF
--- a/scriptmodules/emulators/scummvm.sh
+++ b/scriptmodules/emulators/scummvm.sh
@@ -62,9 +62,7 @@ function configure_scummvm() {
     mkRomDir "scummvm"
 
     local dir
-    mkUserDir "$home/.local"
-    for dir in .config .local/share .cache; do
-        mkUserDir "$home/$dir"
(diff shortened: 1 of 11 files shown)


scriptmodules/emulators/scummvm.sh
```diff
@@ -62,9 +62,7 @@ function configure_scummvm() {
     mkRomDir "scummvm"
 
     local dir
-    mkUserDir "$home/.local"
     for dir in .config .local/share .cache; do
```
psyke83: The helper won't account for `.cache`?

joolswills: It doesn't but probably shouldnt symlink that - cache should remain in home I guess. We can make the .cache folder though. I'll add it.

joolswills: I now create the .cache folder and no longer symlink anything from the .cache as there is no need imho to move this to the config area (can be regenerated etc if needed)

PR #2555, opened 2018-12-06
addEmulator / Runcommand RetroArch commandline detection

joolswills:
helpers / addEmulator - adjust the logic for automatic RA cmd generation
 * match against any cmd parameter that is an .so in $md_inst

runcommand - detect retroarch by emulator id not command string

@hhromic - hopefully this will help https://github.com/RetroPie/RetroPie-Setup/pull/2551


joolswills:
I adjust the runcommand detection also so your script doesn't have to have retroarch in the name.

PR #2558, opened 2018-12-07
advmame, configedit - fix scriptmodule permissions (755 -> 644)

hhromic:
I just noticed that the `advmame` and `configedit` scriptmodules are set to executables. Like all the other scriptmodules, these should also not be executables. Probably someone forgot to set them correctly.
This PR doesn't change any functionality or files, it just fixes the file permissions.

Found doing this:

    $ find scriptmodules -maxdepth 2 -executable -type f
    scriptmodules/supplementary/configedit.sh
    scriptmodules/emulators/advmame.sh


joolswills:
Thanks

PR #2559, opened 2018-12-07
steamlink: update script for 1.0.7

psyke83:
* update dependencies
* update directory locations for new version


CookiePLMonster:
Quoting changelog of the latest build, Build 69:

> Fixed black screen when overscan_scale is enabled

Wouldn't that mean ignore_overscan is now not needed anymore?

psyke83:
Yes, but the override will do no harm. The steamlink package is now part of the official raspberrypi repository too, but it's still version 1.0.7. No need to update the module until the next version bump, at least.

PR #2560, opened 2018-12-13
esthemes - do not append `(not installed)` if a theme is not installed

hhromic:
I noticed that the esthemes menu displays *(not installed)* for every non-installed theme:

esthemes-current

Which makes the menu look a bit cluttered/convoluted. A better approach is what is currently done in the **RetroPie packages menus**, where only the installed packages are marked as *(installed)*:

packages-menu

This PR helps reduce visual clutter in the **esthemes menu** using the same approach as the package managing menus. Example:

esthemes-cleaner


joolswills:
Ta. Will merge, but am in progress of changing this code (Going to split by author).

PR #2563, opened 2018-12-18
correct mame2003-plus artwork path & copy artwork sources

markwkidd:
This PR pertains to a feature of mame2003-plus which had bugs and had thought I would give up on: backdrop artwork. However, the bugs with artwork unexpectedly were solved by an unrelated fix, and now backdrops can be displayed.

@joolswills would you please review this PR? I do not have a RetroPie setup to test it on.

My intentions are to:
  1. Correctly locate the artwork folder within the mame2003-plus SYSTEM folder instead of the SAVE folder
  2. Using the `hiscore.dat` and `cheat.dat` as a model, copy and chown the backdrop artwork files to the right place

In case you are wondering: at the moment there is just one artwork file in the repository folder, but there may be as many as a dozen or two files eventually. Not all that many games have backdrop art to begin with.

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -59,17 +59,21 @@ function configure_lr-mame2003() {
         for mame_sub_dir in cfg ctrlr diff hi memcard nvram; do
             mkRomDir "$mame_dir/$dir_name/$mame_sub_dir"
         done
-
-        # lr-mame2003-plus also has an artwork folder
-        [[ "$md_id" == "lr-mame2003-plus" ]] && mkRomDir "$mame_dir/$dir_name/artwork"
     done
 
     mkUserDir "$biosdir/$dir_name"
     mkUserDir "$biosdir/$dir_name/samples"
 
-    # copy hiscore.dat
+    # copy hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$dir_name/"
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
+    
+    # lr-mame2003-plus also has an artwork folder
+    if [[ "$md_id" == "lr-mame2003-plus" ]]; then
+        mkRomDir "$biosdir/$dir_name/artwork"
+        cp "$md_inst"/metadata/artwork/* "$biosdir/$dir_name"/artwork/
+        chown -R $user:$user "$biosdir/$dir_name/artwork"
+    fi
 
     # Set core options
     setRetroArchCoreOption "${dir_name}-skip_disclaimer" "enabled"


scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -59,17 +59,19 @@ function configure_lr-mame2003() {
...
-
-    # copy hiscore.dat
+    
+    # copy hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$dir_name/"
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
+    
+    # lr-mame2003-plus also has an artwork folder
+    [[ "$md_id" == "lr-mame2003-plus" ]] && mkRomDir "$biosdir/$dir_name/artwork"
+    [[ "$md_id" == "lr-mame2003-plus" ]] && cp "$md_inst/metadata/artwork/*.*" "$biosdir/$dir_name/artwork/"
```
joolswills: quoting *.* won't work.

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -59,17 +59,19 @@ function configure_lr-mame2003() {
...
-
-        # lr-mame2003-plus also has an artwork folder
-        [[ "$md_id" == "lr-mame2003-plus" ]] && mkRomDir "$mame_dir/$dir_name/artwork"
     done
 
     mkUserDir "$biosdir/$dir_name"
     mkUserDir "$biosdir/$dir_name/samples"
-
-    # copy hiscore.dat
+
```
joolswills: trailing whitespace.

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -59,17 +59,19 @@ function configure_lr-mame2003() {
...
-    # copy hiscore.dat
+    
+    # copy hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$dir_name/"
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
+    
+    # lr-mame2003-plus also has an artwork folder
+    [[ "$md_id" == "lr-mame2003-plus" ]] && mkRomDir "$biosdir/$dir_name/artwork"
+    [[ "$md_id" == "lr-mame2003-plus" ]] && cp "$md_inst/metadata/artwork/*.*" "$biosdir/$dir_name/artwork/"
+    [[ "$md_id" == "lr-mame2003-plus" ]] && chown $user:$user "$biosdir/$dir_name/artwork/*.*"
```
joolswills: An `if` block would be better for these lines.

scriptmodules/libretrocores/lr-mame2003.sh
markwkidd: fix, thanks!

scriptmodules/libretrocores/lr-mame2003.sh
markwkidd: Thanks you -- I only deal with Bash occasionally. I've tried to follow your suggestion here and removed the `*.*` but I'm not positive I understand what you mean by quoting it.

Can I use a single `*` and still keep it in the quotes as in my most recent commit?

scriptmodules/libretrocores/lr-mame2003.sh
markwkidd: Thank you -- I've added an `if` block with the latest commit.

scriptmodules/libretrocores/lr-mame2003.sh
hhromic: @markwkidd , you need to leave the globbing (`*`) outside of the quotes, otherwise they will be interpreted as asterisks instead of expanding. In fact it is good practice to quote the minimum possible (around the variables). For example:
```
cp "$md_inst"/metadata/artwork/* "$biosdir/$dir_name/artwork/"
```

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -63,15 +63,18 @@ function configure_lr-mame2003() {
...
     
     # lr-mame2003-plus also has an artwork folder
-    [[ "$md_id" == "lr-mame2003-plus" ]] && mkRomDir "$biosdir/$dir_name/artwork"
-    [[ "$md_id" == "lr-mame2003-plus" ]] && cp "$md_inst/metadata/artwork/*.*" "$biosdir/$dir_name/artwork/"
-    [[ "$md_id" == "lr-mame2003-plus" ]] && chown $user:$user "$biosdir/$dir_name/artwork/*.*"
+    if (( "$md_id" == "lr-mame2003-plus" ))
+    then
+        mkRomDir "$biosdir/$dir_name/artwork"
+        cp "$md_inst/metadata/artwork/*" "$biosdir/$dir_name/artwork/"
+        chown $user:$user "$biosdir/$dir_name/artwork/*.*"
```
hhromic: here you should also just use `*` instead of `*.*` :)

scriptmodules/libretrocores/lr-mame2003.sh
markwkidd: Thanks @hhromic ! Now I understand what's up with the asterisk and I've pushed a commit to move it outside the double quotes

scriptmodules/libretrocores/lr-mame2003.sh
markwkidd: Thanks @hhromic !

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -59,17 +59,22 @@ function configure_lr-mame2003() {
...
     mkUserDir "$biosdir/$dir_name"
     mkUserDir "$biosdir/$dir_name/samples"
 
-    # copy hiscore.dat
+    # copy hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$dir_name/"
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
+    
+    # lr-mame2003-plus also has an artwork folder
+    if (( "$md_id" == "lr-mame2003-plus" ))
```
hhromic: one more little comment, here you should use square brackets `[[`/`]]` instead of round ones `((`/`))`

scriptmodules/libretrocores/lr-mame2003.sh
hhromic: happy to help!  and thanks you too for the great `lr-mame2003-plus` core, it's a delight hands down!

scriptmodules/libretrocores/lr-mame2003.sh
markwkidd: got it, thanks one more time! :)

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -59,17 +59,22 @@ function configure_lr-mame2003() {
...
     mkUserDir "$biosdir/$dir_name/samples"
 
-    # copy hiscore.dat
+    # copy hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$dir_name/"
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
+    
+    # lr-mame2003-plus also has an artwork folder
+    if [[ "$md_id" == "lr-mame2003-plus" ]]
+    then
```
joolswills: ` if [[ "$md_id" == "lr-mame2003-plus" ]]; then` to match style of rest of code.

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -59,17 +59,22 @@ function configure_lr-mame2003() {
...
+        mkRomDir "$biosdir/$dir_name/artwork"
+        cp "$md_inst"/metadata/artwork/* "$biosdir/$dir_name"/artwork/
+        chown $user:$user "$biosdir/$dir_name"/artwork/*
```
joolswills: `chown -R $user:$user "$biosdir/$dir_name/artwork/"

joolswills:
Please also squash these commits down.

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -69,11 +69,10 @@ function configure_lr-mame2003() {
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
     
     # lr-mame2003-plus also has an artwork folder
-    if [[ "$md_id" == "lr-mame2003-plus" ]]
-    then
+    if [[ "$md_id" == "lr-mame2003-plus" ]] then
```
hhromic: almost there, you need to put `;` between `]]` and `then` :)

markwkidd:
Lord I thought I was just going to be able to do some copy-paste here but instead I appreciate this tutorial.

I've squashed my commits and hopefully fixed the syntax on my if statement.

joolswills:
I appreciate your contribution, but I think next time unless you are able to test the changes on a PC, it would be best just to open an issue requesting the changes needed as may save time just to implement it rather than doing a detailed review.

mkerryfuller:
I'm a bit lost.  On a Pi, using RetroPie.  

See an artwork directory under bios/mame2003-plus and one under config/roms/arcade/mame2003-plus

Just compiled src using Retropie update at 2:00 pm EST US.   Based on the git thread above, I believe I should put my artwork in a zip file named after the rom, in the BIOS/mame2003-plus/artwork folder.

Doesn't work.  artwork is enabled in Retroarch options and I have overlays disabled.   Yes, I know to restart also.

What am I missing?


BTW, great work on the vector multiplier and overlay colors.  I'm new to your emu but that's what sucked me in... good speed with the extras.   ; )

markwkidd:
@mkerryfuller when you build mame2003-plus from source, do you see `warrior.zip` automatically appear in the new artwork folder? (It should)

If so, could you try to play the `warrior` romset and see if the backdrop appears?

mkerryfuller:
I was going to include that in my previous reply, but hit send too quick.

No, the artwork did not get installed.  I assumed the one you mentioned in the thread would be the Warrior one, and so I looked for it.

I ended up tossing my Warrior.zip backdrop from Mr. Do in there just to test... it didn't work.

mkerryfuller:
PS.  if you saw my post on the Retropie forum, I have a fresh install of Raspbian, Retropie installed on top of it, and your emu (only emu on the system).

markwkidd:
@mkerryfuller do you mind moving this conversation to either the retropie forum thread or the mame2003-plus github issue tracker?

PR #2564, opened 2018-12-19
lr-mame2003-plus: fix installation of artwork in $biosrom instead of $romdir

hhromic:
The correct function to use is `mkUserDir()` instead of `mkRomDir()`.
Also did some minor code cosmetic clean-up.

As reported in the forum posts:
* https://retropie.org.uk/forum/post/171213
* https://retropie.org.uk/forum/post/171515

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -67,11 +67,11 @@ function configure_lr-mame2003() {
     # copy hiscore.dat and cheat.dat
     cp "$md_inst/metadata/"{hiscore.dat,cheat.dat} "$biosdir/$dir_name/"
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
-    
+
     # lr-mame2003-plus also has an artwork folder
     if [[ "$md_id" == "lr-mame2003-plus" ]]; then
-        mkRomDir "$biosdir/$dir_name/artwork"
-        cp "$md_inst"/metadata/artwork/* "$biosdir/$dir_name"/artwork/
+        mkUserDir "$biosdir/$dir_name/artwork"
+        cp "$md_inst/metadata/artwork/"* "$biosdir/$dir_name/artwork/"
         chown -R $user:$user "$biosdir/$dir_name/artwork"
     fi
 


hhromic:
@markwkidd   is `$biosdir` really the correct place to put artwork? shouldn't it be in `$romdir` ? Alongside the existing `cfg ctrlr diff hi memcard nvram` dirs? 
Let me know to adjust the PR accordingly.

joolswills:
Thanks. Actually I probably shouldn't have accepted the PR as it was untested. I will go back to being strict again :-)

scriptmodules/libretrocores/lr-mame2003.sh
```diff
@@ -67,12 +67,12 @@ function configure_lr-mame2003() {
...
     chown $user:$user "$biosdir/$dir_name/"{hiscore.dat,cheat.dat}
-    
+
     # lr-mame2003-plus also has an artwork folder
     if [[ "$md_id" == "lr-mame2003-plus" ]]; then
-        mkRomDir "$biosdir/$dir_name/artwork"
+        mkUserDir "$biosdir/$dir_name"/artwork
         cp "$md_inst"/metadata/artwork/* "$biosdir/$dir_name"/artwork/
-        chown -R $user:$user "$biosdir/$dir_name/artwork"
+        chown -R $user:$user "$biosdir/$dir_name"/artwork
```
joolswills: BTW this isn't actually how we generally style things - the previous line is.

joolswills: the cp line is also not in line with other code for the same reason. It's not a major issue, but just saying.

hhromic: no problem, will revert it ..  I like to minimise the quotes scope to where is really necessary :)

joolswills: I don't think there is a right or wrong here, but it's not how we do it elsewhere so better to stick with what we use currently.

hhromic: ok should be more like the other scriptmodules now... I gave instructions in the forum to test the PR, so when some feedback comes from the users you can merge it

hhromic: Thinking it better, maybe this PR should be merged ASAP and adjusted later if necessary (unlikely). Users are getting their romdirs polluted if they use the current version:

> Now there is a /home/pi/RetroPie/roms/home/pi/RetroPie/BIOS/mame2003-plus/artwork path on my Pi.

joolswills:
Thanks.

hhromic:
Thanks for the clarification @markwkidd ! Then everything should be fine now.

PR #2565, opened 2018-12-20
steamlink: switch to xinit+matchbox session & misc updates

psyke83:
When steamlink is launched without a running Xorg server,
DualShock controllers erroneously act as both a gamepad and virtual
mouse, which causes major input conflicts in-game.

Launching from an Xorg session fixes this issue, but steamlink also
requires a window manager running in order to set the window focus
property (to allow joystick events to be processed).

Additional changes:
* switch to repository version of steamlink package
* update Help text to reflect new onscreen welcome message
* remove overscan override (issue fixed upstream).

### DIFF
--- a/scriptmodules/ports/steamlink.sh
+++ b/scriptmodules/ports/steamlink.sh
@@ -14,17 +14,14 @@ rp_module_desc="Steam Link for Raspberry Pi 3"
 rp_module_licence="PROP https://steamcommunity.com/app/353380/discussions/0/1743353164093954254/"
 rp_module_section="exp"
 rp_module_flags="!mali !x86 !kms !rpi1 !rpi2"
-rp_module_help="Streaming games to your computer with Steam.\n\nIf you experience a black screen during startup, you may need to disable the overscan_scale setting in your Pi's /boot/config.txt."
+rp_module_help="Stream games from your computer with Steam"
 
 function depends_steamlink() {
-    getDepends python3-dev curl xz-utils libinput10 libxkbcommon-x11-0
+    getDepends python3-dev curl xz-utils libinput10 libxkbcommon-x11-0 matchbox
 }
 
 function install_bin_steamlink() {
-    local ver="1.0.7"
-    local url="http://media.steampowered.com/steamlink/rpi"
-    wget -qO "$__tmpdir/steamlink_"$ver"_armhf.deb" ""$url"/steamlink_"$ver"_armhf.deb"
-    dpkg -i "$__tmpdir/steamlink_"$ver"_armhf.deb"
+    aptInstall steamlink
 }
 
 function remove_steamlink() {
@@ -32,6 +29,7 @@ function remove_steamlink() {
 }
 
 function configure_steamlink() {
+    local sl_script="$md_inst/steamlink_xinit.sh"
     local sl_dir="$home/.local/share/SteamLink"
     local valve_dir="$home/.local/share/Valve Corporation"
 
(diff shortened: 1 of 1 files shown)


hhromic:
Got a chance to test, this is the list of packages that are pulled alongside `matchbox`:
```
The following additional packages will be installed:
  gnome-icon-theme gtk-update-icon-cache hicolor-icon-theme libatk1.0-0 libatk1.0-data libfakekey0 libgtk2.0-0 libgtk2.0-common libmatchbox1 libpangoxft-1.0-0
  librsvg2-common libstartup-notification0 libxcomposite1 libxft2 libxsettings-client0 matchbox-common matchbox-desktop matchbox-keyboard matchbox-panel
  matchbox-panel-manager matchbox-window-manager
Suggested packages:
  gvfs ttf-dejavu
Recommended packages:
  libgail-common libgtk2.0-bin matchbox-keyboard-im
The following NEW packages will be installed:
  gnome-icon-theme gtk-update-icon-cache hicolor-icon-theme libatk1.0-0 libatk1.0-data libfakekey0 libgtk2.0-0 libgtk2.0-common libmatchbox1 libpangoxft-1.0-0
  librsvg2-common libstartup-notification0 libxcomposite1 libxft2 libxsettings-client0 matchbox matchbox-common matchbox-desktop matchbox-keyboard matchbox-panel
  matchbox-panel-manager matchbox-window-manager
0 upgraded, 22 newly installed, 0 to remove and 0 not upgraded.
Need to get 15.2 MB of archives.
After this operation, 44.5 MB of additional disk space will be used.
```

Perhaps a proper solution should be addressed upstream so bloating the system is not required?
I understand the Steam Link developers are very open to suggestions/feedback.

joolswills:
I wouldn't call it bloat. And other retropie packages use Xorg so it's not an issue.

psyke83:
@hromic,

The dependencies of the actual steamlink debian package don't list Xorg components either, but that's because it only lists the absolute bare essentials for the script's bootstrapping to function. If you look at `~/.local/share/SteamLink/steamlinkdeps.txt` and how the script implements custom dependency checking & installation via `steamlinkdeps`, you'll see that steamlink really does need a minimal installation of Xorg's libraries (even though it's possible for it to run without an Xorg session, likely thanks to SDL2's rpi video driver).

As for the issue of the window manager pulling in unnecessary dependencies, I was planning to use a different workaround to the window focus issue that didn't require any WM:

```
#!/bin/bash

/usr/bin/steamlink &
SL_PID=$!

echo "pid: $SL_PID"
while kill -0 $SL_PID; do
    xdotool search --onlyvisible --sync . windowfocus
    sleep 3
done
```

This seemed to work well enough, although one time the screen went blank and stopped responding to controller inputs. I'm not sure if that because I didn't configure the dpms suspend via xset, or if xdotool somehow focused the wrong window and thus disabled controller input, but since I later noticed that some other scriptmodules (minecraft & micropolis) already use matchbox, I opted for the safer alternative rather than the above script which is a bit of a hack.

hhromic:
> The dependencies of the actual steamlink debian package don't list Xorg components either, but that's because it only lists the absolute bare essentials for the script's bootstrapping to function.

Indeed. When SL was just launched for the RPI I tried it and ran nicely without any need for Xorg at all, so definitively is not necessary to run a full-blown Xorg session for it.

However, if there are issues without Xorg, I guess that's unavoidable for now. As you say, probably because they are assuming people use SL under a typical Xorg desktop such as Raspbian Full (not Lite).

I'm not particularly impressed by their integration engineering to be honest, these scripts and dep handling are not very elegant. Also I find it ridiculous that SL requires Xorg + WM to properly handle controllers!

> As for the issue of the window manager pulling in unnecessary dependencies, I was planning to use a different workaround to the window focus issue that didn't require any WM:

Yes indeed that looks hackish. I'm not against installing a WM and Xorg, if that is what SL really seems to need, but requiring it just for proper controller support is disappointing. Anyway the moonlight scriptmodule that I'm preparing is very lightweight so will be a good alternative.

Does this issue only happen with DualShock controllers?

psyke83:
I only have DS3 controllers to test, so I'm not certain if others would be impacted.

The SDL2 controller mapping is identical when launched with or without Xorg:
```
Added controller with mapping 050000004c0500006802000000800000,PS3 Controller,a:b0,b:b1,back:b8,dpdown:b14,dpleft:b15,dpright:b16,dpup:b13,guide:b10,leftshoulder:b4,leftstick:b11,lefttrigger:b6,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b12,righttrigger:b7,rightx:a3,righty:a4,start:b9,x:b3,y:b2,
```

The issue manifests only when the native `hid-sony` kernel driver is in use, but not on the userspace `sixad` wrapper (which we want to phase out anyway). My initial guess is that `hid-sony`'s mapping of the L2/R2 buttons via `HID_GD_POINTER` is confusing SDL2 or steamlink's own code into thinking that it should use a virtual mouse mapping for the device, but I'm uncertain as to why the behaviour is different when launched in an X session. I haven't had time to properly troubleshoot further or even confirm my guess as to the issue, though.

PR #2566, opened 2018-12-20
fix the copy up for lr mame2003+

grant2258:
The * in the last commit was placed wrong and not needed with a recursive copy

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -71,7 +71,7 @@ function configure_lr-mame2003() {
     # lr-mame2003-plus also has an artwork folder
     if [[ "$md_id" == "lr-mame2003-plus" ]]; then
         mkUserDir "$biosdir/$dir_name/artwork"
-        cp "$md_inst/metadata/artwork/"* "$biosdir/$dir_name/artwork/"
+        cp -r "$md_inst/metadata/artwork" "$biosdir/$dir_name"
         chown -R $user:$user "$biosdir/$dir_name/artwork"
     fi
 


hhromic:
I was thinking on doing it like that too, however @markwkidd did not and I thought it was for a reason.
Why you say the wildcard is wrong? it still works but I agree it doesn't copy recursively.  However the current content on that folder is just a single file. I agree the recursive approach is more robust.

grant2258:
I just like future proofing if sub folders arrive we would need to update the script again. The can fix the * placement or use this its down to there personal preference at the end of the day

hhromic:
Ah I see, so the current script is not broken. That's a relief.

I agree that this enhancement you propose is a good idea. I would just advise on improving the commit log message before merging. Something like should do:

    lr-mame2003-plus: install artwork assets recursively

Thanks for the help.

grant2258:
https://github.com/RetroPie/RetroPie-Setup/blob/49c3fa54f2fbddf546f982563511cb0c2b554db1/scriptmodules/libretrocores/lr-mame2003.sh#L74

look where the star is

hhromic:
It looks odd, but is not wrong. See this example:
```sh
$ mkdir source
$ mkdir target
$ touch source/file1 source/file2
$ SRC="source"
$ TRG="target"
$ cp -v "$SRC/"* "$TRG"
'source/file1' -> 'target/file1'
'source/file2' -> 'target/file2'
```

It is not working like that for you?

grant2258:
yes it works fine 

```
cp "$md_inst/metadata/artwork/"* "$biosdir/$dir_name/artwork/" 
```
should be 

```
cp "$md_inst/metadata/artwork/*" "$biosdir/$dir_name/artwork/" 
```

hhromic:
That's incorrect. If you put the globbing (`*`) inside the quotes, it does not get expanded:
```
$ cp -v "$SRC/*" "$TRG/"
cp: cannot stat 'source/*': No such file or directory
```
As you see, `cp` is trying to copy a file literally named `*` instead of expanding the actual file names.

grant2258:
i never checked it it copied the files ill check now

grant2258:
pi@retropie:~/RetroPie-Setup $ cp -v logs/* test/
'logs/rps_2018-12-16_211006.log.gz' -> 'test/rps_2018-12-16_211006.log.gz'
'logs/rps_2018-12-17_201232.log.gz' -> 'test/rps_2018-12-17_201232.log.gz'
'logs/rps_2018-12-17_201245.log.gz' -> 'test/rps_2018-12-17_201245.log.gz'
pi@retropie:~/RetroPie-Setup $ cp -v "logs/"* test/
'logs/rps_2018-12-16_211006.log.gz' -> 'test/rps_2018-12-16_211006.log.gz'
'logs/rps_2018-12-17_201232.log.gz' -> 'test/rps_2018-12-17_201232.log.gz'
'logs/rps_2018-12-17_201245.log.gz' -> 'test/rps_2018-12-17_201245.log.gz'

grant2258:
usual jools attitude got to love it :)

hhromic:
@grant2258   you are stepping on your own foot. You are using the wildcard (star) **outside** the quotes, contrary to your own suggestion to put in **inside**, as you said in https://github.com/RetroPie/RetroPie-Setup/pull/2566#issuecomment-448971561

The scriptmodule is not currently broken, but I agree it is not recursively copying either.
When more assets are available, we can revisit to make it recurive.

grant2258:
not i big deal in the big picture I do agree

joolswills:
If you are going to make personal comments, I will be happy to block you from this project.

PR #2568, opened 2018-12-23
steamlink: update xorg dependency & reduce matchbox install

psyke83:
* Add missing xorg dependency
* Replace matchbox with matchbox-window-manager to eliminate unnecessary
  desktop/gtk dependencies.


psyke83:
Tagging @hromnic re: our previous discussion on matchbox.

Installing just the window manager component seems to work fine and should reduce the install footprint, but I'm testing from a well-used card that already has a pre-existing installation of the full Pixel desktop. I'd appreciate if anyone happens to have a fresh install can test and confirm that no additional dependencies are necessary (I don't have any spare SD cards at the moment).

42Swampy:
For me it worked. I never had a full Pixel desktop installaton.

joolswills:
thanks.

PR #2574, opened 2019-01-08
Update abuse.sh

PabOu-be:
Had to provide the datadir because it was looking for the data in $md_inst/share/games/abuse/


retro24:
Hello,abuse is not shown in the retropie-setup menu anymore for installation,is been removed?

PR #2584, opened 2019-01-20
Modifications for Skyscraper 3.x release; fix reported issued

cmitu:
Prepare for the 3.x verion of **Skyscraper**, which has some notable changes for scraping - most important one being that scrape/gather of resources (artwork, info) and gamelist generation are separate actions, so they need to be run separately.

 I refactored the GUI to make sure that scraping and gamelist generation are 2 separate actions and moved back (to sub-menus) the options for scraping and game list generation. Overview of changes:

1. UI Refactoring and additions.
    * Added sections for gathering and game list generations, each with actions and options
    * Added a separate section for advanced options and moved them back from the main menu
    * Added editing of the `aliasMap.csv` file (introduced in the 2.x series)
    * Added an update action - similar to the one in the `scraper` module. User must first run 'Check for Update' and if a new release is detected', the item becomes 'Update to x.y.z' that can be run to update Skyscraper from source.
    * LocalDB replaced with 'Cache' in description and help strings
    * Added the `esgamelist` as a scraping source, introduced in a previous version.
    * Account for the cache folder change (`dbs` moved to `cache`). This means we no longer create this folder automatically, will be created on **Skyscraper**'s first run.

2. Bug fixes  
   * Modified the listing of available ROM systems so that:
        * when a ROM system has only sub-folders (but no direct ROM file), the system would appear correctly
        *  when the `roms` folder is a symlink, listing of the sub-folders would fail, so that the user wouldn't be able to select a system for scraping.
    

The GUI script should work with both 2.x and 3.x versions, installation also.
I still need to update the Wiki with the new UI and _modus operandi_, links to the - moved - docs on the **Skyscraper** repository and re-organizing a bit the list of options available.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -39,56 +39,67 @@ function install_skyscraper() {
         'artwork.xml.example2'
         'artwork.xml.example3'
         'artwork.xml.example4'
-        'ARTWORK.md'
         'tgdb_developers.json'
         'tgdb_publishers.json'
         'mameMap.csv'
+        'aliasMap.csv'
         'hints.txt'
-        'dbs'
         'import'
         'resources'
     )
 }
 
+# Get the location of the cached resources folder. In v3+, this changed to 'cache'.
+# Note: the cache folder might be unavailable during first time installations
+function _cache_folder_skyscraper() {
+    if [[ -d "$configdir/all/skyscraper/dbs" ]]; then
+        echo "dbs"
+    else
+        echo "cache"
+    fi
+}
+
 # Purge all Skyscraper caches
 function _purge_skyscraper() {
     local platform
+    local cache_folder=$(_cache_folder_skyscraper)
 
     while read platform; do
-        # Find any folders underneath the Local DB cache and clear them
+        # Find any sub-folders of the cache folder and clear them
         _clear_platform_skyscraper "$platform"
-    done < <(find "$configdir/all/skyscraper/dbs" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
+    done < <(find "$configdir/all/skyscraper/$cache_folder" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
 }
 
 function _clear_platform_skyscraper() {
     local platform="$1"
+    local cache_folder=$(_cache_folder_skyscraper)
 
-    [[ ! -d "$configdir/all/skyscraper/dbs/$platform" ]] && return
+    [[ ! -d "$configdir/all/skyscraper/$cache_folder/$platform" ]] && return
 
     # Remove any folder underneath the platform, try to keep any customized files
-    pushd "$configdir/all/skyscraper/dbs/$platform"
+    pushd "$configdir/all/skyscraper/$cache_folder/$platform"
     find . -maxdepth 1 -mindepth 1 -type d -print0 -exec rm -fr {} \;
     rm -f db.xml
     popd
 }
 
-# Purge by cached platform
 function _purge_platform_skyscraper() {
     local options=()
+    local cache_folder=$(_cache_folder_skyscraper)
     local system
 
     while read system; do
         # If there is no 'db.xml' file underneath the folder, skip it, it means folder is empty
-        [[ ! -f "$configdir/all/skyscraper/dbs/$system/db.xml" ]] && continue
+        [[ ! -f "$configdir/all/skyscraper/$cache_folder/$system/db.xml" ]] && continue
 
         # Get the size on disk of the system and show it in the select list
-        local size=$(du -sh  "$configdir/all/skyscraper/dbs/$system" | cut -f1)
+        local size=$(du -sh  "$configdir/all/skyscraper/$cache_folder/$system" | cut -f1)
         options+=("$system" "$size" OFF)
-    done < <(find "$configdir/all/skyscraper/dbs" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
+    done < <(find "$configdir/all/skyscraper/$cache_folder" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
 
     # If not folders are found, show an info message instead of the selection list
     if [[ ${#options[@]} -eq 0 ]] ; then
-        printMsgs "dialog" "Nothing to delete ! No cached systems found in \n$configdir/all/skyscraper/dbs "
+        printMsgs "dialog" "Nothing to delete ! No cached systems found in \n$configdir/all/skyscraper/$cache_folder."
         return
     fi
 
@@ -103,7 +114,7 @@ function _purge_platform_skyscraper() {
 
 function _get_ver_skyscraper() {
     if [[ -f "$md_inst/Skyscraper" ]]; then
-        echo $("$md_inst/Skyscraper" -h | grep 'Running Skyscraper'  | cut -d' '  -f 3 2>/dev/null)
+        echo $("$md_inst/Skyscraper" -h | grep 'Running Skyscraper'  | cut -d' '  -f 3 | tr -d v 2>/dev/null)
     fi
 }
 
@@ -113,11 +124,11 @@ function _latest_ver_skyscraper() {
 
 # List any non-empty system found in the ROM parent folder
 function _list_systems_skyscraper() {
-    find "$romdir"  -mindepth 2 -maxdepth 2 -type f -not -name '+*' | sed 's%/[^/]*$%%' | sort -u
+    find "$romdir/" -mindepth 2 -type f -not -name '+*' | sed "s%$romdir/%%; s%/.*$%%" | sort -u
 }
 
 function remove_skyscraper() {
-    # On removal of package, purge the Local DB
+    # On removal of the package, purge the cache
     _purge_skyscraper
 }
 
@@ -133,18 +144,21 @@ function configure_skyscraper() {
     # Handle the cases where the user has an existing Skyscraper installation.
     if [[ -d "$home/.skyscraper" && "$local_config" != "$configdir/all/skyscraper" ]]; then
         # We have an existing Skyscraper installation, but not handled by this scriptmodule.
-        # Since the $HOME/.skyscraper folder will be moved, make sure the Local DB and Import folders are moved separately
+        # Since the $HOME/.skyscraper folder will be moved, make sure the 'cache' and 'import' folders are moved separately
         local f_size
-        f_size=$(du --total -sm "$home/.skyscraper/dbs" "$home/.skyscraper/import" 2>/dev/null | grep total | cut -f 1 )
-        printMsgs "console" "INFO: Moving the Local DB and Import folders to new configuration folder (total: $f_size Mb)"
+        local cache_folder="dbs"
+        [[ -d "$home/.skyscraper/cache" ]] && cache_folder="cache"
+
+        f_size=$(du --total -sm "$home/.skyscraper/$cache_folder" "$home/.skyscraper/import" 2>/dev/null | tail -n 1 | cut -f 1 )
+        printMsgs "console" "INFO: Moving the Cache and Import folders to new configuration folder (total: $f_size Mb)"
 
         local folder
-        for folder in dbs import; do
+        for folder in $cache_folder import; do
             mv "$home/.skyscraper/$folder" "$home/.skyscraper-$folder" && \
                 printMsgs "console" "INFO: Moved "$home/.skyscraper/$folder" to "$home/.skyscraper-$folder""
         done
 
-        # When having an existing installation, changes are the gamelist is in the ROMs folder
+        # When having an existing installation, chances are the gamelist is generated in the ROMs folder
         # Create a GUI config file with this setting pre-set.
         iniConfig " = " '"' "$configdir/all/skyscraper.cfg"
         iniSet "use_rom_folder" 1
@@ -152,8 +166,8 @@ function configure_skyscraper() {
 
     moveConfigDir "$home/.skyscraper" "$configdir/all/skyscraper"
 
-    # When migrating an existing installation, make sure we move the Local DB and Import folder back the new conf folder
-    for folder in dbs import; do
+    # Move the Cache and Import folders back the new conf folder
+    for folder in $cache_folder import; do
         if [[ -d "$home/.skyscraper-$folder" ]]; then
             printMsgs "console" "INFO: Moving "$home/.skyscraper-$folder" back to configuration folder"
             mv  "$home/.skyscraper-$folder" "$configdir/all/skyscraper/$folder"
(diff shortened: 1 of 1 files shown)


cmitu:
For now I think we'll let it in the `exp` still (https://retropie.org.uk/forum/topic/19588/skyscraper-now-officially-part-of-retropie-please-test/151). I'll pinged @muldjord on the forum, maybe he want to do a bit of testing before.

cmitu:
@joolswills Ready for review and merge.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -133,27 +144,30 @@ function configure_skyscraper() {
     # Handle the cases where the user has an existing Skyscraper installation.
     if [[ -d "$home/.skyscraper" && "$local_config" != "$configdir/all/skyscraper" ]]; then
         # We have an existing Skyscraper installation, but not handled by this scriptmodule.
-        # Since the $HOME/.skyscraper folder will be moved, make sure the Local DB and Import folders are moved separately
+        # Since the $HOME/.skyscraper folder will be moved, make sure the 'cache' and 'import' folders are moved separately
         local f_size
-        f_size=$(du --total -sm "$home/.skyscraper/dbs" "$home/.skyscraper/import" 2>/dev/null | grep total | cut -f 1 )
-        printMsgs "console" "INFO: Moving the Local DB and Import folders to new configuration folder (total: $f_size Mb)"
+        local cache_folder="dbs"
+        [ -d "$home/.skyscraper/cache" ] && cache_folder="cache"
```
hhromic: Should use double brackets `[[` and `]]`

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -241,8 +250,13 @@ function _scrape_skyscraper() {
     fi
 
 
+    # If 2nd parameter is set, use the cache for scrapin and not configured scrape source.
+    # Scraping from cache means we can omit '-s' from the parameter list.
+    if [ -z $2 ]; then
```
hhromic: also should use double-brackets here `[[` and `]]`

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -469,89 +419,85 @@ function gui_skyscraper() {
...
                 [[ $i -ge 10 ]] && online="Local"
-                options+=(3 "Scrape from ${s_source_names[$i]} ($online)")
+                options+=(3 "Gather source - ${s_source_names[$i]} ($online) -->")
                 source_found=1
             fi
         done
 
-        if [[ ! $source_found ]]; then
-            options+=(3 "Scrape from Screenscraper (Default)")
+        if [[ $source_found -ne 1 ]]; then
```
hhromic: not sure if `-ne` or `!=` is preferred

hhromic: after checking the RetroPie source code, looks like `-ne` is preferred for numbers indeed, so please ignore this comment :)

muldjord:
@joolswills Do you think it would be possible to review this before or during the weekend? I'd like to get Skyscraper 3.0.0 out at Sunday if possible. This script needs updating before I do so. Thanks!

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -133,27 +144,30 @@ function configure_skyscraper() {
...
         # We have an existing Skyscraper installation, but not handled by this scriptmodule.
-        # Since the $HOME/.skyscraper folder will be moved, make sure the Local DB and Import folders are moved separately
+        # Since the $HOME/.skyscraper folder will be moved, make sure the 'cache' and 'import' folders are moved separately
         local f_size
-        f_size=$(du --total -sm "$home/.skyscraper/dbs" "$home/.skyscraper/import" 2>/dev/null | grep total | cut -f 1 )
-        printMsgs "console" "INFO: Moving the Local DB and Import folders to new configuration folder (total: $f_size Mb)"
+        local cache_folder="dbs"
+        [[ -d "$home/.skyscraper/cache" ]] && cache_folder="cache"
+
+        f_size=$(du --total -sm "$home/.skyscraper/$cache_folder" "$home/.skyscraper/import" 2>/dev/null | grep total | cut -f 1 )
```
joolswills: total is always on the last line right ? So no need to grep every row. eg `| tail -n1 | cut -f 1` ?
`

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -241,8 +250,13 @@ function _scrape_skyscraper() {
     fi
 
 
+    # If 2nd parameter is set, use the cache for scrapin and not configured scrape source.
+    # Scraping from cache means we can omit '-s' from the parameter list.
+    if [[ -z $2 ]]; then
```
joolswills: please quote the `$2`

note the comment might read better if it was worded like "if 2nd parameter is not set blah, else use the cache for scraping", since the logic checks if it is set. minor but. oh and should it be scraping rather than scrapin ?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -339,62 +353,25 @@ function _open_editor_skyscraper() {
...
-            options+=(2 "Cache covers (Disabled)")
-        fi
-
-        if [[ "$cache_wheels" -eq 1 ]]; then
-            options+=(3 "Cache wheels (Enabled)")
-        else
-            options+=(3 "Cache wheels (Disabled)")
-        fi
+        local cmd=(dialog --backtitle "$__backtitle" --help-button --colors --no-collapse --default-item "$default" --ok-label "Ok" --cancel-label "Back" --title "Advanced options" --menu "    EXPERT - edit configurations\n" 14 50 5)
+        local -a options=()
```
joolswills: there is no need for `local -a`. local is enough.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -457,7 +407,7 @@ function _gui_advanced_skyscraper() {
 
 function gui_skyscraper() {
     if pgrep "emulationstatio" >/dev/null; then
-        printMsgs "dialog" "This scraper must not be run while Emulation Station is running or the scraped data will be overwritten. \n\nPlease quit from Emulation Station, and run RetroPie-Setup from the terminal"
+        printMsgs "dialog" "This scraper must not be run while Emulationstation is running or the scraped data will be overwritten.\n\nPlease quit Emulationstation and run RetroPie-Setup from the terminal:\n\n sudo \$HOME/RetroPie-Setup/retropie_setup.sh"
```
joolswills: actually - it's technically `EmulationStation`

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -469,89 +419,85 @@ function gui_skyscraper() {
...
 
-    s_source=([1]=screenscraper [2]=arcadedb [3]=thegamesdb [4]=igdb [5]=openretro [6]=worldofspectrum)
-    s_source+=([10]=localdb [11]=import)
+    s_source=([1]=screenscraper [2]=arcadedb [3]=thegamesdb [4]=openretro [5]=worldofspectrum)
+    s_source+=([10]=esgamelist [11]=import)
 
-    s_source_names=([1]=ScreenScraper [2]=Arcade Database [3]=TheGamesDB [4]=IGDB [5]="OpenRetro" [6]="World of Spectrum")
-    s_source_names+=([10]="LocalDB Cache" [11]="Import Folder")
+    s_source_names=([1]=ScreenScraper [2]=ArcadeDB [3]=TheGamesDB [4]="OpenRetro" [5]="World of Spectrum")
+    s_source_names+=([10]="Emulationstation Gamelist" [11]="Import Folder")
```
joolswills: these would be more readable if not done as long lines (and then no += is needed).

```
array=(
    [1] = "Blah"
...
```

etc

joolswills: (as you have done elsewhere anyway) with the help strings.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -469,89 +419,85 @@ function gui_skyscraper() {
...
 
     ver=$(_get_ver_skyscraper)
 
     while true; do
         [[ -z "$ver" ]] && ver="v(Git)"
 
-        local cmd=(dialog --backtitle "$__backtitle"  --colors --cancel-label "Exit" --help-button --no-collapse --cr-wrap --default-item "$default" --menu "   Skyscraper: game scraper by Lars Muldjord ($ver)\\n \\n" 21 65 10)
+        local cmd=(dialog --backtitle "$__backtitle"  --colors --cancel-label "Exit" --help-button --no-collapse --cr-wrap --default-item "$default" --menu "   Skyscraper: game scraper by Lars Muldjord ($ver)\\n \\n" 22 60 13)
 
         local -a options=(
```
joolswills: no `-a` needed.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -469,89 +419,85 @@ function gui_skyscraper() {
...
+            options+=(V "Download videos (Enabled)")
         else
-            options+=(8 "Use ROM folders for gamelist & media (Disabled)")
+            options+=(V "Download videos (Disabled)")
         fi
 
         options+=(A "Advanced options -->")
 
+        # Show different options, depending on the previous check action
+        if [[ "$latest_ver" > "$ver" && -n "$latest_ver" ]]; then
```
joolswills: That's not a safe comparison for versions

eg (2.10 is a higher version than 2.2)
```
 [[ 2.2 > 2.10 ]] && echo "true"
```


use compareVersion from helpers.

scriptmodules/supplementary/skyscraper.sh
cmitu: OK, will do

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -562,22 +508,22 @@ function gui_skyscraper() {
...
                     elif [[ $? -eq 2 ]]; then
-                        printMsgs "dialog" "Scraping was aborted"
+                        printMsgs "dialog" "Gathering was aborted"
                     fi
                     ;;
 
                 3)
-                    # Scrape options have a separate dialog
+                    # Scrape source options have a separate dialog
                     declare -a s_options=()
```
joolswills: why declare -a rather than local ?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -658,3 +610,195 @@ function gui_skyscraper() {
         fi
     done
 }
+
+function _gui_cache_skyscraper() {
+    local db_size
+    local cache_folder=$(_cache_folder_skyscraper)
+    local -A help_strings_cache
```
joolswills: for associate arrays we tend to use declare over local. Better documented in bash.

scriptmodules/supplementary/skyscraper.sh
cmitu: That clears up the confusion :). Will fix everywhere.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -658,3 +610,195 @@ function gui_skyscraper() {
...
+
+    help_strings_cache=(
+        [1]="Toggle whether screenshots are cached locally when scraping.\n\nSkyscraper option: \Zb--noscreenshots\Zn"
+        [2]="Toggle whether covers are cached locally when scraping.\n\nSkyscraper option: \Zb--nocovers\Zn"
+        [3]="Toggle whether wheels are cached locally when scraping.\n\nSkyscraper option: \Zb--nowheels\Zn"
+        [4]="Toggle whether marquees are cached locally when scraping.\n\nSkyscraper option: \Zb--nomarquees\Zn"
+        [5]="Force the refresh of resources in the local cache when scraping.\n\nSkyscraper option: \Zb--refresh\Zn"
+        [P]="Purge \ZbALL\Zn all cached resources."
+        [S]="Purge all cached resources for a chosen system."
+#        [V]="Removes all non-used cached resources for a chosen system (Vacuum)"
```
joolswills: this commented line is needed ?

scriptmodules/supplementary/skyscraper.sh
cmitu: Sure, no problem.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -658,3 +610,195 @@ function gui_skyscraper() {
...
+
+        options+=("-" "PURGE cache commands")
+        # options+=(V "Vacuum chosen system")
+        options+=(S "Purge chosen system")
+        options+=(P "Purge all systems (!)")
+
+        local choice=$("${cmd[@]}" "${options[@]}" 2>&1 > /dev/tty)
+
+        if [[ -n "$choice" ]]; then
+            local default=$choice
```
joolswills: prefer `local default="$choice"`

scriptmodules/supplementary/skyscraper.sh
cmitu: It's an option that will only work in the 3.x version. 
I removed it when I realized the users might no upgrade to it right away, I commented it thinking I'll enable it on a future update. The corresponding action is left untouched.
If you prefer to remove it, it's fine.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -658,3 +610,195 @@ function gui_skyscraper() {
...
+                    fi
+            esac
+        else
+            break
+        fi
+    done
+}
+
+function _gui_generate_skyscraper() {
+    local -A help_strings_gen
```
joolswills: as above.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -658,3 +610,195 @@ function gui_skyscraper() {
...
+        if [[ "$use_rom_folder" -eq 1 ]]; then
+            options+=(3 "Use ROM folders for game list & media (Enabled)")
+        else
+            options+=(3 "Use ROM folders for game list & media (Disabled)")
+        fi
+
+        local choice=$("${cmd[@]}" "${options[@]}" 2>&1 > /dev/tty)
+
+        if [[ -n "$choice" ]]; then
+            local default=$choice
```
joolswills: as above.

cmitu:
@joolswills Thank you for the update. I pushed an update incorporating your comments.

muldjord:
@joolswills You can release now. I need the new script before I can release Skyscraper 3.0.0.

Thank you for your work!

joolswills:
If I release now won't it mess people up who have the older version installed (the gui stuff for it)

joolswills:
ah - it's backward compatible ?

muldjord:
No, the script is compatible with both versions

joolswills:
Done. thanks to both of you for your contributions.

muldjord:
I will get right on the Skyscraper release, thanks guys!

muldjord:
Skyscraper 3.0.0 now released. I've added a mention in my release notes about the importance to update the RP script prior to installing it.

hhromic:
> @joolswills I'll keep it in mind for the next update - the extra `-n` check being at the beginning is indeed better to short-circuit the test, since this would be called any time the menu is displayed.

The second instance of `compareVersion` should probably be patched soon anyway, maybe @joolswills can just quickly modify it instead of another PR?

cmitu:
I'll send a PR tomorrow, I have a spelling mistake to fix and I also found a small corner case on upgrades, when the user must re-select the scraping source to have it saved (due to the removed scraping source modules in 3).

PR #2586, opened 2019-01-24
Skyscraper - properly check version numbers this time (amend previous module change)

cmitu:
This should fix version checking using `compareVersions` fix (discussed [here](https://github.com/RetroPie/RetroPie-Setup/pull/2584#commitcomment-32040176)) - thanks @hhromic for spotting the 2nd missing check.

It also fixes a spelling mistake after gathering using selecting systems and properly set the scrape source when the user has previously scraped with a source that was removed in v3 (i.e. 'localdb' or 'igdb').

On the issue of `compareVersions` - this could possibly be optimized to use `sort -V` instead of calling `dpkg --compare-versions` - see https://stackoverflow.com/questions/16989598/. Might be worth looking into it to replace it in the helper.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -490,6 +490,7 @@ function gui_skyscraper() {
         if [[ $source_found -ne 1 ]]; then
             options+=(3 "Gather from - Screenscraper (Online) -->")
             scrape_source="screenscraper" # default scraping source if none found
+            iniSet "scrape_source" "$scrape_source"
         fi
 
         options+=(4 "Cache options and commands -->")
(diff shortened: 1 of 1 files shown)


hhromic:
LGTM :+1:

hhromic:
> On the issue of `compareVersions` - this could possibly be optimized to use `sort -V` instead of calling `dpkg --compare-versions` - see https://stackoverflow.com/questions/16989598/. Might be worth looking into it to replace it in the helper.

Comparing versions is rather complicated if we don't know the nature of the versioning system for the numbers we are comparing. I.e. semantic versioning? major-minor-patch, year-month? single number?

Using `dpkg` is somehow robust here because it uses Debian's versioning logic which covers many of these cases. I wouldn't change the helper. Moreover, I don't think  `dpkg` has such a performance impact anyway to micro-optimise it :)

I think is best to just use it as-is, it has been used for some years in RetroPie, it is robust enough and nobody has complained about its performance (yet).

joolswills:
Basically what @hhromic said. But I looked into this in detail when coming up with a solution. As this isn't something called in a loop, speed isn't the main priority, but sort -V would be much messier to implement also (to implement all the different operators). There is no way I'm going to change this code anyway, but appreciate the input :-)

hhromic:
Note that my loop-based speed test shows `dpkg` being faster than `sort` here, without the cumbersome logic required (which requires even more external programs to be called). So yes, I wouldn't change any code regarding the versions comparator.

joolswills:
:-) yeah - I actually read it wrong. All good then.

PR #2589, opened 2019-01-29
helpers / gitPullOrClone - ensure git branch command runs in $dir and print HEAD status

hhromic:
* Minor Bug: the `git branch -D $commit` command should run in `$dir`, otherwise it fails if `$dir` is different than `$PWD`, for example:

      function sources_module() {
        # here $PWD is 'tmp/build/module'
        gitPullOrClone "/other" https://github.com/user/repo branch commit
      }

* Enhancement: print the status (branch/commit) of HEAD after any operation. It is useful to know exactly at which branch/commit the pull/clone operation left the requested repository, specially on calls without branch/commit arguments where this information is not directly visible.

Tested checking-out the sources of a number of scriptmodules and all is working fine on my side.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -368,9 +368,13 @@ function gitPullOrClone() {
 
     if [[ -n "$commit" ]]; then
         printMsgs "console" "Winding back $repo->$branch to commit: #$commit"
-        git branch -D "$commit" &>/dev/null
+        git -C "$dir" branch -D "$commit" &>/dev/null
         runCmd git -C "$dir" checkout -f "$commit" -b "$commit"
     fi
+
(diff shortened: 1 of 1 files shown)


hhromic:
I'm not a big fan of using pushd/popd to be honest :)
For me these are better suited for interactive shell usage than inside a script. For instance, one has to keep in mind the state of $PWD for debugging. I think it is much more clear to use `git -C <dir>` explicitly. However I also understand repeating too much is not ideal.

Perhaps a better idea could be to use an alias variable, e.g. `gitCmd=git -C $dir` and then just use `$gitCmd pull`, `$gitCmd log`, `$gitCmd branch` and so on and so forth :)

PR #2590, opened 2019-01-30
lr-scummvm - "fix quit" patch is now merged upstream

hhromic:
* The "fix quit" patch is now merged upstream so it is no longer needed here

Ref: https://github.com/libretro/scummvm/pull/130


joolswills:
Thanks!

PR #2592, opened 2019-02-03
configedit: improve joystick detection

psyke83:
Certain hid drivers that enumerate multiple nodes (such as hid-sony)
may not advertise the device name in the parent node of udev DEVPATH.

Fix this by using a more reliable method to query the device name
through the /sys/class userspace IO.

Fixes erroneous "not connected" message on DualShock 3/4 controllers
when using the hid-sony driver.

### DIFF
--- a/scriptmodules/supplementary/configedit.sh
+++ b/scriptmodules/supplementary/configedit.sh
@@ -113,7 +113,7 @@ function _joypad_index_configedit() {
                 # get joystick device paths
                 while read -r dev; do
                     if udevadm info --name=$dev | grep -q "ID_INPUT_JOYSTICK=1"; then
-                        paths+=("$(udevadm info --name=$dev | grep DEVPATH | cut -d= -f2)")
+                        paths+=("$(udevadm info --name=$dev --query=name)")
                     fi
                 done < <(find /dev/input -name "js*")
 
@@ -121,7 +121,7 @@ function _joypad_index_configedit() {
                     # sort by path
                     IFS=$'\n'
                     while read -r path; do
-                        devs_name+=("$(</$(dirname sys$path)/name)")
+                        devs_name+=("$(cat /sys/class/$path/device/name)")
                     done < <(sort <<<"${paths[*]}")
                     unset IFS
                 fi


psyke83:
Note: I only have a DS3 controller to test, but I can at least verify that this change doesn't cause any problems when using the ```ps3controller``` wrapper instead of ```hid-sony```.

joolswills:
Thanks.

ExpatUK:
Using Sixaxis now with the latest Kodi build on RetroPie (18.1) results in no input unless using the old sixad driver package.

The only thing in the log is for some reason Kodi can see the controller, but it is trying to grab input events on keypushes from a non-existent eventid, i.e.

DS3 controller pairs and assigns /dev/input/event3.
Kodi recieves no input and log states:
`ERROR: open_restricted - failed to open /dev/input/event4 (No such device)`

Posting here as it seemed relevant, but may be a separate issue.

psyke83:
@ExpatUK 

See: https://github.com/xbmc/xbmc/issues/15588#issuecomment-466726664

Once merged, installing the `sixaxis` scriptmodule from #2263 will install a udev rule to workaround the issue (since it's unlikely that Raspbian stretch will see a systemd version bump any time soon).

ExpatUK:
Thanks for this. I made a udev rule and got it working, then took a git clone from your repo and compiled it and have been very much enjoying the non-erratic controller behaviour - but the disconnect functionality isn't currently working, is that a known issue? Sorry - I'm aware this is probably not the best place to post this but I'm new to github. Is there a quick fix? Typing disconnect from the bluetoothctl terminal works as expected.

psyke83:
The issue is likely that you didn't install `libevdev-tools`, which would be installed by the associated scriptmodule. By default, the axes emit constant output due to jitter that needs to be corrected by `libevdev-tweak-device` via the calibration function of the script. Without this calibration, the timeout  won't work as expected as it believes that there's constant axis movement.

This isn't a good place for discussion; see the forum topic if you need any further assistance: https://retropie.org.uk/forum/topic/15967/sony-dualshock-3-call-for-testers-sixaxis-script-module

PR #2593, opened 2019-02-03
custombluez: Implement custom BlueZ scriptmodule

psyke83:
BlueZ <5.48 does not support Shanwan or Gasia pairing.

This scriptmodule will build & install newer BlueZ 5.50-1 (debian) packages,
needed to allow sixaxis pairing with third-party DualShock 3 controllers
(and improve DualShock 4 compatibility).

The source includes no custom modifications from upstream except
for the inclusion of Raspbian's patches (which are not
intrusive enough to warrant concern about breaking non-Pi systems).


joolswills:
Thanks :-)

PR #2594, opened 2019-02-03
bluetooth: improve BlueZ 5.50+ & sixaxis compatibility

psyke83:
All changes are compatible with Raspbian stretch's default
BlueZ version.

bluez_cmd_bluetooth:
* enable default-agent during general scan
* fix compatibility with BlueZ 5.50-1 userspace tools
* improved reliability when sending multiple commands

sixaxis:
* ensure both authorization & trust is performed to account for
  different BlueZ behaviour between versions
* force-disconnect Bluetooth connection upon first registration to
  avoid controller hang caused by USB input claim conflict

general:
* increase scan time to 20 seconds
* silence sixad messages

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -46,34 +46,37 @@ function get_script_bluetooth() {
     echo "$name"
 }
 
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

PR #2596, opened 2019-02-04
updated - retroarch to v1.7.6 (and refresh patches)

joolswills:


hhromic:
@joolswills Thanks for updating retroarch so quickly in RetroPie!

I want to take the opportunity to ask/propose a small enhancement in usability. As you know, every time you open the RGUI menu for the first time in each RA session, you are greeted with a *"Welcome to RetroArch!"* message/popup that stays for a couple of seconds on screen.  I realised that this becomes quite annoying when you need to access the menu quickly after re-starting RetroArch, e.g. when testing settings.

Fortunately, there is an option to disable this welcome message (`rgui_show_start_screen`) that can be used. However, due to the paradigm of not saving settings on exit, it is necessary to either edit the `retroarch.cfg` file manually or to use the advanced config editor.

So, do you feel like setting this to `false` by default in RetroPie?  Or you think users might complain and therefore a "poll" should be proposed in the forums?  Or maybe is not worth it to bother at all? :)

hhromic:
Another RGUI option that would be interesting/worth to re-visit/consider is `menu_linear_filter` to disable the linear filter applied to the menu. This filter is kind of blurry and when disabled makes RGUI look much more crispy and clear.

joolswills:
I'm happy to disable the start message. The filter is more subjective. I'll see how it looks.

cmitu:
SInce the new RGUI has also themes now, maybe we can create a RetroPie theme and add it to RGUI
* https://github.com/libretro/RetroArch/pull/7950
* https://github.com/libretro/retroarch-assets/pull/285

hhromic:
@cmitu you read my mind, I was already experimenting with a wallpaper made from the official logo just to see how it would look, of course with reddish colours.

Poor-man's preview (note I have the linear filter OFF):

image

I'm sure there are great artists in the forum that would take a shot on this.

hhromic:
Ah I was also thinking on using that theme (Brogramer) as a base, colourful but more balanced.
Honestly I state that I am no artist here, but was just poking around to see the potential.
I def like your idea of showing the RP logo smaller in a corner so is just a "signature" instead of an old windows 95 wallpaper like I did :)   Placing it above the clock is perfect because is not a busy area.
Are you going to give it a try and make a mockup? :)

cmitu:
I don't have a mockup, but let's decide first if the change would be beneficial (not in this PR, obviously).

cmitu:
@joolswills I bet you didn't expect the 'Pimp by RGUI' contest in the PR ?

hhromic:
@joolswills thanks for adding the `rgui_show_start_screen` option :+1:

PR #2597, opened 2019-02-05
add freechaf scriptmodule to libretrocores

recompileorg:
Everything should be in order this time.


hhromic:
You should squash your commits and give the result a proper commit message, for example:

    add freechaf libretro core to experimental

PR #2601, opened 2019-02-06
helpers: add dkmsManager helper function

psyke83:
Introduce new DKMS helper function to consolidate DKMS management in
the relevant scriptmodules (customhidsony, mkarcadejoystick and xpad).

The helper will properly remove old versions of DKMS modules to prevent
"dkms.conf not found" errors occurring when removing a related scriptmodule
but an older DKMS module version was present.

Also fix the update hooks and ensure proper module reload when needed.


joolswills:
Lovely! Thanks.

PR #2603, opened 2019-02-06
lr-hatari - fix build after recent changes upstream

hhromic:
* Use better linker flags instead of patching Makefile to find libcapsimage

Forum: https://retropie.org.uk/forum/topic/21127/lr-harari-error-installing-from-source/


joolswills:
Thanks.

PR #2605, opened 2019-02-06
Add kodi virtual filesystems | Use kodi-standalone instead

diogomsantos:
In Kodi 18, some components where split from the Kodi binary, the kodi-vfs-* binary addons. For people to use, for example, the integrated nfs, sftp, etc, filesystems and open compressed archives, they need to have them installed.
"kodi-standalone" can now be used instead of "kodi", it doesn't crash on exit anymore.


joolswills:
Thanks.

PR #2607, opened 2019-02-07
joy2key: move delay into joy2key script

psyke83:
Reference: https://retropie.org.uk/forum/topic/21139/error-and-gui-terminates-back-to-es-when-launching-retropie-setup-direct-on-machine-libudev-but-ssh-is-ok/33

The previous workaround did not account for runcommand; by placing the
delay inside the script, we can avoid having to slow down script
execution at the expense of potentially missing inputs for one second
(but even on a Core i7 system, joy2key with this delay is ready for
intercepting inputs immediately by the time the main menu appears)


psyke83:
(this solves the problem for me, but maybe wait for feedback on the forum before merging)

psyke83:
Whoops sorry, it doesn't work as expected.

PR #2608, opened 2019-02-07
dkmsManager: remove update_hook

psyke83:
This might cause issues due to unexpected module rebuilds on script update.


joolswills:
Thanks.

PR #2609, opened 2019-02-08
Added RapidJSON as dependency for Emulationstation

cmitu:
Accompanying https://github.com/RetroPie/EmulationStation/commit/9ce4419a5a1af248032ceffb78432c5dbb3ca226, it won't build without it.

Since the `dev` will switch to `-stable`, I havent' added checks to see if we're on `-dev` or `-stable`.
Also removed the `vlc`, it's not needed on runtime or build, and it pulls un-needed dependencies.

### DIFF
--- a/scriptmodules/supplementary/emulationstation.sh
+++ b/scriptmodules/supplementary/emulationstation.sh
@@ -126,8 +126,9 @@ function depends_emulationstation() {
         libboost-system-dev libboost-filesystem-dev
         libboost-date-time-dev libfreeimage-dev libfreetype6-dev
         libcurl4-openssl-dev libasound2-dev cmake libsdl2-dev libsm-dev
-        libvlc-dev libvlccore-dev vlc
+        libvlc-dev libvlccore-dev rapidjson-dev
     )
+
 
     isPlatform "x11" && depends+=(gnome-terminal)
     getDepends "${depends[@]}"


hhromic:
Thanks a lot for the `vlc` purging! so it wasn't necessary all this time?
I always was so dissappointed by the amount of deps emulationstation needed to pull. I thought it depended dynamically on `libvlc*` stuff.  Does this mean ES is actually statically linked to VLC libs?

joolswills:
VLC is needed for some of its dependencies. Probably something in vitro history.

I'll add the new dependency but not removing VLC for now

hhromic:
Right, the commit is this: 48282c3

The required dependency was `vlc-nox`, which is no longer available on Ubuntu.
I will just leave this info here for future investigation on how to slim-down EmulationStation deps.

joolswills:
There was more information elsewhere - can't remember. It was easier to just include vlc (originally vlc-nox which had a few less) dependencies than all the other packages. I don't think this is an issue and if for example some other packages are renamed, it's safer to stick with vlc.

joolswills:
vlc-bin is not enough afair. I don't think we need to change this.

joolswills:
I looked into this (and remember I had done this before). You need libvlc-bin and vlc-plugin-base which also pulls in vlc-data so there's less savings than you may think (and it's safer to just pull in vlc - incase they change packaging again).

hhromic:
@joolswills  thanks for looking into it, I was also going to take a look myself next week when I had more time avail, so I really appreciate your time.

So looks like it is a lost cause to try to slimdown all these deps with the current ES implementation. Unfortunately it would then require a deeper decoupling in ES. Such a pity as none of VLC is necessary on an RPI using omxplayer :(

Again, thanks for checking and leaving the explanation here for future reference.

joolswills:
VLC is necessary on a RPI depending on theme. Some themes won't work with omxplayer as it will not display correctly (since it can't be displayed under things).

hhromic:
Ah good point, indeed omxplayer displays on a higher dispmanx layer. I didn't know that some themes layered the videos under other elements, requiring VLC that renders on a texture.
I guess then that there is really nothing to be done unless another (lighter) video framework is used in ES.

tomaz82:
As a sidenote, this script still pulls all the boost libraries? Those shouldn't be needed any more.

cmitu:
@tomaz82 yes, it's used by both the - now - `stable` and the `dev`. While `rapidjson` libs are needed for the new `dev`, the `boost` ones are needed for `stable`.
I guess that once the `dev` becomes `stable`, they can be erased (finally).

cmitu:
> I didn't know that some themes layered the videos under other elements, requiring VLC that renders on a texture.

It's not only that, `omxplayer` works only on a RPI, removing `vlc` (render-to-texture) rendering would make the `video` tag broken on Windows/Ubuntu or other ARM (non-RPI) based distros.

hhromic:
@cmitu I was being confusing, sorry, I didn't mean to remove VLC completely, but just for the RPI platform where `omxplayer` is available, accelerated and small enough.
I actually thought that VLC was added in the first place to support non-RPI platforms such as the ones you mention, but now I realise that VLC is used more than just supporting more platforms.

cmitu:
@hhromic no worries. Actually `vlc` can still be a viable candidate -performance wise - since it's been patched by the RPF guys to use the [HW decoder](https://www.raspberrypi.org/blog/raspbian-update-november-2018/). Alas, the patch introduced the 'switch the blue-with-red' bug which makes it right now unusable on the Pi (still unsolved).

hhromic:
Thanks for removing the obsolete `libboost` dependencies :)
```
The following packages will be REMOVED:
  libboost-date-time-dev* libboost-date-time1.62-dev* libboost-date-time1.62.0* libboost-filesystem-dev* libboost-filesystem1.62-dev* libboost-filesystem1.62.0*
  libboost-serialization1.62-dev* libboost-serialization1.62.0* libboost-system-dev* libboost-system1.62-dev* libboost-system1.62.0* libboost1.62-dev*
0 upgraded, 0 newly installed, 12 to remove and 0 not upgraded.
After this operation, 120 MB disk space will be freed.
```

Not bad of a saving (120 MB) !

PR #2610, opened 2019-02-10
fbalpha - changed version number

LodanZark:
The current fbalpha version it's 0.2.97.44, the 0.2.97.43 version it's now known as fbalpha2018:
https://github.com/barbudreadmon/fbalpha2018


joolswills:
.44 is still wip. I will update when .44 is finalised.

PR #2611, opened 2019-02-11
Makes time display 12h/24h according to locale in motd greeter

Gemba:
Just a small PR. Display the time in the locale of the retropie user.


joolswills:
Thanks.

PR #2618, opened 2019-02-16
running git fetch before checkout to prevent a stale local repo

Gemba:
Fixes: `error: pathspec '...' did not match any file(s) known to git`
when running `./retropie_package.sh <module>` and a local `.git/` is present for that <module>.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -352,6 +352,7 @@ function gitPullOrClone() {
 
     if [[ -d "$dir/.git" ]]; then
         pushd "$dir" > /dev/null
+        runCmd git fetch
         runCmd git checkout "$branch"
         runCmd git pull
         runCmd git submodule update --init --recursive
@@ -1222,7 +1223,7 @@ _EOF_
 ## @param cmd commandline to launch
 ## @brief Adds a new emulator for a system.
 ## @details This is the primary function for adding emulators to a system which can be
-## switched between via the runcommand launch menu 
+## switched between via the runcommand launch menu
 ##
 ##     addEmulator 1 "vice-x64" "c64" "$md_inst/bin/x64 %ROM%"
 ##     addEmulator 0 "vice-xvic" "c64" "$md_inst/bin/xvic %ROM%"


hhromic:
@Gemba do you have a concrete set of steps to reproduce this issue? In the code, the `git pull` command in the sequence is essentially a `fetch` and a `merge`.  The fetch command might need to be done before the checkout as you propose (then the pull command is redundant), but it would be nice to fully understand the error you report too. Thanks!

Gemba:
I came across this while recompiling skyscraper. IIRC I had this effect since 2.9.5 to 3.0.3 from skyscraper. It should be reproducable since a checkout of a local repo in my example of skyscraper of 3.0.2 to 3.0.3 without fetching the latest branch/tag information will surely show this effect.

As I ran into it the first time I simply restarted the retropie_packages.sh script, of course it then worked since the git pull was performed before the checkout of the new tag. HTH

joolswills:
It should only do a checkout if there is already no .git folder, so not sure how to reproduce this issue. Please can you provide steps to reproduce.

cmitu:
Here's what I think happened. @Gemba used `retropie_packages.sh` directly with `source`, `build`, `install`, but without `clean`. 
The 1st time this happened, `skyscraper` was on 2.9.5, then he tried to do the same when `3.0.3` was released, but the `source` step didn't run a new _git fetch_, so the `3.0.3` branch was not present (hence the error).

I'd say this is not the intended method for running an installation through `retropie_packages.sh`. 
In this particular case, the `skyscraper` module is always using the latest tagged release to install the package (the tagged release is determined dynamically at runtime through the Github API). Other packages my be using a fixed tagged release or a named branch (`stable`, etc.), so you'd not encounter this problem using `source`, `build`, `install` or have a different way to get the sources (i.e. direct download) so the `source` step is not always guaranteed to get the latest correct source tree without `clean` first.

hhromic:
Managed to reproduce this now, thanks to the comments from @cmitu .
1. `sudo ./retropie_packages.sh skyscraper sources`
2. Edit `scriptmodules/supplementary/skyscraper.sh` and force a non-existing branch in `gitPullOrClone`. As @cmitu says, this can happen with non-static branches such as with skyscraper where they might not be in the original pull/clone (done in 1).
3. Try again `sudo ./retropie_packages.sh skyscraper sources` and it will fail because naturally the new branch is not in the current state and must be re-fetched to show up.

The correct solution as @Gemba suggest should be to do a fetch first before any checkout/merge. This will ensure any new remote branch is fetched before attempting to check it out. Moreover, I think the later `git pull` is not necessary (but I might be wrong).

> I'd say this is not the intended method for running an installation through `retropie_packages.sh`.
In this particular case, the `skyscraper` module is always using the latest tagged release to install the package (the tagged release is determined dynamically at runtime through the Github API). Other packages may be using a fixed tagged release or a named branch (`stable`, etc.), so you'd not encounter this problem using `source`, `build`, `install`. However, they may have a different way to get the sources (i.e. direct download) so the `source` step is not always guaranteed to get the latest correct source tree without `clean` first.

While it is true that this workflow is not regular, I think this should be addressed as `gitPullOrClone` was designed to handle existing local git repositories (hence why it checks for an existing `.git`) and the issue reported is a valid bug IMHO. Always running `clean` first is not strictly mandatory/assumed and is actually a very common work pattern for debugging scriptmodules.

hhromic:
Wrong, doing `git fetch` doesn't help here actually. Here is why.

The original clone is done using `git clone --recursive --depth 1 --branch <branch>` which is only set to track the particular branch `<branch>`. If a different branch is requested later, the `git fetch` command doesn't fetch it and the checkout fails anyway. For example for skyscraper:
```
$ git clone --recursive --depth 1 --branch 3.0.2 "https://github.com/muldjord/skyscraper"
$ git -C skyscraper fetch
$ git -C skyscraper checkout 3.0.4
error: pathspec '3.0.4' did not match any file(s) known to git.
$ git -C skyscraper config list
remote.origin.fetch=+refs/tags/3.0.2:refs/tags/3.0.2
```
This means that as @cmitu says, using dynamic branches on local git repos is not currently supported by the current design of `gitPullOrClone`.

joolswills:
it was designed to handle existing repos, but not really if a branch is changed and the source is still there. But sources are cleared usually by retropie-setup - we don't leave them around in most cases. 

This function is very old (although has changed somewhat over the years), and already contains a bunch of logic. Just got to decide the best way of handling such situations. I'm happy for suggestions and improvements, but as it's a function that is used often, we need a fair amount of testing if major changes happen :-) Cheers.

Gemba:
As @cmitu outlined I got into this state by using undocumented targets of `retropie_packages.sh`, while not touching any supplied scripts.
I am not freakin' out on this, overall it is a corner case in which not many retropie user will run into I assume. 
By the time I could rethink the situation with your information, for now I can live with the workaround (run script twice).

hhromic:
@joolswills if the interest to improve `gitPullOrClone` is there, then I can offer myself to give it a try as I finally have more time nowadays :)

As you mention, it's a fundamental function, so I will prepare a suitable test suite and then improve the function accordingly. Also, seems that the function was designed with older versions of git in mind too, i.e. `1.x` series, so this can be further modernised aswell considering the current `2.x` series (`2.1.4-2.1` in Jessie and `2.11.0-3` in Stretch) that have improved commands nowadays.

Anyway, maybe we can just close this PR for now and I will be sending a new one soon with a revised `gitPullOrClone` addressing this issue while keeping compatibility.

joolswills:
ok. closing. thanks.

PR #2623, opened 2019-02-19
helpers - fix printMsgs() return value

hhromic:
Without an explicit `return` command, functions in bash default to returning the exit value of the last command executed.

The current implementation of `printMsgs()` uses `[[..]]` tests for deciding on the output type (dialog, heading, console), however if the last test (for heading) is false, then this will be the final exit value to be returned from `printMsgs()`. In other words, for all cases except "heading", it always will return `1` (false).

This can be solved by using if-elif-else blocks or, more elegantly, a switch-case block as in this patch. A correct return value is obtained by returning after dialog/echo on error.

---

Example/Tests:
```
# current implementation, "OK" is not printed, except if using "heading"
$ source scriptmodules/helpers.sh
$ printMsgs "console" "Hello World" && echo OK
Hello World

# using implementation in this PR, "OK" is printed properly except on errors
$ source scriptmodules/helpers.sh
$ printMsgs "console" "Hello World" && echo OK
Hello World
OK
$ printMsgs "console" "Hello World" >&- && echo OK  # force echo to fail
-bash: echo: write error: Bad file descriptor
```

I discovered this bug when testing `gitPullOrClone` that was always returning "false" no matter if the git operations fail or not, because of the last `printMsgs` in the function.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -28,6 +28,7 @@ function printMsgs() {
         [[ "$type" == "console" ]] && echo -e "$msg"
         [[ "$type" == "heading" ]] && echo -e "\n= = = = = = = = = = = = = = = = = = = = =\n$msg\n= = = = = = = = = = = = = = = = = = = = =\n"
     done
+    return 0
 }
 
 ## @fn printHeading()


joolswills:
it's not enough just to put a return at the end of this function ?

joolswills:
and / or adding a return to the affected function also.

hhromic:
> it's not enough just to put a return at the end of this function ?

Thought of that too, it would need to be a `return $retval` and then handling `$retval` through the function which is more cumbersome IMHO. You can also think of just `return 0` but that will obscure/ignore errors. My proposed solution, while I agree is a bit more lines of code, is simpler, cleaner correct and less error-prone. I hope you see it the same.

> and / or adding a return to the affected function also.

That in my opinion would be just moving the problem further away from its source, i.e. `printMsgs` not returning with standard convention values (0=OK, <>0=Error). I really think this behaviour should be addressed here in this function, making any artificial returns in other functions unnecessary.

I acknowledge that this function is never directly checked in any place for its return value, but given how bash propagates return values, it can introduce unwanted/obscure effects such as other functions unintentionally returning "false" if `printMsgs` was one of the last functions called in them.

I hope this better motivates the PR :)

joolswills:
the printMsg functions can't really fail though (well it's unlikely to happen), and even if it does things should carry on so return 0 is fine. Not sure I want that function to ever fail :)

I really meant "and". not "or". making sure the git function always returns something.

scriptmodules/helpers.sh
```diff
@@ -20,13 +20,19 @@
 function printMsgs() {
     local type="$1"
     shift
-    if [[ "$__nodialog" == "1" && "$type" == "dialog" ]]; then
-        type="console"
-    fi
+    [[ "$__nodialog" == "1" && "$type" == "dialog" ]] && type="console"
```
joolswills: Although more compact, I prefer the readability of the original code.

hhromic:
> the printMsg functions can't really fail though (well it's unlikely to happen), and even if it does things should carry on so return 0 is fine. Not sure I want that function to ever fail :)

The return value is never checked by any current code so it won't stop any execution at the moment, however it at least will not propagate incorrect return values (currently `1` on non-heading messages), which is the main purpose of this PR :)

> I really meant "and". not "or". making sure the git function always returns something.

I'm surely considering for the improved `gitPullOrClone` to also return proper values on OK/error conditions, basically to respect git's own return values. Again this is not currently checked by any other code but is good practice to keep behaviour consistent, specially when used from other functions.

I will revert the compact code as you suggest. Thanks for considering!

joolswills:
I would still rather a single return in this function Vs the || return. It's not an important function and should never return an error code imho even if somehow the echo fails.

hhromic:
@joolswills sure no problem, updated the PR to just `return 0` at the end of the function so it doesn't return `1` inadvertently, which is good.

joolswills:
btw it may just be easier to do an issue for small changes like this as more time was spent reviewing this for one line of code than it would have been just to report the issue and for me to add the line :-)

joolswills:
I didn't have an issue with your use of the case/esac change btw. but it's not important either way.

PR #2624, opened 2019-02-20
Mali-DRM support

dsx724:
This pull request adds initial platform support for KMS/DRM systems using ARM's Mali Utgard 4xx generation binary blobs for acceleration. It does not use the legacy FBDEV and exhibits tear free video output. SDL 2.0.8 was swapped out for SDL 2.0.9 because 2.0.9 fixes a resource locking bug when emulationstation exits and retroarch starts. We upstream that fix for that a few months ago but did not get around to submitting this pull into RetroPie.

I would suggest making targets more generic so that we don't riddle the code base with board specific targets when they really should be architecture specific. These changes should support armhf and arm64 systems.

One area of uncertainty is include and library file paths. We elected to use /usr/local/ since these blobs are usually not native to distributions. We currently have a few mali-drm-gles3 platforms based on newer Mali Midgard and Mali Bitfrost as well and we will make mali-drm-gles3 targets for those as well.

With regards to Lima and Panfrost projects, both are targeting GLES2. When they are complete, we can just add another target drm-gles2 instead of mali-drm-gles2.

Please review and let us know if any changes are necessary. We have some other unfinished patches in the mali-drm-gles2-next branch that we will clean up and send pull requests for in March base on the feedback and edits needed for this.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -390,3 +390,9 @@ function platform_vero4k() {
     __platform_flags="arm armv7 neon vero4k gles"
 }
 
+function platform_armv8-mali-drm-gles2() {
+    __default_cflags="-O2 -march=armv8-a+crypto+crc -mtune=cortex-a53 -ftree-vectorize -funsafe-math-optimizations"
+    __default_asflags=""
+    __default_makeflags="-j`nproc`"
+    __platform_flags="armv8 kms gles"
+}


dsx724:
We have tested this on Libre Computer Board AML-S905X-CC, ALL-H3-CC H3, ALL-H3-CC H5, and ROC-RK3328-CC based on Linux 4.19.

joolswills:
Not had a chance to look at this yet, but will when I have some free time. Actually we already lean towards architecture for logic Vs actual target. Eg. Mali / Kms etc.

As more "systems" are added, we will define this more.

Thanks for the contribution. Appreciated.

microplay:
> ALL-H3-CC H5 (ARMv8)

dsx724 have you a workable version with gl support for cc h5?

dsx724:
@mpcore2 H5's Mali GPU has some slight differences from others. You get kernel messages spamming pixel core failures and graphical glitches. Hopefully someone in sunxi gets a change to look at it and resolve it.

joolswills:
This will need to be rebased due to switching to SDL 2.0.9 - but should simplify it also for the SDL2 part.

I agree with generic targets and the platform flags should be used for most logic where possible. There are improvements needed in this area which I plan to address after retropie 4.5.

Apologies again for delays.i am keen to get this included

dsx724:
Is RetroPie planning to move to 2.0.9? We can submit the handle release to the RetroPie SDL 2.0.8 if there's no plans to go to 2.0.9 and rework this patchset.

joolswills:
we have already moved to 2.0.9 which is why this PR needs rebasing. Thanks.

dsx724:
@joolswills I will respin this next week. Meantime can you comment about these dependencies that I encountered?

sdl2: libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxrandr-dev libxss-dev libxt-dev libxxf86vm-dev libgl1-mesa-dev
emulationstation: rapidjson-dev libfreeimage-dev libcec-dev libp8-platform-dev

PR #2625, opened 2019-02-20
WIP: bluetooth: improve scan timeout functionality

psyke83:
Ensure that bluetoothctl will never cause the script to hang due to
infinitely repeating status change messages.

Also reduce timeout to 15 seconds and clarify DualShock instructions;
the PS button needs to be pressed if the controller is still scanning
during cable connect to ensure it binds to the USB connection.


joolswills:
thanks.

PR #2627, opened 2019-02-23
pegasus-fe: Added Raspberry Pi 3 release

mmatyas:
There is now a Raspberry Pi 3-optimized build of Pegasus, which might provide improved performance.

### DIFF
--- a/scriptmodules/supplementary/pegasus-fe.sh
+++ b/scriptmodules/supplementary/pegasus-fe.sh
@@ -40,13 +40,7 @@ function install_bin_pegasus-fe() {
     # find out which platform's package we'll need
     local platform
     isPlatform "x11" && platform="x11"
-    if isPlatform "rpi"; then
-        if isPlatform "armv6"; then
-            platform="rpi1"
-        else
-            platform="rpi2"
-        fi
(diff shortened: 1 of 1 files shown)


joolswills:
This could be simplified by using something like.. if isPlatform "rpi" then platform="$__platform" or similar. On phone so untested.

mmatyas:
I see, but is that alright? `$__platform` seems kind of an internal variable to me; is it guaranteed that its value will match rpi1/2/3 (ie. it's not a field in `$__platform_flags` or somewhere else instead)?

hhromic:
@mmatyas on RPI devices, `$__platform` is guaranteed to be one of `rpi1`, `rpi2` or `rpi3`:
https://github.com/RetroPie/RetroPie-Setup/blob/a571418bbd19b74f05eedf162d8303dfdabebbb6/scriptmodules/system.sh#L221-L246

joolswills:
It's fine to use that variable as @hhromic says above. Thanks.

mmatyas:
All right then, I've just pushed an update with this change.

joolswills:
Thanks.

PR #2628, opened 2019-02-23
Fix: downloadAndExtract doesnt function how the comment describes

sigboe:
The comment says you can pass more arguments the the extraction program. But you cant do that $opts only grabs the third argument to the function. 

This change grabs all arguments from the third argument and after to an array. Then passes $opts as an array to the extraction programs. A very simple change really

Please review and accept, I really need this feature to not make cumbersome workaround in my upcoming script module.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -63,10 +63,10 @@ function bluez_cmd_bluetooth() {
     mkfifo "$fifo"
     exec 3<>"$fifo"
     local line
-    while read -t "$2"; do
+    while true; do
         _slowecho_bluetooth "$1" >&3
         # collect output for specified amount of time, then echo it
-        while read -t "$2" -r line; do
+        while read -r line; do
             printf '%s\n' "$line"
             # (slow) reply to any optional challenges
             if [[ -n "$3" && "$line" =~ $3 ]]; then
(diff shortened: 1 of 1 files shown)


sigboe:
Only thing to note is that the first argument has to be the number of components to strip `--strip-components`, if you don't want to use that function, you can have the number be `0`.

sigboe:
If you want I can add a check, if the first element in the array is not a number, then I can add a 0 in front of the array.

sigboe:
I added the a commit to the pull request to not fail if the user does not follow the documentation in the user would dare to not put it there them selves.

joolswills:
Thanks. I prefer not to have the safety code but your fix is generally ok except I would use "shift 2" and @$ rather than selecting array range.

sigboe:
Ops, I closed this one by accident, it autoclosed when I removed the commits, opened a new pull request with the changes you wanted here. https://github.com/RetroPie/RetroPie-Setup/pull/2630

joolswills:
BTW, Instead of removing the commits, you can add another commit with a correct, the squash the commits into one and force push to your branch. This PR would then update. (for future reference).

sigboe:
@joolswills  Oh, tanks for the tip. I'm only Jr. Sysadmin at work, so I need to learn all this stuff.

PR #2595, opened 2019-02-04
add hexen2 scriptmodule to ports

tpo1990:
I have successfully created hexen2 script module for RetroPie project from source port: Hammer of Thyrion.
http://uhexen2.sourceforge.net/

PR #2630, opened 2019-02-25
downloadAndExtract() Allow sending additional options to the unzip pr…

sigboe:
downloadAndExtract() Allow sending additional options to the unzip program, like how the description of the function says.

I tried to edit the previously opened [pull request](https://github.com/RetroPie/RetroPie-Setup/pull/2628), but when I deleted all the commits the pull request got closed. :smile: 

I have made the changes the way @joolswills asked.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -947,11 +947,14 @@ function applyPatch() {
 ## @details Download and extract an archive, optionally stripping off a number
 ## of directories - equivalent to the tar `--strip-components parameter`. For
 ## zip files, the strip parameter can contain additional options to send to unzip
+## or tar. If you send additional parameters to tar, you need to start with the
+## number of components to strip, this number can be 0.
 ## @retval 0 on success
 function downloadAndExtract() {
     local url="$1"
     local dest="$2"
-    local opts="$3"
+    shift 2
+    local opts=( "$@" )
 
     local ext="${url##*.}"
     local cmd=(tar -xv)
@@ -973,7 +976,7 @@ function downloadAndExtract() {
             local tmp="$(mktemp -d)"
             local file="${url##*/}"
             runCmd wget -q -O"$tmp/$file" "$url"
-            runCmd unzip $opts -o "$tmp/$file" -d "$dest"
+            runCmd unzip "${opts[@]}" -o "$tmp/$file" -d "$dest"
             rm -rf "$tmp"
             ret=$?
     esac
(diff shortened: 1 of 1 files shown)


joolswills:
There is a minor change needed, but I think it might just be easier for me to make the changes myself than update this PR and squash etc. But thanks for the heads up.

joolswills:
BTW checking the original code code - this worked before due to the fact `$opts` is not quoted when sent to zip. with your changes it should be quoted though. What problem were you having with the function? Are you sure it's broken and if so can you give an example?

joolswills:
And the documentation for the function would need to be updated also with your changes. Please give me your example that doesnt work and I will adjust the function as needed.

joolswills:
allowing additional arguments is more flexible of course, as you could include parameters with spaces then, but I may be inclined if changing it to multiple arguments, so change the way the tar code works with the stripping of components.

sigboe:
@joolswills I was unable to only extract one file from a tar archive. Beacuse it didn't pass the next argument into tar.

joolswills:
ok. I will extend it to support passing additional arguments to tar - your PR title didn't mention this but instead said it was a problem with unzip parameters.

hhromic:
@joolswills the scriptmodule for hexen2 in PR #2595 also would benefit from this improvement. See my comments left to the author https://github.com/RetroPie/RetroPie-Setup/pull/2595#issuecomment-467061776, in particular for the `game_data_hexen2()` function.
In that function, the author also needs to extract only one folder from within the tgz file.

sigboe:
@joolswills Here is an example of what I need to do without this PR

```
downloadAndExtract "http://constexpr.org/innoextract/files/snapshots/innoextract-${innoversion}/innoextract-${innoversion}-linux.tar.xz" "$md_inst/innotemp" 2
		mv "$md_inst/innotemp/armv6j-hardfloat/innoextract" "$md_inst/"
rm -rf "$md_inst/innotemp"
```

Here is what I can do with this PR

    	isPlatform "arm" && downloadAndExtract "http://constexpr.org/innoextract/files/snapshots/innoextract-${innoversion}/innoextract-${innoversion}-linux.tar.xz" "$md_inst" 3 innoextract-${innoversion}-linux/bin/armv6j-hardfloat/innoextract

You see I just use the following argument to say what file I want. Instead of extracting many and pick the file I want. I can even drill down deeper, since there are several files of the same name (in directories for different architectures), and they overwrite each other if I strip 3 without picking one to extract instead of extracting everyone.

sigboe:
@joolswills 

1. I updated both the comment, and quoted the array for unzip, like you wanted.
2. I fixed a typo in the tar cmd array (ops it was not in my previous PR)
3. I squashed it.
4. I learned to squash a commit (but I still need to google it the next time lol)

joolswills:
Thanks. Appreciate the changes - however, I may not use this PR as I am considering to re-factor the arguments so the strip components part is not included at all (and adjust our other modules etc).

hhromic:
@joolswills was thinking exactly on the same :), that it should be more generic. Great you want to improve it.

joolswills:
https://github.com/RetroPie/RetroPie-Setup/pull/2634

I just threw this together quickly. I will need to double check everything looks and test before merging.

sigboe:
This has been implemented in another way now.

PR #2631, opened 2019-02-25
fix prefer screenshots when using mame downloader

fillier:
The default mame settings are set **after** the other options. The result is that the "prefer screenshots" option is overridden. Moved the default settings before the custom options and reworked the params to prevent duplicates.


joolswills:
Thanks.

Issue #1805, opened 2016-12-27
Look into workaround(s) for X permissions issue


joolswills:
https://retropie.org.uk/forum/topic/6304/desktop-from-ports-fails-to-load-4-1

Due to xserver update.

joolswills:
Related issues

https://bugzilla.redhat.com/show_bug.cgi?id=1177513
https://bugs.launchpad.net/ubuntu/+source/xinit/+bug/1562219

Note that running `startx 2>myerror.log` from terminal fails without the startx patch from the redhat bugtracker. It isn't enough to fix it when launching from ES though as it will get a permission denied on /dev/vt0. - might be able to use openvt or something as a workaround (want to avoid adjusting vt permissions which has been a suggested workaround)

joolswills:
Have a solution and a fix.

With the older xserver and the xwrapper we were able to allow anybody to start X. Now we can't do that - but we can start X on the TTY/VT that we were logged in on that ES was launched from. We can save that before launching ES (because when ES is launched, it is not attached to a tty), and then pass it in to xinit or startx on launch via runcommand.

for example instead of `startx` we do `startx -- vt1 -keeptty`

PR #2633, opened 2019-02-26
pegasus-fe: Fixed launching X11 applications

mmatyas:
[Similarly to ES](https://github.com/RetroPie/RetroPie-Setup/blob/34426d09db82e3d59941db0b7a47d67a681a2cd9/scriptmodules/supplementary/emulationstation.sh#L217), the [X11 permission workaround](https://github.com/RetroPie/RetroPie-Setup/issues/1805#issuecomment-269387535) is also necessary for Pegasus to properly launch X11 applications. Based on the ES package code, this patch creates a similar launcher script on install to set up the TTY before launch.

### DIFF
--- a/scriptmodules/supplementary/pegasus-fe.sh
+++ b/scriptmodules/supplementary/pegasus-fe.sh
@@ -60,6 +60,28 @@ function install_bin_pegasus-fe() {
     # download and extract the package
     printMsgs "console" "Download URL: ${asset_url}"
     downloadAndExtract "${asset_url}" "$md_inst"
+
+    # create launcher script
+    cat > /usr/bin/pegasus-fe << _EOF_
+#!/bin/bash
+
+if [[ \$(id -u) -eq 0 ]]; then
+    echo "Pegasus should not be run as root. If you used 'sudo pegasus-fe' please run without sudo."
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/pegasus-fe.sh
```diff
@@ -60,6 +60,28 @@ function install_bin_pegasus-fe() {
...
+export TTY="\${tty:8:1}"
+
+clear
+"$md_inst/pegasus-fe" "\$@"
+_EOF_
+    chmod +x /usr/bin/pegasus-fe
+}
+
+function remove_pegasus-fe() {
+    rm -rfv /usr/bin/pegasus-fe
```
joolswills: Is -r needed?

mmatyas: Not really actually, but now that you mention it, how about `v`? Would you prefer just `-f`?

joolswills: `-f` on it's own is fine. can use -v if you want it to be verbose.

mmatyas: Ok, changed to just `-f`, that seems more common with the other scripts.

PR #2634, opened 2019-02-26
helpers / downloadAndExtract - allow additional parameters for both t…

joolswills:
…ar and unzip

 * change existing calls which used --strip-components
 * use multiple parameters for arguments which should allow additional arguments with spaces
 * implements #2630

### DIFF
--- a/scriptmodules/admin/crosscomp.sh
+++ b/scriptmodules/admin/crosscomp.sh
@@ -61,19 +61,19 @@ function sources_crosscomp() {
             ;;
     esac
 
-    downloadAndExtract "ftp://gcc.gnu.org/pub/gcc/infrastructure/isl-${pkgs[isl]}.tar.bz2" isl 1
-    downloadAndExtract "http://www.bastoul.net/cloog/pages/download/count.php3?url=./cloog-${pkgs[cloog]}.tar.gz" cloog 1
+    downloadAndExtract "ftp://gcc.gnu.org/pub/gcc/infrastructure/isl-${pkgs[isl]}.tar.bz2" isl --strip-components 1
+    downloadAndExtract "http://www.bastoul.net/cloog/pages/download/count.php3?url=./cloog-${pkgs[cloog]}.tar.gz" cloog --strip-components 1
 
-    downloadAndExtract "https://ftp.gnu.org/gnu/binutils/binutils-${pkgs[binutils]}.tar.gz" binutils 1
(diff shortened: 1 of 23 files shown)


hhromic:
Good stuff!
Made a quick extractor for every call of `downloadAndExtract`:
```
grep "downloadAndExtract .*" -o -h -r \
    | sed -e "s/(.*)downloadAndExtract /downloadAndExtract /" \
    > dae-calls.txt
```

Tested the standard cases and some exotic ones, all working fine:

    downloadAndExtract <url> <dir>
    downloadAndExtract <url> <dir> --strip-components 1
    downloadAndExtract <url> <dir> -j
    downloadAndExtract <url> <dir> -j -LL

Also found you missed just one call to patch :)
https://github.com/RetroPie/RetroPie-Setup/blob/8af428cb5098a8d0112f4383717b27404b3dc9de/scriptmodules/emulators/hatari.sh#L30

Idea: how about some `printMsgs` for progress report? Thinking on something like this:
```diff
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -972,7 +972,9 @@ function downloadAndExtract() {
             is_tar=0
             local tmp="$(mktemp -d)"
             local file="${url##*/}"
+            printMsgs "console" "Running: wget -q -O\"$tmp/$file\" \"$url\""
             runCmd wget -q -O"$tmp/$file" "$url"
+            printMsgs "console" "Running: unzip ${opts[*]} -o \"$tmp/$file\" -d \"$dest\""
             runCmd unzip "${opts[@]}" -o "$tmp/$file" -d "$dest"
             rm -rf "$tmp"
             ret=$?
@@ -982,6 +984,7 @@ function downloadAndExtract() {
         mkdir -p "$dest"
         cmd+=(-C "$dest" "${opts[@]}")

+        printMsgs "console" "Running: ${cmd[*]} < <(wget -q -O- \"$url\")"
         runCmd "${cmd[@]}" < <(wget -q -O- "$url")
         ret=$?
     fi
```

joolswills:
Thanks for your thorough testing. Will fix the missed call. Will consider more verbosity :-)

sigboe:
Looking forward to this merging!

joolswills:
I will merge this now. Verbosity can wait (I don't think it needs to be quite as verbose as mentioned, but I may add something).

hhromic:
Sure, I was just using that while testing to see what parameters were given to each program. Of course it doesn't need to be so verbose for the end-user. Take for example `gitPullOrClone`, it can just say from where it's fetching the archive. Nice improvement in any case!

PR #2635, opened 2019-02-28
Added OpenBOR 3400 Scriptmodule

crcerror:
This is OpenBOR with PAK support
So no extraction steps must be taken to play older modules
This version can reside next to all OpenBOR branches and just adds the binary to be selected as emulator.
This version is very similar to the one in orignal branch.... but does not need extracted files.

@hhromic We make it easy step by step okay?
Thanks for your help

### DIFF
--- a/scriptmodules/ports/openbor-3400.sh
+++ b/scriptmodules/ports/openbor-3400.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="openbor-3400"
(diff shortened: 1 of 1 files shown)


joolswills:
I had a wip port of openbor. I did check this PR but wanted to revisit my module beforehand.

I can't promise this will be accepted but I appreciate the work.

crcerror:
@joolswills I strip the filename with echo because I want to use a file structure [like I've posted here](https://retropie.org.uk/forum/topic/19326/openbor-6xxx-openbeta-testphase/80). Then all branches and installs can reside side by side.
I would not through away the old fork that RetroPie uses now because it is incredible fast and can be usefull for older platforms.
Thank you

crcerror:
@joolswills May I ask the status of your code review?

joolswills:
No ETA. As I said I want to revisit my own openbor work first. It's not going to happen anytime soon. You can always submit this to the retropie extras repo.

joolswills:
(Also I am not sure I want to include this currently).

crcerror:
Whether this module finds its place in the official repo or not is not so important. I will continue to maintain the main version 6.xxx in the RetroPie forum. The 3.400 version presented here has the big advantage to directly access PAK files (and load them via CLI). I think the OpenBOR engine is a powerful tool for retrogaming. I choosed in first instance this version because it can replace (or be added) to current (RetroPie-)version.

PR #2639, opened 2019-03-05
Add "Jump 'n Bump" port to experimental

hhromic:
I found this little game recently and decided to make a scriptmodule for RetroPie.
It's a very simple but fun multiplayer game, ideal for kids -- and not so kids too :)

About the Game:
https://en.wikipedia.org/wiki/Jump_%27n_Bump

Also found it was previously requested in the RetroPie Forum:
https://retropie.org.uk/forum/topic/12370/jump-n-bump-as-a-port

Scriptmodule Features:

* Installs extra levels from https://wiki.debian.org/Games/Jumpnbump
* Creates a launcher for selecting levels and starting net play
* Configurable options such as "nogore" (no blood for kids)
* Installs tools for creating own levels (gobpack, jnbpack, jnbunpack)

Considerations:

* When waiting for network players or connecting to a server, the game can't be killed (CTRL+C). This is not a runcommand nor RetroPie limitation, but the game is ignoring signals while waiting. Must be addressed upstream.
* While the game supports joysticks for gameplay, it can't currently be used without a keyboard because you still need it for exiting via ESC or F12, i.e. no joystick support for this action.

Configurable Options (`$md_conf_dir/jumpnbump/options.cfg`):

    nogore=0|1           play without blood (set as default on first install)
    noflies=0|1          disable flies
    scaleup=0|1          play with doubled resolution (800x512)
    musicnosound=0|1     play with music but without sound

Let me know if you want to make adjustments.

### DIFF
--- a/scriptmodules/ports/jumpnbump.sh
+++ b/scriptmodules/ports/jumpnbump.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="jumpnbump"
+rp_module_desc="Jump 'n Bump, play cute bunnies jumping on each other's heads - Modernization fork"
+rp_module_help="Copy custom game levels (.dat) to $romdir/ports/jumpnbump"
+rp_module_licence="GPL2 https://gitlab.com/LibreGames/jumpnbump/raw/master/COPYING"
+rp_module_section="exp"
+rp_module_flags=""
+
+function depends_jumpnbump() {
+    getDepends libsdl2-dev libsdl2-net-dev libsdl2-mixer-dev libbz2-dev zlib1g-dev
+}
+
+function sources_jumpnbump() {
+    gitPullOrClone "$md_build" https://gitlab.com/LibreGames/jumpnbump.git
+}
+
+function build_jumpnbump() {
+    make clean
+    CFLAGS="$CFLAGS -fsigned-char" make PREFIX="$md_inst"
+}
+
+function install_jumpnbump() {
+    make PREFIX="$md_inst" install
+    strip "$md_inst"/bin/{gobpack,jnbpack,jnbunpack,jumpnbump}
+}
+
+function game_data_jumpnbump() {
+    local tmpdir="$(mktemp -d)"
+    local compressed
+    local uncompressed
+
+    # install extra levels from Debian's jumpnbump-levels package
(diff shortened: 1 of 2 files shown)


hhromic:
Btw, in the forum the user mentions that it segfaults with sound. I fixed that too (required building with signed char). Module and game tested.

scriptmodules/ports/jumpnbump.sh
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function build_jumpnbump() {
+    make clean
+    CFLAGS="$CFLAGS -fsigned-char" make PREFIX="$md_inst"
+}
+
+function install_jumpnbump() {
+    make PREFIX="$md_inst" install
+    strip "$md_inst"/bin/{gobpack,jnbpack,jnbunpack,jumpnbump}
```
joolswills: Please can you strip the binaries in the build stage?

scriptmodules/ports/jumpnbump.sh
```diff
@@ -0,0 +1,66 @@
...
+    # install extra levels from Debian's jumpnbump-levels package
+    local tmpdir="$(mktemp -d)"
+    downloadAndExtract "https://salsa.debian.org/games-team/jumpnbump-levels/-/archive/master/jumpnbump-levels-master.tar.bz2" "$tmpdir" --strip-components 1 --wildcards "*.bz2"
+    for compressed in "$tmpdir"/*.bz2; do
+        uncompressed="${compressed##*/}"
+        uncompressed="${uncompressed%.bz2}"
+        bzcat "$compressed" > "$romdir/ports/jumpnbump/$uncompressed"
+    done
+    chown -R $user:$user "$romdir/ports/jumpnbump"
+    rm -rf "$tmpdir"
```
joolswills: This might be worth putting in a game data function?

hhromic: Didn't notice a game data function was a RetroPie practice :) Done now.
I also made it a bit smarter by only installing levels that are not found already in the user roms folder.

scriptmodules/ports/jumpnbump.sh
hhromic: As I commented in the moonlight PR, stripping binaries should be done at the install phase as a good practice. However, I agree that for this particular module it won't affect doing it in the build phase as you request. If you still want to move it there and I will do it.

hhromic:
> The launch script as it's complicated may be better put into md_inst.

Not sure what you mean, the launch script is already being installed to `$md_inst` :)

joolswills:
I must have misread :)

scriptmodules/ports/jumpnbump.sh
joolswills: I'm ok then with leaving it.

joolswills:
thanks.

hhromic:
Thanks for merging.

At some point I was experimenting on using joy2key to control the launch script via gamepad (until I realised the game can't be closed without a keyboard anyway). Then I realised that I needed to copy/paste the `start/stop_joy2key` functions from `runcommand` to be able to start/stop joy2key in the same way.

So I was wondering maybe these two functions can be externalised into a `joy2key.sh` sourceable script for other scripts (and runcommand) to import them and not having to copy/paste redundant code?
For example it can be installed in `/opt/retropie/supplementary/runcommand/joy2key.sh` alongside `joy2key.py`. What do you think? are you interested on this idea?

PR #2640, opened 2019-03-05
Add moonlight gamestream client to experimental

hhromic:
After a lot of testing, debugging and fine-tuning in the forum, I think this is ready to be merged now.
I'm also adding a commit to remove the obsolete/deprecated Limelight module.

RetroPie Forum Thread:
https://retropie.org.uk/forum/topic/19872/moonlight-request-for-comments-and-testing

Let me know if you want to make any adjustments.


joolswills:
I've not got time to do a full review but as I'm very keen to ditch the broken limelight I'll merge and if anything comes up we can tweak later.

Thanks for a your work on this.

hhromic:
Thanks for merging. The module passed many iterations of refinement so all should be up and running quite stably by now. In any case as you say we can tune later if necessary.

Btw, if you wonder why I do `strip` in the install phase here, it is becasuse the `make install` command adjusts the binary rpaths, so I prefer to do the stripping at the very last step. GNU Make also prefers to do stripping as an install target (`install-strip`), see [here](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html), hence why I adopted the same approach.

PR #2642, opened 2019-03-06
skyscraper: small improvements to the UI

cmitu:
Small improvements and additions, based on the 3.x version and the feedback from the forum.
* removed the _Gather for all platforms_ entry from the menu, which should be now simpler. The gather action should now be similar to the gamelist generation. 
* added a warning before starting the actual online scraping, so it doesn't surprise the user(s). Mentioned the `Ctrl+C` shortcut to stop the scraping.
* added the _Vacuum_ action to the cache operations and modified the _Purge_ operations to use Skyscraper's options now.
* replaced all _system_ verbiage with _platform_, to be more consistent with Skyscraper's docs.

This version requires the 3.x version (released back on 23 Jan 2019). I'll update the wiki after the PR gets accepted.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -72,15 +72,17 @@ function _purge_skyscraper() {
 
 function _clear_platform_skyscraper() {
     local platform="$1"
+    local mode="$2"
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

PR #2644, opened 2019-03-09
lr-reicast - enable for RPI

joolswills:
* redirect stdin from </dev/null to avoid segfault on RPI
 * keep disabled for kms / mali targets for now
 * ref: https://retropie.org.uk/forum/topic/20042/reicast-vs-lr-reicast-perfomance


joolswills:
Not tested this PR yet. I have tested building it with the changes above manully. I want to look into the reason for needing the stdin redirect (also need to test if it works at this position in the emulators.cfg launch line before the --config option).

cmitu:
It seems to be related to running from `runcommand`, if I exit ES and launch manually (without the redirection), then it works fine.

cmitu:
Seems similar to https://github.com/libretro/reicast-emulator/issues/365.

steeeb:
Same needed for Tinkerboard (I pushed a change to the Makefile for Tinkerboard and lr-reicast works great).

joolswills:
Thanks. Will enable for Kms and merge later.

PR #2645, opened 2019-03-09
joy2key: fix sys.exit() by avoiding naked exceptions

psyke83:
The sys.exit() function actually calls an exception (SystemExit), but our
script handles naked exceptions which can result in the script not terminating
properly. Fix by ensuring that all exceptions specify the relevant exception type only.

This seems to resolve the issue of the leftover process on my Pi and Ubuntu 18.10, but perhaps wait before merging until it's confirmed by others.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -44,6 +44,7 @@ function printHeading() {
 function fatalError() {
     printHeading "Error"
     echo -e "$1"
+    joy2keyStop
     exit 1
 }
 
@@ -1043,9 +1044,8 @@ function joy2keyStart() {
     [[ -z "$__joy2key_dev" ]] || pgrep -f joy2key.py >/dev/null && return 1
 
     # if joy2key.py is installed run it with cursor keys for axis/dpad, and enter + space for buttons 0 and 1
-    if "$scriptdir/scriptmodules/supplementary/runcommand/joy2key.py" "$__joy2key_dev" "${params[@]}" & 2>/dev/null; then
+    if __joy2key_ppid=$$ "$scriptdir/scriptmodules/supplementary/runcommand/joy2key.py" "$__joy2key_dev" "${params[@]}" & 2>/dev/null; then
         __joy2key_pid=$!
-        sleep 1
         return 0
     fi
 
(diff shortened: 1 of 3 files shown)


psyke83:
This needs a bit more work. Closing for a bit.

crcerror:
@psyke83 Do you think it might be related with this one?
https://retropie.org.uk/forum/topic/18772

psyke83:
Explanation of behaviour without PR:

* Add debug:
```
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -131,6 +131,7 @@ def get_button_codes(dev_path):
     return btn_codes
 
 def signal_handler(signum, frame):
+    print 'signal_handler'
     close_fds(js_fds)
     sys.exit(0)
```

* Make sure controller is connected.

* Launch and interrupt (ctrl+c) after 1 second:
```
conn@satellite:~/RetroPie-Setup$ ~/RetroPie-Setup/scriptmodules/supplementary/runcommand/joy2key.py /dev/input/jsX kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20
^Csignal_handler
conn@satellite:~/RetroPie-Setup$
```

* Launch and immediately interrupt (ctrl+c until process exits):
```
conn@satellite:~/RetroPie-Setup$ ~/RetroPie-Setup/scriptmodules/supplementary/runcommand/joy2key.py /dev/input/jsX kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20
^Csignal_handler
^Csignal_handler
conn@satellite:~/RetroPie-Setup$
```

As you can see, signal_handler may be called but fail to exit the script.

psyke83:
> @psyke83 Do you think it might be related with this one?
> https://retropie.org.uk/forum/topic/18772

This issue causes orphaned joy2key processes to be left behind, so yes, it could be related. When I have time, I'll try to reproduce the issue with openbor.

joolswills:
Thanks for debugging this. Works well here on a setup that sometimes left joy2key running. Will merge - we can revisit if anyone notices any problems. Cheers.

PR #2646, opened 2019-03-10
dosbox: update to SVN 4194 with joystick patch

### LINKED ISSUE #2643: DOSBox requires the fully bindable Joystick Patch to be able to use the 4th and 5th axis
The current DOSBox have an ancient issue where only axis 0-3 can be bound, making the right stick and right trigger on playstation 4 controllers and an xbox controller useless. This put a halt to the keyboard 2 joypad project that attempts to add joypad controls to all DOS actiongames.

This issue have long since been patched in most popular releases of DOSBox, such as ECE and Daum, but the patch is missing in Retropie. It can however be manually patched sucessfully. Please see dudleydes post here; https://retropie.org.uk/forum/topic/21561/dosbox-does-not-detect-axis-4-or-5-on-ps4-and-xbox-360-controller

We hope you can include the patch so that it can be installed using the setup script rather than the command line.

Here is a post with the patch:
https://www.vogons.org/viewtopic.php?f=32&t=27452&p=314840

psyke83:
Patch via: https://www.vogons.org/viewtopic.php?f=32&t=27452&p=314840

Should resolve #2643.

### DIFF
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -24,9 +24,10 @@ function depends_dosbox() {
 
 function sources_dosbox() {
     local revision="$1"
-    [[ -z "$revision" ]] && revision="4006"
+    [[ -z "$revision" ]] && revision="4194"
 
     svn checkout https://svn.code.sf.net/p/dosbox/code-0/dosbox/trunk "$md_build" -r "$revision"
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks.

joolswills:
Updated binaries are being rebuilt and will be available shortly.

PR #2647, opened 2019-03-10
fixing edge case where multiplayer is broken for dreamcast

Adam7288:
Changed grep pipe so that it parses the last match for the controller mapping file. In the case where xbox userspace controllers were found, there was a userspace #2 which was matching for the first and second controller. This led to only the second controller being recognized.


joolswills:
Thanks.

PR #2650, opened 2019-03-11
retropiemenu: fix printing of own IP address

cmitu:
Fixes the IP addresss detection when the `ip` adds extra info to the route output (ex. _uid_). On Ubuntu 18.04:

    ip route get 8.8.8.8

outputs (2 lines):
```
8.8.8.8 via 10.0.1.1 dev ens160 src 10.0.1.88 uid 1000 
    cache 
```
In this case, the IP shown in the **Show IP** menu entry is `0` .

The change gets only the next value after the `src` field in the `ip` output.

### DIFF
--- a/scriptmodules/supplementary/retropiemenu.sh
+++ b/scriptmodules/supplementary/retropiemenu.sh
@@ -151,7 +151,7 @@ function launch_retropiemenu() {
             mc
             ;;
         showip.rp)
-            local ip="$(ip route get 8.8.8.8 2>/dev/null | awk '{print $NF; exit}')"
+            local ip="$(ip route get 8.8.8.8 2>/dev/null | sed -n '1{s/.*src //; s/\s.*//;p}')"
             printMsgs "dialog" "Your IP is: $ip\n\nOutput of 'ip addr show':\n\n$(ip addr show)"
             ;;
         *.rp)


joolswills:
Sorry but I'd prefer not to use such a complicated RE to extract it. I will look at other options.

joolswills:
Also there are other places this is used, so I will probably make a helper function.

joolswills:
This is a bit simpler - `ip route get 8.8.8.8 | grep -oP "(?<=src\s)[^\s]+"`

I will leave this PR open until I have implemented the changes as a reminder.

hhromic:
I was going to propose this one based on the same idea :)

    ip route get 8.8.8.8 | grep -oP "src \K[0-9.]+"

Another alternative if making a function, is to use Bash's builtin Regex matcher and capture the group with `$BASH_REMATCH`, however I don't think it's worth it compared to the very simple `grep -o` solution. I also don't think this can be done elegantly without a regex capture.

cmitu:
I'm going to play the devil's advocate here (I do agree the `sed` one line is a bit opaque) and argue for the simplicity of `sed`'s operations. The `sed` command can be read in `awk` fashion:
``` awk
# process only 1st line
1 {
   # delete everything up to and including `src `
    s/.*src //;
   # delete everything from the 1st space to the end.
   # In retrospect, `.` matches also newline, so we can probably drop the 1 at the beginning.
    s/\s.*//;
   # print what's left
    p
}
```

hhromic:
Sed does not replace regexs as it also uses them for matching :)
Thats why I think grep is better here because instead of manipulating the output line you just output the matching string of interest and discard the rest. This is why the grep invocation is also simpler.

crcerror:
`hostname -I`

from the manpage
```
-I, --all-ip-addresses
    Display all network addresses of the host. This option enumerates all configured addresses on all
    network interfaces. The loopback interface and IPv6 link-local addresses are omitted. Contrary to
    option -i, this option does not depend on name resolution. Do not make any assumptions about
    the order of the output. 
```

joolswills:
Hostname -I can be unreliable afair. There was a specific reason I didn't use it anyway.

hhromic:
The main problem is that `hostname -I` (or `hostname -i`) can return more than one IP address separated by spaces, and as the manual says, in an arbitrary order. This can happen if the user has more funky configurations like Ethernet to a private network *and* WiFi to the Internet or VPNs. Therefore you would need extra logic to find out which one is the "actual external address".

The `ip route get <knownIP>` method on the other hand directly gives you the source interface IP address used to reach a known address on the Internet, which is most probably the IP you want to report to the user. I don't see a problem using Google's DNS for this. In a more technical aspect, *all* addresses should be reported but that would confuse the users more.

hhromic:
fwiw, the official Raspbian image uses `hostname` to report the local IP addresses:
```
# Print the IP address
_IP=$(hostname -I) || true
if [ "$_IP" ]; then
  printf "My IP address is %s\n" "$_IP"
fi
```
But then it reports *all* the addresses and doesn't select any one in particular :)

psyke83:
FYI: `hostname -I` will print both the address for your IPv4 and IPv6 connections if your service and router is IPv6-enabled, unlike `ip route` which only shows the local IPv4 address.

hhromic:
@psyke83 good point indeed. I guess in the end is all about what is the intended behaviour. If "one" externally routable IP address is desired or "all" interface IP addresses.

For the case of "one", you made me realise that there is also the possibility for a user to be using *only* IPv6 networking. In that case we would need to check if we actually have an IPv4 address to report or try to get an IPv6 one. For example something like this draft snippet (using Google's public DNS):
```
# try to get an externally routable source IPv4 address
_IPV4=$(ip -4 route get 8.8.8.8 2>/dev/null | grep -oP "src \K[^\s]+")
if [[ -z "$_IPV4" ]]; then
  # if there is no IPv4 address, try to obtain an IPv6 address instead
  ip -6 route get 2001:4860:4860::8888 2>/dev/null | grep -oP "src \K[^\s]+"
  return
fi
echo "$_IPV4"
```

joolswills:
@hhromic I think your last snippet may well be the way to go for now. Rather than just copy it into a function if you would like to do it, I'm happy for that also (a getIPAddress helper function to be used where needed).

hhromic:
Sure, no problem :+1: . Will send a PR soon.

PR #2651, opened 2019-03-13
joy2key: protect against IndexError and KeyError exceptions

psyke83:
Various IndexError or KeyError exception could occur if 'a', 'b', 'x' or 'y'
mappings were not available due to arrays accessing unexpected elements.

### DIFF
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -98,11 +98,10 @@ def get_button_codes(dev_path):
     biggest_num = 0
     i = 0
     for btn in list(btn_map):
-        if i > len(dev_button_codes)-1:
+        if i >= len(dev_button_codes):
             break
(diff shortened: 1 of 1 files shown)


psyke83:
Waiting for feedback from: https://retropie.org.uk/forum/topic/21616/error-after-updating-retropie-setup-to-4-4-9-a161774a/14

The easiest way to reproduce these exceptions is either to remove all controller mappings, or comment the 'a' 'b' 'x' 'y' entries. It looks like @cmitu's controller

psyke83:
This fixes the fatal errors that were causing the script to actually break, but not the `bad file descriptor` non-fatal warning. I'll address the latter in another patch, after I manage to reproduce it on one of my systems... and if anyone can point out how to reliably reproduce this warning, it's help me a lot. Thanks.

joolswills:
Thanks for debugging this.

joolswills:
I will test latest code and debug if I see it.

PR #2654, opened 2019-03-15
Blacklist indiana jones n64 in a way that picks a config that runs

### LINKED ISSUE #2649: Remove "infernal" from mupen64plus blacklist
With a current source-compiled version of RSP HLE, GLideN64, and the new dynarec compiler, "Indiana Jones and the Infernal machine" starts up fine and runs at close to 30fps on an RPi3.

If nobody beats me to it, I can do it, I just wanted to open the issue before because I won't have time until the middle of the week.

timfel:
Fixes #2649

### DIFF
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -175,7 +175,6 @@ function testCompatibility() {
     local blacklist=(
         naboo
         body
-        infernal
     )
 
     # these games do not run with gles2n64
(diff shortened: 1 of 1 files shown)


timfel:
This depends on a recent version of mupen64 and the glide video plugin. A normal source install through the script works, I think the binaries need to be updated

joolswills:
Binaries were updated within the last 2 weeks. But will check / rebuild if needed (then merge after). Thanks.

joolswills:
thanks,

PR #2655, opened 2019-03-15
WIP: joy2key: avoid bad file descriptor errors & ensure graceful exit

psyke83:
* Catch OSError exceptions for bad fdesc errors on os.close
* Halt main loop using bool ("running") instead of terminating via
  sys.exit, to minimize chance of errant exceptions when script is
  interrupted.
* ensure tty_fd is also closed before script exits.


psyke83:
Waiting for feedback; in case a testcase is needed, this might help: https://retropie.org.uk/forum/post/184076

psyke83:
Needs more work; closing for the moment.

PR #2656, opened 2019-03-16
joy2key: daemonize to prevent early-init related errors

psyke83:
joy2key:
* Protect signal_handler from concurrent calls via multiple kill requests
* Daemonize to background as soon as signal handlers are registered;
  guarantees that joy2keyStart/start_joy2key returns only when the script
  is in a functional state (and thus the stop calls can't run too early).
* Ensure tty_fd is also closed on exit.

helper/runcommand:
* Don't run script in background
* Remove parent pid check (no longer needed)
* Remove joy2key PID upon kill to avoid unbalanced stop calls (although
  the script now protects from this via signal_handler)

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1044,8 +1044,8 @@ function joy2keyStart() {
     [[ -z "$__joy2key_dev" ]] || pgrep -f joy2key.py >/dev/null && return 1
 
     # if joy2key.py is installed run it with cursor keys for axis/dpad, and enter + space for buttons 0 and 1
-    if __joy2key_ppid=$$ "$scriptdir/scriptmodules/supplementary/runcommand/joy2key.py" "$__joy2key_dev" "${params[@]}" & 2>/dev/null; then
-        __joy2key_pid=$!
+    if "$scriptdir/scriptmodules/supplementary/runcommand/joy2key.py" "$__joy2key_dev" "${params[@]}" 2>/dev/null; then
+        __joy2key_pid=$(pgrep -f joy2key.py)
         return 0
     fi
 
@@ -1057,6 +1057,8 @@ function joy2keyStart() {
 function joy2keyStop() {
     if [[ -n $__joy2key_pid ]]; then
         kill $__joy2key_pid 2>/dev/null
+        __joy2key_pid=""
+        sleep 1
     fi
 }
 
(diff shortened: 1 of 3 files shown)


psyke83:
This seems to work very well, but I highly recommend not merging until it's been fully tested.

Again, this script has been useful in diagnosing problems:
```
#!/bin/bash
SLEEP_TIME=$1
[[ -z "$SLEEP_TIME" ]] && SLEEP_TIME=1

function joy2keystart()
{
    ~/RetroPie-Setup/scriptmodules/supplementary/runcommand/joy2key.py /dev/input/jsX kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20 &
}

function joy2keystop()
{
    kill $(pgrep -f joy2key.py) 2>/dev/null
}

function joy2keytoggle()
{
    joy2keystart
    sleep "$SLEEP_TIME"
    joy2keystop
    joy2keystop
}

while true; do
    echo "loop"
    joy2keytoggle
done

```
The `bad file descriptor` errors were happening due to unbalanced joy2keyStop invocations in admin.sh; it was being called twice in fairly quick succession, leading to joy2key's signal handler to be triggered multiple times, causing an attempt to close the same fds multiple times. This script emulates the sam condition by calling the stop function twice, which almost always guarantees to trigger the error.

Additionally, you can reproduce various other warnings related to killing the process too soon - such as the `libudev` error that was mentioned before by some Ubuntu 18.04+ users - by specifying the sleep to a low value such as 0.1 or 0.2, etc.

If you're going to test the script against this PR, make sure to remove the ampersand from joy2keystart so that it properly emulates the new behaviour of the PR. The start/stop/stop process should be 100% reliable, and even if you specify a sleep of 0, it shouldn't emit any of the aforementioned errors or warnings that were previously caused by a kill signal arriving before the script fully initialized.

psyke83:
I pushed a small update to add a sleep call to the stop functions. This is necessary if the stop/start is going to be called consecutively; the most obvious example of this is seen in `wikiview`, where joy2key is restarted with different button mappings.

From repeated testing on my hardware, a Pi 3 needs just 0.3 seconds after kill to guarantee the process terminates completely, and an i7 needs 0.1 seconds. A 0.5 sleep period might have been sufficient, but to account for slower devices, I left it at a full second just to be safe.

joolswills:
This is working well for me. Thanks.

PR #2658, opened 2019-03-18
skyscraper: fix user errors on vacuum/purge actions

cmitu:
Purge and Vacuum actions should not be runa as 'root' - but as the RetroPie user.


joolswills:
Thanks.

PR #2666, opened 2019-03-27
helpers - add getIPAddress() function

hhromic:
I created a cleaner version of the snippet proposed before.

* the function obtains an externally routable source IP address for the machine
* an IPv4 route is tried first and, if not found, then an IPv6 route is tried
* the function also handles queries for specific devices (used in `wifi.sh`)
* replace all instances of `ip route (...)` with new `getIPAddress` function

I also found one more usage of `ip route (...)` in the `bashwelcometweak.sh` scriptmodule, however it is used from within the installed `.bashrc` snippet, therefore it does not have access to the helpers library.
(( btw, that `.bashrc` snippet does not seem to use any scriptmodule variable, perhaps it is better to move it outside the scriptmodule into `$md_data` for a cleaner code? ))

I tested the function in different machines I have around and seems to be working okay in different scenarios, however I do not have any publicly routable IPv6 machine so I couldn't test that part of the function easily. Nevertheless the logic is simple enough and should still work. Once merged I will pay attention to the forum in case someone reports something odd with that.

Closes #2650

### DIFF
--- a/scriptmodules/admin/builder.sh
+++ b/scriptmodules/admin/builder.sh
@@ -73,7 +73,7 @@ function chroot_build_builder() {
     mkdir -p "$md_build"
 
     # get current host ip for the distcc in the emulated chroot to connect to
-    local ip="$(ip route get 8.8.8.8 2>/dev/null | awk '{print $NF; exit}')"
+    local ip="$(getIPAddress)"
 
     local dist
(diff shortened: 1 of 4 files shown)


joolswills:
Looks great. Thanks.

PR #2671, opened 2019-04-01
jzintv - update to 20181225 version

joolswills:
* repack source archive so it unpacks to jzintv not jzintv-20181225-src and remove dlls from bin subfolder
 * remove now unneeded subMakefile changes
 * override CC/CXX in Makefile to use just gcc rather than hardcoded version
 * clear WARNXX to avoid incompatibility with -Wc++1z-compat and older g++

### DIFF
--- a/scriptmodules/emulators/jzintv.sh
+++ b/scriptmodules/emulators/jzintv.sh
@@ -21,17 +21,15 @@ function depends_jzintv() {
 }
 
 function sources_jzintv() {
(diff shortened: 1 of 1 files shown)


cmitu:
Looks good - I'll give it a test on an x86 system to see how it runs (installation works fine on both the Pi and Ubuntu).

cmitu:
Plays ok also on Ubuntu (18.04 x86_64), I managed to beat the Astrosmash record from the Pi.

PR #2673, opened 2019-04-03
lr-mame2003(-plus): update core options suffix

cmitu:
The `mame2003` and `mame2003-plus` core options are now better aligned with the other cores by prefixing the options with `<core_name>_` instead of `<core_name>-`.

* mame2003-libtetro - https://github.com/libretro/mame2003-libretro/blob/2961cbd34383f6a0d4d8a3362bb79911607158f6/src/mame2003/mame2003.c#L182-L220
* mame2003-plus-libretro : https://github.com/libretro/mame2003-plus-libretro/blob/927c5578dd5daa161b880876119444e0a414bdb4/src/mame2003/mame2003.c#L185-L223

This issue came up recently in the forums (https://retropie.org.uk/forum/topic/21828) - so I updated the Wiki, the PR also updates the core options on installation.

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2003.sh
+++ b/scriptmodules/libretrocores/lr-mame2003.sh
@@ -76,9 +76,9 @@ function configure_lr-mame2003() {
     fi
 
     # Set core options
-    setRetroArchCoreOption "${dir_name}-skip_disclaimer" "enabled"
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

PR #2675, opened 2019-04-04
lr-pcsx-rearmed - set make parameters manually according to platform

hhromic:
The libretro Makefile does not select appropriate cpu/gpu flags.
Therefore, set ARCH, HAVE_NEON, BUILTIN_GPU and USE_DYNAREC manually.

Was:
> The libretro Makefile does not override an empty platform parameter, therefore we must at least start with "unix" for non-RPI2/3 devices that are not ARM nor NEON-enabled, i.e. "generic-x11" machines.
Turns out the previous check was actually necessary for passing "unix" as default when platform is left empty. The libretro makefile does not override an empty platform parameter as we thought (mpff).
> 
> Sorry I didn't catch this in the previous PR. The "generic-x11" platform is the only affected one and not many people use it there in favour of beetle psx.

### DIFF
--- a/scriptmodules/libretrocores/lr-pcsx-rearmed.sh
+++ b/scriptmodules/libretrocores/lr-pcsx-rearmed.sh
@@ -26,15 +26,19 @@ function sources_lr-pcsx-rearmed() {
 }
 
 function build_lr-pcsx-rearmed() {
-    local platform
-    if isPlatform "rpi2" || isPlatform "rpi3"; then
-        platform="$__platform"
-    else
-        isPlatform "arm" && platform+="armv"
-        isPlatform "neon" && platform+="neon"
+    local params=()
+
+    if isPlatform "arm"; then
+        params+=(ARCH=arm USE_DYNAREC=1)
+        if isPlatform "neon"; then
+            params+=(HAVE_NEON=1 BUILTIN_GPU=neon)
+        else
+            params+=(HAVE_NEON=0 BUILTIN_GPU=peops)
+        fi
     fi
-    make -f Makefile.libretro platform="$platform" clean
-    make -f Makefile.libretro platform="$platform"
+
+    make -f Makefile.libretro "${params[@]}" clean
+    make -f Makefile.libretro "${params[@]}"
     md_ret_require="$md_build/pcsx_rearmed_libretro.so"
 }
 


joolswills:
I don't think this is right. The Makefile does set unix if platform isn't set.
```
ifeq ($(platform),)
        platform = unix
        ifeq ($(shell uname -a),)
                platform = win
        else ifneq ($(findstring MINGW,$(shell uname -a)),)
                platform = win
        else ifneq ($(findstring Darwin,$(shell uname -a)),)
                platform = osx
        else ifneq ($(findstring win,$(shell uname -a)),)
                platform = win
        endif
endif
```

joolswills:
ah. it worked before your last commit I think. Checking.

hhromic:
Yes, I thought the same.
However we are passing `platform=` if nothing matched before (armv, neon, rpi2/3), and that is not overriden by the makefile (no `override` command). The logic in the makefile works if no platform is passed at all, but that would require more logic in the scriptmodule.

joolswills:
yeah sorry. that's correct. I might change the logic here.

hhromic:
I think it is safe to start with `unix` as platform. That is the most basic platform supported by RetroPie anyway, and is what the makefile itself defaults to if no platform is provided (apart from `win`).

joolswills:
We don't do this elsewhere though. I would like to use the logic in the makefile. I'll change it.

hhromic:
Another scriptmodule that has to deal with this is `lr-snes9x` where you can see it changes the entire make invocation depending if platform was populated or not. I particularly don't like that approach very much, as it is cleaner to properly populate platform.

joolswills:
Sorry. closed by accident. offtopic - I'm going to be getting rid of the make clean in all modules at some point. just a heads up. it's not needed and would be more useful to be able to resume building.

hhromic:
Ah that sounds good!
I myself very often have to comment-out the clean lines in modules to debug. Appreciated!
I implement it just to follow the lead from other scriptmodules.

joolswills:
I appreciate your argument. Although I don't think there is a good or bad way really. you could argue hardcoding "unix" is bad - if they renamed it in the makefile it would break. And I seem to remember this has happened before with some flag naming relating to unix/linux

hhromic:
Yes agree there is no right or bad way, just stubborn makefiles O=)

I made a change in the logic now to only pass platform if populated, similar to how I do it for the `lr-scummvm` scriptmodule. Hope that is a better approach.

joolswills:
I prefer something like this - needs testing though. I don't like the cpu param duplication using the built in rpi1/rpi2 configs in the makefile. And this technically is more flexible (unless they rename flags).

```
function build_lr-pcsx-rearmed() {
    local params=()
    if isPlatform "arm"; then
        params+=(ARCH=arm USE_DYNAREC=1)
        if isPlatform "neon"; then
            params+=(HAVE_NEON=1 BUILTIN_GPU=neon)
        else
            params+=(BUILTIN_GPU=peops)
        fi
    fi
    make -f Makefile.libretro "${params[@]}" clean
    make -f Makefile.libretro "${params[@]}"
    md_ret_require="$md_build/pcsx_rearmed_libretro.so"
}
```

hhromic:
Mmm I see what you mean, okay I will give it a try on my machines and report back how that works. I will test the core itself with some games too.

joolswills:
Adjusted above to include HAVE_NEON=0

Seems to build on rpi and linux. had to add HAVE_NEON=0 to avoid the autodetect

hhromic:
Ah yes the autodetect of HAVE_NEON I had to handle it before, sent a patch upstream to force it when platform is armv, but in this case platform is not being used so needs to be set.

joolswills:
I'm still 50/50 on this one. I don't like the hardcoded CPU flags in the makefile. in other modules we just set platform=armvneon etc. For this one I think settings the flags directly might be useful so we don't need to rely on some hardcoded configs in the makefile. so with this we still use the platform logic, but the cpu / gpu features are controlled by us.

joolswills:
ill do a PR.

joolswills:
https://github.com/RetroPie/RetroPie-Setup/pull/2676/files

joolswills:
Maybe this is better? 

```
function build_lr-pcsx-rearmed() {
    local params=()
    isPlatform "arm" && platform+="armv"
    isPlatform "neon" && platform+="neon"
    [[ -n "$platform" ]] && params+=(platform="$platform")
    make -f Makefile.libretro "${params[@]}" clean
    make -f Makefile.libretro "${params[@]}"
    md_ret_require="$md_build/pcsx_rearmed_libretro.so"
}
```

joolswills:
like your code above, but removing the rpi2/rpi3 configs, which I don't think offer any benefit to us.

joolswills:
Since it's just a modification of your original changes, feel free to adjust this PR and I'll close mine (if the code above works).

hhromic:
Yes looks more clean indeed, for rpi2/3 will simply use the generic unix+armv+neon and should produce the same binary (more or less). Will test it soon and report back.

joolswills:
Thanks.

hhromic:
I did some testing now and I think I found an argument to manually create and pass the parameters for the make invocation as you suggested before.

When using the generic `platform=armvneon`, the FPU flag is set to `-mfpu=neon` by the Makefile, and takes precedence over the flag passed by RetroPie (because it is appended to `CFLAGS` and the compiler uses the last flag). As reference this is how the FPU is set by RetroPie and the makefile for each platform:

RetroPie flags for platforms:
* rpi1: `-mfpu=vfp`
* rpi2: `-mfpu=neon-vfpv4`
* rpi3: `-mfpu=neon-fp-armv8`

Makefile flags for platforms:
* rpi2: `-mfpu=neon-vfpv4`
* rpi3: `-mfpu=neon-fp-armv8`
* neon: `-mfpu=neon`

Therefore, for example when building for RPI3 with `platform=armvneon`, the actual FPU flag used is the generic `-mfpu=neon` instead of the more precise `-mfpu=neon-fp-armv8` given by RetroPie. This is not ideal because it might be not fully utilising the full NEON instruction set available.

To test, I also built binaries with and without that flag hardcoded in the Makefile and they differ in checksum. However I couldn't notice any big performance impact, so it might be not really important after all.

I checked more carefully and the only parameters we need to deal with are `ARCH`, `HAVE_NEON`, `BUILTIN_GPU` and `USE_DYNAREC`. They seem easy to handle as you showed in your snippet. I can workout the scriptmodule logic from your snippet and more testing and update the PR if you think it is worth it. Shame on the Makefile indeed.

Let me know and I will be happy to help.

hhromic:
I pushed now a commit that manually set the make parameters based on your snippet. I added an extra case for when platform is not ARM, i.e. generic-x11, to disable dynarec and neon. All builds and seems to be working fine on my tests.

I instrumentalised the Makefile to print the relevant variables according to each platform to see how they are set. I also show the actual compiler flags being used. As you can see, they are set correctly now.

For `__platform=rpi3`:

	platform is "unix"
	USE_DYNAREC is "1"
	HAVE_NEON is "1"
	BUILTIN_GPU is "neon"

	cc -O2 -march=armv8-a+crc -mtune=cortex-a53 -mfpu=neon-fp-armv8 -mfloat-abi=hard -ftree-vectorize -funsafe-math-optimizations -pipe -DGIT_VERSION=\"" 588bf79"\" -fPIC -Wall -Iinclude -ffast-math -O2 -DNDEBUG -Ideps/zlib -DFRONTEND_SUPPORTS_RGB565 -DHAVE_LIBRETRO -DNO_FRONTEND -c -o libpcsxcore/cdriso.o libpcsxcore/cdriso.c

For `__platform=rpi2`:

	platform is "unix"
	USE_DYNAREC is "1"
	HAVE_NEON is "1"
	BUILTIN_GPU is "neon"

	cc -O2 -mcpu=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard -ftree-vectorize -funsafe-math-optimizations -pipe -DGIT_VERSION=\"" 588bf79"\" -fPIC -Wall -Iinclude -ffast-math -O2 -DNDEBUG -Ideps/zlib -DFRONTEND_SUPPORTS_RGB565 -DHAVE_LIBRETRO -DNO_FRONTEND -c -o libpcsxcore/cdriso.o libpcsxcore/cdriso.c

For `__platform=rpi1`:

	platform is "unix"
	USE_DYNAREC is "1"
	HAVE_NEON is "0"
	BUILTIN_GPU is "peops"

	cc -O2 -mcpu=arm1176jzf-s -mfpu=vfp -mfloat-abi=hard -pipe -DGIT_VERSION=\"" 588bf79"\" -fPIC -Wall -Iinclude -ffast-math -O2 -DNDEBUG -Ideps/zlib -DFRONTEND_SUPPORTS_RGB565 -DHAVE_LIBRETRO -DNO_FRONTEND -c -o libpcsxcore/cdriso.o libpcsxcore/cdriso.c

For `__platform=generic-x11`:

	platform is "unix"
	USE_DYNAREC is "0"
	HAVE_NEON is "0"
	BUILTIN_GPU is "peops"

	cc -O2 -pipe -DGIT_VERSION=\"" 588bf79"\" -fPIC -Wall -Iinclude -ffast-math -O2 -DNDEBUG -Ideps/zlib -DFRONTEND_SUPPORTS_RGB565 -DHAVE_LIBRETRO -DNO_FRONTEND -c -o libpcsxcore/cdriso.o libpcsxcore/cdriso.c

joolswills:
Thanks btw for the testing.

hhromic:
> I don't think we should force any flags for non arm. The makefile defaults suffice. That's the only change from my PR I think?

Yes that is the only change. I removed the extra case for non-ARM in this PR and force-pushed in case you don't want to re-open yours. I don't mind either way. I will be testing the core with games during my free time today, if I found something I will report back but all should be fine.

joolswills:
Thanks. Happy to stick with this for now.

PR #2683, opened 2019-04-19
lr-pcsx-rearmed fails to build

AbiuPlayer:


hhromic:
I'm curious why you need this only for ODROID. On RPI2 this is not necessary and the `-mcpu` flag is the same for example. Can you elaborate the fix more?  Is `-marm` implied in gcc on Raspbian and not in ODROID?

hhromic:
More importantly, why no other scriptmodule seem to need this? Do all others set `-marm` in their builds?

AbiuPlayer:
Not only for odroid but also for allwinner mali based device.

AbiuPlayer:
Tinkerboard already have -marm arguments in default_cflags

hhromic:
I just test-compiled on RPI2 and builds fine without explicitly passing `-marm`. I guess the compiler in these boards you mention simply doesn't imply this flag by default.
I'm just amused that no other scriptmodule requires it or perhaps it hasn't been tested recently.

hhromic:
For the record, the reason this core worked before is because we were passing `platform=armv` and the Makefile does set `-marm` for that platform. Now that we configure it directly, we need to pass it manually too. If doing this globally or only for this core is probably subject for some discussion. In my opinion, `-marm` seems to be more desirable as a system-wide flag (as in this PR).

hhromic:
After giving this another round of thinking, I think is not good to force `-marm` at the system level. This core's DRC in particular doesn't work with Thumb instructions, as stated in the `configure` script:

    # must disable thumb as recompiler can't handle it
    if check_define __thumb__; then
      CFLAGS="$CFLAGS -marm"
    fi

This is not necessary the case for any other component of RetroPie, hence making it global might not be a good idea. Thumb instructions (the default) should be better than ARM mode in many cases. The fact that the Tinker board is forcing `-marm` doesn't mean it is correct, and maybe it isn't correct after all.

To fix the particular problem of this PR, maybe better would be to just pass `-marm` in the scriptmodule instead of setting it global.

hhromic:
I submitted a PR upstream now to address settings the `-marm` flag properly: https://github.com/libretro/pcsx_rearmed/pull/273

This will set `-marm` if we are building with `ARCH=arm`, so compilers that default to Thumb instructions will not use them for this core. Once accepted, no change should be necessary in RetroPie.

@AbiuPlayer  can you test the PR upstream?  thanks.

hhromic:
The PR upstream has been merged now.
@AbiuPlayer can you please confirm that the core now builds for you without this PR ? Thanks.

RetroPie should not need to add `-marm` in any platform's default cflags.

joolswills:
Thanks @hhromic

PR #2684, opened 2019-04-24
drivers - add xpadneo Xbox Wireless Gamepad driver (Bluetooth)

cmitu:
The `xpadneo` provides better support for the XBox One S - and later - Bluetooth controllers (https://github.com/atar-axis/xpadneo).

There have been some cases in the forums where installing it solved the issue of unresponsive or mis-behaving Xbox One, plus it support Rumble and Trigger Force Feedback.

The pre/post DMKS scripts included in the module handle the module configuration and parameters (they also add/remove  `disable_ertm=1` for bluetooth), so this is a step less for users to perform when installing.

### DIFF
--- a/scriptmodules/supplementary/xpadneo.sh
+++ b/scriptmodules/supplementary/xpadneo.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="xpadneo"
+rp_module_desc="Linux Driver for Xbox One Wireless Gamepad"
+rp_module_help="Enhanced Linux driver for Xbox One Wireless Gamepad (which is shipped with the Xbox One S)."
+rp_module_licence="GPL3 https://raw.githubusercontent.com/atar-axis/xpadneo/master/LICENSE"
+rp_module_section="driver"
+rp_module_flags="noinstclean !mali"
+
+function _version_xpadneo() {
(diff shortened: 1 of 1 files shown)


movisman:
I would just like to mention that i've been running this driver for a few months now and it works very well indeed. It does indeed take care of disabling the ERTM value, also rumble etc all works without issue.

It would be good to implement this in RetroPie, the author updates the driver quite regularly as well.

Cheers

movisman:
PS. Will the script contain functionality to update the driver?

Thanks

cmitu:
@movisman Running the installation it will get the latest version of the driver, it's always installing from source, just like the other drivers.

movisman:
@cmitu

Thanks. I was just wondering because the readme points users to use a specific update.sh script for updating the driver, as opposed to the regular install.sh for a fresh installation, I think like skyscraper does. I am assuming installing over the top will still work, I haven't looked inside either scripts since I first installed a number of months back.

cmitu:
@movisman the `update.sh` script checks the version installed with the online (git) version. If they're different, it just runs uninstall + install. The scriptmodule will do the install un-conditionally, without checking the versions.

movisman:
@cmitu
Ah ha, I see. That makes sense, thanks a lot for the info. Definitely good to get this driver added to the modules, it works really well for me.

EpicLPer:
Apparently it doesn't work anymore since the Kernel updated and it's not taken care of at the original Repository. Which is sad cause so many people actually said this one works better than anything else out there :(

cmitu:
@EpicLPer which kernel version are you referring to ? I don't see anything in the upstream repository reported about this.

movisman:
Just for reference, I am still using this driver on an up to date Ubuntu Server 18.04.3 instance, and it works with no issue for me.

rafal-prasal:
Just did
installed newest raspbian image 2020-02-13
updated it on 2020-02-23
intalled xpad-neo from github
connected my xbox one controller via bluetooth
disabled ertm as it does not work here properly

controller works like a charm.

joolswills:
I merged my PR for xpadneo, but please submit any fixes if I missed something. Thanks and sorry for the accidental duplication.

PR #2689, opened 2019-05-02
Update lr-stella due to rebrand

LeePorte:
stella has rebranded to become stella2014. This commit reflects that
change and now enable the build and install to complete.


joolswills:
Thanks. However, it requires a little more than just renaming. I will sort this out later / over the weekend.

LeePorte:
It ran without issue from my branch, though I'm interested to learn if there are additional changes required.

Every day is a learning day :)

joolswills:
I will rename the module and shift configs.

joolswills:
see https://github.com/RetroPie/RetroPie-Setup/pull/2691

PR #2690, opened 2019-05-03
Update lr-stella.sh to match new .so name

raomin:
Stella has been rebranded stella2014. See https://github.com/libretro/stella2014-libretro/commit/163a1b283b400bd504fdbcd5bce390ed77b73b92


raomin:
Ooops. There is one already for this. I close this one.

PR #2703, opened 2019-05-14
(FBAlpha) Update build script

barbudreadmon:


joolswills:
Thanks.

cmitu:
Would it make sense to also rename the RA core config path due to the [core name change](https://github.com/libretro/FBAlphaFB/commit/e8b51b749d51feec45fda1b850faaffec9d6d533) ? Existing core overrides will not work due to this path change. 
Maybe sym-linking the existing `FB Alpha` folder if found under `$HOME/.config/retroarch/config`.

PR #2705, opened 2019-05-15
sdlpop - fix quicksave and in-game settings saving

tsoliman:
The game attempts to write some files to its working directory but they aren't
writable and so they're being redirected to the /opt/retropie/conf directory


joolswills:
Thanks

PR #2708, opened 2019-05-19
Add lr-quasi88 - NEC PC-8801 emulator for libretro

SuperFromND:
I've tested this core and can confirm that it runs games properly.  As this core is still a WIP according to [the README.md in the repository](https://raw.githubusercontent.com/libretro/quasi88-libretro/master/README.md) and missing features like tape support and games that use more than two disks, I have marked it as experimental.

Note for testing: the filenames for the BIOS files are case sensitive at the moment, so `N88.ROM` will not be detected, but `n88.rom` will.


joolswills:
Thanks.

PR #2711, opened 2019-05-22
Prevent welcome message when using scp or rsync

rob-marr:
When running this command with `rsync`:
`rsync -a pi@retropie:/opt/retropie/configs/all/retroarch/screenshots/ ~/`

I got the following error:
```
tput: No value for $TERM and no -T specified
protocol version mismatch -- is your shell clean?
```

This is because the bash welcome script runs and outputs data that rsync is not expecting, I found the change made prevents this.

### DIFF
--- a/scriptmodules/supplementary/bashwelcometweak.sh
+++ b/scriptmodules/supplementary/bashwelcometweak.sh
@@ -124,8 +124,9 @@ function retropie_welcome() {
     done
     echo -e "\n$out"
 }
-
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Haven't noticed this issue (have copied to and from retropie with rsync) but will test and get back to you.

joolswills:
I cannot reproduce this issue when using rsync to copy from a retropie machine with the bash welcome msg.

PR #2719, opened 2019-05-29
lr-beetle-dc - renamed from lr-reicast

joolswills:


joolswills:
Going to merge this after testing. If the repo renames again we can do another rename. Not ideal, but preferable over broken packages.

barbudreadmon:
Final name is flycast

joolswills:
Thanks!

PR #2721, opened 2019-05-31
add mupen64plus "next" for libretro to experimental

hhromic:
This scriptmodule has been under testing in the forum since Feb/2019 without major problems. Features:

* This core uses up-to-date GLideN64 plugin
* Scriptmodule sets a different namespace (`mupen64plus-next-`) for the core options to avoid clashing with the older `lr-mupen64plus` core options, i.e. they both can co-exist.

Forum Topic: https://retropie.org.uk/forum/topic/21093/
Upstream: https://github.com/libretro/mupen64plus-libretro-nx

### DIFF
--- a/scriptmodules/libretrocores/lr-mupen64plus-next.sh
+++ b/scriptmodules/libretrocores/lr-mupen64plus-next.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-mupen64plus-next"
+rp_module_desc="N64 emulator - Mupen64Plus + GLideN64 for libretro (next version)"
+rp_module_help="ROM Extensions: .z64 .n64 .v64\n\nCopy your N64 roms to $romdir/n64"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/libretro/mupen64plus-libretro-nx/master/LICENSE"
+rp_module_section="exp"
+
+function depends_lr-mupen64plus-next() {
+    local depends=(flex bison libpng-dev)
+    isPlatform "x11" && depends+=(libglew-dev libglu1-mesa-dev)
+    isPlatform "x86" && depends+=(nasm)
+    isPlatform "rpi" && depends+=(libraspberrypi-dev)
+    getDepends "${depends[@]}"
+}
+
+function sources_lr-mupen64plus-next() {
+    gitPullOrClone "$md_build" https://github.com/libretro/mupen64plus-libretro-nx.git GLideN64
+}
+
+function build_lr-mupen64plus-next() {
+    local params=()
+    if isPlatform "rpi"; then
+        params+=(platform="$__platform")
+    elif isPlatform "mali"; then
+        params+=(platform="odroid")
+    else
+        isPlatform "arm" && params+=(WITH_DYNAREC=arm)
+        isPlatform "neon" && params+=(HAVE_NEON=1)
+        isPlatform "gles" && params+=(FORCE_GLES=1)
+        isPlatform "kms" && params+=(FORCE_GLES3=1)
+    fi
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

joolswills:
This doesn't build currently on the RPI 1. Fails during the dynarec sources due to unsupported instructions. Will provide a log when next on the PC. You can test with `sudo __platform=rpi1 ./retropie_packages.sh` etc (you probably know this).

hhromic:
Thanks for the report. I will check it soon as well when i get to my PC too. I don't really expect this core to perform very well on RPI0/1 so perhaps we should consider not enabling it for these devices. Nevertheless I will check it out.

joolswills:
```
cc -O2 -mcpu=arm1176jzf-s -mfpu=vfp -mfloat-abi=hard -pipe -std=gnu11 -O2 -DNDEBUG -fsigned-char -ffast-math -fno-strict-aliasing -fomit-frame-pointer -fvisibility=hidden  -DGIT_VERSION=\"" 96608ba"\" -DCORE_NAME=\""mupen64plus-next"\" -DUNDEF_GL_GLEXT_PROTOTYPES -DOS_LINUX -I./mupen64plus-core/subprojects/md5 -I./mupen64plus-core/subprojects/minizip -I./mupen64plus-core/subprojects/xxhash -D__LIBRETRO__ -DUSE_FILE32API -DM64P_PLUGIN_API -DM64P_CORE_PROTOTYPES -D_ENDUSER_RELEASE -DSINC_LOWER_QUALITY -DTXFILTER_LIB -D__VEC4_OPT -DMUPENPLUSAPI  -I/opt/vc/include -I/opt/vc/include/interface/vcos -I/opt/vc/include/interface/vcos/pthreads -I./custom -I./custom/mupen64plus-core -I./custom/android/include -I./custom/GLideN64 -I./GLideN64/src -I./GLideN64/src/osal -I./mupen64plus-core/src -I./mupen64plus-core/src/api -I./custom/mupen64plus-core/plugin/audio_libretro -I./libretro-common/include -I./libretro -I./GLideN64/src/inc -I./xxHash  -fPIC  -DVC  -DEGL -DHAVE_OPENGLES -DHAVE_OPENGLES2 -DGLES2  -DNEW_DYNAREC=3 -DDYNAREC -I./mupen64plus-core/src/asm_defines/ -c mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.S -o mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.o
mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.S: Assembler messages:
mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.S:177: Error: selected processor does not support `movw r3,#:lower16:fp_restore_candidate' in ARM mode
mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.S:178: Error: selected processor does not support `movt r3,#:upper16:fp_restore_candidate' in ARM mode
Makefile:442: recipe for target 'mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.o' failed
make: *** [mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.o] Error 1
/home/pi/RetroPie-Setup
Could not successfully build lr-mupen64plus-next - N64 emulator - Mupen64Plus + GLideN64 for libretro (next version) (/home/pi/RetroPie-Setup/tmp/build/lr-mupen64plus-next/mupen64plus_next_libretro.so not found).
```

hhromic:
I had a chat with m4xw (main upstream dev) and he confirmed to me that the newer dynarec uses these armv7+ instructions (`movw`, `movt`) and currently won't work for armv6 devices unfortunately. In this case the older lr-mupen64plus core can be used on lower-end RPIs at the moment. He also confirmed to me that there are no plans at the time to remove the old mupen64plus core repository so we can safely keep that one for RPI0/1 devices.

Therefore I think the best thing to do is indeed to flag this core as rpi2/3 (armv7) only and keep the older core covering the older armv6 devices.

joolswills:
Thanks. Yep - that's sounds like the best solution. As you said, N64 emulation doesn't perform well in the rpi1. Cheers for the update.

joolswills:
Hope you don't mind, but as I was doing a couple of minor things I added the armv6 exclusion. Thanks again for looking into this.

hhromic:
> Hope you don't mind, but as I was doing a couple of minor things I added the armv6 exclusion.

No problem at all, thanks for handling it quickly :)

PR #2724, opened 2019-06-02
Fix version number requirements

EricFromCanada:
Update dialog to state that the minimum Ubuntu is 16.04 (upon which Linux Mint 18 [is based](https://en.wikipedia.org/wiki/Linux_Mint_version_history)), and set the [Debian version](https://askubuntu.com/a/445496) to 9 for Ubuntu 16 & 17 / Mint 18 and 10 for Ubuntu 18 & 19 / Mint 19.

This fixes the issue I encountered where the installer would attempt to install ES 2.7.6 instead of stable on Mint 18.3 and Ubuntu 16.04, and fail due to missing Boost libraries.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -84,7 +84,7 @@ function get_os_version() {
                 __os_release=10
             fi
 
-            if compareVersions "$__os_debian_ver" lt 8; then
+            if compareVersions "$__os_debian_ver" lt 9; then
                 error="You need Raspbian/Debian Stretch or newer"
             fi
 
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks. Looks ok, but I need to do a few tests.

joolswills:
I don't want to merge this as it is currently sorry - We don't want to support anything below 18.04

joolswills:
I realise we have code in place that says you need 16.04 or newer but I need to change this. I will make the minimum 18.04

joolswills:
Scrap the above posts. I am getting confused on which are the LTS releases ;-) I will go over this though.

joolswills:
the PR is fine. apologies for my confusion.

scriptmodules/system.sh
```diff
@@ -84,7 +84,7 @@ function get_os_version() {
                 __os_release=10
             fi
 
-            if compareVersions "$__os_debian_ver" lt 8; then
+            if compareVersions "$__os_debian_ver" lt 9; then
```
joolswills: This was actually intentional - I will need to revert this part. Although we don't support versions less than 9, we don't stop retropie-setup from launching (the main setup script has an additional warning).

PR #2725, opened 2019-06-03
gamecondriver: update the scriptmodule for kernels 4.15+

cmitu:
As reported in [the forum](https://retropie.org.uk/forum/topic/22384/), the `gamecondriver` package doesn't build in Kernels 4.15+ due to the deprecation of the `setup_timer` function. This affects the new Linux Kernel included in Raspbian Stretch, which has been upgraded to the 4.19.x stable branch (and will likely be the base for the upcoming Raspbian Buster release).

I added a fix for both modules (a few lines) and submitted them upstream:
 * https://github.com/marqs85/gamecon_gpio_rpi, accepted in https://github.com/marqs85/gamecon_gpio_rpi/pull/2
* https://github.com/marqs85/db9_gpio_rpi, accepted in https://github.com/marqs85/db9_gpio_rpi/pull/4

The modification to the `gamecondriver` change the source of the package to the Github repositories and - consequently - the building/removing of the drivers. I had to use `dkms` directly for installation since `dkmsManager` expects the module source (including the _dkms.conf_ file) to be in `$md_inst`.

### DIFF
--- a/scriptmodules/supplementary/gamecondriver.sh
+++ b/scriptmodules/supplementary/gamecondriver.sh
@@ -10,14 +10,10 @@
 #
 
 rp_module_id="gamecondriver"
-rp_module_desc="Gamecon & db9 drivers"
+rp_module_desc="Gamecon & Db9 drivers GPIO drivers"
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks for this. Was thinking maybe the old debs should be removed maybe in the depends function, but only looked at this briefly. Will review ASAP.

joolswills:
Merging. Thanks.

PR #2732, opened 2019-06-09
psx: really add CHD as extension

cmitu:
Fix slip-ul in `platforms.cfg`.


joolswills:
I missed that too! Thanks.

Issue #2738, opened 2019-06-16
Cannot build from source on vero4k due to mame4all


DBMandrake:
While attempting to install Retropie on an OSMC Vero4k it seems that the installation from source fails due to lack of platform support in mame4all:

```
= = = = = = = = = = = = = = = = = = = = =
Installing dependencies for 'mame4all' : MAME emulator MAME4All-Pi
= = = = = = = = = = = = = = = = = = = = =

Did not find needed package(s): libsdl1.2-dev rbp-userland-dev-osmc. I am trying to install them now.
Reading package lists...
Building dependency tree...
Reading state information...
The following additional packages will be installed:
  libcaca-dev libdrm-amdgpu1 libdrm-dev libdrm-exynos1 libdrm-freedreno1
  libdrm-nouveau2 libdrm-omap1 libdrm-radeon1 libdrm-tegra0 libgl1-mesa-dev
  libglu1-mesa libglu1-mesa-dev libpthread-stubs0-dev libpulse-dev
  libpulse-mainloop-glib0 libslang2-dev libx11-dev libx11-xcb-dev libxau-dev
  libxcb-dri2-0-dev libxcb-dri3-dev libxcb-glx0-dev libxcb-present-dev
  libxcb-randr0-dev libxcb-render0-dev libxcb-shape0-dev libxcb-sync-dev
  libxcb-xfixes0-dev libxcb1-dev libxdamage-dev libxdmcp-dev libxext-dev
  libxfixes-dev libxshmfence-dev libxxf86vm-dev mesa-common-dev
  rbp-userland-osmc x11proto-damage-dev x11proto-dri2-dev x11proto-fixes-dev
  x11proto-gl-dev x11proto-input-dev x11proto-kb-dev x11proto-xext-dev
  x11proto-xf86vidmode-dev xtrans-dev
Suggested packages:
  libxcb-doc libxext-doc
Recommended packages:
  libx11-doc
The following NEW packages will be installed:
  libcaca-dev libdrm-amdgpu1 libdrm-dev libdrm-exynos1 libdrm-freedreno1
  libdrm-nouveau2 libdrm-omap1 libdrm-radeon1 libdrm-tegra0 libgl1-mesa-dev
  libglu1-mesa libglu1-mesa-dev libpthread-stubs0-dev libpulse-dev
  libpulse-mainloop-glib0 libsdl1.2-dev libslang2-dev libx11-dev
[...]

PR #2740, opened 2019-06-19
Exclude modules harmful to Vero4k and build for armv7

hissingshark:
Changes:

1. Build armv7 binaries from now on.  This is harmless enough and may iron out a few quirks in Reicast.  It definitely resolves the mystery dynarec bug in Mupen64plus triggered by the addition of 64 Disk Drive support .  I have a [PR out there to optionally disable 64 Disk Drive support](https://github.com/mupen64plus/mupen64plus-core/pull/649) in builds that are affected, but it hasn't had any attention.  That is beyond our control, but this is not.
2. There are only 2 modules in "main" that are totally toxic to the Vero4k. Excluding these means users can safely perform a the bulk "basic install" without breaking their systems #2738 .

Many thanks,


joolswills:
Thanks. Probably would be best to group these into different commits (but can still be done under one PR).

hissingshark:
Good point.  I will do in future.  Many thanks again.

PR #2741, opened 2019-06-24
pegasus-fe: Fix gradient banding

mmatyas:
Qt applications on the Pi, like Pegasus, tend to have less detailed color gradients due to some framebuffer format related reasons. A kind user pointed out these banding issues can be fixed by defining an env var before launching Pegasus (mmatyas/pegasus-frontend#467).


joolswills:
Thanks.

PR #2742, opened 2019-06-25
image - install newer kernel from buster repos onto stretch to fix sony bt issues

joolswills:
proposed solution to get a newer kernel onto Stretch with the Sony BT fix.

if a new kernel/firmware is released for Stretch soon this won't be needed.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -421,7 +421,7 @@ function update_packages_gui_setup() {
     rps_logInit
     {
         rps_logStart
-        [[ "$update_os" -eq 1 ]] && apt_upgrade_raspbiantools
+        [[ "$update_os" -eq 1 ]] && rp_callModule raspbiantools apt_upgrade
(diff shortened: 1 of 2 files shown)


joolswills:
I threw this code together quite quickly so I may do a cleanup before merge. Still testing it.

psyke83:
Since you've created a separate function, perhaps it might be prudent to put a call to the function in the setup update function for users on an older image who upgrade manually? (If yes, you'd need to put a version check against the installed packages to make sure the function doesn't unnecessarily update on each invocation)

joolswills:
Thanks. Makes sense. When they they a os update from retropie I guess is best.

joolswills:
I have reworked this and made a new function which can be used to manually install any firmware version directly from the repos. Might be useful or not in the future. I also made it run during apt_upgrade as you pointed out (which means we don't need to call it in the image creation as apt_upgrade function is already run).

This is not yet tested.

PR #2744, opened 2019-06-27
Fkms initial support

psyke83:
* Use the GL driver for EmulationStation; can be switched to GLESv2 after support is implemented
* Disable dispmanx on retroarch for mesa targets; should be possible to be enabled after upstream changes, but disable for now to avoid build error.
* Remove hard block on f/kms detection (don't even print error for RPI4)
* Requires SDL2 built with OpenGL support.

Should allow basic bringup for with fkms driver (default for rpi4).


psyke83:
Related PR: https://github.com/RetroPie/SDL-mirror/pull/15

joolswills:
Thanks!

PR #2747, opened 2019-06-27
system - added rpi4 detection / system config

joolswills:


joolswills:
(Not ready for merging yet.)

scriptmodules/system.sh
```diff
@@ -309,6 +312,13 @@ function platform_rpi3() {
     __platform_flags="arm armv8 neon rpi gles"
 }
 
+function platform_rpi4() {
+    __default_cflags="-O2 -march=armv8-a+crc -mtune=cortex-a72 -mfpu=neon-fp-armv8 -mfloat-abi=hard -ftree-vectorize -funsafe-math-optimizations"
+    __default_asflags=""
+    __default_makeflags="-j2"
+    __platform_flags="arm armv8 neon rpi gles"
```
psyke83: As per the discussion with @hhromic, it might be a good idea to deprecate the plain "gles" tag for all platforms to help distinguish what revision is supported; RPI1/2/3 will instead use "gles2", and rpi4 should be "gles3".

Since the `isPlatform` function uses a regex match, it shouldn't interfere with the existing scripting, and we can extend support to target gles3 builds on supported hardware configurations (like rpi4) in future.

joolswills:
I agree.

scriptmodules/system.sh
joolswills: isPlatform "gles" won't match against gles2 and gles3 though - so we would need to add gles2 and gles3 as additional flags to keep existing scripts

joolswills: I think it might be enough just to use gles and gles3 as flags (so gles3 systems get the extra gles3 flas. We can make gles default to mean it has gles2 support). What do you think ?

psyke83: Seems fine to me. Recent Mesa releases have deprecated GLES1 headers, and I'm not aware of any hardware that we target which doesn't support at least GLES2.

PR #2748, opened 2019-06-28
Added config for the Raspberry Pi 4.

raidensnake:
I've managed to create a way for the raspberry pi 4 to compile and be detected.


joolswills:
Thanks but we already have this code ready. We are working on RPI4 support - it's just not all in the main repo yet.

https://github.com/RetroPie/RetroPie-Setup/pull/2747

PR #2750, opened 2019-06-29
Add .zip ROM extension support for Dreamcast

cmitu:
Add `.zip` as valid Dreamcast system extension, for Naomi/Atomiswave ROM support.
Naomi/Atomiswave ROM can be run with `lr-flycast`.


joolswills:
thanks.

PR #2751, opened 2019-06-30
lr-mupen64plus: buster & fkms support

psyke83:
Tested on RPI against fkms and legacy driver.

Legacy driver seems same as stretch, and the fkms version seems to run at similar speed - but there seems to be some tearing, as though vsync is not working correctly.

The mesa workaround may not be suitable for upstream submission due to it having potential to break on older distributions, so it's patched only for newer versions that have the different typedef definition.


joolswills:
Thanks

PR #2754, opened 2019-07-02
bashwelcometweak - add getIPAddress() function

hhromic:
In Buster, IPv6 seems to be enabled by default, therefore the `bashwelcometweak` displays the IPv6 localhost instead of preferring IPv4 as done in the more recent function implemented in `helpers.sh` in 1e849710 .

For now, just copy the newer function from `helpers.sh` in the `bashwelcometweak`. However, in the future, code duplication can be avoided by deploying a public library with RetroPie utilities.


joolswills:
thanks.

PR #2755, opened 2019-07-03
jzintv: fix building on buster

cmitu:
Removed `-Wall` from both c/c++ compilation options, allowing it to build on Buster.


joolswills:
Thanks. I'd like to check this as previously it looks like the warnings were overridden by changing the var.

PR #2756, opened 2019-07-03
Added wiringpo as dependency to controlblock and powerblock modules.

petrockblog:
I have added the wiring package as dependency to the PowerBlock and ControlBlock modules. Their corresponding drivers have had an update recently.


joolswills:
Thanks. Note that this may not be available on some RPI distros. I will disable this module on non rpi targets also.

petrockblog:
All right - thanks for the info!

PR #2757, opened 2019-07-04
DOSbox FKMS support & FKMS-related fixes

psyke83:
* DOSBox:  update to r4252 & add fkms support
* DOSBox-sdl2: unblock & confirm fkms support.
* retroarch: add missing GLESv2 mesa headers.
* helpers: don't install custom SDL1 build for fkms targets


joolswills:
Thanks.

PR #2758, opened 2019-07-04
ppsspp: buster & fkms support

psyke83:
* Implement fkms support (treated as a generic GLES2 target)
* Update build dependencies to enable additional emulator features

Observations on rpi3 via fkms:
* performance seems roughly on par with the legacy driver
* the core and standalone version have tearing; this may be an issue unique to rpi3's fkms overlay and not rpi4, but I can't be sure until someone tests. Same issue observed with `lr-pcsx-rearmed` and `lr-mupen64plus`.
* load & save of save states now appears to be stable in `lr-ppsspp`; building against the most recent tag, `v1.8.0`, causes the instability to return.
~~* sound is currently not working in `lr-ppsspp`, but it's a known issue and unrelated to this PR: https://github.com/libretro/ppsspp/issues/36~~

### DIFF
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -17,8 +17,9 @@ rp_module_section="opt"
 rp_module_flags=""
 
 function depends_ppsspp() {
-    local depends=(cmake libsdl2-dev libzip-dev)
-    isPlatform "rpi" && depends+=(libraspberrypi-dev)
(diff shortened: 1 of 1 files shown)


psyke83:
Let me know if you're not happy with the libretro module being deprecated and I can undo that part of the change. Having them merged is very convenient, as it means we need only one build to produce both modules.

psyke83:
Updated the PR to undo the merging of modules. 

The libretro core now has working sound due to an upstream fix, so everything is in reasonably good shape, apart from the vertical sync issue that's common to other fkms builds for rpi3.

PR #2759, opened 2019-07-04
lr-mupen64plus-next: buster & fkms support

psyke83:
EGL detection needs to be forced for RPI3 via FKMS, otherwise the
core will segfault.

Seems to work well, apart from graphical tearing.

### DIFF
--- a/scriptmodules/libretrocores/lr-mupen64plus-next.sh
+++ b/scriptmodules/libretrocores/lr-mupen64plus-next.sh
@@ -20,18 +20,24 @@ function depends_lr-mupen64plus-next() {
     local depends=(flex bison libpng-dev)
     isPlatform "x11" && depends+=(libglew-dev libglu1-mesa-dev)
     isPlatform "x86" && depends+=(nasm)
-    isPlatform "rpi" && depends+=(libraspberrypi-dev)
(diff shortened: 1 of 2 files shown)


scriptmodules/libretrocores/lr-mupen64plus-next.sh
```diff
@@ -20,18 +20,24 @@ function depends_lr-mupen64plus-next() {
...
+    # HACK: force EGL detection on FKMS
+    isPlatform "mesa" && applyPatch "$md_data/0001-force-egl.patch"
 }
 
 function build_lr-mupen64plus-next() {
     local params=()
-    if isPlatform "rpi"; then
+    if isPlatform "videocore"; then
         params+=(platform="$__platform")
+    elif isPlatform "videocore"; then
```
dankcushions: I think this check should be for "mesa"?

psyke83: Well-spotted. The fallback defines at the end of the if block happened to be close enough to work for a GLES2 target, so the build still worked. It's fixed properly now, and to be sure it's OK I just tested a rebuild with success. Thanks.

PR #2760, opened 2019-07-04
vice: don't depend on 'checkinstall'

cmitu:
`checkinstall` is not available on Buster and it's not used by the scriptmodule, so it can be removed.

### DIFF
--- a/scriptmodules/emulators/vice.sh
+++ b/scriptmodules/emulators/vice.sh
@@ -17,7 +17,7 @@ rp_module_section="opt"
 rp_module_flags=""
 
 function depends_vice() {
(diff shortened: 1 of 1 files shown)


psyke83:
I've no issue with removing a useless dependency, but I do see it available in Raspbian's repo:
```
pi@rpi:~ $ apt-cache policy checkinstall
checkinstall:
  Installed: 1.6.2-4
  Candidate: 1.6.2-4
  Version table:
 *** 1.6.2-4 500
        500 http://raspbian.raspberrypi.org/raspbian buster/main armhf Packages
        100 /var/lib/dpkg/status
```
It's definitely used by `ps3controller` and `sixaxis` to build their requisite packages. Maybe I should phase them out if the package is obsoleted on most other repositories, though.

cmitu:
It's not available in the 'other' Buster: https://packages.debian.org/search?keywords=checkinstall and https://tracker.debian.org/news/886126/checkinstall-removed-from-testing/.

joolswills:
Will merge. Looks like buster is in Debian unstable still? But if it's not going to make it into official Debian buster I guess we will have to switch. Thanks.

PR #2765, opened 2019-07-07
raspbiantools - downgrade kernel/firmware instead for stretch bt fix

joolswills:
* there are additional issues with the newer kernel including overscan problems
 * add option to pass parameter to apt-mark to hold/unhold and hold downgraded packages

### DIFF
--- a/scriptmodules/supplementary/raspbiantools.sh
+++ b/scriptmodules/supplementary/raspbiantools.sh
@@ -16,7 +16,7 @@ rp_module_flags="!x11 !mali"
 
 function apt_upgrade_raspbiantools() {
     # install a newer kernel/firmware for stretch to resolve sony bt issues
-    kernel_bt_fix_raspbiantools
+    stretch_fix_raspbiantools
 
     aptUpdate
     apt-get -y dist-upgrade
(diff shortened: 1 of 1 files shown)


hhromic:
Looks good to me. I perhaps would now rename the `kernel_bt_fix_raspbiantools()` function to something more generic as it is not only fixing bluetooth now :)

psyke83:
Finally I had the opportunity to test the function on my stretch install, before & after this PR.

The `1.20190620+1-1` firmware did fix the PS3 pairing issue - as does this downgrade to `1.20190401-1` . The additional check you added in this PR to not reinstall the same version over and over again seems to be working as intended. Finally, I verified that the latest `customhidsony` is still working correctly against the 4.14 kernel, so there shouldn't be any issues regarding controllers.

I'm assuming that you'll update the function to remove the package hold after a stable stretch firmware is released?

Considering users' issues with overscan, composite, etc., I've no problem with you merging this. Thanks.

joolswills:
Yeah. Can unhold when (if?) Stretch gets a new kernel without these issues.

I renamed the function to just stretch_fix_raspbiantools. Will merge.

PR #2766, opened 2019-07-07
dxx-rebirth: update for fkms and buster

psyke83:
* Update dependencies
* Enable OpenGL + SDL2 build for non-gles targets
* Enable fkms/mesa target, also using SDL2+GL (due to upstream
  source requiring changes to support GLESv2 builds)

### DIFF
--- a/scriptmodules/ports/dxx-rebirth.sh
+++ b/scriptmodules/ports/dxx-rebirth.sh
@@ -10,14 +10,21 @@
 #
 
 rp_module_id="dxx-rebirth"
-rp_module_desc="DXX-Rebirth (Descent & Descent 2) build from source"
+rp_module_desc="DXX-Rebirth (Descent & Descent 2) source port"
 rp_module_licence="NONCOM https://raw.githubusercontent.com/dxx-rebirth/dxx-rebirth/master/COPYING.txt"
 rp_module_section="opt"
-rp_module_flags="!mali !kms"
+rp_module_flags="!mali"
 
 function depends_dxx-rebirth() {
-    local depends=(libphysfs1 libphysfs-dev libsdl1.2-dev libsdl-mixer1.2-dev scons)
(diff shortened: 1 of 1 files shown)


psyke83:
Tested on RPI3/fkms + buster. In GL+SDL2 mode, the D1+2 appears to run at a perfectly smooth 60fps.

scriptmodules/ports/dxx-rebirth.sh
```diff
@@ -10,14 +10,21 @@
...
+rp_module_desc="DXX-Rebirth (Descent & Descent 2) source port"
 rp_module_licence="NONCOM https://raw.githubusercontent.com/dxx-rebirth/dxx-rebirth/master/COPYING.txt"
 rp_module_section="opt"
-rp_module_flags="!mali !kms"
+rp_module_flags="!mali"
 
 function depends_dxx-rebirth() {
-    local depends=(libphysfs1 libphysfs-dev libsdl1.2-dev libsdl-mixer1.2-dev scons)
-    isPlatform "rpi" && depends+=(libraspberrypi-dev)
+    local depends=(libpng-dev libphysfs1 libphysfs-dev libsdl1.2-dev libsdl-mixer1.2-dev scons)
```
hhromic: I remember @joolswills asking to not include library packages in the dependencies but intsead the `-dev` packages of them. In this case `libphysfs1` should be removed in favour of the existing and correct `libphysfs-dev`. Of course this is not your fault in this PR and seems to be vestigial from older code but now that you are on it, why not amend it? :)

psyke83: I've updated the patch to do this - thanks for spotting.

I've also changed the build configuration to specify `words_need_alignment=1` for all arm targets. This is already enabled when `raspberrypi` is selected, but I'm guessing it might be needed for other generic arm boards too.

PR #2767, opened 2019-07-07
Temporarily work around libretro/mame2010-libretro#128

c0d3h4x0r:


joolswills:
Seem to have two PRs with the same changes or? I'll close this one. Thanks.

c0d3h4x0r:
When it comes to managing forks/branches/pull-requests using git and GitHub, I'm still trying to figure out how it all works.  Somehow my PR for the temporary work around got intermixed here with the subsequent commit to update RetroArch to v 1.7.7, which was for a different issue entirely.

The problem is that right now, updating lr-mame2010 from source is busted, thus the reason for the temporary workaround in my commit c4858f7.

PR #2768, opened 2019-07-07
Update Retroarch to 1.7.7

c0d3h4x0r:


joolswills:
Thanks but we are holding off on this for now. I don't want to checkout a specific commit - we will probably wait till the next stable.

c0d3h4x0r:
It sounds like you are commenting on my commit c48548f7, which somehow got erroneously picked up and intermixed with my commit 1454929, which is the actual change I was proposing here.  Please consider taking only commit 1454929, as RetroArch 1.7.7 *is* a stable release and includes a number of nice new features.

c0d3h4x0r:
It seems like my original mistake was commiting c4858f7 against **retroarch-1.7.7** rather than against a separate branch.  I'm guessing I might be able to repair/fix this pending PR by committing an undo of c4858f7 to my **retroarch-1.7.7** branch and then squashing/combining all four of those commits into one.  Does that sound right?

c0d3h4x0r:
Okay, it took some serious wrangling and liberal usage of the --force flag, but I believe I've finally got this pull request fixed up to reflect ONLY the proposed changes to update RetroArch to 1.7.7.

psyke83:
Jools already said he doesn't want to switch to the new release yet, but I should also point out that the PR in its current state would also cripple retroarch for rpi fkms/rpi4 targets. There's no need to explicitly disable opengl for gles platforms.

We already have a `gl` platform flag defined for some generic targets, but it's almost never used. Maybe we will need to start using it more, but 1) it needs to be renamed to `opengl` (as `isPlatform` does a regex match, so e.g. `gl` would be erroneously detected on a `gles`-only platform), and 2) we still need to take into account that some platforms want to have both GLES & GL support, not either/or (mostly Pi).

c0d3h4x0r:
`--disable-opengl1 --disable-opengl_core` is necessary to avoid OpenGL-related compilation errors in the RetroArch 1.7.7 codebase.  See https://github.com/libretro/RetroArch/issues/8770 for more information.

c0d3h4x0r:
> Jools already said he doesn't want to switch to the new release yet

It doesn't sound to me like that's what he was saying:

>I don't want to checkout a specific commit - we will probably wait till the next stable.

His wording sounded to me like he was referring to the originally-intermixed change (unrelated to this issue, but it snuck its way into the original list of commits for this pull request, which I have since fixed) to `lr-mame2010.sh` to effectively pin it to a particular commit number as a workaround for https://github.com/libretro/mame2010-libretro/issues/128.

joolswills:
Sorry. @psyke83 is correct. We will update later. We may skip 1.7.7. upstream called it alpha. I heard they retagged but we may wait. 

Appreciate the PR, but I would recommend starting a discussion first on the forum to see what is needed. Some code may be intentionally not the most recent, or code may already be being worked on.

Cheers.

c0d3h4x0r:
That's fair.  Thanks for explaining.  But in the meanwhile, I'll continue using my local build of 1.7.7 :)

PR #2769, opened 2019-07-09
NO NOT MERGE: gl4es scriptmodule

psyke83:
Would appreciate any feedback, but keep in mind it's not ready to be merged due to deliberate `rp_callModule gl4es install_bin || rp_callModule gl4es` bypass that's needed to ensure that the module builds (as binaries aren't available, obviously).

I've included a scriptmodule that installs neverball and neverputt to demonstrate the library being used.

When a scriptmodule includes the `gl4es` tag, it will enable the translation library on generic `gles` targets (such as RPI with the legacy driver), but not RPI using `mesa` due to the GL 2.1 hardware support being present.

Note: SDL2 needs to be compiled with OpenGL support even on `videocore` to work with `gl4es`; my [other](https://github.com/RetroPie/RetroPie-Setup/pull/2770) PR will account for this.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -246,6 +246,15 @@ function getDepends() {
                 packages+=("$required")
                 continue
             fi
+
+            # gl4es translation library installation for gles targets
+            if [[ "$required" == "libgl1-mesa-dev" ]] \
+              && isPlatform "gles" \
+              && hasFlag "${__mod_flags[$md_idx]}" "gl4es" \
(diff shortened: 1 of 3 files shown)


romanbsd:
Nice! How is the performance? I tried gl4es with SuperModel3: https://sourceforge.net/p/model3emu/code/HEAD/tree/trunk/
the gl4es is initialized, but supermodel3 still fails with "OpenGL initialization failed". Any ideas what can be wrong?

psyke83:
Closing for now as it may have limited utility, and I don't really use the legacy driver often enough to test properly.

PR #2771, opened 2019-07-09
mupen64plus: debian buster & initial fkms support

psyke83:
Note: fkms targets uses the GLideN64 video plugins only, as the older
plugins are much less accurate and not as well-maintained. Additionally,
these targets will build the omx plugin, but will not be used by default
due to the better compatibility offered by the SDL audio plugin.

* apply armv8 optimization based on platform flag (rpi3/4)
* remove kms blocking flag and add initial fkms build support
* consolidate launcher script execution arguments
* skip "emplace" patch for fkms targets (no crash observed)

### DIFF
--- a/scriptmodules/emulators/mupen64plus/0002-GLideN64-force-egl.patch
+++ b/scriptmodules/emulators/mupen64plus/0002-GLideN64-force-egl.patch
@@ -1,0 +1,13 @@
+diff --git a/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp b/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
+index 9f2de97..745133f 100644
+--- a/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
++++ b/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
+@@ -145,7 +145,7 @@ void GLInfo::init() {
+ 	texture_barrierNV = Utils::isExtensionSupported(*this, "GL_NV_texture_barrier");
+ 
+ 	ext_fetch = Utils::isExtensionSupported(*this, "GL_EXT_shader_framebuffer_fetch") && !isGLES2 && (!isGLESX || ext_draw_buffers_indexed) && !imageTextures;
+-	eglImage = (Utils::isEGLExtensionSupported("EGL_KHR_image_base") || Utils::isEGLExtensionSupported("EGL_KHR_image"));
++	eglImage = true; //(Utils::isEGLExtensionSupported("EGL_KHR_image_base") || Utils::isEGLExtensionSupported("EGL_KHR_image"));
+ 
+ #ifdef OS_ANDROID
+ 	eglImage = eglImage &&
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -14,11 +14,12 @@ rp_module_desc="N64 emulator MUPEN64Plus"
 rp_module_help="ROM Extensions: .z64 .n64 .v64\n\nCopy your N64 roms to $romdir/n64"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/mupen64plus/mupen64plus-core/master/LICENSES"
 rp_module_section="main"
-rp_module_flags="!mali !kms"
+rp_module_flags="!mali"
 
 function depends_mupen64plus() {
     local depends=(cmake libsamplerate0-dev libspeexdsp-dev libsdl2-dev libpng-dev fonts-freefont-ttf)
     isPlatform "rpi" && depends+=(libraspberrypi-bin libraspberrypi-dev)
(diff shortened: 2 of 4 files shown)


scriptmodules/emulators/mupen64plus/0002-GLideN64-force-egl.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp b/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
+index 9f2de97..745133f 100644
+--- a/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
++++ b/GLideN64/src/Graphics/OpenGLContext/opengl_GLInfo.cpp
+@@ -145,7 +145,7 @@ void GLInfo::init() {
+ 	texture_barrierNV = Utils::isExtensionSupported(*this, "GL_NV_texture_barrier");
+ 
+ 	ext_fetch = Utils::isExtensionSupported(*this, "GL_EXT_shader_framebuffer_fetch") && !isGLES2 && (!isGLESX || ext_draw_buffers_indexed) && !imageTextures;
+-	eglImage = (Utils::isEGLExtensionSupported("EGL_KHR_image_base") || Utils::isEGLExtensionSupported("EGL_KHR_image"));
++	eglImage = true; //(Utils::isEGLExtensionSupported("EGL_KHR_image_base") || Utils::isEGLExtensionSupported("EGL_KHR_image"));
```
psyke83: I'm reasonably happy with the PR, except for this hack. @hhromic has been doing some digging into this, so I'd appreciate if a better solution could be found.

(the hack does seem to work absolutely fine, though, so if nobody objected, we could just merge this as-is and then remove/change the patch later when a more elegant fix is found).

psyke83:
Reopened the PR and made some changes. I'd appreciate any testers to look for issues, but it likely needs some changes, so it's not ready to be merged.

To recap:

* This will require the %XRES% and %YRES% variable substitution in runcommand (via https://github.com/RetroPie/RetroPie-Setup/pull/2774). KMS doesn't support framebuffer scaling, so we have to run at the monitor resolution and rely on the internal rendering resolution being a reasonable value (`nativeResFactor = 1` will ensure it renders 4:3 content at 320x240, for ex).
* Most of the games I tested work well, but `nativeResFactor = 1` causes graphical distortions in both Zelda games.
* To fix Zelda, you can setting the native resolution factor to `2` (640x480), but that's too slow on RPI3. It's possible to hack the source and force a more reasonable intermediate resolution upscale  - 1.15x of the game's original resolution [like so](https://gist.github.com/psyke83/938c0b08d9dda08c76c063fe2621bf05), which resolves all the rendering errors, but it's not suitable for upstreaming, as the the factor is supposed to provide a clean upscale. I'll file a bug upstream soon to see if it can be diagnosed/fixed properly, or at least have a way to specify a custom rendering resolution that's different from the window resolution.
* I'm not certain exactly why, but launching from an xinit session seems to increase performance. I've temporarily added an X and non-X emulator entry so you can compare the difference while checking the PR.

joolswills:
Have changed this to use the new fkms_rpi4 branch - as this is a development branch I'm happy to merge this if you are.

psyke83:
There are some small issues that prevent me feeling 100% happy with the PR. If we were to drop the omx audio plugin (since it's not even set by default in favour of the sdl plugin), I could shave off a few lines where the `VC=1` param exception is needed. I left it in because it works, and presumably might perform a bit better than sdl, but the RPI4 should have sufficient CPU crunching power to offset any possible advantage, and the sdl plugin has better maintenance as an official plugin.

Additionally, the section that cleans up videocore entries is handy when switching between fkms and videocore, but it's an extra ~10 lines being added to facilitate something that end users shouldn't be doing (switching between drivers).

If you agree, I can slough off those two bits from the PR and slim it down a bit.

dankcushions:
agree re: OMX, although you can even benchmark with the --nospeedlimit launch param (once my mupen64plus vsync PR is accepted).

psyke83:
@dankcushions 

Actually - you reminded me that your PR will be obsoleted by this patchset, too (my PR set threaded rendering for both targets before you opened yours).

That's another part of the PR I can remove (so that the decision can be made to enable threading in yours).

joolswills:
If you're happy with this please feel free to merge. As it's on our dev branch we can revisit if needed. Wasn't sure so will leave to you. Thanks.

psyke83:
The PR is OK*, but I'd prefer to wait until the two recent mupen64plus commits to master have been merged into fkms_rpi4, so I can ensure that the PR includes those changes in the correct place, considering that the PR rearranges the entire section related to ini configuration.

*As a reminder, performance is bad via kmsdrm, and I haven't yet had time to determine the reason, hence the reason why there's an emulator entry that launches via an xinit session included. I notice the same performance regression when testing on Ubuntu 19.04 x86_64 via kmsdrm vs X (using Intel graphics), so it's likely not an EGL/GLES or Pi firmware issue.

PR #2773, opened 2019-07-10
Pass platform through to make commands (at least for Raspberry Pi).

c0d3h4x0r:
This is a partial fix for https://github.com/libretro/mame2010-libretro/issues/130

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2010.sh
+++ b/scriptmodules/libretrocores/lr-mame2010.sh
@@ -24,6 +24,7 @@ function build_lr-mame2010() {
     make clean
     local params=()
     isPlatform "arm" && params+=("VRENDER=soft" "ARM_ENABLED=1")
+    isPlatform "rpi" && params+=("platform=$__platform")
     make "${params[@]}" ARCHOPTS="$CFLAGS" buildtools
     make "${params[@]}" ARCHOPTS="$CFLAGS"
     rpSwap off


c0d3h4x0r:
Please read https://github.com/libretro/mame2010-libretro/issues/130 — it should answer all of your questions.

dankcushions:
given that mame2003's makefile had similar requirements, i presumed we passed platform in by default: https://github.com/libretro/mame2003-libretro/blob/master/Makefile#L166

but i guess we don't? maybe we should? in the case of mame2003 i think we do all/most of those compiler flags by default, so maybe it doesn't make a difference. i guess it depends on whether you want to be in control of the compiler flags, or let the upstream makefile. personally i would probably want the makefiles to be the master.

in the case of mame2010, it looks like you could use the ARM_ENABLED flag that we DO pass through already to avoid these x86 specific paths, but that would require a further change to https://github.com/libretro/mame2010-libretro/pull/132/files, when it probably already solves the issue for lakka and so forth, who presumably use the platform flag.

joolswills:
we intentionally don't pass platform flags in most cases as we want to handle compiler flags etc ourselves. I would prefer to handle any other makefile flags ourselves rather than using the platform=

joolswills:
I checked out the makefile. In this case passing just platform=rpi would be sufficient (so we avoid the cpu flags). although probably we could just set signed-char (if it's needed) and FORCE_DRC_C_BACKEND = 1 if that's the actual fix (The other changes look just like whitespace changes - would be better in another commit btw - makes things more readable).

c0d3h4x0r:
lr-mame2010 maintainers have accepted my pull request for the related changes on their side, so now it’s just up to you to decide whether you want to take this fix or come up with your own different approach.  I’ve done my part by making you aware of the issue and proposing a reasonable (and working) fix.

ginkage:
Just to add my 5p, this does fix the aforementioned error for me on RPi4 with mame2010.
The core itself is still painfully slow, but at least this fix makes it work.

joolswills:
Closing as should have been resolved in 77005667

PR #2774, opened 2019-07-11
runcommand KMS x/y res support & yquake2 update

psyke83:
* add variable replacement to pass framebuffer x/y res
* update yquake2 for fkms targets


joolswills:
Thanks.

PR #2775, opened 2019-07-12
Fix 2x build errors, 1x gitPullOrClone bug 

hissingshark:
**1st commit:**
Reicast have restructured their branches.
Presently we pull the default branch which was `master`.  It is now `alpha` (and for a short time was `develop`) and breaks the build with new paths and renaming.
Recommendation [here](https://github.com/reicast/reicast-emulator/issues/1669) is to adopt the `beta` branch until further notice.

---
**2nd commit**
Builds fail on Vero4K/OSMC/Debian at the installation of the python `evdev` dependancy:

> Collecting evdev
>   Downloading https://files.pythonhosted.org/packages/fe/6e/3fa84a43571dec4d00dc26befccc9867b6b3263651531cbc1345f718860f/evdev-1.2.0.tar.gz
> Building wheels for collected packages: evdev
>   Running setup.py bdist_wheel for evdev ... error
>   Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-EavkPG/evdev/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/tmpfnOGqwpip-wheel- --python-tag cp27:
>   usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
>      or: -c --help [cmd1 cmd2 ...]
>      or: -c --help-commands
>      or: -c cmd --help
>   
>   error: invalid command 'bdist_wheel'
>   
>   ----------------------------------------
>   Failed building wheel for evdev
>   Running setup.py clean for evdev
> Failed to build evdev

Fixed by installing `wheel` first.
This may be a local package/user configuration issue so I've made the fix platform specific right now.  I doubt it would hurt to apply this generically so I could change that if you think appropriate.

Many Thanks,

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -20,6 +20,7 @@ function depends_reicast() {
     local depends=(libsdl2-dev python-dev python-pip alsa-oss python-setuptools libevdev-dev libasound2-dev libudev-dev)
     isPlatform "vero4k" && depends+=(vero3-userland-dev-osmc)
     getDepends "${depends[@]}"
+    isPlatform "vero4k" && pip install wheel
     pip install evdev
 }
 
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -24,7 +24,7 @@ function depends_reicast() {
 }
 
 function sources_reicast() {
-    gitPullOrClone "$md_build" https://github.com/reicast/reicast-emulator.git
+    gitPullOrClone "$md_build" https://github.com/reicast/reicast-emulator.git master
 }
 
 function build_reicast() {
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -367,7 +367,7 @@ function gitPullOrClone() {
         if [[ "$__persistent_repos" -ne 1 && "$repo" == *github* && -z "$commit" ]]; then
             git+=" --depth 1"
         fi
-        [[ "$branch" != "master" ]] && git+=" --branch $branch"
+        git+=" --branch $branch"
         printMsgs "console" "$git \"$repo\" \"$dir\""
         runCmd $git "$repo" "$dir"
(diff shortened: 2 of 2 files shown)


cmitu:
Unfortunately, this doesn't work - again - due to the source tree changes. The `Makefile` moved to `reicast/linux`.

hissingshark:
Oh for crying out loud... Which is back to the original problem... 

Beta should have been ok and I think is meant to be the way to go. Rather than specify Master, which is on the way out, I'll keep pointing to Beta and change the path to match.

hissingshark:
Nope.  `Beta` is still too unstable for building it seems. I hate to think what `alpha` would do.
I think we need to languish at `master` for now then, which at least builds.

Unfortunately it doesn't build in this context, because if you specify `master`, then `gitPullOrClone` ignores it, assuming that it will be the default branch and that gets us `alpha`...  So to build from within RetroPie I need to amend `gitPullOrClone`, although I suspect there was a reason for defining that behaviour?

hissingshark:
I can't see what the benefit of ignoring a branch setting of `master` was, so I've changed it to not be ignored.

**If** we wanted the ability to let the repo offer its "default branch" then we could:
1. remove the initial `[[ -z "$branch" ]] && branch="master"` and add `master` to every gitPullOrClone call in every submodule that didn't specify a branch and the default was confirmed not to be master...
2. Or easier, add a "use default branch" argument!

Anyway, seems I've found a gotcha along the way.  It's fixed, provided I'm not misunderstanding the purpose and most importantly I can confirm Reicast builds again...

joolswills:
The for branch fix seems ok. I think some logic got changed during some other overhauls and this case scenario was lost. I may rework the function myself but I appreciate the heads up. Short on time right now but will get back to you. Thanks.

PR #2777, opened 2019-07-13
More FKMS updates

psyke83:
Cannonball works fullspeed; tyrquake is not quite fullspeed (GL at native res), but it's not too bad.

### DIFF
--- a/scriptmodules/ports/darkplaces-quake.sh
+++ b/scriptmodules/ports/darkplaces-quake.sh
@@ -13,11 +13,12 @@ rp_module_id="darkplaces-quake"
 rp_module_desc="Quake 1 engine - Darkplaces Quake port with GLES rendering"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/xonotic/darkplaces/master/COPYING"
 rp_module_section="opt"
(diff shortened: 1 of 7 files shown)


scriptmodules/ports/darkplaces-quake.sh
```diff
@@ -13,11 +13,12 @@ rp_module_id="darkplaces-quake"
 rp_module_desc="Quake 1 engine - Darkplaces Quake port with GLES rendering"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/xonotic/darkplaces/master/COPYING"
 rp_module_section="opt"
-rp_module_flags="!mali !kms"
+rp_module_flags="!mali"
 
 function depends_darkplaces-quake() {
     local depends=(libsdl2-dev libjpeg-dev)
     isPlatform "rpi" && depends+=(libraspberrypi-dev)
```
joolswills: Should this be switched to videocore  or?

psyke83: Yep, somehow missed that one. Patch has been updated, thanks!

joolswills:
Thanks

PR #2778, opened 2019-07-13
ThreadedVideo = True for GLiden64

dankcushions:
Better performance. You can benchmark by changing the launch string in `/opt/retropie/emulators/mupen64plus/bin/mupen64plus.sh` to:
` SDL_VIDEO_RPI_SCALE_MODE=1 "$rootdir/emulators/mupen64plus/bin/mupen64plus" --noosd --nospeedlimit --windowed $RES --rsp ${RSP_PLUGIN}.so --gfx ${VIDEO_PLUGIN}.so --audio dummy --configdir "$configdir/n64" --datadir "$configdir/n64" "$ROM"`
but note that it won't be able to go faster than vsync (ie, you won't see an improvement) unless you have https://github.com/gonetz/GLideN64/pull/2070

True:
mario64 title screen: 95-105 VI/S

False:
mario64 title screen: 72-85 VI/S

See https://github.com/gonetz/GLideN64/pull/2014 for details on the change.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -242,6 +242,8 @@ function configure_mupen64plus() {
         iniSet "ShowFPS " "True"
         iniSet "fontSize" "14"
         iniSet "fontColor" "1F1F1F"
+        # Enable Threaded GL calls
+        iniSet "ThreadedVideo" "True"
 
         # Disable gles2n64 autores feature and use dispmanx upscaling
         iniConfig " = " "" "$md_conf_root/n64/gles2n64.conf"


psyke83:
This is a good change, IMO. It improves performance, and while I did observe overheating when running at an unlocked framerate, it maxes out at 72C for Super Mario running at full speed even with 2x native resolution scaling.

Tested with my Pi3B, but I expect that the thermal behaviour will be the same on Pi3B+, as the default behaviour is to limit CPU frequency to 1200Mhz when reaching 60C before doing actual throttling at the 80C range like a normal 3B model.

@joolswills - if/when you merge this, could you also rebase the `fkms_rpi4` branch against latest master so I can verify the mupen64plus_buster PR to make sure this doesn't conflict?

joolswills:
Thanks. Will update fkms branch in next day or two.

PR #2779, opened 2019-07-14
runcommand: Add onbeforelaunch user script

tsoliman:
I'm trying to use xboxdrv as a more powerful joy2key for actual emulators that are launched by runcommand. (the Retroflag GPi case has a built-in xbox 360 controller and no USB)

The original solution that someone else did was to launch it in `runcommand-onstart.sh` and kill it in `runcommand-onend.sh`. However, that runs into a problem with all of the stuff that happens between start and launch. Specifically the art and the launch configuration screens. They're not accessible because the keymapper for the emulator is active.

So I needed a hook to act as a wrapper for the emulator itself.

There's currently no need for a `runcommand-onafterlaunch.sh` because not much happens between the launch the `runcommand-onend.sh`

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -1064,6 +1064,8 @@ function runcommand() {
         COMMAND+=" -- vt$TTY -keeptty"
     fi
 
+    user_script "runcommand-onbeforelaunch.sh"
+
     local ret
     launch_command
     ret=$?


cmitu:
> However, that runs into a problem with all of the stuff that happens between start and launch. Specifically the art and the launch configuration screens. They're not accessible because the keymapper for the emulator is active.

Can you give an example that reproduces this problem ? What's the content of the `onstart` script that triggers this behavior ?

tsoliman:
The content of onstart is a conditional call to start xboxdrv which takes over the built-in Xbox 360 controller and intercepts its button presses and turns them to keyboard keys (e.g. WSAD and such).
The content of onend is to pkill xboxdrv, thus returning the controller to its natural state.

The onstart activation of xboxdrv is conditional on the emulator being launched (userscripts get command line args as you know) which means xboxdrv only launches when the emulator is (say) a stand-alone emulator like pcsx-rearmed or pifba.

So here's the scenario: I'm running a pi zero and so the default emulator for neogeo is lr-fbneo which works fine with the cyclone dynarec. But some ROMs have compatibility problems and so I need to use pifba with them. So I press A to skip the "loading" launch art I created and press A again go into the runcomand configuration screen and make a rom-specific emulator override, and quit the runcommand menu. Now this rom will launch with the standalone pifba. I launch it and it works fine, but I would like set the get into the runcomand menu again, but I can't. This is because onstart has been told it is about to launch pifba and so it is running xboxdrv which makes the controller buttons act like keyboard keys. I'm now stuck. I can't skip the art and I can't press A to reconfigure runcommand because A is now keyboard_z or whatever pifba expects it to be. 

So the problem is that the xboxdrv remapping is occcuring before the art and menu parts of runcomand instead of being "just before the emulator starts".

tsoliman:
Here's a snipped gist of the actual userscript

```
case $2 in
    pifba)
        sudo /opt/retropie/supplementary/xboxdrv/bin/xboxdrv --detach-kernel-driver --silent --config xboxdrv-fba.ini &
    ;;
esac
```

SinisterSpatula:
Hi, just wanted to add, that I'm the one that's using xboxdrv in onstart currently and would love to see this implemented.  Another option would be pushing the call to runcommand-onstart.sh further down in the process of launching.  So that it is the last thing called right before launch.

tsoliman:
> I'm the one that's using xboxdrv in onstart currently

You mean you are _also_ using xboxdrv in onstart 😉

> Another option would be pushing the call to runcommand-onstart.sh further down in the process of launching. So that it is the last thing called right before launch.

For your one xboxdrv use case maybe? However, I need the existing `on-start` behavior for the existing use cases to continue to work (meaning it runs where it does today, before the art and menu). I also need `on-beforelaunch` as it is proposed in this PR for both xboxdrv/joy2key, and screenshotting.

tsoliman:
I should have probably linked this from the beginning: https://github.com/RetroPie/RetroPie-Setup/wiki/Universal-Controller-Calibration-&-Mapping-Using-xboxdrv#3a-key-mapping-and-launch-fundamentals

SinisterSpatula:
Good Gravy!  Ok, sorry I had assumed you were talking about what I was doing.  I need to stop making such broad assumptions.  I'll just go find a nice hole to crawl into now. lol

tsoliman:
Don't worry about it. This is a problem both of us are running into and both of us want it fixed :)

joolswills:
I'm not too keen on the additional script but I appreciate the issue. I will look into it but it may have to wait a bit. Thanks.

tsoliman:
@joolswills thanks for looking at this.
Can you help me understand the concern?
My understanding is that If the script isn't there then user_script is a noop.

tsoliman:
I've moved all my use cases to `runcommand-onbeforelaunch`. So @SinisterSpatula was right in that we can probably just move the existing user script to the new location instead of having an additional script. My preference (for backwards compat) is introducing an additional script, but then again I don't understand the concern so I'll wait 😄

tsoliman:
Any update on this issue?

joolswills:
No -I will consider changing the hooks (there was another request), but this won't happen until after fkms work.

I cannot promise it will be implemented like this.

tsoliman:
Thanks for the update. Good luck with the fkms work.

joolswills:
Btw I didn't mean to start my rely with No. Was copy and paste fail from the other related issue. Thanks.

PR #2780, opened 2019-07-14
dosbox: ensure freepats soundfonts are configured

cmitu:
In Buster, the default soundfont configured in Timidity is `fluidr3_gm`. Without it, `timidity` logs an error during the `runcommand` startup for `dosbox` and the emulator is not started:
```
   /etc/timidity/fluidr3_gm.cfg: No such file or directory
   timidity: Error reading configuration file.
   Please check /etc/timidity/timidity.cfg
```

We could use the `fluid-soundfont-gm` as default (and remove `freepats`), that would require to modifying `/etc/timidity/timidity.cfg` . On the other hand, the `fluid-soundfont-gm` is considerably larger than `freepats` (149MB vs 34MB).

### DIFF
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -87,12 +87,16 @@ function midi_synth() {
     case "\$1" in
         "start")
             timidity -Os -iAD &
-            until [[ -n "\$(aconnect -o | grep TiMidity)" ]]; do
+            sleep 1
+            pgrep timidity || timidity -Os -iAD -c /etc/timidity/freepats.cfg &
+            local timeout=5
+            until [[ -n "\$(aconnect -o | grep TiMidity)" || \$timeout -le 0 ]]; do
                 sleep 1
+                timeout=\$((timeout-1))
             done
             ;;
         "stop")
-            killall timidity
+            killall timidity 2>/dev/null
             ;;
         *)
             ;;


psyke83:
Grepping /usr/bin/timidity, I see `%s/.timidity.cfg`. This looks like a user override (i.e. `/home/pi/.timidity.cfg`). Perhaps we could try that?

I don't mind using fluidsynth, but as you said, the package is considerably larger, and may also affect memory usage. I do recall the MIDI playback being quite different as well.

cmitu:
I agree, I'll see if we can override with a user configuration file and keep using `freepats`.

cmitu:
Unfortunately the user config (`$HOME/.timidity.cfg`) is not an override, but an addition to the main `.cfg` and it's read after the system config. 
I have added a helper function to deal with the system config (`/etc/timidify/timidity.cfg`) and modified also the `+Start DosBox.sh` script to not loop indefinitely, just in case there is an error during `timidity` startup.

hhromic:
Be aware that `dosbox` is not the only module using freepats, perhaps `_config_soundfonts_freepats_dosbox()` can be migrated to `helpers` and be run from any module using it.

cmitu:
@hhromic Thanks for the pointer - I'll take a look. I see `zdoom` and `eduke` have `freepats` and `timidity`, I'll give it a test.

scriptmodules/emulators/dosbox.sh
```diff
@@ -54,6 +54,20 @@ function build_dosbox() {
     md_ret_require="$md_build/src/dosbox"
 }
 
+# helper method to ensure that 'freepats' soundfonts are configured for Timidity
+function _config_soundfonts_freepats_dosbox()
```
psyke83: Are you searching for the package name via dpkg? I don't think it's correct, as it's `fluid-soundfont-gm` and  `fluidr3_gm.cfg`.

Could you also use our helper functions for this as well as editing the file? i.e.:
`
# helper method to ensure that 'freepats' soundfonts are configured for Timidity
function _config_soundfonts_freepats_dosbox()
{
    hasPackage fluid-soundfont-gm && exit 0

    iniConfig " " "" "/etc/timidity/timidity.cfg"
    iniSet "source" "/etc/timidity/freepats.cfg"

    systemctl is-active --quiet timidity && systemctl restart timidity
}
`
The last line would also restart any running timidity daemon instance, just in case.

psyke83: Also forgot to mention: the `exit 0` will also prematurely end the configure_dosbox function if that package is installed, so it should return instead.

cmitu: I forgot about the `hasPackage` - thanks for the pointers.

cmitu:
So, there are 2 additional scriptmodules including `freepats` and/or `timidity`
* (ports) **zdoom** . By default, it uses the system's MIDI device, whatever it is. It has options to switch to different MIDI devices, including `timidity`. When selected, it uses `timidity` directly to play the MIDI files. If the config file is incorrect (the issue at hand), then playing the background music sound fails.    
**zdoom** does however has an [OPL software synth](https://zdoom.org/wiki/OPL_synth_emulation)' option built in, which works without any soundfonts installed.

* (ports) **eduke32**. It supports MIDI playback through a built-in SDL_Mixer based synth or through the `EDUKE_MUSIC_CMD` environment variable. During testing, I used `pmidi` (with `timidity` started in the background) to play music, of course this doesn't work when `timidity` configuration fails.

Both `zdoom` and `eduke32` are not using by default `timidity`, they must be configured specifically. For **eduke32** there is no configuration option to choose the MIDI playback command, **zdoom** has the options in the _Sound Configuration_ screen.

joolswills:
I have edited the base for this to point to the new branch for fkms/rpi4/buster. Hope that's ok - let me know and I'll merge. Thanks.

psyke83:
I've revisited this, and think that the solution being implemented here may be a bit overkill. The point of my `midi_synth` function was to avoid the need to install `timidity-daemon` or otherwise make any system modifications outside the scope of `/opt/retropie`.

It seems to me that appending the configuration will work OK, and we don't even need to use `~/.timidity.cfg`. Simply changing `timidity -Os -iAD &` to `timidity -Os -iAD -c /etc/timidity/freepats.cfg &` should be sufficient. The only downside is that if `/etc/timidity/fluidr3_gm.cfg` is already installed, it may be used before freepats, but I don't think that's a huge deal?

cmitu:
That's a better solution to this problem.
The `fluid` sound fonts should have a better quality than `freepats`, so even if they're loaded before, it's ok. I'll add the modifications to the PR, thank you for the suggestion.

cmitu:
Based on @psyke83's suggestion, I've removed the helper and just start `timidity` a 2nd time if it's not starting with the default settings. 
Unfortunately `timidity` doesn't set an error code if a config error is found - probably because it forks itself in the background before any configuration is read - so 
```
timidity -Os -iAD || to timidity -Os -iAD -c /etc/timidity/freepats.cfg
```
always starts 2 instances. I've added a check before trying to stard the 2nd time, so we don't have 2 instances running.

cmitu:
I'll close this, seems it wasn't a big problem.

PR #2781, opened 2019-07-15
mkarcadejoystick: update the scriptmodule for kernels 4.15+ 

cmitu:
Similar to https://github.com/RetroPie/RetroPie-Setup/pull/2725, the driver fails buiding on 4.15+ Linux kernels.

I've submitted the patch to the - closest - upstream I could find, in the [Recalbox](https://gitlab.com/recalbox/mk_arcade_joystick_rpi/merge_requests/71) repository for the driver, but there's no response so far.

Reported in https://retropie.org.uk/forum/topic/22859, tested by the reporting user.

### DIFF
--- a/scriptmodules/supplementary/mkarcadejoystick.sh
+++ b/scriptmodules/supplementary/mkarcadejoystick.sh
@@ -17,7 +17,7 @@ rp_module_section="driver"
 rp_module_flags="noinstclean !x86 !mali"
 
 function _version_mkarcadejoystick() {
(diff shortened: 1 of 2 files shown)


joolswills:
I wonder if we should maybe fork this repo? Saves a patch and it's not an actively developed driver.

cmitu:
> I wonder if we should maybe fork this repo?

Probably yes, but I though of giving it a shot for the Recalbox repo, since it received some updates a few weeks ago.

PR #2782, opened 2019-07-16
Introduce ioquake3 scriptmodule

psyke83:
* Built against upstream ioquake3 source
* Suitable for Mesa (RPI3/4) and X11 targets
* Mesa targets use GL1 rendering - this works reasonably well, but
  has occasional rendering errors for some models (on RPI3).
* KMS (incl. Mesa) targets are configure to run at the display resolution.


joolswills:
Looks good thanks. Will merge after 5.1 image :-)

PR #2786, opened 2019-07-17
image: boot partition should be 256Mb for Buster

cmitu:
The recommendation for Buster (Pi4 related probably) is for a larger boot partition:
* https://www.raspberrypi.org/blog/buster-the-new-version-of-raspbian/#comment-1510398

Upstream `pi-gen` has been updated also with the new size:
* https://github.com/RPi-Distro/pi-gen/commit/0308e92705bc9a22b7574ab5acaa7623e9050b8f

### DIFF
--- a/scriptmodules/admin/image.sh
+++ b/scriptmodules/admin/image.sh
@@ -208,7 +208,7 @@ function create_image() {
 
     # make image size 300mb larger than contents of chroot
     local mb_size=$(du -s --block-size 1048576 $chroot 2>/dev/null | cut -f1)
-    ((mb_size+=300))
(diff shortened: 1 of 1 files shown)


joolswills:
Why the other increase btw?

joolswills:
Oh sorry. Due to boot increase. Did I use that in the calculations. I forget sorry.

cmitu:
Yes, you reserved 300Mb to cover for the `boot` partition, I just added the difference from 64 to 256 (192).

joolswills:
Actually I just thought maybe this should be dependent on which distro we are building but it's not an issue I guess.

PR #2787, opened 2019-07-19
Add detection for Vero4k+

DBMandrake:
Add detection for the Vero4k+. This is a minor hardware revision of the Vero4k which upgrades Ethernet to Gigabit and improves onboard flash performance. In other regards it is the same and should be treated as a minor revision of the same platform for the purposes of RetroPie and emulators.

This patch prevents it being reported as an unknown platform when trying to use retropie_setup.sh.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -268,7 +268,7 @@ function get_platform() {
             "Rockchip (Device Tree)")
                 __platform="tinker"
                 ;;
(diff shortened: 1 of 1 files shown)


DBMandrake:
I've updated the PR to make it a lot simpler and less redundant.

joolswills:
Thanks. Yeah was going to suggest that (or vero4k*)

PR #2790, opened 2019-07-23
vice: add fullscreen parameter for the (f)kms

cmitu:
Using the RPI `(f)kms` video driver, the emulator window is not fullscreen, but setting the`VICIIFullscreen` resource will make it start in fullscreen.


cmitu:
I re-targeted the `fkms` branch and fixed the logic error of the first attempt.

PR #2798, opened 2019-07-31
xroar: disable any X11 dependency for non-X11 platforms

cmitu:
The latest version will add an unwanted dependency on X11 if compiled without `--disable-kbd-translation` and fails to run on a Pi system. Reported in:
* https://retropie.org.uk/forum/topic/22623/gpi-case-xroar-emulator-crashes-solved/11
* (possibly also affects) https://retropie.org.uk/forum/topic/23084/

Other changes:
* SDL2 support is added and picked up if present by `configure`, so add the explicit dependency for `libsdl2-dev`.
* correct the typo in the start command line (`-ccr` vs `--ccr`).

I've also tested on a `(f)kms` enabled system and it looks fine, but we can remove the `!kms` flag later. In the 1st topic, the poster said the SDL2 version was slower, a `-sdl1` version can also be added if desired.

### DIFF
--- a/scriptmodules/emulators/xroar.sh
+++ b/scriptmodules/emulators/xroar.sh
@@ -17,7 +17,7 @@ rp_module_section="opt"
 rp_module_flags="!mali !kms"
 
 function depends_xroar() {
-    getDepends libsdl1.2-dev automake
+    getDepends libsdl2-dev automake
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

PR #2803, opened 2019-08-08
gles2n64.conf does not not like spaces around the '='

dankcushions:
iniSet permits this if you put it all in the first parameter. this, in conjunction with https://github.com/ricrpi/mupen64plus-video-gles2n64/pull/28 (now accepted) will allow gles2n64 to run at low resolution scaled to screen, as was the intention (rather than full resolution of display, as it currently is)

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -244,7 +244,7 @@ function configure_mupen64plus() {
         iniSet "fontColor" "1F1F1F"
 
         # Disable gles2n64 autores feature and use dispmanx upscaling
-        iniConfig " = " "" "$md_conf_root/n64/gles2n64.conf"
+        iniConfig "=" "" "$md_conf_root/n64/gles2n64.conf"
         iniSet "auto resolution" "0"
 
         addAutoConf mupen64plus_audio 1


scriptmodules/emulators/mupen64plus.sh
```diff
@@ -245,7 +245,8 @@ function configure_mupen64plus() {
 
         # Disable gles2n64 autores feature and use dispmanx upscaling
         iniConfig " = " "" "$md_conf_root/n64/gles2n64.conf"
-        iniSet "auto resolution" "0"
+        # Note: does not like spaces around the '='
+        iniSet "auto resolution=0"
```
hhromic: I don't really like this change. It's an ugly hack :)

You should instead do (notice I removed the spaces in the first parameter):

    iniConfig "=" "" "$md_conf_root/n64/gles2n64.conf"

To make every `iniSet` call to not use spaces around the `=` as it is needed.
Then you leave the `iniSet` call as it was. This is the way the `inifuncs` is meant to be used :)

dankcushions: oh, yes definitely that is better! i didn't notice it had that ability. will change

dankcushions: done! now doing it the proper way

hhromic: Looks good! I will assume you tested all so will approve the change :)

joolswills:
Thanks.

PR #2806, opened 2019-08-10
Fix bad Bluetooth input lag when all registered Bluetooth devices are…

c0d3h4x0r:
… already attached and the Bluetooth connection mode is set to "background".  Under those conditions, the original background script would repeatedly run "hcitool con" without any sleep time in between consecutive calls.  Debugging the script was difficult, and locating where the script came from when viewing processes with "htop" etc was also difficult, due to the script not being implemented by a file on disk -- so I've moved it out to its own file, added some helpful output (so that when it is run manually for debugging purposes, you can easily see what it's doing), and updated its logic to be as efficient as possible.  Due to the efficiency improvements, it's also now safe to decrease the retry interval from 10 seconds to 5 seconds.  Original repro was found with two 8bitdo SN30 Pro Bluetooth controllers both registered/paired and simultaneously connected.  The fix was validated using the same configuration, with all possible combinations of connection state (0/1/2 controllers connected at any given time out of a total of 2 registered controllers).

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -327,8 +327,7 @@ function boot_bluetooth() {
             while read mac_address; read device_name; do
                 macs+=($mac_address)
             done < <(list_registered_bluetooth)
-            local script="while true; do for mac in ${macs[@]}; do hcitool con | grep -q \"\$mac\" || { echo \"connect \$mac\nquit\"; sleep 1; } | bluetoothctl >/dev/null 2>&1; sleep 10; done; done"
-            nohup nice -n19 /bin/sh -c "$script" >/dev/null &
+            nohup nice -n19 /usr/bin/env bash "$md_data/bluetooth_background.sh" "${macs[@]}" >/dev/null &
             ;;
     esac
 }
--- a/scriptmodules/supplementary/bluetooth/bluetooth_background.sh
+++ b/scriptmodules/supplementary/bluetooth/bluetooth_background.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+registered_macs=("$@")
+
+function contains() {
+    local argc=$#
+    local value=${!argc}
+    for (( argi=1; argi < $argc; argi++ )); do
+        if [ "${!argi}" == "$value" ]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
+while true; do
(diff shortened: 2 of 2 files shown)


joolswills:
It sleeps for 10s between calls with the current script - I just did a test and it seems to be ok. Please can you provide details to reproduce the issue. 

Btw background shouldn't be needed for most controllers including the 8bitdo ones. It's a hack to workaround problems with some devices. I added it due to issues with old 8bitdo firmware but it's not needed on any of my 8bitdo devices now. (N30 pro. Sfc30 pro, etc)

c0d3h4x0r:
It does not sleep 10 seconds if all registered/paired devices are already connected.  It takes the first side of the “or” and doesn’t execute any of the rest, due to one of the “done” statements being in the wrong spot relative to the “sleep”.

Without using the background script, neither my ikbc Bluetooth keyboard nor my 8bitdo SN30 Pro controllers will reconnect to the Pi (3B+) after they fall asleep.  My 8bitdo SN30 Pro controllers are already running the latest firmware.  But before my change here, with the background script running, there’s massive input lag due to the background script making all those “hcitool con” calls.

(sent from my mobile phone)

On Aug 10, 2019, at 4:32 PM, Jools Wills <notifications@github.com<mailto:notifications@github.com>> wrote:


It sleeps for 10s between calls with the current script - I just did a test and it seems to be ok. Please can you provide details to reproduce the issue.

Btw background shouldn't be needed for most controllers including the 8bitdo ones. It's a hack to workaround problems with some devices. I added it due to issues with old 8bitdo firmware but it's not needed on any of my 8bitdo devices now. (N30 pro. Sfc30 pro, etc)

joolswills:
I will look into it. Sorry but at this time I don't want to use your replacement code.

c0d3h4x0r:
Ah shit, you're right -- I got my operator precedence incorrect.  I ran the following quick test, which busts a hole right through my hypothesis:

```
pi@retropie:~ $ true || { echo bob; echo foo; } | grep foo; echo bar;
bar
```

But still, there's a clear problem using the original code, and the problem is very clearly resolved by my new code.  But I guess I still don't understand exactly *why*.

One quick observation: the original code calls `hcitool con` once per registered device per iteration, which is inefficient.  My new code only calls it once per iteration, reusing the result for each device check.  But I'd be very surprised if that alone is sufficient to cause problems, given I only have 3 registered devices.

I'll keep digging on this.

c0d3h4x0r:
Question: why is there a `sleep 1` inside the compound statement that echos the commands to `bluetoothctl`?

c0d3h4x0r:
Also, why bother echoing `quit`?  It's also unnecessary, as far as I can tell.

The following appears to work fine (runs the `connect` command and then immediately exits back to the bash prompt):

```
pi@retropie:~ $ echo "connect 00:00:00:00:00:00" | bluetoothctl
Agent registered
[bluetooth]# connect 00:00:00:00:00:00
Device 00:00:00:00:00:00 not available
pi@retropie:~ $
```

c0d3h4x0r:
I updated to the latest stable Raspbian kernel, and now my ikbc Bluetooth keyboard and 8bitdo SN30 Pro Bluetooth controllers reconnect without using the "background" mode.  However, when I enable "background" mode (either with or without my changes), I still experience bad input lag.  I'm closing this pull request because it doesn't appear to clearly solve anything after all.

PR #2810, opened 2019-08-14
Add runcommand hook just after retroarch config generation

ChloeTigre:
This hook allows some auto-generated config to be inserted on the fly in /dev/shm/retroarch.cfg.
This allows the resolution switch system at https://github.com/ChloeTigre/es-rgbpi to embed a proper display configuration dynamically.


joolswills:
Thanks. I will consider changing the hooks (there was another request), but this won't happen until after fkms work.

ChloeTigre:
since this was implemented another way', closing it.

PR #2811, opened 2019-08-15
Add ecwolf port

tpo1990:
ECWolf is an advanced source port for Wolfenstein 3D, Spear of Destiny, and Super 3D Noah's Ark based off of the Wolf4SDL code base. ECWolf pulls a substantial amount of code from ZDoom and aims to provide a wide array of mod editing capabilities without the need to modify the source code.

### DIFF
--- a/scriptmodules/ports/ecwolf.sh
+++ b/scriptmodules/ports/ecwolf.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. This may have to wait until we have completed fkms/rpi4 work though.

tpo1990:
Thank you. No problem. It will wait until rpi4 work is complete.

tpo1990:
I have replaced the `downloadAndExtract` from source with `gitPullOrClone` which is a much better approach of pulling from the ecwolf bitbucket repo and pushed the change into this PR.

This has already been tested as working before updating the commits.

crcerror:
@joolswills 7 years later ... How do you think about it? Merge or leave?

Meanwhile ecwolf is a github repro but be sure I will never step into this script again

PR #2686, opened 2019-04-30
Update Solarus broken game links

TechnologyClassroom:
I updated Solarus to 1.6.0 and fixed the broken Solarus game source links.  I could not find the game source on the Solarus page so I pulled it from gitlab release branches.

PR #2813, opened 2019-08-19
solarus - update and modernise scriptmodule

hhromic:
General changes:

* Use the official repositories that are now in GitLab
* Use latest HEAD for OpenGL ES support and other required features
* Build with RPATH-replacement for proper dynamic library linking
* Moved Solarus from `ports` to `emulators` to be used as a system
* Added official theme assets to Carbon and Simple ES themes
* Configuration is symlinked to `$md_conf_root/solarus`
* Added a shell script launcher to assist with runtime options

Added scriptmodule GUI:

* Option to configure joypad axis deadzone
* Option to configure joypad buttons combo for quitting

Solarus quests (games) changes:

* Quests (games) no longer need to be "built"
* Do not ship games/quests anymore as the contain copyrighted material
* Users just have to download and place `.solarus` files in `$romdir/solarus`

Users should update their Carbon and Simple themes to get the new assets.

Closes #2686 and Closes #2453

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -206,6 +206,9 @@ sg-1000_fullname="Sega SG-1000"
 snes_exts=".7z .bin .bs .smc .sfc .fig .swc .mgd .zip"
 snes_fullname="Super Nintendo"
 
+solarus_exts=".solarus .zip"
+solarus_fullname="Solarus Engine"
+
 ti99_exts=".ctg"
 ti99_fullname="TI99"
 
--- a/scriptmodules/ports/solarus.sh
+++ b/scriptmodules/ports/solarus.sh
@@ -10,64 +10,133 @@
 #
 
 rp_module_id="solarus"
-rp_module_desc="solarus - An Open Source Zelda LttP Engine"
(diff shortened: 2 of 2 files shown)


joolswills:
Looks good. It doesn't need to be moved from ports to emulators to be a system though. Technically I think it should still remain in ports.

hhromic:
Modified it now to remain in ports. I had to add `setConfigRoot ""` so the "emulator" gets configured in `$configdir/solarus` instead of `$configdir/ports/solarus`, otherwise `runcommand` won't find the config. 

The only other "port as a system" that does this is `love`, also another game engine.

All tested again and ready to merge.

joolswills:
Yeah. There are others that should be in ports also really. Like AGS. Scummvm.

Historical mostly. Moving is a pain as it has a different install location.

joolswills:
Please merge if you're happy. Thanks.

hhromic:
Okay I added the Solarus system to `platforms.cfg` as well, like AGS and Löve. So they are all consistently implemented at least.
Everything seems to be working fine on my end so I'll merge. If anything arises, I will update.
Thanks for the feedback!

PR #2815, opened 2019-08-21
$biosdir needs to be initialised in runcommand launch scripts

dankcushions:


hhromic:
What this change really addresses?
The correct display of the dialog box message if there are no BIOSes found (which uses `$biosdir`)?

dankcushions:
precisely. it also tidies up the code a little.

hhromic:
Just wanted to confirm. I agree with your changes :)

joolswills:
Where does it hardcode it twice? I only see one change here?

Actually really this should be passed in from calling script so it's not hardcoded at all. But I'm also not sure we need this "helper" dialog.

joolswills:
Ah I see. It referenced it before so it wasn't working (missed that). I think I'd just remove the dialog still.

dankcushions:
You don't think it's useful? It was useful to me as I didn't have the BIOS files in the right place (they were changed to a /dc/ subfolder since the last time I used it. It just had a small bug :)

joolswills:
We don't do it for other platforms that need bios files so it seems inconsistent. If we are to have this we shouldn't hardcode this at all in the file though.

joolswills:
I'll merge it, as it had a bug, but I think I'd rather remove it still tbh.

PR #2816, opened 2019-08-21
Flycast has different compile options for Raspberry Pi 4

dankcushions:
see https://github.com/libretro/flycast/blob/master/Makefile#L166

### DIFF
--- a/scriptmodules/libretrocores/lr-flycast.sh
+++ b/scriptmodules/libretrocores/lr-flycast.sh
@@ -30,8 +30,14 @@ function sources_lr-flycast() {
 function build_lr-flycast() {
     make clean
     if isPlatform "rpi"; then
+        local params=()
+        if isPlatform "rpi4"; then
+            params+=("platform=rpi4")
+        else
+            params+=("platform=rpi")
+        fi
         # MAKEFLAGS replace removes any distcc from path, as it segfaults with cross compiler and lto
-        MAKEFLAGS="${MAKEFLAGS/\/usr\/lib\/distcc:/}" make platform=rpi
+        MAKEFLAGS="${MAKEFLAGS/\/usr\/lib\/distcc:/}" make "${params[@]}"
     else
         make
     fi


scriptmodules/libretrocores/lr-flycast.sh
```diff
@@ -31,7 +31,13 @@ function build_lr-flycast() {
     make clean
     if isPlatform "rpi"; then
         # MAKEFLAGS replace removes any distcc from path, as it segfaults with cross compiler and lto
-        MAKEFLAGS="${MAKEFLAGS/\/usr\/lib\/distcc:/}" make platform=rpi
+        MAKEFLAGS="${MAKEFLAGS/\/usr\/lib\/distcc:/}"
```
hhromic: This won't work as expected because the variable `MAKEFLAGS` is not exported, i.e. `make` will not see it if not given as an invocatio-variable (like is done before this PR).

You either do:

    VARIABLE=foo command

or:

    export VARIABLE=foo
    command

However, I think a better solution is to adopt the following idiom (used in other scriptmodules):

    local -a params=()
    if isPlatform "rpi4"; then
        params+=("platform=rpi4")
    else
        params+=("platform=rpi")
    fi
    MAKEFLAGS="${MAKEFLAGS/\/usr\/lib\/distcc:/}" make "${params[@]}"

dankcushions: thanks for the correction! changes made and seem to test fine (although I don't have a pi4 so I'm presuming the new compile options are good since they were tested upstream). Squashed and good to go

scriptmodules/libretrocores/lr-flycast.sh
```diff
@@ -30,8 +30,14 @@ function sources_lr-flycast() {
 function build_lr-flycast() {
     make clean
     if isPlatform "rpi"; then
+        local params=()
```
hhromic: Just a small extra change. The `local` declarations should be in the main function body scope, not inside `if` blocks. Otherwise `params` will not be initialised if `isPlatform "rpi"` is `false`.
Simply move the local declaration to the beginning of the function and that will do it.

dankcushions: but params are not used if not rpi?

hhromic: Oh you are right, didn't notice that the `make` invocation is inside the `if` block and there is yet *another* `make` invocation later down below. That maybe can be improved but I agree it is not the purpose of this PR, so I won't bug about it.

In bash there is not really strong scoping, i.e. no `if` block scoping, so putting `local`s at the beginning of the functions is more good practice for legibility, i.e. "hey this function has the following locals", than required programming syntax.

The majority (if not all?) of scriptmodules in RetroPie using `local params=()` declare it at the beginning of functions so I would stick to that for now.

joolswills:
If parameterising this is definitely get rid of one of the makes.

hhromic:
The problem is that I'm not sure how to properly parameterise passing env-variables to the `make` invocation, e.g. the `MAKEFLAGS` env-variable which is not a parameter for `make`.

Perhaps something like this?

    if condition; then
        MAKEFLAGS=MAKEFLAGS="${MAKEFLAGS/\/usr\/lib\/distcc:/}"
        if isPlatform "rpi4"; then
            params+=("platform=rpi4")
        else
            params+=("platform=rpi")
        fi
    fi
    MAKEFLAGS="$MAKEFLAGS" make "${params[@]}"

That would work? It is just from the top of my head, didn't test.

joolswills:
The MAKEFLAGS change can be included - no need to parameterise that. This is one of those times making the change is quicker than doing a review. :-)

joolswills:
I'm happy to just sort this later if needed Vs fixing up this PR.

joolswills:
Plus it should be against the branch not master.

dankcushions:
I won't be offended :) just something I noticed in my travels. I can attempt fixing it up later if not

hhromic:
> This is one of those times making the change is quicker than doing a review. :-)

I like these discussions tbh (yes I'm a masochist sometimes) and I think is good for @dankcushions (and me) to learn more as well :)

> Plus it should be against the branch not master.

Ah yes, I'm not yet used that there is now an RPI4-specific branch.
You are 100% right it should be PRed to that one.

Let's decide quickly on closing or not the PR so at least it doesn't pile-up.

joolswills:
I'll merge then do the changes. I changed the base to the rpi4 branch.

PR #2817, opened 2019-08-22
amiberry - put all the platform / bin logic in a single function

joolswills:


scriptmodules/emulators/amiberry.sh
```diff
@@ -16,6 +16,27 @@ rp_module_licence="GPL3 https://raw.githubusercontent.com/midwan/amiberry/master
 rp_module_section="opt"
 rp_module_flags="!x86"
 
+function _get_platform_bin_amiberry() {
+    local choice="$1"
+    local amiberry_bin="$__platform-sdl2"
+    local amiberry_platform="$__platform-sdl2"
+    if isPlatform "rpi" && ! isPlatform "kms"; then
+        amiberry_bin="$__platform-sdl1"
```
hhromic: RPI0/1/2/3 would get `sdl1` here?
Aren't at least RPI2/3 good enough for `sdl2`?
You know better, just wondering if it is a vestigial decision from the old past.

joolswills: There's a discussion on a previous amiberry ticket. I may switch to SDL2 but I'm still testing (I discussed this the other day in our dev chat).

PR #2818, opened 2019-08-23
esthemes: add MiniJawn theme

cmitu:
Added with author's permission - https://retropie.org.uk/forum/topic/22621.
The theme is designed for small screens (like the one present in the RetroFlag's GPICase)

Screenshot


joolswills:
Thanks.

PR #2819, opened 2019-08-24
Add FUNDING Github metadata

hhromic:
* Adds a "sponsors" button in Github

You might need to check if funding links are enabled on the repository settings.


joolswills:
Thanks.

PR #2829, opened 2019-09-02
lr-nestopia - fix build after upstream repository changes

cmitu:
Seems like the upstream repository was re-created as a standalone repository (instead of a fork) and some re-organization has taken place. The files no longer in the repository have been removed from `md_ret_files` to allow installation from source of the core.


joolswills:
Thanks.

PR #2832, opened 2019-09-04
solarus - minor improvements to scriptmodule

hhromic:
Minor improvements, does not change behaviour but adds safety checks.

* From acba46e0: "disable on aarch64 (only armhf libluajit-5.1-dev currently available in odroid-c2 repo)"
* Previously forgot to use `md_ret_require` to validate that the build phase generates the binary correctly.


joolswills:
Cheers.

PR #2836, opened 2019-09-08
image: really set boot partition to 256MiB

tsoliman:
The Pi4 Buster recommendation is actually 256MiB but the script was creating
a much smaller boot partition (240MiB)  because of two bugs:

- parted mkpart units defaults to MB not MiB for input (256MiB is 268MB)

- parted's parameters are `start` `end` not `start` `size`. We need to specify
  260 as an end because the start is 4

I changed the default unit to MiB because while explicitly specifying 260MiB
is possible, specifying units on mkpart disables parted's helpful range of
sloppiness.

### DIFF
--- a/scriptmodules/admin/image.sh
+++ b/scriptmodules/admin/image.sh
@@ -218,9 +218,10 @@ function create_image() {
     printMsgs "console" "partitioning $image ..."
     parted -s "$image" -- \
         mklabel msdos \
-        mkpart primary fat16 4 256 \
+        unit mib \
+        mkpart primary fat16 4 260 \
         set 1 boot on \
-        mkpart primary 256 -1
+        mkpart primary 260 -1s
 
     # format
     printMsgs "console" "Formatting $image ..."


tsoliman:
In case you were wondering why I changed `-1` to `-1s`:
`-1` meant "1 unit from the end", the default unit being MB, (changing the unit to MiB made it mean `1MiB from the end`) we really want to mean "the end" which is what `-1s` means (the last sector, regardless of default unit)

hhromic:
I'm not saying your solution is wrong, but I wonder why not adopt how it's done by the RPI foundation in `pi-gen`?
https://github.com/RPi-Distro/pi-gen/blob/master/export-image/prerun.sh

tsoliman:
I think this approach is cleaner than pi-gen's. 

- parted is much smarter than fdisk when it comes to its "helpful" sloppiness.

- It is also more scriptable.  Note the way the interactive prompts in fdisk are being fed keystrokes in that script. 

- Telling the tool the units we want to roughly hit and letting it correct with the knowledge of what we are trying to do is much better functionally and more readable than converting to bytes ourselves and passing bytes to it. 

All of this is out of scope for the PR, though. I'd be happy to discuss elsewhere.

hhromic:
That's great and you give very good points imho.
I don't think there is need to discuss further from my side as I agree with all you said.
Thanks for the PR.

joolswills:
Thanks. Looks good. From my memory -1 worked when I wrote this - as the default unit was `s` as I recall. Maybe something changed.

joolswills:
I probably remember wrong - makes more sense to specify anyway,

PR #2841, opened 2019-09-11
duke3d overhaul and ionfury

psyke83:
eduke32:
* Updated to r8090 with patches for:
  - startup delay on ARM devices (r6918)
  - fix skybox render error with r_useindexedcolortextures disabled and don't
    explicitly force-disable this setting for the commercial release (r7424)
  - glTexImage2D/glTexSubImage2D uses internal format incompatible with VC4/VC6 (r6776)
  - disabled netcode due to r7242 introducing a major bump in memory
    requirements for the executable to even dynamically link on RPI3; will
    investigate later, or worst case scenario perhaps we can provide an alternate
    port launcher with revision r7241 or earlier for netplay only. (r7242)
* enforce vsync for KMS targets
* Set r_useindexedcolortextures for Mesa devices to avoid palette corruption
  on startup splash.
* Disable GTK support for non-X11 targets
* Ensure GL support is really enabled for non-X11 GL targets
* Enable GL support for Mesa targets (tested on RPI3 & 4)
* Add full support for official Megaton addons:
  - to install base game, place duke3d.grp and DUKE.RTS in roms/ports/eduke32.
  - to install addons, simply place the "addon" folder from the Megaton edition
    to roms/ports/eduke32 and reinstall the module to generate the launcher items.
  - note: case sensitivity didn't appear to matter when copying the files directly
    from my Steam installation.

ionfury:
* written as a skeleton script that defers to eduke32 functions when possible.
* all generic improvements to eduke32 also apply
* note: performance is not yet satisfactory on RPI3 or RPI4; it's possible to
  launch and play, albeit running at ~10-15fps. This doesn't appear to be an issue
  with Pi drivers per se, as people with older spec PCs complain about performance on
  the eduke32 forums. Optimization may hopefully improve in future builds.
* Installation:
  - Copy fury.def, fury.grp and fury.grpinfo to roms/ports/ionfury


psyke83:
Tested on RPI3 (legacy + fkms), RPI4 and i965 (X11 & KMS). All working fine with caveats:
* for Duke, OpenGL mode is a bit sluggish on RPI3/4, but you can still select the classic renderer in the options to get excellent performance (appears to be 60fps; RPI3 fkms runs faster than RPI3 with legacy driver).

Ion Fury is unplayable on RPI3, but may be considered borderline playable on RPI4 using the classic renderer. I feel that including this is still beneficial, since we can now build against the latest version of eduke32, and perhaps if the developers see that there is actually interest for a niche platform like RPI, it may help spur further optimizations for low end devices. It runs full speed on my laptop using the i965 driver.

When this is merged we can cut the entire part of the wiki related to addons, as moving forward it will only require a simple drag and drop of the `addons` folder from the Megaton edition of Duke3D.

joolswills:
Thanks - looks great.

PR #2842, opened 2019-09-12
atari800 - switch to using github repo and update to v4.1.0

joolswills:
* added automake dependency
 * update patch - we only change the compiler name to gcc now


joolswills:
Added basic archive support also using launch scripts.

Issue #2669, opened 2019-03-30
Updating DraStic to 2.5.0.4


DanTheMan827:
There's a Linux version of DraStic 2.5.0.4 that's fully SDL2 and fixes compatibility for games that use the microphone.

Is there any particular reason that this isn't used?

https://github.com/Retro-Arena/binaries/blob/master/odroid-xu4/drastic.tar.gz?raw=true

quicksilver7837:
There have been previous permission issues with distribution of drastic. I'll reach out to exophase (the author of drastic) on the drastic forums and see if he's ok with it.

DanTheMan827:
there's also xperia64

quicksilver7837:
https://drastic-ds.com/viewtopic.php?f=5&t=5610

I made a post on their forums. Lets hope I get a response.

robertybob:
Exophase said yes so that's great news!

2play:
@quicksilver7837 there was a request also and the devs forum and approved the use of that binary 
See here
https://drastic-ds.com/viewtopic.php?t=5610

mrcmunir:
If someone wants to try it under a pure distribution of aarch64 64bits for some devboards that are based on debian or ubuntu it's possible with multilib 

There workaround drastic run on aarch64

sudo dpkg --add-architecture armhf
sudo apt-get update
sudo apt-get install libc6: armhf libstdc ++ 6:armhf

Install small dependencies 
sudo apt-get install libsdl2-2.0-0:armhf
sudo apt-get install libasound2-plugins:armhf
sudo apt-get install zlib1g: armhf
sudo apt-get install libgtk3-nocsd0:armhf

joolswills:
This version is now included.

theofficialgman:
> If someone wants to try it under a pure distribution of aarch64 64bits for some devboards that are based on debian or ubuntu it's possible with multilib
> 
> There workaround drastic run on aarch64
> 
> sudo dpkg --add-architecture armhf
> sudo apt-get update
> sudo apt-get install libc6:armhf libstdc++6:armhf
> 
> Install small dependencies
> sudo apt-get install libsdl2-2.0-0:armhf
> sudo apt-get install libasound2-plugins:armhf
> sudo apt-get install zlib1g:armhf
> sudo apt-get install libgtk3-nocsd0:armhf

[...]

PR #2843, opened 2019-09-13
drastic: update to 2.5.0.4 with RPI4 compatibility

psyke83:
Permission for new release has been granted; see: https://drastic-ds.com/viewtopic.php?f=5&t=5610

Please note that while RPI4 (and RPI3/fkms) has been confirmed as working by me,
the developers have not explicitly advertised this release as supporting
RPI4; it works only because it is built as a generic arm binary that is
not dependent on the proprietary Broadcom GLES libraries.

Please keep the above in mind if reporting any bugs upstream.

Closes https://github.com/RetroPie/RetroPie-Setup/issues/2669


joolswills:
Thanks.

joolswills:
Will cherry pick this to master also.

PR #2844, opened 2019-09-13
cgenius: Added support for ROM arguments

crcerror:
Just a minor enhancement
We can launch keen episodes by using single shell scripts call rather then to use the cgenius GUI

### DIFF
--- a/scriptmodules/ports/cgenius.sh
+++ b/scriptmodules/ports/cgenius.sh
@@ -41,7 +41,7 @@ function install_cgenius() {
 }
 
 function configure_cgenius() {
-    addPort "$md_id" "cgenius" "Commander Genius" "pushd $md_inst; ./CGeniusExe; popd"
+    addPort "$md_id" "cgenius" "Commander Genius" "pushd $md_inst; ./CGeniusExe dir=games/%ROM%; popd"
 
     mkRomDir "ports/$md_id"
 


hhromic:
How does `%ROM%` work here when the component is installed as a port?
Does this not break normal installations?

crcerror:
Yes, have to provide a valid path as I wrote in the request - yes you're right.
But you can remove the `games` path then it would work.

```
addPort "$md_id" "cgenius" "Commander Genius" "pushd $md_inst; ./CGeniusExe dir=%ROM%; popd"
```
The changes above will provide usual GUI launch just by calling the provided shell file created by the script module.

But we can also use a shell file this now...
```
"/opt/retropie/supplementary/runcommand/runcommand.sh" 0 _PORT_ "cgenius" "games/keen1 - Marooned on Mars (special) (1990)"
```

I changed the [Retropie forum entry: Cgenius Launch Menu](https://retropie.org.uk/forum/topic/23148) according this new suggestion. Thanks for pointing me in right direction and finding a generic solution.

hhromic:
I'm still confused :(

Without this PR (original behaviour), the scriptmodule will create a `Commander Genius.sh` script in the `ports` directory that simply launches `./CGeniusExe`, right? In this case `%ROM%` is assigned to `/home/pi/RetroPie/ports/Commander Genius.sh` by ES but is not used.

With this PR, the scriptmodule will still create a `Commander Genius.sh` script in the `ports` directory that now launches `./CGeniusExe dir=%ROM%`, independent to any shellscripts you add later for the individual episodes. If the user clicks this default shellscript, and because ES assigns `%ROM%` like before, it would be then called as

    ./CGeniusExe dir=/home/pi/RetroPie/ports/Commander Genius.sh

is this true?

This according to how `addPort` works: https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/helpers.sh#L1194

crcerror:
No then you need to assign the script inside the addPort
```
addPort "$md_id" "[system]" "[title]" "[command emulators.cfg]" "[ROM to load]"
```

as the [ROM to load] part is empty the %ROM%-call is empty and the usual graphical GUI will be loaded. This is `$5` in runcommand-script

hhromic:
Ah, so you are saying that in the specific case of CGenius, the command

    ./CGeniusExe dir=

is equivalent to previous command

    ./CGeniusExe

That is, the GUI is shown when `dir=` ?

crcerror:
Yes it is.
CGenius seems to ignore an invalid (=empty) argument.
But the PR i've done is (of course) wrong
```
./CGeniusExe dir=games/
```
will lead to a "No gamedata found" error message.
But thank you for let me (re)-think about it.

PR #2848, opened 2019-09-17
mkarcadejoystick: change the source repository & add Pi4 support

cmitu:
Since upstream is no longer active, fork the upstream repo and update the source in a `retropie` branch to include:

* compatibility with Linux kernels >= 4.15 (added as a patch in https://github.com/RetroPie/RetroPie-Setup/pull/2781)
* compatibility with the Pi 4B model. 

The module uses a similar approach as the [gamecon_gpio_rpi](https://github.com/marqs85/gamecon_gpio_rpi) module to initialize the PI's [peripherals address](https://www.raspberrypi.org/documentation/hardware/raspberrypi/peripheral_addresses.md), the modifications for Pi 4B support are similar to the ones in https://github.com/marqs85/gamecon_gpio_rpi/pull/4.

I don't have a proper hardware setup to fully test this, the module compiles and loads on a Pi 3B (Raspbian Stretch) and a Pi 4B (Raspbian Buster). The original `gamecon_gpio_rpi` changes (sent upstream) have been tested [here](https://retropie.org.uk/forum/topic/22384/gamecon-driver-fails-to-build-on-kernel-4-19-x/23).

I also updated the removal function to delete all module versions (similar to the other GPIO modules) and bumped the version.


Superpiffer:
What about the hotkeybtn branch? Any chance to upgrade it too?

cmitu:
@Superpiffer we never used that branch - so the Pi4 suport was not added to it. Are you installing the driver manually ?

Superpiffer:
I need an extra hotkey button for my new rpi3 arcade machine so I'm using the hotkeybtn branch. In order to use it I edited the mkarcadejoystick script from stable (with the original repository):
`gitPullOrClone "$md_inst" https://github.com/recalbox/mk_arcade_joystick_rpi`
to
`gitPullOrClone "$md_inst" https://github.com/recalbox/mk_arcade_joystick_rpi hotkeybtn`

After that I installed this driver normally from retropie_setup.sh and it compile and works properly. I don't own a rpi4 so I don't know if the current patch is compatible also with this branch.

cmitu:
> After that I installed this driver normally from retropie_setup.sh and it compile and works properly. I don't own a rpi4 so I don't know if the current patch is compatible also with this branch.

There isn't any patch for that branch, since it's not used in RetroPie. We didn't support that branch before this update, so the situation hasn't changed. 

I initially submitted the previous modifications (https://github.com/RetroPie/RetroPie-Setup/pull/2781) to the (current) upstream repo, on the the _hotkey_ branch, but I haven't received a response.  Since there was no interest in integrating the changes upstream, the modifications in this PR only touch the main branch.

If there is an interest in adding compatibility for future Linux kernel versions and/or PI4, I could add it, but I don't have a way to test it right now. If you intend to get a PI4 and use it with the gamepad/joystick, then let me know and - when you can test it - we can add the necessary modifications to the _hotkey_ branch. I'd prefer if you'd use the forum instead of the Github repo, though.

Superpiffer:
My only fear is to have a non functional button on my newly built arcade machine in future versions of RetroPie. I'll submit such a request on the forum in the future if necessary, at the moment I could only test my rpi3 with a newer kernel...
Thanks and sorry for the github post!

PR #2849, opened 2019-09-17
splashscreen: RPI4/fkms support

psyke83:
* Avoid VT2 switch in KMS mode, otherwise EmulationStation will run on the
  llvmpipe software rendering path.
* Explicitly invoke --no-osd for omxplayer (set by default on RPI4, but needs
  to be set for earlier board revisions in KMS mode).

### DIFF
--- a/scriptmodules/supplementary/splashscreen/asplashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen/asplashscreen.sh
@@ -10,6 +10,7 @@ do_start () {
     local config="/etc/splashscreen.list"
     local line
     local re="$REGEX_VIDEO\|$REGEX_IMAGE"
+
     case "$RANDOMIZE" in
         disabled)
             line="$(head -1 "$config")"
@@ -32,7 +33,7 @@ do_start () {
         while ! pgrep "dbus" >/dev/null; do
             sleep 1
         done
-        omxplayer -o both -b --layer 10000 "$line"
+        omxplayer --no-osd -o both -b --layer 10000 "$line"
     elif $(echo "$line" | grep -q "$REGEX_IMAGE"); then
         if [ "$RANDOMIZE" = "disabled" ]; then
             local count=$(wc -l <"$config")
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -12,7 +12,7 @@
 rp_module_id="splashscreen"
 rp_module_desc="Configure Splashscreen"
 rp_module_section="main"
-rp_module_flags="noinstclean !x86 !osmc !xbian !mali !kms"
+rp_module_flags="noinstclean !x86 !osmc !xbian !mali"
 
 function _update_hook_splashscreen() {
     # make sure splashscreen is always up to date if updating just RetroPie-Setup
@@ -39,8 +39,8 @@ function install_bin_splashscreen() {
 [Unit]
 Description=Show custom splashscreen
 DefaultDependencies=no
-Before=local-fs-pre.target
-Wants=local-fs-pre.target
+After=console-setup.service
+Wants=console-setup.service
 ConditionPathExists=$md_inst/asplashscreen.sh
 
 [Service]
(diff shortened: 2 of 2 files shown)


scriptmodules/supplementary/splashscreen/asplashscreen.sh
```diff
@@ -10,6 +10,13 @@ do_start () {
     local config="/etc/splashscreen.list"
     local line
     local re="$REGEX_VIDEO\|$REGEX_IMAGE"
+    local splash_vt=2
+
+    # switching VT in KMS mode will break graphics acceleration on RPI4
+    if grep -q okay /proc/device-tree/soc/v3d@7ec00000/status 2> /dev/null || grep -q okay /proc/device-tree/soc/firmwarekms@7e600000/status 2> /dev/null; then
```
hhromic: it's guaranteed that `v3d@7ec00000` is stable across devices/setups?
There isn't a more robust way to check this?

psyke83: It's stable across devices. There may be a more optimal way to detect it, but I'd prefer to leave it this was as it's the precise method that raspi-config uses: https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config#L74-L80

hhromic: Ah! if `raspi-config` uses it like that, then I def agree is good to leave it as-is. Thanks for clarifying!

psyke83:
Don't merge this yet; it seems to break console-setup customization made by the consolefont scriptmodule. Will troubleshoot when I have time.

joolswills:
There was a reason we didn't display on vt1 btw, but I can't remember - but I have a feeling I have a branch with some code changes. Maybe it interfered with boot msgs, but it's very old code so worth a revisit. 

The losing acceleration on terminal switch feels like something we should feed upstream. At least to get some information as to why if it's by design.

joolswills:
My vt comments unrelated to rpi4 of course. Just came to mind. I'm not against overhauling this code anyway. Thanks.

psyke83:
@joolswills 

Sure, but it may be complicated to pin down the cause. What I know about the issue is:
* It only happens on RPI4's V3D driver, which exposes `/dev/dri/renderD128` (which doesn't exist for VC4 of older boards).
* It's not a problem unique to `fbi` and `EmulationStation`. If you are in a VT different to VT1 (with no graphical apps running) and start the `lightdm` service, anything you launch in the Xorg session will not be accelerated. People have been also having the issue intermittently while testing [Kodi](https://www.raspberrypi.org/forums/viewtopic.php?f=38&t=246837&p=1533309#p1533309), too.
* A workaround is to add your user to the `render` group, but that's probably a bad idea, as systemd is supposed to grant access to the render group when needed via ACLs.

cmitu:
I have a setup using `omxiv` instead of `fbi`, which uses OpenMAX (dispmanx) to show images, similar to `omxplayer`. It doesn't use KMS/DRM at all and it behaves similarly to the way `fbi` works without KMS active. Would that be a solution ?

psyke83:
Updated PR and changed base to master. To be clear, the service is now delayed until after `console-setup.service`, as that seems to be what's responsible for prematurely interrupting `fbi`. You'll see some boot log before the splash starts, but it will persist until EmulationStation is almost finished loading (in my case).

It seems that after recent Raspbian updates, EmulationStation isn't getting stuck in software rendering mode, either. Tested on RPI3 and 4 a few times, and I couldn't reproduce the issue, but others should verify that it's OK.

cmitu:
I gave it a quick test, but looks like the SW renderer is still used by EmulationStation on PI4 with the _fkms_ overlay:
```
...
lvl2:   Creating window...
lvl2:   Created window successfully.
lvl2:   GL vendor:   VMware, Inc.
lvl2:   GL renderer: llvmpipe (LLVM 9.0.1, 128 bits)
lvl2:   GL version:  3.1 Mesa 19.3.2
lvl2:   Checking available OpenGL extensions...
lvl2:    ARB_texture_non_power_of_two: ok
...
```
I have the MESA _19.3.2-1~bpo10+1~rpt1_ packages installed, libSDL2 is still at _2.0.10+4rpi_.

joolswills:
Closing as the other splashscreen PR is integrated.

PR #2851, opened 2019-09-21
kodi: unblock installation for fkms (buster)

psyke83:
As of version 2:18.4-4~buster, pipplware's Kodi package ships with
kodi-rbpi_v8 built to target Mesa/KMS for RPI4, but also works on
RPI3 in fkms mode.

The v7 binary is still present, so the packages will work for older
board revisions with the legacy graphics drivers.

Thanks to Rascas for providing the packages & updating the launcher
script to support RPI3 in fkms mode, and popcornmix (built from his
leia_pi4 Kodi branch).


joolswills:
Assumed this could be merged. Thanks.

PR #2852, opened 2019-09-22
splashscreen: added KMS/RPI4 support (2)

cmitu:
This commit changes the `splashscreen` from using `fbi` to [omxiv](https://github.com/HaarigerHarald/omxiv), an OpenMax image viewer for Raspberry PI that bypasses the KMS/DRM layer (similar to `omxplayer`). I added a separte module for it, using a modified version which adds a few `fbi`-compatible features (randomisation or images, exiting after a slideshow, reading the image list from a text file).

It's backwards compatible with non-KMS setups (I tested on a RPI3 using the _master_ RetroPie branch).

Notes:
 - on `kms` enabled systems, there is a video mode initialisation done after boot, when the DRM module gets loaded, so I added an extra dependency for the `systemd` service on module loading. Otherwise, the splashscreen gets cut off after the DRM initialisation. This doesn't happen for `omxplayer` since it's started later, after the `dbus` service.
- `omxiv` needs a version of the `libraspberrypi-dev` package released after Sept. 2018 (because it needs [this commit](https://github.com/raspberrypi/userland/commit/58f4b168d3368928e80d72f846a99ba926191f02) for `OMX_DISPLAY_ALPHA_FLAGS_MIX`. If we intend to support older versions, I'll see if we can modify `omxiv` to not depend on this flag (it's only used for alpha blending).

This is an alternative approach to implement https://github.com/RetroPie/RetroPie-Setup/pull/2849. It might be worth finding a generic KMS alternative in the future, it could benefit more systems than the Raspberry Pi.

### DIFF
--- a/scriptmodules/supplementary/omxiv.sh
+++ b/scriptmodules/supplementary/omxiv.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="omxiv"
+rp_module_desc="OpenMAX image viewer for the Raspberry Pi"
+rp_module_licence="GPL2 https://raw.githubusercontent.com/cmitu/omxiv/master/LICENSE"
+rp_module_flags="!x86 !osmc !xbian !mali"
+
+function depends_omxiv() {
+    getDepends libraspberrypi-dev libpng-dev libjpeg-dev
+}
+
+function sources_omxiv() {
+    gitPullOrClone "$md_build" https://github.com/retropie/omxiv.git
+}
+
+function build_omxiv() {
+    make clean
+    make ilclient
+    make
+    md_ret_require="omxiv.bin"
+}
+
+function install_omxiv() {
+    make install INSTALL="$md_inst"
+}
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -12,7 +12,7 @@
 rp_module_id="splashscreen"
 rp_module_desc="Configure Splashscreen"
 rp_module_section="main"
-rp_module_flags="noinstclean !x86 !osmc !xbian !mali !kms"
+rp_module_flags="noinstclean !x86 !osmc !xbian !mali"
 
 function _update_hook_splashscreen() {
     # make sure splashscreen is always up to date if updating just RetroPie-Setup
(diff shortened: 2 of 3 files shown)


cmitu:
I just noticed that on a cold boot on PI4, the delay added to the service is not enough. I'll investigate, but a simple workaround to make it work is to add the delay (waiting for `dbus`) used for `omxplayer`.

cmitu:
I've decided to drop the service override on `fkms`, since the `asplashscreen` service should start early and it's required to start before `local-fs-pre`. I've added a 1.5 sec delay when the `fkms` overlay is detected, it seems just enough for `vc4drmfb` to initialise and not mess up the splashscreen.

joolswills:
I'm happy for this to be merged and thanks for your work on this. Are the changes in @psyke83 's previous PR needed also.

Are you ok if I pull your fork in so it's managed under RetroPie?

cmitu:
> I'm happy for this to be merged and thanks for your work on this. Are the changes in @psyke83 's previous PR needed also.

I'll take a look and adjust if needed.

> Are you ok if I pull your fork in so it's managed under RetroPie?

Sure, just clone the original repo (https://github.com/HaarigerHarald/omxiv) and I'll add the extra commits from my repository.

cmitu:
I've changed the service dependency, as in @psyke83's PR for the splashscreen, and removed the delay when displaying an image. Seems to work fine on the PI4 (Buster), the splashscreen doesn't seem to get interrupted by the KMS switch.

 Since the image is now displayed in a _dispmanx_ layer, it will stay on-top on Emulationstation when using KMS, so I lowered the display delay from 20 sec to 12, otherwise on the PI4 the splashscreen would be displayed too long. On a non-KMS platform (i.e. Pi3 and older), Emulationstation will interrupt the splashscreen, as before.
 
 I'll squash the commits.

joolswills:
Thanks for your work on this. Can probably work around the time to display. Configurable length / script to stop it when es loads etc. Nice work on getting round having to have a delay before launch as things like that are always unreliable.

Merging.

cmitu:
We can make the delay configurable and also separate the conf. parameters from the script. On update, the configuration is always reset due to the `asplashscreen.sh` being overwritten.

2play:
@cmitu 
I was playing around on a pi3B setup. I have enabled full kms.
All installed ok but splashscreen wont is blank either a static or a video.
When I enable fkms (dtoverlay=vc4-fkms-v3d) I can see video splash but ES is laggy.

Any suggestion on what to change so splash can run under full kms
dtoverlay=vc4-kms-v3d?

Kind regards

cmitu:
@2play please use the forum for support questions.

2play:
@cmitu HI Cristi
No worries. I thought just wanted to brought this up  for your ref and work. 
I dont think anyone noticed that or tested pi3 and full kms splashscreen

joolswills:
It's not supported.

2play:
@joolswills thank you. I thought so. I noticed today on pi4 is fksm and kms has been removed from raspi-config
so it makes sense
Tx again

PR #2853, opened 2019-09-22
retroarch: update to 1.7.9

cmitu:
Some changes since 1.7.6:
 - upstream dropped the `video_shader` parameter and replaced it instead with a CLI option (`--set-shader`), but in my tests this is not behaving the same - the shader from the command line overrides any game overrides, whereas before, the shader saved for the game would override the one in the config file. Also, for some reason the default for enabling the Video Shaders has been changed to _false_. I added a patch to bring back reading the `video_shader` parameter when every other auto-shader preset (game/core/content/global) is not present.  
 ~~This modification impacts the **Configuration Editor**, which has been used so far to set the shader per system/globally through the individual system's `retroarch.cfg` files.~~

- the default compile options enable support for OpenGL1, I disabled them on the Pi when the BRCM libs are used since it produces a linking error (missing `-lGL`).

~~If `rpi-update` has been used, the update pulls in an extra `egl.pc` that's not part of `libraspberrypi-dev` and this confuses the configuration script, which tries to use the BRCM EGL libs (thanks @psyke83). That's the reason for the `pkgconfig` override.~~ Not needed anymore since #2875.

Things needed:
~~ - decide how to tackle the lack of `video_shader` in the config files w.r.t. Configuration Editor.~~
 - do we need Wayland / Vulkan support on the PI ?


scriptmodules/helpers.sh
```diff
@@ -882,6 +882,20 @@ function setRetroArchCoreOption() {
...
+function setRetroArchConfigOption() {
+    local option="$1"
+    local value="$2"
+    iniConfig " = " "\"" "$configdir/all/retroarch.cfg"
+    iniGet "$option"
+    if [[ -z "$ini_value" ]]; then
+        iniSet "$option" "$value"
+    fi
+}
+
```
joolswills: Not convinced about adding this function - also as it only sets a value if it isn't already set, the description/name isn't sufficient.

joolswills:
>the joypad sorting is dropped upstream, so the patch included in RetroPie is not needed anymore. Remains to be seen if the new joypad order is compatible with the previous behavior.

It wasn't from my testing - actually it seemed wrong on the 1.7.6 when I did some testing the other day. I didn't get a chance to debug further.

joolswills:
I think that the behaviour with the video shaders needs to be addressed upstream first also (if possible as it sounds like it's not ideal). Do you have a reference to this change and any discussion?

cmitu:
> I think that the behaviour with the video shaders needs to be addressed upstream first also (if possible as it sounds like it's not ideal). Do you have a reference to this change and any discussion?

https://forums.libretro.com/t/cfg-file-parameter-video-shader-not-working/
https://forums.libretro.com/t/1-78-video-shader-option-not-working-anymore/
https://www.libretro.com/index.php/retroarch-1-7-8-released/ - section _Shader usability changes_.

scriptmodules/helpers.sh
cmitu: It's similar to the options settings function (`setRetroArchCoreOption`), since there were 4 similar options needed to be set, I thought of using a function. We can re-use the previous option though, not a problem.

joolswills: Its not used outside of this module, so can do a hidden function in the module for it if you need to de-duplicate.

cmitu: sure, will modify with a local routine.

cmitu: Should be fixed now, I added a local routine in the scriptmodule script.

joolswills:
I saw you put the video shader settings back in. Thanks. Let me know if you want any testing. 

Much appreciated.

cmitu:
I have to test it in conjuction with the RetroPie configuration editor, I've only tested briefly by running RetroArch from the command line.

cmitu:
OK, I've tested the re-introduction of the `video_shader` config option and it looks ok. 

Currently, the logic for loading an 'auto-shader' in RetroArch is to try to find - and load - :

* (a game specific preset shader) from `$CONF/shaders/preset/$CORE_NAME/$BASENAME.$EXT`
* (a content dir preset shader) from `$CONF/shaders/preset/$CORE_NAME/$CONTENT_DIR.$EXT`
* (a core preset shader) from `$CONF/shaders/preset/$CORE_NAME.$EXT`
* (a global preset shader) from `$CONF/shaders/preset/global.$EXT`

where `$EXT` is one of the supported shader preset extensions (`.glslp`, `.slangp`, maybe `.hlslp` ?).

The patch adds the option to load - after trying every other option from the list above - the shader preset specified in the `video_shader` parameter, so it shouldn't interfere with any shader preset saved from the RetroArch (per-game/core/content-dir/global).

The optioned offered in RetroArch with the removal of this parameter from the config file is to use a command line parameter, `--set-shader <preset_path>`. From my tests this acts as a hard override to any 'auto-shader' present (per-game/core/content-dir/global), so it doesn't have the same behavior as the patch added here.

cmitu:
There is a also a corner-case bug with preset saving, reported upstream (https://github.com/libretro/RetroArch/issues/9724).

scriptmodules/emulators/retroarch.sh
```diff
@@ -56,16 +55,16 @@ function build_retroarch() {
...
+    isPlatform "dispmanx" && ! isPlatform "kms" && params+=(--enable-dispmanx --disable-opengl1)
     isPlatform "mali" && params+=(--enable-mali_fbdev)
-    isPlatform "kms" && params+=(--enable-kms)
+    isPlatform "kms" && params+=(--enable-kms --enable-egl)
     isPlatform "arm" && params+=(--enable-floathard)
     isPlatform "neon" && params+=(--enable-neon)
     isPlatform "x11" && params+=(--enable-vulkan)
     isPlatform "vero4k" && params+=(--enable-mali_fbdev --with-opengles_libs='-L/opt/vero3/lib')
-    ./configure --prefix="$md_inst" "${params[@]}"
+   ./configure --prefix="$md_inst" "${params[@]}"
```
joolswills: Indent looks wrong

joolswills:
Minor indent issue I noticed. Apart from that, I'm happy to merge. Thanks for the work on this.

scriptmodules/emulators/retroarch.sh
cmitu: indent fixed - should be fine now.

joolswills:
Thanks.

cmitu:
Now that 1.8.2 is out, I could re-test and re-base against it.

joolswills:
I just noticed it was a PR against the fkms branch. I switched it. You may want to check / switch other PRs (I'll go through them eventually :-) ).

joolswills:
Let's do that as a new PR. I've tested this already and works well. I may have missed something but as it's a core component I'd like to test 1.8.2 and I'm keen to get this in.

joolswills:
Sorry. I have to revert for now. Fails to build on stretch - struct member issue in sdl2_common.c. probably a quick fix but I don't have any time right now.

cmitu:
@joolswills I'll take a look. Maybe this one - https://github.com/libretro/RetroArch/issues/9583 ?

joolswills:
Yep. That's the one.

cmitu:
I'll sort it out and then re-submit the PR.

joolswills:
Thanks.

PR #2854, opened 2019-09-22
cgenius.sh: Update to allow joystick game saving

JimHume:
As per cgenius recommendation at https://github.com/gerstrong/Commander-Genius/issues/309#issuecomment-358072833

Update the cgenius `cmake` command to include `-DNOTYPESAVE=on` by default.


joolswills:
Thanks.

PR #2855, opened 2019-09-22
N64: GLideN64: highres mode needs to be paired with a nativeResFactor=2 to work

dankcushions:
you can pass nativeResFactor=2 via a command line argument, so I have added that (and the ability to pass additional parameters to mupen64plus).

~~I think there's an argument that we shouldn't be using windowed mode with gliden64, as it demands you specify a resolution, but n64 resolutions aren't always factors of 320x240, so it's possible to get unwanted scaling/scaling artifacts. however, the alternative is to run fullscreen (letting gliden64 upscale from the true resolution) which removes the performance gain we get from our SDL upscaler. might be something to consider with the pi4...~~
ignore - gliden64 overrides the resolution when using `nativeResFactor`>0, so it doesn't matter what we send.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -179,8 +179,14 @@ function configure_mupen64plus() {
         isPlatform "videocore" && addEmulator 1 "${md_id}-auto" "n64" "$md_inst/bin/mupen64plus.sh AUTO %ROM%"
         for res in "${resolutions[@]}"; do
             local name=""
-            [[ "$res" == "640x480" ]] && name="-highres"
-            addEmulator 1 "${md_id}-GLideN64$name" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM% $res"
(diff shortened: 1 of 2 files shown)


dankcushions:
I can't edit my post for some reason, but:

> I think there's an argument that we shouldn't be using windowed mode with gliden64, as it demands you specify a resolution, but n64 resolutions aren't always factors of 320x240, so it's possible to get unwanted scaling/scaling artifacts. however, the alternative is to run fullscreen (letting gliden64 upscale from the true resolution) which removes the performance gain we get from our SDL upscaler. might be something to consider with the pi4...

i'm not sure this is true. i think resolution is ignored when you use nativeResFactor > 0, and i also don't think it's mandatory anyway (but am unable to test right now), so probably ignore that...

joolswills:
Is this still relevant? If so, please rebase and I'll check. Thanks.

dankcushions:
rebased and I will re-test tonight. I believe this is still needed for highres mode to work properly

joolswills:
Thanks.

PR #2857, opened 2019-09-27
reicast: update for rpi4 & generic kms

psyke83:
reicast: update for rpi4 & generic kms

* Add support for RPI4 and generic KMS targets
* Allow configuration of rpi3 platform in favour of rpi2 platform
* Move build configuration into separate function to avoid duplication between
  build and install functions.
* avoid duplication of addEmulator definitions due to audio backend options
* Set alsa as default audio backend for mesa (omx doesn't work, and oss backend via
  aoss causes emulator stuttering).

Patch details (may be removed after upstream is fixed):
* Add SDL/gles2 sub-platform for rpi4 to allow kmsdrm compatibility
* Ensure vsync is enforced for SDL, otherwise KMSDRM will run at unlocked framerate.
* Fix signal handler conflict by disabling SDL2's exit signal handler, and ensuring that
  reicast gracefully cleans up by stopping emulation and invoking SDL_Quit() during exit.

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -19,6 +19,7 @@ rp_module_flags="!armv6 !mali"
 function depends_reicast() {
     local depends=(libsdl2-dev python-dev python-pip alsa-oss python-setuptools libevdev-dev libasound2-dev libudev-dev)
     isPlatform "vero4k" && depends+=(vero3-userland-dev-osmc)
+    isPlatform "mesa" && depends+=(libgles2-mesa-dev)
     getDepends "${depends[@]}"
     isPlatform "vero4k" && pip install wheel
     pip install evdev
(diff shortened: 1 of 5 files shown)


steeeb:
Does adding < /dev/null when called via Runcommand fix it?

Had the same issue on Tinkerboard with is KMS.

psyke83:
@steeeb 

That didn't work, but did help me to figure out a workaround - using the SDL_INPUT_LINUX_KEEP_KBD environment variable. This has the disadvantage of causing keyboard inputs to leak into the background terminal, so it's not ideal, but I've updated the PR with the workaround until I can investigate a proper fix.

psyke83:
OK, I've resolved the problem (see `0003-fix-sdl2-sighandler-conflict.patch`). I've tested on KMS with RPI3, RPI4 and i965, but didn't try it against the legacy driver, but I'm fairly sure that it will also work OK (making the `CON:` runcommand workaround obsolete).

Please review and let me know if there are any problems. I'm not sure whether to pass along any of the patches upstream yet, since the master branch seems dead, and the newer branches don't yet have working SDL2.

joolswills:
Thanks. Confirmed working.

PR #2859, opened 2019-10-01
N64: Used vcgencmd to use nearest neighbour scaling for SDL upscales.

dankcushions:
I never liked the blurry scaling that we get as a side-effect of our SDL upscale, so I found a way to leverage vcgencmd to apply a better looking (IMO) nearest neighbour scaler. this is the same way that Retraorch does it with the dispmanx video driver. see https://github.com/libretro/RetroArch/issues/2641

It's also (probably unmeasurably) faster :)

Screenshots! From my phone as I couldn't get the internal screenshot button to work. Ignore the debug info in the corner :)

Before:
bilinear

After:
nearest

opinions may differ as to what looks better, but IMO the look the emulator should be controlled by the emulator config (in this case the various n64 video plugins do support AA, etc), and our upscaling should not apply additional effects.

### DIFF
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -12,7 +12,7 @@
 AUDIO_PLUGIN="mupen64plus-audio-sdl"
 VIDEO_PLUGIN="$1"
 ROM="$2"
-[[ "$3" -ne 0 ]] && RES="$3"
+[[ "$3" != 0 ]] && RES="$3"
 [[ "$4" -ne 0 ]] && RSP_PLUGIN="$4"
 PARAMS="${@:5}"
 [[ -n "$RES" ]] && RES="--resolution $RES"
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -176,20 +176,27 @@ function configure_mupen64plus() {
     isPlatform "kms" && resolutions=("%XRES%x%YRES%")
 
     if isPlatform "rpi"; then
-        isPlatform "videocore" && addEmulator 1 "${md_id}-auto" "n64" "$md_inst/bin/mupen64plus.sh AUTO %ROM%"
-        for res in "${resolutions[@]}"; do
-            local name=""
-            local nativeResFactor=""
-            if [[ "$res" == "640x480" ]]; then
-                name="-highres"
-                nativeResFactor=2
-            else
-                nativeResFactor=1
-            fi
-            addEmulator 1 "${md_id}-GLideN64$name" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM% $res 0 --set Video-GLideN64[UseNativeResolutionFactor]\=$nativeResFactor"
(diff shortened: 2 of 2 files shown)


dankcushions:
Hold fire on this for now. I want to review the possible scaling modes (info here: https://forum.kodi.tv/showthread.php?tid=224884) - it might be there's a nicer middle-ground between nearest-neighbour (which will have visible scaling artefacts on some low resolutions), and non-visible blurring at high resolutions. If I find a better one I should also suggest that for the retroarch dispmanx driver.

dankcushions:
i tried a few alternative modes via that link and it seems the 'sharpest' non-nearest-neighbour scaling mode is still blurry AF. i think we just stick with nearest neighbour (which is what retroarch does by default anyway).

dankcushions:
how about this: https://github.com/RetroPie/RetroPie-Setup/pull/2859/commits/44072c2d43c8a7ba7cf18f382cb4d05baeb8e6bc

i added a new entry to the mupen64plus.cfg that we can use to control the smoothing. at least now users can control it independently

joolswills:
I'd like to revisit this after we have merged the fkms branch back of that's ok.

dankcushions:
I've rebased this with current master, tested it on rpi3, and fixed a wee bug :) works for me! look good? if so i will squash, etc.

dankcushions:
Scrap that, just noticed the .cfg option isn't working. One moment...

hhromic:
Does this distinguish properly between RPI<4 and RPI4 ?
The script seems to check if the processor is BCM which I'm not sure is enough?
Not sure if `vcgencmd` works the same way with RPI4 and the (F)KMS drivers.
Just asking to confirm you covered that case :)

dankcushions:
@hhromic yeah the previous logic didn't make that distinction, presumably because the SDL hint just does nothing if code not there to retrieve it. i've put some safety code around it the new stuff :) not got my rpi4 hooked up to test, mind

scriptmodules/emulators/mupen64plus/mupen64plus.sh
```diff
@@ -436,12 +436,20 @@ getAutoConf mupen64plus_texture_packs && useTexturePacks
 
 if [[ "$(sed -n '/^Hardware/s/^.*: \(.*\)/\1/p' < /proc/cpuinfo)" == BCM* ]]; then
     WINDOW_MODE="--windowed $RES"
-    SDL_VIDEO_RPI_SCALE_MODE=1
-    # If a raspberry pi is used lower resolution to 320x240 and enable SDL dispmanx scaling mode 1
+    if isPlatform "videocore"; then
```
hhromic, line 439: I don't think you can use `isPlatform` here because this script does not run under RetroPie's scripting environment.
You probably will need to copy the detection logic from the `system.sh` library to re-use here.

https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/system.sh#L291-L296

I know is not very elegant but `system.sh` is not meant to be imported by 3rd-party scripts afaik.

dankcushions: yup, i forgot. i put that stuff behind the iniGet instead. the ini option isn't set unless it's videocore

PR #2865, opened 2019-10-09
add .hdm for x68000

nakatamaho:
title says all. i tested and worked with some stock .hdm files.


nakatamaho:
thank you very much!

PR #2866, opened 2019-10-10
pegasus-fe: Raspberry Pi 4 support

mmatyas:
Pegasus is now functional on the Pi 4, and the build should work there too. There's also a second commit that adds SDL2 as a dependency, which is used for gamepad handling on Pi 0-3 (not on Pi 4 yet, but will be there too eventually).

### DIFF
--- a/scriptmodules/supplementary/pegasus-fe.sh
+++ b/scriptmodules/supplementary/pegasus-fe.sh
@@ -22,6 +22,7 @@ function depends_pegasus-fe() {
         gstreamer1.0-libav
         gstreamer1.0-plugins-good
         jq
+        libsdl2-dev
         policykit-1
     )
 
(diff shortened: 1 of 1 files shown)


mmatyas:
PS: The pull request is against `RetroPie:master`, would you prefer the fkms branch?

hhromic:
I assume in current RetroPie it will not build anymore without the new SDL2 dependency? If so, then I would suggest to merge it in `master` for the moment indeed.
Btw, thanks for this front-end, looks amazing!

mmatyas:
The Pi releases are now built with SDL2, yes. However, there is a configuration option to disable it, you can manually create a custom build if that turns out to be a problem.

HerbFargus:
Ran some quick tests  on my pi 4 and it seems to work well.

joolswills:
Thanks

PR #2867, opened 2019-10-11
Update dolphin depends

raelgc:
Since [May, 2018](https://dolphin-emu.org/blog/2018/05/02/legend-dolphin-lens-between-worlds/#epilogue-tldr) Dolphin switched to a QT UI only.

So, I changed the `depends` to include GTK dependency only if the build is using the old 5 branch.

I tested this on Ubuntu 18.04 with GCC7 and it worked properly.

### DIFF
--- a/scriptmodules/emulators/dolphin.sh
+++ b/scriptmodules/emulators/dolphin.sh
@@ -17,9 +17,9 @@ rp_module_section="exp"
 rp_module_flags="!arm"
 
 function depends_dolphin() {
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Will test and merge.

joolswills:
Will maybe remove older branch support later. Thanks.

raelgc:
@joolswills Sorry for my late answer on this: my only concern about remove the old GCC support is with Ubuntu 16.04. It has only more 1 year and few months of support, so if you agree, I'll create the PR to remove it when it's no more supported.

joolswills:
No worries. I'll remove before then. We are not obligated to support old LTS versions as long as Ubuntu.

PR #2868, opened 2019-10-12
Fix pm2 installation in virtualgampead.sh

diego09310:
I had problems installing virtual gamepad because pm2 installation was throwing errors (Method Not Allowed). Installing it using the official installation script instead of npm worked for me

### DIFF
--- a/scriptmodules/supplementary/virtualgamepad.sh
+++ b/scriptmodules/supplementary/virtualgamepad.sh
@@ -50,7 +50,9 @@ function sources_virtualgamepad() {
 }
 
 function install_virtualgamepad() {
-    npm install pm2 -g --unsafe-perm
+    if ! [[ -x "$(command -v pm2)" ]]; then
+        wget -qO- https://getpm2.com/install.sh | bash
+    fi
     cd "$md_inst"
     sudo -u $user npm install
     sudo -u $user npm install ref


joolswills:
Please can you post a full log? Was this on Stretch? I'll test it, but I think I'd like to use npm still if there's a workaround.

diego09310:
Yes, npm is probably better, but wasn't able to make it work (not really used to npm on the raspberry, though).

Full log:
[virtualgamepad.log](https://github.com/RetroPie/RetroPie-Setup/files/3721253/virtualgamepad.log)

cmitu:
I think the issue is that the NPM package repository doesn't accept requests from node v4 clients.
A simple fix is to install the next LTS node version. Replacing
```
https://deb.nodesource.com/setup_4.x
```
with 
```
https://deb.nodesource.com/setup_8.x
```
will not produce any errors and the installation works.  
Tested on Raspbian Stretch (RetroPie 4.5.1), where `nodejs` v8 is available, but doesn't provide `npm`, so the scriptmodule will try and install the LTS version from deb.nodesource.com.

diego09310:
I think I tried with `https://deb.nodesource.com/setup_12.x` but didn't work, maybe because node 12 is not available for raspbian or because npm goes separately, I'll try to test this during the weekend if I have time

cmitu:
One other thing that could be improved is removing the `systemd` service when removing this package. Otherwise the unit keeps failing on boot multiple times. This can be done by simply adding a
```
pm2 unstartup
```
in the `remove_virtualgamepad` function, before removing the APT NodeSource repository.

user747:
also had success using https://deb.nodesource.com/setup_8.x 
also changing the default listening port from 80 to something else got the page loading for me, and also running the pm2 systemd service as root solved a blinking player 1-2 controller issue. 

works pretty well overall

joolswills:
I don't think this is an issue anymore. Latest repo npm seems to work (see e8251fdf)

PR #2871, opened 2019-10-13
BufferSwapMode=2 (On buffer update) is most performant.

dankcushions:
Small performance boost:

mario64 title screen (with --nospeedlimit)

**bufferswapmode=0**
78-98 VI/sec (mainly around low 80s) 

**bufferswapmode=1**
85-97 (mainly around low-high 80s)

**bufferswapmode=2**
86-96 (mainly around mid-high 80s) 

This is the value that mupen64plus-fz (android port of mupen64plus that targets similar hardware) uses: https://github.com/fzurita/mupen64plus-ae/blob/master/app/src/main/java/paulscode/android/mupen64plusae/persistent/GLideN64Prefs.java#L194


joolswills:
Thanks.

PR #2873, opened 2019-10-14
scummvm - bump version to latest stable (2.1)

hhromic:
User in the forum tested new version with different games and all seems to be working fine.
Ref: https://retropie.org.uk/forum/topic/23778/scummvm-2-1-0

@joolswills once this is merged I would suggest to rebuild the binary with the updated version as well, considering how long it takes to build from source. Thanks!


joolswills:
Thanks.

PR #2877, opened 2019-10-23
N64: GLideN64: FPS Counter hack no-longer works, and redundant since zelda* is in the GLideN64 blacklist

dankcushions:
See https://github.com/gonetz/GLideN64/issues/1688#issuecomment-545045823

If this eventually gets fixed (I doubt there is much appetite to fix GLES2/videocore bugs, though), the FPS hack may become relevant again, but there are apparently even more issues with Zelda*, hence it is in the GLideN64 blacklist.

I suggest we retire this FPS Counter hack (which affects all games ran in GLideN64) for now, as it does nothing.


joolswills:
thanks. yeah - followed the discussion. Makes sense to remove.

PR #2878, opened 2019-10-23
N64: GLideN64: set Texture Cache to 128MB for Pi 2+, 64MB for Pi 1/0 (RPI)

dankcushions:
The recommended value is 3/4 VRAM, so let's calculate it 👍 I don't see any performance change, but it was fun to do anyway!

Also done a little cleanup with the FBEmulation setting. I think unless we're using that setting in a blacklist, there's no need to force it to enabled at launch. Users should have the ability to play with these settings after install if they so wish.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -231,7 +231,13 @@ function configure_mupen64plus() {
         # Bilinear filtering mode (0=N64 3point, 1=standard)
         iniSet "bilinearMode" "1"
         # Size of texture cache in megabytes. Good value is VRAM*3/4
-        iniSet "CacheSize" "50"
+        local gpu_mem
+        if isPlatform "armv6"; then
+            gpu_mem=64
+        else
+            gpu_mem=128
+        fi
+        iniSet "CacheSize" "$gpu_mem"
         iniSet "EnableFBEmulation" "True"
         # Use native res
         iniSet "UseNativeResolutionFactor" "1"
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -232,7 +232,6 @@ function configure_mupen64plus() {
         iniSet "bilinearMode" "1"
         # Size of texture cache in megabytes. Good value is VRAM*3/4
         iniSet "CacheSize" "50"
-        # Disable FB emulation until visual issues are sorted out
         iniSet "EnableFBEmulation" "True"
         # Use native res
         iniSet "UseNativeResolutionFactor" "1"
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -253,8 +253,6 @@ function testCompatibility() {
                 config_version=$(<"$configdir/n64/GLideN64_config_version.ini")
             fi
             iniSet "configVersion" "$config_version"
-            # Size of texture cache in megabytes. Good value is VRAM*3/4
-            iniSet "CacheSize" "50"
             # Set native resolution factor of 1
             iniSet "UseNativeResolutionFactor" "1"
             for game in "${GLideN64NativeResolution_blacklist[@]}"; do
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -255,8 +255,6 @@ function testCompatibility() {
             iniSet "configVersion" "$config_version"
             # Size of texture cache in megabytes. Good value is VRAM*3/4
             iniSet "CacheSize" "50"
-            # Enable FBEmulation if necessary
-            iniSet "EnableFBEmulation" "True"
             # Set native resolution factor of 1
             iniSet "UseNativeResolutionFactor" "1"
             for game in "${GLideN64NativeResolution_blacklist[@]}"; do


joolswills:
will need rebasing after last commit i guess ? Would like to get some comment on @gizmo98 on this as he's more familiar than me with this code, but makes sense.

dankcushions:
I messed up the rebase somehow... bear with me

dankcushions:
Rebased properly now!

psyke83:
This won't be a good idea once the RPI4 branch is merged into master.

When booted via the FKMS overlay, the `gpu_mem` setting is no longer relevant, as the VC4/V3D drivers use the CMA heap to allocate memory instead.

To illustrate, 3B FKMS with `gpu_mem` set:
```
pi@rpi3:~ $ cat /boot/config.txt | grep gpu_mem
gpu_mem_256=128
gpu_mem_512=256
gpu_mem_1024=256
pi@rpi3:~ $ vcgencmd get_mem gpu
gpu=256M
pi@rpi3:~ $ dmesg | grep Reserved
[    0.000000] cma: Reserved 256 MiB at 0x1ec00000
```
vs 4B with no `gpu_mem` configured:
```
pi@rpi4:~ $ cat /boot/config.txt | grep gpu_mem
pi@rpi4:~ $ vcgencmd get_mem gpu
gpu=76M
pi@rpi4:~ $ dmesg | grep Reserved
[    0.000000] cma: Reserved 256 MiB at 0x000000001ec00000
```
Although `vcgencmd` reports 76MB for the 4B, this value is bogus for the VC4/V3D drivers which actually use the CMA heap for GPU memory allocation (which is 256MB by default)

From `/boot/overlays/README`:
```
Name:   vc4-fkms-v3d
Info:   Enable Eric Anholt's DRM VC4 V3D driver on top of the dispmanx
        display stack.
Load:   dtoverlay=vc4-fkms-v3d,<param>
Params: cma-256                 CMA is 256MB (needs 1GB)
        cma-192                 CMA is 192MB (needs 1GB)
        cma-128                 CMA is 128MB
        cma-96                  CMA is 96MB
        cma-64                  CMA is 64MB
```

In a nutshell, `vcgencmd get_mem gpu` will report bogus values for FKMS sessions, and thus the calculation will potentially be invalid; in the case of the 4B, it would erroneously calculate a 56MB buffer instead of 192MB.

dankcushions:
it [seems ](https://www.raspberrypi.org/forums/viewtopic.php?p=1491512&sid=d487552384542da0e53c42adc32f3cdf#p1491512)the pi4 (VC6) uses CMA and "IOMMU" to dynamically allocate VRAM: 
...however, the CMA part isn't for 3d operations, but only for framebuffer allocations, so i assume it's not going to be available for GL operations (?).

as for the pi3 + mesa/fkms, i just fired up and updated my vanilla raspbian lite buster image, selected mesa driver, and i get similar results to your pi4:
````
pi@raspberrypi:~ $ cat /boot/config.txt | grep gpu_mem
pi@raspberrypi:~ $ vcgencmd get_mem gpu
gpu=76M
pi@raspberrypi:~ $ dmesg | grep Reserved
[    0.000000] cma: Reserved 256 MiB at 0x1ec00000
````
so, by default the pi3 + fkms presumably uses CMA to allocate its 3d VRAM.

i looked into CMA, but i don't understand [this thread](https://www.raspberrypi.org/forums/viewtopic.php?t=234348) - i thought CMA dynamically allocated, so what does 'Reserved 256 MiB' mean - it will never go above that? never go below? i don't know if the hwm/lwm configs still work: https://elinux.org/RPiconfig#CMA_-_dynamic_memory_split - maybe i need to ask in the RPI forums for some clarification.

as for pi4s - my ultimate intention is to create an entirely new set of mupen64plus defaults for the pi4 (and it probably doesn't need the plugin switching/blacklist stuff and can just use GLideN64 for all). pi3 + msea may also be able to support some more aggressive defaults. so i think the merge of fkms branch with this will necessitate some platform specific IF blocks in these scripts. i hope this will make more sense when i start working on the fkms branch.

however, what we use for the `CacheSize` for the pi4 is a bit semantic i guess. 1/2 * total system ram?

**Comment by psyke83** [comment]
> it [seems ](https://www.raspberrypi.org/forums/viewtopic.php?p=1491512&sid=d487552384542da0e53c42adc32f3cdf#p1491512)the pi4 (VC6) uses CMA and "IOMMU" to dynamically allocate VRAM:
> ...however, the CMA part isn't for 3d operations, but only for framebuffer allocations, so i assume it's not going to be available for GL operations (?).

I'm fairly certain that the CMA allocation is used by the GL driver directly.

> 
> as for the pi3 + mesa/fkms, i just fired up and updated my vanilla raspbian lite buster image, selected mesa driver, and i get similar results to your pi4:
> 
> ```
> pi@raspberrypi:~ $ cat /boot/config.txt | grep gpu_mem
> pi@raspberrypi:~ $ vcgencmd get_mem gpu
> gpu=76M
> pi@raspberrypi:~ $ dmesg | grep Reserved
> [    0.000000] cma: Reserved 256 MiB at 0x1ec00000
> ```
> 
> so, by default the pi3 + fkms presumably uses CMA to allocate its 3d VRAM.

Yes. The `gpu_mem` being reported as 256MB was due to it having RetroPie's `gpu_mem` [configuration](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/admin/image.sh#L105-L107).

> 
> i looked into CMA, but i don't understand [this thread](https://www.raspberrypi.org/forums/viewtopic.php?t=234348) - i thought CMA dynamically allocated, so what does 'Reserved 256 MiB' mean - it will never go above that? never go below? i don't know if the hwm/lwm configs still work: https://elinux.org/RPiconfig#CMA_-_dynamic_memory_split - maybe i need to ask in the RPI forums for some clarification.

That link is quite outdated, but the CMA heap does appear to be dynamically allocated. The `gpu_mem` allocation appears to be static, which results in a permanent memory reduction:

With `gpu_mem_1024=256`:
```
[    0.000000] Memory: 501940K/786432K available (8192K kernel code, 629K rwdata, 2176K rodata, 1024K init, 821K bss, 22348K reserved, 262144K cma-reserved)
```

With no `gpu_mem` allocation:
```
[    0.000000] Memory: 684640K/970752K available (8192K kernel code, 629K rwdata, 2176K rodata, 1024K init, 821K bss, 23968K reserved, 262144K cma-reserved)
```
 
> as for pi4s - my ultimate intention is to create an entirely new set of mupen64plus defaults for the pi4 (and it probably doesn't need the plugin switching/blacklist stuff and can just use GLideN64 for all). pi3 + msea may also be able to support some more aggressive defaults. so i think the merge of fkms branch with this will necessitate some platform specific IF blocks in these scripts. i hope this will make more sense when i start working on the fkms branch.

Yes, I've already disabled the blacklist/whitelist stuff for fkms builds, which applies to 3B and 4B alike when firmware kms is active.

> however, what we use for the `CacheSize` for the pi4 is a bit semantic i guess. 1/2 * total system ram?

Don't forget that 4B comes in 2GB and 4GB varieties - so that would likely be overkill.

**Comment by dankcushions** [comment]
would something like this be acceptable (within the fkms branch)?

```
        # Size of texture cache in megabytes. Good value is VRAM*3/4
        local gpu_mem
        if isPlatform "videocore"; then
            gpu_mem="$(vcgencmd get_mem gpu | grep -o '[0-9]*')"
        elif isPlatform "fkms"; then
            gpu_mem="$(dmesg | grep 'cma: Reserved' | grep -o '[0-9]*' | sed -n 3p)"
        else
            gpu_mum=128
        fi
        iniSet "CacheSize" "$(( gpu_mem * 3/4 ))"
```

**Comment by joolswills** [comment]
Unless there is some measured improvement I would probably not bother. Using dmesg like that is going to be unreliable.

**Comment by psyke83** [comment]
I agree that this is probably not worth the effort. You can't rely on dmesg due to the possibility of the log being flushed at the line you're looking for. There is also `/proc/meminfo`, but the calculation is a bit messier as you'd need to use awk and also rely on a bash builtin to convert the output to megabytes:

```gpu_mem=$(($(cat /proc/meminfo | grep CmaTotal | awk '{ print $2 }')/1024))```

Something to keep in mind as well: the `kms` platform tag is not suitable here. As far as I know, determining the GPU memory via the CMA reservation applies only to the RPI boards using the Mesa driver. For that reason you should have used the `mesa` tag so as not to break all other targets that might run in a kms context (or at least isolate to `rpi` first).

**Comment by dankcushions** [comment]
fair. cachesize will definitely be a problem if users select higher resolutions (eg, PC users, or pi4/odroid users). in the gliden64 issue tracker they typically talk about it being 500MB or more.

i could change it to 128MB which is what mupen64plus-fz uses (targets similar android hardware), but i think eventually it will need to become a calculation of some sort, once we start dealing with hardware that can do 2x+ resolutions.

**Comment by psyke83** [comment]
I'm ignorant as to whether the `CacheSize` would make such a difference, but I suppose if you wanted a simpler change, you could just do:

```
        local gpu_mem
        if isPlatform "armv6"; then
            gpu_mem=64
        else
            gpu_mem=128
        fi
        iniSet "CacheSize" "$gpu_mem"
```

Maybe Jools could chime in on whether we have any other low-memory targets that don't run on armv6 and thus would be affected by this?

joolswills:
Just pi 0/1 I can think of.

dankcushions:
done!

joolswills:
Thanks.

PR #2879, opened 2019-10-24
Fix colour sticking at white if you're logged in as root

TauPan:
I often log in as root to my retropie box (via ssh with a pubkey). Since the account doesn't have colours in its PS1, the last one used stays on the prompt, which is white and not very readable on my light terminal background. This is a quick fix for that problem.


hhromic:
Regardless of being for root or not, I think what's done in this PR improves correctness indeed.

joolswills:
I wasn't that keen on the commit msg but change is fine. merged

Issue #2749, opened 2019-06-29
Raspberry Pi 4 support


joolswills:
We are working on RPI4 and Raspbian Buster support. There's no ETA but RetroPie 4.5 will be released first as a final Stretch based image (although we may support it for longer than Raspbian do - we are not sure what their future support for Stretch is).

Please comment here or on the forum before submitting RPI4 PRs as you may be repeating work others have already done.

Thanks!

burt111:
Will specifying platform rpi3 and installing from be broke on raspbian buster or does it specifically not work on the new differences seeing the possibility of modifying the setup script to see if it will install when I get my pi4 if the platform option doesn’t work

joolswills:
I don't want this thread to become a support thread. Please use the forum for that. It wont work out of the box on buster yet.

psyke83:
I notice that emulationstation via fkms sometimes crashes during idle (possibly due to controller idle disconnect). I vaguely recall this occurring on stretch when you upgraded to 2.0.9, so it may be the same issue cropping up again.

I notice that the buster raspbian source (2.0.9+dfsg1-1+rpt1) includes several patches that aren't present via upstream:
```
debian/patches/0001-ARM-Create-configure-option-enable-arm-simd-to-gover.patch
debian/patches/0002-ARM-SIMD-assembly-optimization-for-function-BlitRGBt.patch
debian/patches/0003-ARM-SIMD-assembly-optimization-for-function-BlitARGB.patch
debian/patches/0004-SDL_blit-use-a-named-enum-for-required-hardware-feat.patch
debian/patches/0005-ARM-SIMD-assembly-optimization-for-BGR-to-RGB-32bpp-.patch
debian/patches/0006-ARM-assembly-optimization-for-SDL_FillRect.patch
debian/patches/0007-ARM-SIMD-optimization-for-4-4-4-4-to-8-8-8-8-normal-.patch
debian/patches/0008-ARM-Create-configure-option-enable-arm-neon-to-gover.patch
debian/patches/0009-ARM-NEON-assembly-optimization-for-function-BlitRGBt.patch
debian/patches/0010-ARM-NEON-assembly-optimization-for-function-BlitARGB.patch
debian/patches/0011-ARM-NEON-assembly-optimization-for-SDL_FillRect.patch
debian/patches/fix-cross-building-907711.patch
debian/patches/no-libdir.patch
```
[...]

PR #2882, opened 2019-10-28
#2749 fix reicast and RPI4 can do with -j4

rafal-prasal:
reicast for RPI4 needs to be alpha in order to work properly, at least at this time

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -25,12 +25,28 @@ function depends_reicast() {
 }
 
 function sources_reicast() {
-    gitPullOrClone "$md_build" https://github.com/reicast/reicast-emulator.git master
+
+    if isPlatform "rpi4"; then
+        gitPullOrClone "$md_build" https://github.com/reicast/reicast-emulator.git alpha
+    else
+        gitPullOrClone "$md_build" https://github.com/reicast/reicast-emulator.git master
+    fi
 }
 
 function build_reicast() {
-    cd shell/linux
-    if isPlatform "rpi"; then
+
+    if isPlatform "rpi4"; then
+        cd reicast/linux
+        md_ret_require="$md_build/reicast/linux/reicast.elf"
+    else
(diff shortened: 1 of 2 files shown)


rafal-prasal:
Basically makes reicast available for RPI4

https://github.com/RetroPie/RetroPie-Setup/issues/2749

joolswills:
https://github.com/RetroPie/RetroPie-Setup/pull/2857

I'd prefer to wait for this. But thanks. We won't switch to -j4 globally for rpi4.

rafal-prasal:
it is your choice of course, 

in the meantime I would merge that and then replace when #2857 becomes available.

joolswills:
That would just make extra work. Sorry but I'd prefer to use the other PR. I'm in no rush.

psyke83:
Just to clarify some issues re: reicast and RPI4.

If this PR worked successfully for you, it's possible that you might be running an improperly configured RetroPie installation that runs in an Xorg context, which is not going to be the supported configuration for the upcoming release.

The current reicast `alpha` branch (as well as `beta` and `stable`) are not capable of building with SDL2 support for any targets, so they won't be able to run on RPI4 unless you prefix the emulator command with `xinit` (which you didn't do in your PR) or if you already run emulationstation within an Xorg session (which as I said, is not how the upcoming release will work).

Although my PR is working, I'm also not happy with the state of my PR due to the fact that the master branch seems to run very slowly, and keyboard inputs are leaking into the background terminal. We could switch to newer branches and run in xinit, but that will break the`+Start reicast.sh` launcher (as runcommand doesn't know to pass the xinit supplementary arguments to these shell scripts).

It seems that upstream is in the process of refactoring the code related to SDL2 (and have also broken RPI3 and earlier builds). It may be a good idea to hold off on merging anything until those branches have settled and SDL2 support works again (which means that we can run without Xorg). Fortunately, it seems that `lr-flycast` is in a pretty good state and runs faster than the standalone `master` branch.

PR #2883, opened 2019-10-28
image: don't set custom gpu_mem allocation for mesa targets

psyke83:
The RPI VC4/VC6 drivers use the kernel CMA heap allocation for GPU memory
(which defaults to 256MB on 1GB+ targets) instead of the gpu_mem allocation.

Although the gpu_mem allocation is still used, it's needed only for camera and
hardware decoders, but the default allocation (~64-76MB) is sufficient.

Without this change, we are losing ~192MB of user-addressible kernel memory
for no reason.

### DIFF
--- a/scriptmodules/admin/image.sh
+++ b/scriptmodules/admin/image.sh
@@ -100,11 +100,13 @@ function install_rp_image() {
         sed -i "s/quiet/quiet loglevel=3 consoleblank=0 plymouth.enable=0 quiet/" "$chroot/boot/cmdline.txt"
     fi
 
-    # set default GPU mem, and overscan_scale so ES scales to overscan settings.
+    # set default GPU mem (videocore only) and overscan_scale so ES scales to overscan settings.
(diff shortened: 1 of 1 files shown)


psyke83:
Some discussion may be needed to ensure this change is OK:

1. Can we agree on a standard for selecting a `kms` build? The regex match here would match `rpi3-fkms`, for example. I understand that we're not planning to switch to RPI3 to the mesa driver, but if we do in future, it would be good to sort this out.
2. The `config.txt` seems to use different sections to isolate Pi 4B (`[pi4]`) from others (`[all]`). I haven't tested image building to see how this change impacts us.

joolswills:
Why regexp over the isPlatform flags?

joolswills:
Ignore comment above. I forget my own code.

joolswills:
im going to merge this as is for now - I may tweak some of the image creation code later. thanks.

PR #2887, opened 2019-11-01
Lr-parallel-64 - fix building on rpi4/kms

joolswills:
* remove a typedef conflict - may need to upstream a proper fix for this
 * change logic from "tinker" to "kms" for make parameters to make it more general
 * rollback code for "neon" rather than "arm" platforms

I've not tested this at all apart from building on rpi2 and rpi4. Need to check the typedef change on tinkerboard. Probably should be upstreamed but upstream master is broken on arm which also needs fixing.

### DIFF
--- a/scriptmodules/libretrocores/lr-parallel-n64.sh
+++ b/scriptmodules/libretrocores/lr-parallel-n64.sh
@@ -19,6 +19,7 @@ function depends_lr-parallel-n64() {
     local depends=()
     isPlatform "x11" && depends+=(libgl1-mesa-dev)
     isPlatform "rpi" && depends+=(libraspberrypi-dev)
+    isPlatform "kms" && isPlatform "gles" && depends+=(libgles2-mesa-dev)
     getDepends "${depends[@]}"
 }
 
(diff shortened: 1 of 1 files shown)


joolswills:
To clarify. HEAD doesn't build on arm due to some makefile changes or so. Fails at linking stage.

https://github.com/libretro/parallel-n64/issues/544

Probably a simple fix.

scriptmodules/libretrocores/lr-parallel-n64.sh
```diff
@@ -28,18 +28,20 @@ function sources_lr-parallel-n64() {
...
 }
 
 function build_lr-parallel-n64() {
     rpSwap on 1000
     make clean
     local params=()
-    if isPlatform "rpi" || isPlatform "odroid-c1"; then
-        params+=(platform="$__platform")
-    elif isPlatform "tinker"; then
+    if isPlatform "kms"; then
```
psyke83: Can you change this to `mesa`, please?

IMHO, usage of the `kms` platform flag should be architecture-agnostic when possible due to the fact that any modern system is potentially capable of running in KMS mode. Your PR runs fine on x86 kms mode with the flag changed to `mesa` (whereas the PR breaks the x86 kms build as-is). Thanks.

joolswills: Thanks. Will do.

joolswills:
Merging but this may need further platform logic tweaks.

PR #2890, opened 2019-11-05
FPS hack is no longer needed

dankcushions:
As per https://github.com/RetroPie/RetroPie-Setup/commit/96e1d73217678cc0f38bae57c312ce9396d444e8#diff-fe35047c4f73377c42592c2271157752

Small miss from the rebase. You might want to do it as a rebase rather than as a new commit, just highlighting it here :)


joolswills:
Not sure what you mean by doing it as a rebase. We can't rebase fkms_rpi4 if that's what you mean - people are using it.

PR #2891, opened 2019-11-07
Retrieve Atari 800 colour palettes

Neo-Stone:
Retrieves missing Atari 800 / XL / XE colour palettes so they are usable in the emulator.

### DIFF
--- a/scriptmodules/libretrocores/lr-atari800.sh
+++ b/scriptmodules/libretrocores/lr-atari800.sh
@@ -29,6 +29,7 @@ function install_lr-atari800() {
     md_ret_files=(
         'atari800_libretro.so'
         'atari800/COPYING'
+        'atari800/act'
     )
 }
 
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Need to make sure there will be no overlap with upcoming atari800 standalone. Will merge later. There are already issues due to colliding paths for the config so I may look into changing config locations.

Neo-Stone:
Thanks for your comment. Colour palettes for both emulators (lr-atari800 and atari800 standalone) are exactly the same. So retrieving them is only needed be be done once. There should be only one directory to store them. Using them is done by path and file selection inside the appropriate emulator (by pressing F1, selecting the colour palette, and saving the atari config file). This process is exactly the same for standalone as well as libretro emulator. Hope this helps for your testing.

joolswills:
Sorry for the late review but revisiting this, I'm unsure about the reason to copy to the user configuration area - as it will overwrite any user changes on a future update. Thanks.

PR #2892, opened 2019-11-07
VSync patch now merged upstream. Removing.

dankcushions:
see https://github.com/mupen64plus/mupen64plus-core/pull/670


joolswills:
Thanks.

PR #2872, opened 2019-10-14
Generate pcsx2 from source

raelgc:
Actually the Debian/Ubuntu PCSX2 packages are really outdated. I'm changing the script to build the emulator using the fresh code from github.

I'm using the same configuration folders used to build the Ubuntu package.

If the user never installed PCSX2, the first run will display the configuration dialog. If it's installed using the system package, it'll run using existing configurations.

PR #2894, opened 2019-11-10
pcsx2: use upstream PPA for the latest version

cmitu:
After #2872, installing `pcsx2` on 64 bit is no longer possible on Ubuntu, due to `libsdl2-dev:i386` installation conflicts (see https://bugs.launchpad.net/ubuntu/+source/mir/+bug/1808060). 

So will use the PPA maintained by the upstream team (from https://launchpad.net/~pcsx2-team/+archive/ubuntu/pcsx2-daily) and add back the SDL2 downgrade option.

Tested on 18.04, should work on newer versions since the PPA builds up-to-date packages for 19.04 and 19.10 also. Should fix https://retropie.org.uk/forum/topic/23992.

### DIFF
--- a/scriptmodules/emulators/pcsx2.sh
+++ b/scriptmodules/emulators/pcsx2.sh
@@ -17,41 +17,39 @@ rp_module_section="exp"
 rp_module_flags="!arm"
 
 function depends_pcsx2() {
-    # Build dependencies (from the Debian/Ubuntu package: https://github.com/PCSX2/pcsx2/blob/master/debian-packager/control)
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks - looks good. I think the repository should be added in the depends section though.

joolswills:
One thing to note also, but perhaps not a major issue is this may have problems on a Debian system - if Debian doesn't have the packaging issue maybe we can keep the source install for it. Although it may have had problems before too.

cmitu:
@joolswills Yes, I though of that too.  Debian might also have this issue, but at some point they bumped into the same issue, they added a patch, then reverted it.

```


libsdl2 (2.0.8+dfsg1-3.1) unstable; urgency=medium

  * Non-maintainer upload.
  * d/tests/build: Add autopkgtest to check that dynamic linking with
    either pkg-config or sdl2-config works correctly
  * Revert "Make libsdl2-dev coinstallable again"
    (Closes: #909778) (reopens: #909740)

 -- Simon McVittie <smcv@debian.org>  Sun, 30 Sep 2018 16:13:38 +0100

libsdl2 (2.0.8+dfsg1-3) unstable; urgency=medium

  [ Hugh McMaster <hugh.mcmaster@outlook.com> ]
  * Make libsdl2 coinstallable again (Closes: #909740).
```

joolswills:
Thanks. Looks good.

PR #2897, opened 2019-11-10
Fix for getting OS version on Armbian

PRIHLOP:
Without this fix on Armbian OS ```lsb_release -sidrc``` returns an empty value.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -67,7 +67,7 @@ function get_os_version() {
 
     # get os distributor id, description, release number and codename
     local os
-    mapfile -t os < <(lsb_release -sidrc)
+    mapfile -t os < <(lsb_release -s -i -d -r -c)
     __os_id="${os[0]}"
     __os_desc="${os[1]}"
     __os_release="${os[2]}"


joolswills:
Thanks. Seems fine, but will check why their version differs (custom tool mimicking behaviour or different version with different parameter parsing). I need to make sure this doesn't affect other platforms too. Maybe you have some more info?

joolswills:
Also as we don't test on this Debian based os (we probably should support it as it's popular), we can't of course guarantee things will work.

PRIHLOP:
I tested it on Ubuntu 18.04 amd64, it is works fine. All utilities `lsb_release` support this parameters divided by `-`, but may be some distros do not support joint parameters.

PRIHLOP:
> Thanks. Seems fine, but will check why their version differs (custom tool mimicking behaviour or different version with different parameter parsing). I need to make sure this doesn't affect other platforms too. Maybe you have some more info?

Can you test it on your device or environment?
I founded that in Ubuntu amd64 utilite `/usr/bin/lsb_release` writed on Python(and it can parse parameters), but on Armbian used shell script based on this https://gist.github.com/skx/1f9e5240856c5976193a (it can't parse parameters, only chek if it exists).
Way with separateв parameters is more universal, and I think it will work on any platforms.

joolswills:
What a hack. They should just use the correct tool. I will implement this PR manually I think with some different comments in the code. Thanks for the info.

We don't officially support this distro either so I may also add some warning.

joolswills:
Tbh I also think armbian should use a script that parses command line args correctly. This could fail for other software too.

PR #2898, opened 2019-11-11
sdl2: add dependencies to debian packages

psyke83:
This should ensure that an installation of SDL2 from binary on a fresh
installation should not fail due to missing dependencies not being
installed by the debian packages themselves.

Certain board-specific packages that may cause distribution dependency
resolution issues are deliberately omitted, but they should already be
installed (if not, we can do so during script first-time run or basic
install).

Some dependencies have also been reordered to be more consistent with our
board flags; x11 depends are still a mess, but reordering them may
break vero4k.

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -31,18 +31,29 @@ function get_arch_sdl2() {
     echo "$(dpkg --print-architecture)"
 }
 
-function depends_sdl2() {
+function _list_depends_sdl2() {
     # Dependencies from the debian package control + additional dependencies for the pi (some are excluded like dpkg-dev as they are
     # already covered by the build-essential package retropie relies on.
-    local depends=(devscripts debhelper dh-autoreconf libasound2-dev libudev-dev libibus-1.0-dev libdbus-1-dev fcitx-libs-dev)
+    local depends=(libasound2-dev libudev-dev libibus-1.0-dev libdbus-1-dev fcitx-libs-dev libsndio-dev)
     # these were removed by a PR for vero4k support (cannot test). Needed though at least for for RPI and X11
-    ! isPlatform "vero4k" && depends+=(libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxrandr-dev libxss-dev libxt-dev libxxf86vm-dev libgl1-mesa-dev)
-    isPlatform "rpi" && depends+=(libegl1-mesa-dev libgles2-mesa-dev libglu1-mesa-dev libraspberrypi-dev)
-    isPlatform "mali" && depends+=(mali-fbdev)
+    ! isPlatform "vero4k" && depends+=(libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxkbcommon-dev libxrandr-dev libxss-dev libxt-dev libxv-dev libxxf86vm-dev libgl1-mesa-dev)
+    isPlatform "gles" || isPlatform "gl" && depends+=(libegl1-mesa-dev libgles2-mesa-dev)
+    isPlatform "gl" || isPlatform "rpi" && depends+=(libgl1-mesa-dev libglu1-mesa-dev)
     isPlatform "kms" || isPlatform "rpi" && depends+=(libdrm-dev libgbm-dev)
-    isPlatform "x11" && depends+=(libpulse-dev libegl1-mesa-dev libgles2-mesa-dev libglu1-mesa-dev)
+    isPlatform "x11" && depends+=(libpulse-dev libwayland-dev)
+
+    echo "${depends[@]}"
+}
+
+function depends_sdl2() {
+    # install additional packages that are needed, but may be unsuitable as debian package dependencies due to distribution oddities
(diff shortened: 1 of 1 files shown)


psyke83:
Two notes:
* It's possible that $(list_sdl2_depends) may produce duplicate entries, but duplicate packages are automatically filtered out of the resulting package, so it's not a problem.
* You may still want to invoke `sdl2 get_depends` in our script for a binary install to catch the packages that I deliberately omitted (to avoid distro issues/let our script filtering do its magic), but, at least in the case of RPI, it `libraspberrypi-dev` is already installed on an untouched buster lite image IIRC.

scriptmodules/supplementary/sdl2.sh
```diff
@@ -31,18 +31,29 @@ function get_arch_sdl2() {
...
+function depends_sdl2() {
+    # install additional packages that are needed, but may be unsuitable as debian package dependencies due to distribution oddities
+    local depends=(devscripts debhelper dh-autoreconf)
+
+    isPlatform "mali" && depends+=(mali-fbdev)
+    isPlatform "rpi" && depends+=(libraspberrypi-dev)
     isPlatform "vero4k" && depends+=(vero3-userland-dev-osmc)
-    getDepends "${depends[@]}"
+
+    getDepends "$(_list_depends_sdl2) ${depends[@]}"
```
hhromic: Each parameter expansion should be quoted independently, like this:
```suggestion
    getDepends "$(_list_depends_sdl2)" "${depends[@]}"
```

hhromic: Actually, `$(_list_depends_sdl2)` returns a space-separated list, right?
It should not be quoted then.

psyke83: Yes (to the latter). I've updated the PR to ensure that `$(_list_depends_sdl2)` is not quoted, thanks.

scriptmodules/supplementary/sdl2.sh
```diff
@@ -31,18 +31,29 @@ function get_arch_sdl2() {
...
-    local depends=(devscripts debhelper dh-autoreconf libasound2-dev libudev-dev libibus-1.0-dev libdbus-1-dev fcitx-libs-dev)
+    local depends=(libasound2-dev libudev-dev libibus-1.0-dev libdbus-1-dev fcitx-libs-dev libsndio-dev)
     # these were removed by a PR for vero4k support (cannot test). Needed though at least for for RPI and X11
-    ! isPlatform "vero4k" && depends+=(libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxrandr-dev libxss-dev libxt-dev libxxf86vm-dev libgl1-mesa-dev)
-    isPlatform "rpi" && depends+=(libegl1-mesa-dev libgles2-mesa-dev libglu1-mesa-dev libraspberrypi-dev)
-    isPlatform "mali" && depends+=(mali-fbdev)
+    ! isPlatform "vero4k" && depends+=(libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxkbcommon-dev libxrandr-dev libxss-dev libxt-dev libxv-dev libxxf86vm-dev libgl1-mesa-dev)
+    isPlatform "gles" || isPlatform "gl" && depends+=(libegl1-mesa-dev libgles2-mesa-dev)
+    isPlatform "gl" || isPlatform "rpi" && depends+=(libgl1-mesa-dev libglu1-mesa-dev)
     isPlatform "kms" || isPlatform "rpi" && depends+=(libdrm-dev libgbm-dev)
```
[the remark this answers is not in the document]
psyke83: Yes, but the line above should be ok, no? Here's the resulting dependencies on an x86 kms build:
```
Depends: libsdl2-2.0-0 (= 2.0.10+1), libc6-dev, libgl1-mesa-dev, libasound2-dev, libudev-dev, libibus-1.0-dev, libdbus-1-dev, fcitx-libs-dev, libsndio-dev, libx11-dev, libxcursor-dev, libxext-dev, libxi-dev, libxinerama-dev, libxkbcommon-dev, libxrandr-dev, libxss-dev, libxt-dev, libxv-dev, libxxf86vm-dev, libegl1-mesa-dev, libgles2-mesa-dev, libglu1-mesa-dev, libdrm-dev, libgbm-dev
```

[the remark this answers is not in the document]
cmitu: I see. The previous dep are ok.

joolswills:
Thanks. Will test and merge later.

joolswills:
Thanks.

WhitecastlePT:
Don´t know if this is still an issue or not but i tried to install retropie in raspian ( raspberrypi 5.10.63-v7 ) RPI 3B+ and still doesn´t work the basic install.
image

image

cmitu:
@WhitecastlePT Please use the forum for support questions. 
You're trying to install on Raspberry Pi OS Bullseye, which is not supported by RetroPie yet.

WhitecastlePT:
> @WhitecastlePT Please use the forum for support questions. You're trying to install on Raspberry Pi OS Bullseye, which is not supported by RetroPie yet.

Ups my bad then ;) thx

PR #2900, opened 2019-11-13
KMS & X11 modesetting support

psyke83:
This is a possible implementation of modesetting for KMS and X11 targets. Tested as working on `rpi3` (fkms), `rpi4` and `x86` (though for the latter, my laptop's i965 driver shows only the native panel resolution in KMS mode, so it's not very useful in my particular case).

The SDL2 patch can be considered temporary, as @cmitu is working on a more fleshed-out version. I do think that it may be a good idea to provide a means to select by the mode index rather than (or in addition to) SDL2 internally matching a string like `XRESxYRES@REFRESH`, because the Pi has multiple modes with identical main timings, but different subtimings.

KMS modesetting currently works only with retroarch and SDL2 applications (but that accounts for virtually everything that runs in KMS mode). I don't think it's suitable for merging, but I'd appreciate any thoughts on whether this solution (or something like it) is worth implementing.

### DIFF
--- a/scriptmodules/emulators/ags.sh
+++ b/scriptmodules/emulators/ags.sh
@@ -34,10 +34,9 @@ function install_ags() {
 }
 
 function configure_ags() {
-    local binary="$md_inst/bin/ags"
+    local binary="XINIT:$md_inst/bin/ags"
     local params=("--fullscreen %ROM%")
     if ! isPlatform "x11"; then
-        binary="xinit $binary"
         params+=("--gfxdriver software")
     fi
 
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -170,7 +170,7 @@ function configure_retroarch() {
     isPlatform "mesa" && iniSet "video_fullscreen" "true"
 
     # set default render resolution to 640x480 for rpi1
-    if isPlatform "rpi1"; then
+    if isPlatform "videocore" && isPlatform "rpi1"; then
         iniSet "video_fullscreen_x" "640"
         iniSet "video_fullscreen_y" "480"
     fi
--- a/scriptmodules/ports/lincity-ng.sh
+++ b/scriptmodules/ports/lincity-ng.sh
@@ -33,11 +33,10 @@ function remove_lincity-ng() {
 }
 
 function configure_lincity-ng() {
-    if isPlatform "x11"; then
-        addPort "$md_id" "lincity-ng" "LinCity-NG" "/usr/games/lincity-ng"
-    else
-        addPort "$md_id" "lincity-ng" "LinCity-NG" "xinit /usr/games/lincity-ng"
-    fi
+    local binary="XINIT:/usr/games/lincity-ng"
+
+    addPort "$md_id" "lincity-ng" "LinCity-NG" "$binary"
+
     moveConfigDir "$home/.lincity-ng" "$md_conf_root/lincity-ng"
     # fix for wrong config location
     if [[ -d "/lincity-ng" ]]; then
--- a/scriptmodules/ports/micropolis.sh
+++ b/scriptmodules/ports/micropolis.sh
@@ -16,7 +16,7 @@ rp_module_section="opt"
 rp_module_flags="!mali !kms"
 
 function depends_micropolis() {
-    ! isPlatform "x11" getDepends xorg matchbox
+    ! isPlatform "x11" getDepends xorg matchbox-window-manager
 }
 
 function install_bin_micropolis() {
@@ -28,11 +28,10 @@ function remove_micropolis() {
 }
 
 function configure_micropolis() {
-    if isPlatform "x11"; then
-        addPort "$md_id" "micropolis" "Micropolis" "/usr/games/micropolis"
-    else
-        addPort "$md_id" "micropolis" "Micropolis" "xinit $md_inst/micropolis.sh"
-    fi
+    local binary="/usr/games/micropolis"
+    ! isPlatform "x11" && binary="XINIT:$md_inst/micropolis.sh"
+
+    addPort "$md_id" "micropolis" "Micropolis" "$binary"
 
     mkdir -p "$md_inst"
     cat >"$md_inst/micropolis.sh" << _EOF_
--- a/scriptmodules/ports/minecraft.sh
+++ b/scriptmodules/ports/minecraft.sh
@@ -13,10 +13,10 @@ rp_module_id="minecraft"
 rp_module_desc="Minecraft - Pi Edition"
 rp_module_licence="PROP"
 rp_module_section="exp"
-rp_module_flags="!mali !x86 !kms"
+rp_module_flags="!mali !x86"
 
 function depends_minecraft() {
-    getDepends xorg matchbox
+    getDepends xorg matchbox-window-manager
 }
 
 function install_bin_minecraft() {
@@ -29,7 +29,7 @@ function remove_minecraft() {
 }
 
 function configure_minecraft() {
-    addPort "$md_id" "minecraft" "Minecraft" "xinit $md_inst/Minecraft.sh"
+    addPort "$md_id" "minecraft" "Minecraft" "XINIT:$md_inst/Minecraft.sh"
 
     cat >"$md_inst/Minecraft.sh" << _EOF_
 #!/bin/bash
--- a/scriptmodules/ports/steamlink.sh
+++ b/scriptmodules/ports/steamlink.sh
@@ -53,5 +53,5 @@ _EOF_
         chmod +x "$sl_script"
     fi
 
-    addPort "$md_id" "steamlink" "Steam Link" "xinit $sl_script"
+    addPort "$md_id" "steamlink" "Steam Link" "XINIT:$sl_script"
 }
--- a/scriptmodules/ports/uqm.sh
+++ b/scriptmodules/ports/uqm.sh
@@ -81,7 +81,7 @@ function configure_uqm() {
     local params=("-f")
     if isPlatform "kms"; then
         # SDL1 needs xinit, and does not have /usr/games in $PATH
-        binary="xinit /usr/games/$binary"
+        binary="XINIT:/usr/games/$binary"
         # OpenGL mode must be also be enabled for high resolution support
         params+=("-o" "-r %XRES%x%YRES%")
     fi
--- a/scriptmodules/supplementary/mesa-drm.sh
+++ b/scriptmodules/supplementary/mesa-drm.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="mesa-drm"
+rp_module_desc="libdrm - userspace library for drm"
+rp_module_licence="MIT https://www.mesa3d.org/license.html"
+rp_module_section=""
+rp_module_flags=""
+
+function depends_mesa-drm() {
+    local depends=(meson ninja-build libgbm-dev libdrm-dev libpciaccess-dev)
+
+    getDepends "${depends[@]}"
+}
+
+function sources_mesa-drm() {
+    gitPullOrClone "$md_build" https://github.com/psyke83/mesa-drm "runcommand_debug"
+}
+
+function build_mesa-drm() {
+    local params=()
+
+    # for RPI, disable all but VC4 driver to minimize startup delay
+    isPlatform "rpi" && params+=(-Dintel=false -Dradeon=false \
+                           -Damdgpu=false -Dexynos=false \
+                           -Dnouveau=false -Dvmwgfx=false \
+                           -Domap=false -Dfreedreno=false \
+                           -Dtegra=false -Detnaviv=false -Dvc4=true)
+    meson builddir --prefix="$md_inst" "${params[@]}"
+    ninja -C builddir
+
+    md_ret_require="$md_build/builddir/tests/modetest/modetest"
+}
+
+function install_mesa-drm() {
+    md_ret_files=(
+        builddir/libkms
+        builddir/tests/modetest
+    )
+}
--- a/scriptmodules/supplementary/raspbiantools.sh
+++ b/scriptmodules/supplementary/raspbiantools.sh
@@ -27,7 +27,7 @@ function lxde_raspbiantools() {
     aptInstall raspberrypi-ui-mods rpi-chromium-mods gvfs
 
     setConfigRoot "ports"
-    addPort "lxde" "lxde" "Desktop" "startx"
+    addPort "lxde" "lxde" "Desktop" "XINIT:startx"
     enable_autostart
 }
 
--- a/scriptmodules/supplementary/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand.sh
@@ -46,6 +46,12 @@ function install_bin_runcommand() {
         dialog --create-rc "$configdir/all/runcommand-launch-dialog.cfg"
         chown $user:$user "$configdir/all/runcommand-launch-dialog.cfg"
     fi
+
+    # needed for KMS modesetting (debian buster or later only)
+    if compareVersions "$__os_debian_ver" ge 10; then
+        rp_callModule mesa-drm
+    fi
+
     md_ret_require="$md_inst/runcommand.sh"
 }
 
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -614,10 +614,10 @@ function main_menu() {
         fi
 
         if [[ "$EMULATOR" == lr-* ]]; then
-            options+=(
-                8 "Select RetroArch render res for $EMULATOR ($RENDER_RES)"
-                9 "Edit custom RetroArch config for this ROM"
-            )
+            if [[ "$HAS_MODESET" == "tvs" ]]; then
+                options+=(8 "Select RetroArch render res for $EMULATOR ($RENDER_RES)")
+            fi
+            options+=(9 "Edit custom RetroArch config for this ROM")
         elif [[ "$HAS_MODESET" == "tvs" ]]; then
             local fb_emu="$(default_mode get fb_emu)"
             local fb_rom="$(default_mode get fb_rom)"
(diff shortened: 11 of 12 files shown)


cmitu:
IMHI it's worth implementing.

I've noticed the sub-timings differences after you mentioned them on a connected TV, but only as reported by `tvservice` - I'll take a look how they're reported by `modetest`.

It can probably be extended in the future with a `HAS_MODESET=x11` and resort to `xrandr` for resolution switching.

One thing that I think should be added is to also set the RA's resolution (which can do mode-setting on its own) when mode-switching is chosen and  `render_resolution=config` or `render_resolution=output`.

psyke83:
I've updated the PR with lots of fixes as well as X11 support. Seems to be working OK on my systems, but needs more thorough checking.

@cmitu: thanks for the information. Is the `render_resolution` configuration unique to the latest RetroArch? I don't see it in any of my existing configuration files, but I haven't yet tried your retroarch PR.

cmitu:
> @cmitu: thanks for the information. Is the `render_resolution` configuration unique to the latest RetroArch? I don't see it in any of my existing configuration files, but I haven't yet tried your retroarch PR.

No, that's a `runcommand` thing - you can choose in the menu the RA's render resolution (see https://github.com/RetroPie/RetroPie-Setup/blob/6aa73f5791c7f58aaadf78641c1134e740693a80/scriptmodules/supplementary/runcommand/runcommand.sh#L451). When chosen, it's pushed to the `/dev/shm/retroarch.cfg` via `retroarch_append_config` (before launch [here](https://github.com/RetroPie/RetroPie-Setup/blob/6aa73f5791c7f58aaadf78641c1134e740693a80/scriptmodules/supplementary/runcommand/runcommand.sh#L1059)).

When doing KMS mode switching, then we can pass the new W/H to the generated `retroarch.cfg` so RA does the mode switching itself. Something like this (in `retroarch_append_config`):
```
    local dim
    # if our render resolution is "config", then we don't set anything (use the value in the retroarch.cfg)
    if [[ "$RENDER_RES" != "config" ]]; then
        if [[ "$RENDER_RES" == "output" ]]; then
            dim=(0 0)
        else
            dim=(${RENDER_RES/x/ })
        fi
    fi

    # when KMS is enabled and the resolution changed, see if we need to set the RetroArch resolution
    # this will force RetroArch to do the mode-setting itself
    if [[ -n "$FB_NEW" && is_kms_enabled ]]; then
        if [[ "$RENDER_RES" == "config" || "$RENDER_RES" == "output" ]]; then
            dim=("${MODE_CUR[2]}" "${MODE_CUR[3]}")
        fi
    fi
```

psyke83:
@cmitu: the latest commit should hopefully improve KMS support for retroarch. Pinging @hhromic, as he also discussed this issue in the group chat.

psyke83:
Updated the PR again to improve `xinit` support for modesetting (and also to make such emulators work if launched from an Xorg context).

cmitu:
@psyke83 I added my video mode selection hint in https://github.com/psyke83/SDL-mirror/pull/1, just in case we want the KSMDRM overrides part of the SDL2 scriptmodule. Otherwise, it's just an extention to the  `0001-sdl-modesetting.patch` included in this PR.

psyke83:
@cmitu: Thanks. I've modified the PR to temporarily build SDL2 against your branch instead of merging immediately, so as not to potentially cause any issues.

I noticed a discrepancy with `modetest` - it prints extra modes that are inaccessible to SDL2's `kmsdrm` driver, which causes the mode selection not to work properly due to the mode index being different than what SDL2 expects. I've switched over to `modeprint` (also with a patch to make parsing out the require data simpler) as it prints the same list of modes as SDL2, so all modes should now work correctly.

I'm wondering about the retroarch "render res" option. Should we just remove this option entirely, or is there still a valid use case to have a render resolution that's different to the mode set via xrandr/tvservice, etc?

cmitu:
Looking at the origin of the _render resolution_ for RetroArch (https://github.com/RetroPie/RetroPie-Setup/issues/531, by @joolswills) it might be specific to the legacy VC/GLES drivers (https://github.com/libretro/RetroArch/commit/9f40b65fcaa2994b4fbb344fbbe2fdc27596647f). From what I read in the PR, it scales the rendering resolution to the fb resolution (which might have been changed via `tvservice`), similar to the scaling `dispmanx` SDL1 patches.

dankcushions:
one usecase here: https://github.com/RetroPie/RetroPie-Setup/blob/27918be14dada7413a7d2ce0624836b4857b503a/scriptmodules/emulators/retroarch.sh#L169 - pi0/1 has bad performance when rendering RA at >640x480

IMO i think we will end up with a similar condition for pi4 - forcing 1080p so it doesn’t attempt 4k (or you could make it use a 1080p display mode). i guess you also need to account for when the display device is less than 1080p (no point rendering higher res than you can output)?

psyke83:
@cmitu @dankcushions Thanks for the information. I've changed the logic so that it displays/sets the retroarch render resolution only for the `tvs` modesetting case, as this appears to be the only  scenario in which setting a non-native resolution will work as expected.

I'm fairly happy with the state of the PR now, so I've removed the do not merge prefix. 

Pending changes needed before merge:
* I've forked the `mesa-drm` repo to avoid in-script patching (and potentially seeing patch conflicts when upstream's master changes). It might be a good idea to re-fork this into RetroPie as well: https://github.com/psyke83/mesa-drm
* @cmitu's SDL2 patch needs merging (before or after my branch is imported into RetroPie's repo for the `fkms_rpi4` merge).

Thoughts on improving the PR:
* is there a better place to trigger installation of `mesa-drm`? The current behaviour is to re-install the module each time runcommand is updated. This might be a bit overkill.

Of course, let me know if you find any issues. I've been testing on legacy, rpi3/4 kms, x86 kms and x11, and everything seems to be working as expected.

cmitu:
I've only been testing on a PI4 (fkms) so far, and it looks fine. SDL2 mode-switching works for SDL2 based emulators and the Libretro cores launch to the desired resolution. 

Regarding improvements of the PR
* I'm curious also about the dependency triggering. I think that if we package the scriptmodule, the installation would download the binary archive and unpack-it if we use `rp_callModule mesa-drm install` 
The installation for `mesa-drm` will not work on Debian Strech, it packages a `meson` version that's too old.    

Since `modeprint` only needs 2 headers and 1 additional file (`kms.c`), what do you think about including the necessary source files in the scriptmodule and don't install an additional repo ?

* Mode listing shows too much information. I think some of the parameters are redundant - i.e. are dependent on the resolution and can be omitted. Looking at Weston's DRM backend - it only prints the resolution, vrefresh, aspect_ratio and the clock (see [here](https://github.com/wayland-project/weston/blob/e404b72fd52b2dddf5aa4f2d4bfb6ebb79b4b132/libweston/backend-drm/modes.c#L397)).
`modeprint`, `modetest` don't print the aspect ratio, however this information is present in the `flags` member of the `drmModeModeInfo` structure for CEA modes - see [here](https://gitlab.freedesktop.org/wayland/weston/commit/f82ff35e8e67de1d11f0ef2538864881b37993fe) and [how is this implemented](https://github.com/wayland-project/weston/blob/e404b72fd52b2dddf5aa4f2d4bfb6ebb79b4b132/libweston/backend-drm/modes.c#L38-L67) in Weston.

psyke83:
@cmitu,

I'm definitely in favour of making the mode output less cluttered, but regarding the aspect ratio:

Here's a quick patch to check the decoded aspect via both tools:
```
diff --git a/tests/modeprint/modeprint.c b/tests/modeprint/modeprint.c
index fbfda36..c65ab1a 100644
--- a/tests/modeprint/modeprint.c
+++ b/tests/modeprint/modeprint.c
@@ -59,9 +59,9 @@ int fbs;
 static int printMode(struct drm_mode_modeinfo *mode, bool is_crtc, int mode_id, int encoder_id)
 {
        if (debug_modes) {
-               printf("Mode: %s @ %i Hz (%i %i %i %i %i %i %i %i %i %i %i %i %i) %s %i %i\n", mode->name, mode->vrefresh, mode->clock,
+               printf("Mode: %s @ %i Hz (%i %i %i %i %i %i %i %i %i %i %i %i %i -> %i) %s %i %i\n", mode->name, mode->vrefresh, mode->clock,
                   mode->hdisplay, mode->hsync_start, mode->hsync_end, mode->htotal, mode->hskew, mode->vdisplay,
-                  mode->vsync_start, mode->vsync_end, mode->vtotal, mode->vscan, mode->vrefresh, mode->flags,
+                  mode->vsync_start, mode->vsync_end, mode->vtotal, mode->vscan, mode->vrefresh, mode->flags, (mode->flags & DRM_MODE_FLAG_PIC_AR_MASK) >> 19,
                   is_crtc ? "crtc" : "connector", mode_id, encoder_id);
        } else if (full_modes) {
                printf("Mode: %s\n", mode->name);
diff --git a/tests/modetest/modetest.c b/tests/modetest/modetest.c
index e66be66..f5b3c94 100644
--- a/tests/modetest/modetest.c
+++ b/tests/modetest/modetest.c
@@ -224,7 +224,7 @@ static void dump_mode(drmModeModeInfo *mode)
 
        printf(" flags: ");
        mode_flag_str(mode->flags);
-       printf("; type: ");
+       printf(" (%d -> %d); type: ", mode->flags, (mode->flags & DRM_MODE_FLAG_PIC_AR_MASK) >> 19);
        mode_type_str(mode->type);
        printf("\n");
 }
```
The result is that `modetest` seems capable of determining the aspect (at least for some modes):
```
60x768 60 1360 1432 1568 1776 768 771 776 798 84750 flags: nhsync, pvsync (6 -> 0); type: preferred, driver
  1920x1080 60 1920 2008 2052 2200 1080 1084 1089 1125 148500 flags: phsync, pvsync (5 -> 0); type: driver
  1920x1080 60 1920 2008 2052 2200 1080 1084 1089 1125 148500 flags: phsync, pvsync (1048581 -> 2); type: driver
  1920x1080 60 1920 2008 2052 2200 1080 1084 1089 1125 148352 flags: phsync, pvsync (1048581 -> 2); type: driver
  1920x1080 50 1920 2448 2492 2640 1080 1084 1089 1125 148500 flags: phsync, pvsync (1048581 -> 2); type: driver
  1920x1080 30 1920 2008 2052 2200 1080 1084 1089 1125 74250 flags: phsync, pvsync (1048581 -> 2); type: driver
  1920x1080 30 1920 2008 2052 2200 1080 1084 1089 1125 74176 flags: phsync, pvsync (1048581 -> 2); type: driver
  1920x1080 25 1920 2448 2492 2640 1080 1084 1089 1125 74250 flags: phsync, pvsync (1048581 -> 2); type: driver
  1920x1080 24 1920 2558 2602 2750 1080 1084 1089 1125 74250 flags: phsync, pvsync (1048581 -> 2); type: driver
  1920x1080 24 1920 2558 2602 2750 1080 1084 1089 1125 74176 flags: phsync, pvsync (1048581 -> 2); type: driver
  1280x1024 60 1280 1328 1440 1688 1024 1025 1028 1066 108000 flags: phsync, pvsync (5 -> 0); type: driver
  1280x768 60 1280 1344 1472 1664 768 771 778 798 80140 flags: nhsync, nvsync (10 -> 0); type: driver
  1280x720 60 1280 1344 1472 1664 720 723 728 748 74250 flags: phsync, pvsync (5 -> 0); type: driver
  1280x720 60 1280 1390 1430 1650 720 725 730 750 74250 flags: phsync, pvsync (1048581 -> 2); type: driver
  1280x720 60 1280 1390 1430 1650 720 725 730 750 74176 flags: phsync, pvsync (1048581 -> 2); type: driver
  1280x720 50 1280 1720 1760 1980 720 725 730 750 74250 flags: phsync, pvsync (1048581 -> 2); type: driver
  1024x768 75 1024 1040 1136 1312 768 769 772 800 78750 flags: phsync, pvsync (5 -> 0); type: driver
  1024x768 70 1024 1048 1184 1328 768 771 777 806 75000 flags: nhsync, nvsync (10 -> 0); type: driver
  1024x768 60 1024 1048 1184 1344 768 771 777 806 65000 flags: nhsync, nvsync (10 -> 0); type: driver
  832x624 75 832 864 928 1152 624 625 628 667 57284 flags: nhsync, nvsync (10 -> 0); type: driver
  800x600 75 800 816 896 1056 600 601 604 625 49500 flags: phsync, pvsync (5 -> 0); type: driver
  800x600 60 800 840 968 1056 600 601 605 628 40000 flags: phsync, pvsync (5 -> 0); type: driver
  720x576 50 720 732 796 864 576 581 586 625 27000 flags: nhsync, nvsync (10 -> 0); type: driver
  720x576 50 720 732 796 864 576 581 586 625 27000 flags: nhsync, nvsync (524298 -> 1); type: driver
  720x576 50 720 732 796 864 576 581 586 625 27000 flags: nhsync, nvsync (1048586 -> 2); type: driver
  720x480 60 720 736 798 858 480 489 495 525 27027 flags: nhsync, nvsync (524298 -> 1); type: driver
  720x480 60 720 736 798 858 480 489 495 525 27027 flags: nhsync, nvsync (1048586 -> 2); type: driver
  720x480 60 720 736 798 858 480 489 495 525 27000 flags: nhsync, nvsync (10 -> 0); type: driver
  720x480 60 720 736 798 858 480 489 495 525 27000 flags: nhsync, nvsync (524298 -> 1); type: driver
  720x480 60 720 736 798 858 480 489 495 525 27000 flags: nhsync, nvsync (1048586 -> 2); type: driver
  640x480 75 640 656 720 840 480 481 484 500 31500 flags: nhsync, nvsync (10 -> 0); type: driver
  640x480 60 640 656 752 800 480 490 492 525 25200 flags: nhsync, nvsync (524298 -> 1); type: driver
  640x480 60 640 656 752 800 480 490 492 525 25175 flags: nhsync, nvsync (10 -> 0); type: driver
  640x480 60 640 656 752 800 480 490 492 525 25175 flags: nhsync, nvsync (524298 -> 1); type: driver
  720x400 70 720 738 846 900 400 412 414 449 28320 flags: nhsync, pvsync (6 -> 0); type: driver
  1360x768 60 1360 1432 1568 1776 768 771 776 798 84750 flags: nhsync, pvsync (6 -> 0); type: preferred, driver
```
Unfortunately, `modeprint` doesn't work work:
```
Mode: 1360x768 @ 60 Hz (84750 1360 1432 1568 1776 0 768 771 776 798 0 60 6 -> 0) connector 50 0
Mode: 1920x1080 @ 60 Hz (148500 1920 2008 2052 2200 0 1080 1084 1089 1125 0 60 5 -> 0) connector 50 1
Mode: 1920x1080 @ 60 Hz (148352 1920 2008 2052 2200 0 1080 1084 1089 1125 0 60 5 -> 0) connector 50 2
Mode: 1920x1080 @ 50 Hz (148500 1920 2448 2492 2640 0 1080 1084 1089 1125 0 50 5 -> 0) connector 50 3
Mode: 1920x1080 @ 30 Hz (74250 1920 2008 2052 2200 0 1080 1084 1089 1125 0 30 5 -> 0) connector 50 4
Mode: 1920x1080 @ 30 Hz (74176 1920 2008 2052 2200 0 1080 1084 1089 1125 0 30 5 -> 0) connector 50 5
Mode: 1920x1080 @ 25 Hz (74250 1920 2448 2492 2640 0 1080 1084 1089 1125 0 25 5 -> 0) connector 50 6
Mode: 1920x1080 @ 24 Hz (74250 1920 2558 2602 2750 0 1080 1084 1089 1125 0 24 5 -> 0) connector 50 7
Mode: 1920x1080 @ 24 Hz (74176 1920 2558 2602 2750 0 1080 1084 1089 1125 0 24 5 -> 0) connector 50 8
Mode: 1280x1024 @ 60 Hz (108000 1280 1328 1440 1688 0 1024 1025 1028 1066 0 60 5 -> 0) connector 50 9
Mode: 1280x768 @ 60 Hz (80140 1280 1344 1472 1664 0 768 771 778 798 0 60 10 -> 0) connector 50 10
Mode: 1280x720 @ 60 Hz (74250 1280 1344 1472 1664 0 720 723 728 748 0 60 5 -> 0) connector 50 11
Mode: 1280x720 @ 60 Hz (74250 1280 1390 1430 1650 0 720 725 730 750 0 60 5 -> 0) connector 50 12
Mode: 1280x720 @ 60 Hz (74176 1280 1390 1430 1650 0 720 725 730 750 0 60 5 -> 0) connector 50 13
Mode: 1280x720 @ 50 Hz (74250 1280 1720 1760 1980 0 720 725 730 750 0 50 5 -> 0) connector 50 14
Mode: 1024x768 @ 75 Hz (78750 1024 1040 1136 1312 0 768 769 772 800 0 75 5 -> 0) connector 50 15
Mode: 1024x768 @ 70 Hz (75000 1024 1048 1184 1328 0 768 771 777 806 0 70 10 -> 0) connector 50 16
Mode: 1024x768 @ 60 Hz (65000 1024 1048 1184 1344 0 768 771 777 806 0 60 10 -> 0) connector 50 17
Mode: 832x624 @ 75 Hz (57284 832 864 928 1152 0 624 625 628 667 0 75 10 -> 0) connector 50 18
Mode: 800x600 @ 75 Hz (49500 800 816 896 1056 0 600 601 604 625 0 75 5 -> 0) connector 50 19
Mode: 800x600 @ 60 Hz (40000 800 840 968 1056 0 600 601 605 628 0 60 5 -> 0) connector 50 20
Mode: 720x576 @ 50 Hz (27000 720 732 796 864 0 576 581 586 625 0 50 10 -> 0) connector 50 21
Mode: 720x480 @ 60 Hz (27027 720 736 798 858 0 480 489 495 525 0 60 10 -> 0) connector 50 22
Mode: 720x480 @ 60 Hz (27000 720 736 798 858 0 480 489 495 525 0 60 10 -> 0) connector 50 23
Mode: 640x480 @ 75 Hz (31500 640 656 720 840 0 480 481 484 500 0 75 10 -> 0) connector 50 24
Mode: 640x480 @ 60 Hz (25200 640 656 752 800 0 480 490 492 525 0 60 10 -> 0) connector 50 25
Mode: 640x480 @ 60 Hz (25175 640 656 752 800 0 480 490 492 525 0 60 10 -> 0) connector 50 26
Mode: 720x400 @ 70 Hz (28320 720 738 846 900 0 400 412 414 449 0 70 6 -> 0) connector 50 27
Mode: 1360x768 @ 60 Hz (84750 1360 1432 1568 1776 0 768 771 776 798 0 60 6 -> 0) crtc 49 0
```
This seems likely related to the discrepancy between tools in which RPI3 reports 35 modes via `modetest`, but only 27 for `modeprint` (which correspond exactly to what SDL2's `kmsdrm` driver reports). A possible explanation for this is that only the latter two use `drmModeGetConnector` to query connectors. This explains why the mode ids correspond properly, but it seems that the function doesn't copy the same value for `flags` that includes the required aspect ratio encoding.

If you've any ideas on how to remedy this, let me know.

psyke83:
Never mind - I figured out a solution.

The aspect ratio in `modeprint` be decoded if you set `drmSetClientCap(fd, DRM_CLIENT_CAP_ASPECT_RATIO, 1);`. I'll update to clean up the modeline and display the aspect ratio string.

psyke83:
Update: enabling the `DRM_CLIENT_CAP_ASPECT_RATIO` causes `modeprint` to show 35 modes, which means that it will break the mode id coherence with SDL2's driver. It looks like we may need to disable aspect ratio support, or perhaps try patching SDL2 to enable the client capability and see if the extra modes work.

cmitu:
> The question is whether it's worthwhile to do this? Are the modes that are otherwise missing important?

I think it's ok to use `DRM_CLIENT_CAP_ASPECT_RATIO`. While not all modes are reported correctly or reported at all (on my 4k test, `720x576` is reported as `16:9`, though it's clearly 5:4; `tvservice` also incorrectly displays it as 4:3) , it's still more friendly than the while modeline.

I can add the SDL2 patch to my PR branch (https://github.com/psyke83/SDL-mirror/pull/1).

Now I'm thinking about if we can add some `p/i` info in the modeline based on `DRM_MODE_FLAG_INTERLACE` flag.

psyke83:
@cmitu 

I agree re: the aspect cap being set in SDL2. I've updated the patchset to fully support aspect ratio display. Regarding the SDL2 patches, I've switched the temporary commit to this branch that includes both of our patches: https://github.com/psyke83/SDL-mirror/tree/retropie-2.0.10_modeset

I've switched back to `modetest`, as it produces reliable output how that I figured out the reason for the mode id discrepancy, and it already has some functions to print the addition flags in human-readable format.

After looking at your 4K modeprint debug log, it highlighted a problem: the aspect ratio encoding can cause problems with detection of the default mode due to the fact that the crtc modeline doesn't encode any aspect ratio information, so the default mode might never match any of the available modes. To work around this, the aspect is now stripped from each line before being compared to the default crtc mode, which should fix default mode detection.

If you get a chance to test on the 4K TV again, try the latest version of PR again with your Pi booted in 4K mode, and make sure that the "Video Mode" on the first runcommand dialog displays the proper 4K mode when there is nothing selected for the per-rom and per-emulator video mode. If the default mode detection is still broken, the "Video Mode" would be blank, which would mean that content may not launch correctly (as the %XRES% etc. is not correct). Thanks.

cmitu:
@psyke83 I'll be able to test again by the end of the week.

psyke83:
@cmitu - I've added the dependency, and to address your earlier observation re: `meson`, limited installation to debian version 10 (buster) as the minimum requirement. For RPI, there's no good reason to be running in KMS mode on stretch considering that the firmware updates have stopped and RPI4 is not supported, and also shouldn't be an issue for x86 (Ubuntu 18.04 is based on buster). When the KMS tool is not found by runcommand, it silently falls back to tvservice, so everything should continue working as before for stretch/jessie users.

Still open to suggestions about better ways to trigger installation of `mesa-drm` so as to minimize unnecessary updating (even if it's going to be from binary).

Thanks again for the assistance. The PR should be working OK as-is, so I'll try not to make any further changes unless absolutely necessary.

cmitu:
@psyke83 I wasn't thinking that using KMS on Strech is desirable - besides Kernel/Firmware , the other important dependencies like Mesa/DRM are not be updated anymore. I think `modetest` doesn't even show the whole range of resolutions if the kernel version is < 4.19.
The only concern was the `runcommand` upgrading, since failing the `mesa-drm` scriptmodule installation might show errors in the logs and we might get error reports - thanks for adjusting that.

psyke83:
Updated PR to fix edge case of non-standard mode names (e.g. FIXED_MODE) in both KMS and X11. Instead of assuming the mode name contains the X/Y resolution, build the name using the real height/width from xrandr's verbose output. For KMS, `modetest`'s debug output has also been modified to print the height and width to avoid the same issue.

psyke83:
I've changed the base to `master` and refreshed the sdl2 commit so that it will now build SDL2 against the `RetroPie` repo instead of my fork. I've also bumped the version to `+2` so that a rebuild will be triggered due to the modesetting patches (which need merging prior to this PR): https://github.com/RetroPie/SDL-mirror/pull/16

joolswills:
Thanks everyone for this. SDL changes included with a minor change for stretch compatibility. Binaries built. Merging.

PR #2901, opened 2019-11-14
add rpcs3 playstation3 emulator licensed under GPLv2.

nakatamaho:
the title says all.

### DIFF
--- a/scriptmodules/emulators/rpcs3.sh
+++ b/scriptmodules/emulators/rpcs3.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks.

Please be more verbose with details. Title doesn't say all. Would be useful to have an overview.

scriptmodules/emulators/rpcs3.sh
```diff
@@ -0,0 +1,36 @@
...
+rp_module_flags="!arm"
+
+function install_bin_rpcs3() {
+    wget --content-disposition https://rpcs3.net/latest-appimage
+    rpcs3_bin=`ls rpcs3*AppImage`  
+    mv $rpcs3_bin $md_inst/$rpcs3_bin
+    chmod +x $md_inst/$rpcs3_bin
+    mkdir $romdir/ps3
+    touch $romdir/ps3/launch.ps3
+}
```
joolswills: Sorry but don't want a binary install like this. Needs a source build imho. Also what about dependencies? Statically linked?

PR #2902, opened 2019-11-16
Add sdl2 library dependency in cgenius.sh

najdanovicivan:
The library is required to be able to successfully compile it


hhromic:
Thanks, can you please use a proper commit message (see existing commits) and place the new dependency in the correct order (near the other `libsdl` dependencies).

najdanovicivan:
I was trying to submit PR using github web interface for the first time so back to the console and now it's fixed. I had to remove the branch as github does not allow me to amend the commit for some reason and publish new one so that we can have so new PR is created #2903

PR #2903, opened 2019-11-17
cgenius - Fix Dependences

najdanovicivan:
Added libsdl2-ttf-dev

The library is required to be able to successfully compile it


joolswills:
Thanks.

PR #2904, opened 2019-11-18
Mesa unneeded for Vero4K GLES and appears to be harmful.

hissingshark:
`libgles2-mesa-dev` is unneeded on this platform and pulls in an awful lot of extra packages.  Some of these appear to break the GPU drivers.

Many Thanks.


joolswills:
Merging - just to note we may implement another fix to avoid this in the main dependency function. Thanks.

PR #2905, opened 2019-11-20
Adding rpcs3 initial script

raelgc:
It's using the appImage download path, which always download the latest build (see the section "For Linux Users" in the [Download](https://rpcs3.net/download) page).

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -179,6 +179,9 @@ ports_platform="pc"
 ps2_exts=".iso .img .bin .mdf .z .z2 .bz2 .dump .cso .ima .gz"
 ps2_fullname="PlayStation 2"
 
+ps3_exts=".ps3"
+ps3_fullname="PlayStation 3"
+
 psp_exts=".iso .pbp .cso"
 psp_fullname="PlayStation Portable"
 
--- a/scriptmodules/emulators/rpcs3.sh
+++ b/scriptmodules/emulators/rpcs3.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="rpcs3"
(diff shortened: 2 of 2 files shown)


joolswills:
Duplicates #2901

raelgc:
@joolswills Sorry. I missed that PR. 

One note about build the binary: I tried in a first moment. But it'll require to add 2 different PPAs: one for Vulkan SDK, other for Qt5.11 (as Ubuntu 18.04 has only Qt5.9).

If you're fine with that, I can create it.

joolswills:
I think its best to wait. I don't want to accept new stuff currently and may prefer to work on this myself. Users can install manually if they want in the meantime. This has to work not only on Ubuntu but also Debian btw

raelgc:
@joolswills Just in order to help, this were the draft functions to compile it (it'll fail due the missing Qt5.11):

```bash
function depends_rpcs3() {
    # Vulkan SDK Repo
    wget -qO - http://packages.lunarg.com/lunarg-signing-key-pub.asc | apt-key add - &>/dev/null
    echo "deb https://packages.lunarg.com/vulkan/1.1.106 $__os_codename main" >/etc/apt/sources.list.d/lunarg-vulkan-1.1.106.list
    local depends=(cmake build-essential gcc-8 g++-8 vulkan-sdk libasound2-dev libpulse-dev libopenal-dev libglew-dev zlib1g-dev libedit-dev libvulkan-dev libudev-dev git qt5-default libevdev-dev qtdeclarative5-dev qtbase5-private-dev libsdl2-2.0-0 libsdl2-dev)
    getDepends "${depends[@]}"
}

function sources_rpcs3() {
    gitPullOrClone "$md_build" https://github.com/RPCS3/rpcs3.git master
}

function build_rpcs3() {
    mkdir build
    cd build
    CXX=g++-8 CC=gcc-8 cmake .. -DCMAKE_INSTALL_PREFIX="$md_inst"
    make clean
    make
    md_ret_require="$md_build/build/bin/rpcs3"
}
```

raelgc:
Answering your questions in the other PR about the appImage file: it's a self contained file (basically just a zip file with everything the app needs inside it). It even has the Qt libraries included.

joolswills:
Thanks.

raelgc:
@joolswills Sorry to insist on this one: but this is basically the same approach we're using actually with pcsx2: as we cannot compile it (due library issues), we're downloading the latest binary build from the authors. 

I don't see why we should not have it in the experimental section, as actually we don't have any entry for it. The benefit of have it on Retropie-Setup is to run updates just updating from source, etc.

PR #2906, opened 2019-11-21
zesarux: use SDL2 for KMS targets

cmitu:
Adds support for using SDL2 on KMS targets. Other changes
* explicitely `--enable-ssl` for build configuration, otherwise it seems SSL support is not [enabled by default](https://github.com/chernandezba/zesarux/blob/89b077721f30157e825cd2d76a00bd81a97f8493/src/Changelog#L434). 
~~* for PI0/1, run with the `--spectrum-reduced-core` parameter for performance. This disables the following: _Debug t-states, Char detection, +3 Disk, Save to tape, Divide, Divmmc, RZX, Raster interrupts, Audio DAC, Video out to file_~~

### DIFF
--- a/scriptmodules/emulators/zesarux.sh
+++ b/scriptmodules/emulators/zesarux.sh
@@ -12,13 +12,20 @@
 rp_module_id="zesarux"
 rp_module_desc="ZX Spectrum emulator ZEsarUX"
 rp_module_help="ROM Extensions: .sna .szx .z80 .tap .tzx .gz .udi .mgt .img .trd .scl .dsk .zip\n\nCopy your ZX Spectrum games to $romdir/zxspectrum"
-rp_module_licence="GPL3 https://sourceforge.net/p/zesarux/code/ci/master/tree/LICENSE"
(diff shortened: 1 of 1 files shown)


joolswills:
Is it not full speed without that option on rpi0/1 ? Do you have an example where that makes a difference as disabling raster interrupts will affect compatibility etc.

joolswills:
There are other spectrum emulators also for the rpi0/1 so disabling some of the useful advanced features that others don't have may not be desirable.

cmitu:
@joolswills I don't know if it's full speed with/without - just stumbled on the upstream option:
> Spectrum reduced core. It uses less cpu, ideal for slow devices like Raspberry Pi One and Zero.

I can remove it and test later - I only tested on Pi3 and Pi4.

joolswills:
Let's leave that for now. Also it's unrelated to rpi4 support so we can change that on master if theres a need to after we have merged the fkms branch.

joolswills:
This needs rebasing now. If you remove the bit I mentioned I will merge and we can test later if it's worth having a reduced core for rpi0/1

cmitu:
Should be ok now, ready to merge.

joolswills:
Thanks.

PR #2912, opened 2019-12-04
packages - added the ability to set a package section based on platfo…

joolswills:
…rm flags

The module/package variable rp_module_section used to need a single string containing
the section for the module - eg "core" "main" "opt".

With this change, you can provide additional space separated parameters to specify the
section for a specific platform flag. This is useful for example to have a module by default
in "opt", but then promote it to "main" for certain systems.

eg for mame4all (mostly useful for slower armv6 devices) you could do

rp_module_section="opt armv6=main"

which would place it in the main section only for armv6

Note that I just threw this code together and am happy for comments/feedback.
It's been something that I've wanted to sort for a while as it will allow us to have better package selections for rpi1/2 and x86.


joolswills:
I've added some section changes as a second commit to illustrate.

joolswills:
Wrong button :-)

joolswills:
I'm going to merge this. Hopefully there's nothing I have missed.

PR #2914, opened 2019-12-05
Adding WPS function from max2play

blackerking:
Added WPS function of
https://github.com/max2play/webinterface/blob/master/opt/max2play/wps_config.sh

## not tested!  ##


joolswills:
Sorry but I'm afraid I can't accept this. I'm not against the feature idea in general but it needs to be redone to fit with our style guide etc. It would also need to.be thoroughly tested and we currently have other priorities. Thanks.

PR #2919, opened 2019-12-16
Add licence and package flag summary pages

mmatyas:
As discussed on [the forum](https://retropie.org.uk/forum/topic/24210/pi-4-support-progress-and-other-package-infos/), this patch adds two new admin modules for generating visual package informations, for listing licences and platform support based on the package flags.


joolswills:
Thanks. Appreciated - this is probably my fault for not being clearer but my idea was for something like this to be integrated into RetroPie-Setup rather than using external repos - and for example using our arrays with the package information rather than grepping etc. So all the code would be part of RetroPie-Setup.

joolswills:
(The javascript could be pulled in, but I would like all custom code to be included as part of our codebase - and so on).

mmatyas:
Ah I see, sure, that could be done too. Should I just place the static website files in a new directory under `admin`?

joolswills:
Yeah same name as module. You can access via `$md_data`

joolswills:
Happy to give some examples/help if it's useful. Check out `scriptmodules/packages.sh` as for example the code to list all modules etc is in there. The arrays with all the module info are defined in that file also.

mmatyas:
Ok, updated, now everything happens locally and using the module infos.

hhromic:
This is really fantastic, thanks for the contribution!

joolswills:
Thanks.

PR #2921, opened 2019-12-18
stats - merge licences and pkgflags into a single module

joolswills:
* allow configuration of upload host/port
 * minor changes regarding build directory relating to merge

@mmatyas I wanted to merge these as I felt they would work in a single module. I don't think I broke anything but please take a look

Also

https://retropie.org.uk/stats/licences/
https://retropie.org.uk/stats/pkgflags/

Cheers.

We probably want to tweak markup later. Also maybe the colours.

Thanks again for your work on this


joolswills:
We may need to do some more processing. The pkgflags doesnt handle !platform values or something I think. Eg. x86 shows as 100% but there are some non x86 packages.

cmitu:
The licences page has an error showing the last update date:

image

joolswills:
Yeah, I saw that. I don't think this is related to this PR (I didn't t even check the html/js yet), but there's a few issues. Feel free to fix. Also GPL shouldn't be red probably :-)

scriptmodules/admin/stats.sh
```diff
@@ -0,0 +1,85 @@
...
+        mod_id="${__mod_id[$idx]}"
+        mod_desc="${__mod_desc[$idx]}"
+        mod_licence="${__mod_licence[$idx]}"
+        echo "$mod_section;$mod_id;$mod_desc;$mod_licence;" >> "$dest/packages.csv"
+    done
+
+    local git_hash=$(git -C "$scriptdir" log -1 --format=%h)
+    local git_date=$(git -C "$scriptdir" log -1 --format=%cd --date=iso-strict)
+    local git_branch=$(git -C "$scriptdir" rev-parse --abbrev-ref HEAD)
+    echo "$git_hash;$git_date;$git_branch;" > commit.csv
```
cmitu: This should be 
```
echo "$git_hash;$git_date;$git_branch;" > "$dest/commit.csv"
```

that's why the 'licences' page doesn't show the revision correctly.

joolswills:
Oops. Thanks.

joolswills:
Fixed. And updated on the above URLs. Cheers

joolswills:
I just reworked it removing most of the vars and the git code duplication. Was done quickly so I may have left a bug or two.

mmatyas:
Nice, yeah the script seems fine to me. However I've just diffed the `packages.csv` against the grep generated one (eg. [this](https://mmatyas.github.io/retropie-pkglicences/packages.csv)), and noticed that our version contains only the packages available for x86 (ie. 221 lines vs 192 lines), which also explains the 100% progress there (I suppose `${__mod_idx[@]}` only contains the available packages?).

There's also one difference, the description of `resetromdirs` is `Reset ownership/permissions of $romdir`, where `$romdir` becomes `/home/jools/RetroPie/roms` in this version. Not sure if that's an issue.

GPL got red as it's the dominant color in [its logo](https://en.wikipedia.org/wiki/File:GPLv3_Logo.svg). I was trying to balance between making it error-red but not losing contrast with the rest of the colors, but  feel free to change it if you don't like it :)

Also now that both pages are generated from the same script, they could also be changed to share a common CSV file, as they only differ in one column anyway. I can make that in a follow up patch if you'd like to.

joolswills:
>Nice, yeah the script seems fine to me. However I've just diffed the packages.csv against the grep generated one (eg. this), and noticed that our version contains only the packages available for x86 (ie. 221 lines vs 192 lines), which also explains the 100% progress there (I suppose ${__mod_idx[@]} only contains the available packages?).

ah yes - I didn't think of that. I will come up with a solution - which might be reverting back to the grep :-) Or maybe some additional packaging code / or logic change as it would be useful to have all package info available. I will probably add a parameter to existing functions so we can obtain a full list of packages etc.

> There's also one difference, the description of resetromdirs is Reset ownership/permissions of $romdir, where $romdir becomes /home/jools/RetroPie/roms in this version. Not sure if that's an issue.

also well spotted - as the values are "sourced" this will end up being the romdir on my system - will keep this in mind with the possible changes I mentioned.

Sorry about this!

Regarding the csv stuff - please feel free to do a PR - I will merge this for now, so you can do a PR against the new locations, and I will fix up the problems you noticed later. Thanks!

joolswills:
Please re-check the urls - does that looks any better - if so I'll do a PR with my changes.

joolswills:
It might without the duplicates - I'll fix that :-)

joolswills:
Might be better now - I will continue later. Thanks.

PR #2929, opened 2019-12-19
pegasus-fe: fix license abbrev

hhromic:
The license abbreviation of this module should be `GPL3` for consistency.

@mmatyas is this okay for you?
Btw this discrepancy was found using your recent license viewer tool :)


mmatyas:
A bit late, but sure, it's fine :)

PR #2931, opened 2019-12-22
jzintv: enable for the kms platform.

cmitu:
Similar to other SDL1 based emulators, pass the resolution arguments to the emulator, to be expanded by `runcommand` when started. Enable the `dispmanx` SDL1 videodriver only for the Raspberry Pi platforms that don't use the KMS driver.


joolswills:
Will hold off on this as it works with dispmanx sdl1 backend without having to render at a higher res. I will submit a PR and we can decide the best route.

joolswills:
Closing as went with dispmanx solution temporarily but this could be revisited for non rpi4 Kms targets maybe.

PR #2932, opened 2019-12-22
emulationstation: enable the PI bits on supported platforms

cmitu:
On the PI4 system, the `omxplayer` and the additional audio devices don't get enabled automatically since it's not using the legacy GLES library. Use the parameter added in https://github.com/RetroPie/EmulationStation/pull/573 to enable them.
Added an explicit dependency on `omxplayer`, it wasn't needed before since it got installed with the `splashscreen` scriptmodule.

EDIT: forgot to mention, this needs https://github.com/RetroPie/EmulationStation/pull/573 to be cherry-picked to the ES stable branch.


joolswills:
Thanks.

PR #2938, opened 2019-12-26
retroarch: update to 1.7.9.2

cmitu:
_This is a small modification on top of #2853, which was reverted because of a bug in RetroArch's build system._
Some changes since 1.7.6:
 - upstream dropped the `video_shader` parameter and replaced it instead with a CLI option (`--set-shader`), but in my tests this is not behaving the same - the shader from the command line overrides any game overrides, whereas before, the shader saved for the game would override the one in the config file. Also, for some reason the default for enabling the Video Shaders has been changed to _false_.    
I added a patch to bring back reading the `video_shader` parameter when every other auto-shader preset (game/core/content/global) is not present.  

- the default compile options enable support for OpenGL1, disable them on the Pi when the BRCM libs are used since it produces a linking error (missing `-lGL`).
- disabled `vulkan` and `wayland` for non-X11 platforms (fixes previous PR's build with RetroArch < 1.8.0 and SDL not having Wayland enabled - see https://github.com/libretro/RetroArch/issues/9583).

Things needed:


scriptmodules/emulators/retroarch.sh
```diff
@@ -56,14 +55,15 @@ function build_retroarch() {
         params+=(--disable-ffmpeg)
     fi
     isPlatform "gles" && params+=(--enable-opengles)
-    # Temporarily block dispmanx support for fkms until upstream support is fixed
-    isPlatform "dispmanx" && ! isPlatform "kms" && params+=(--enable-dispmanx)
     isPlatform "rpi" && isPlatform "mesa" && params+=(--disable-videocore)
+     # Temporarily block dispmanx support for fkms until upstream support is fixed
```
joolswills: Indent.

joolswills:
Builds fine now btw thanks.

joolswills:
The xmb monochrome folder has the extra assets for the ticks on the on-screen messages. Please can you detail what the other assets are needed for by default.

cmitu:
For the extra assets, we'll need (from https://github.com/libretro/retroarch-assets):
* the `menu_widgets` folder
* the `ozone` folder
* (optional) the `rgui` folder, which has some theme presets and backgrounds that can be used.

joolswills:
Thanks. I meant more specifically what are ozone and menu_widgets needed for (so I can test).

cmitu:
The new assets are for for the new notification interface. Basically any notification message should show up with the new font and icons
- controller config loaded
- shader loaded
- remap applied/saved.

cmitu:
@joolswills should be rebased now. I've added the `rgui_aspect_ratio_lock` also to the default rp-dist config, the other settings options that have been added were already present.

scriptmodules/emulators/retroarch.sh
```diff
@@ -108,7 +108,7 @@ function install_xmb_monochrome_assets_retroarch() {
     local dir="$configdir/all/retroarch/assets"
     [[ -d "$dir/.git" ]] && return
     [[ ! -d "$dir" ]] && mkUserDir "$dir"
-    downloadAndExtract "$__archive_url/retroarch-xmb-monochrome.tar.gz" "$dir"
+    downloadAndExtract "$__binary_base_url/retroarch-xmb-monochrome.tar.gz" "$dir"
```
joolswills: this shouldn't be included here if it's been rebased over latest master

joolswills:
You seem to have 2 commits (second commit mentioning being rebased which shouldn't be in a commit msg) and a commit from master included.

joolswills:
as I want to change the assets maybe it's easiest I rebase this and commit to my fork?

joolswills:
https://github.com/RetroPie/RetroPie-Setup/compare/master...joolswills:retroarch_1.7.9.2?expand=1

joolswills:
If you're happy I can do a new PR (place your info in it - I put it in the commit as I think it's useful having all the changes described in the commit, and not just in the PR).

cmitu:
> If you're happy I can do a new PR (place your info in it - I put it in the commit as I think it's useful having all the changes described in the commit, and not just in the PR).

Sounds good.

cmitu:
Closing in favour of #2941.

PR #2939, opened 2019-12-26
Fix pcsx2 chain dependency

raelgc:
After install [`Dolphin`](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/dolphin.sh), PCSX2 will be removed, due packages dependencies.

User should be able to just reinstall PCSX2, but after request the reinstall, system will be lead to a broken dependencies state in `apt`.

The reason: Dolphin **build** requires `portaudio19-dev` and the PCSX2 package is missing a dependency on `libjack-jackd2-0:i386` (and both packages are not compatible).

This PR is just changing the PCSX2 script to explicitly require `libjack-jackd2-0:i386`, so when we request to install/reinstall, `apt` will be able to solve the dependencies.

This is valid for Ubuntu and Debian.

### DIFF
--- a/scriptmodules/emulators/pcsx2.sh
+++ b/scriptmodules/emulators/pcsx2.sh
@@ -38,6 +38,9 @@ function depends_pcsx2() {
         # On Ubuntu, add the PCSX2 PPA to get the latest version
         [[ -n "${__os_ubuntu_ver}" ]] && add-apt-repository -y ppa:pcsx2-team/pcsx2-daily
         dpkg --add-architecture i386
+        # For some reason, this dependency is not explicity, and can lead to broken system state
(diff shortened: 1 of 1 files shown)


raelgc:
I'll send a PR to PCSX2 too, maybe they can fix it directly in their dependencies.

nlsdg:
Is this still relevant? The PR on the PCSX2 repo has been closed in september 2020 without any code change. But it mentions that portaudio is now optional.

raelgc:
Yeah, aside this, pcsx2 has now `amd64` builds, which probably is no more conflicting with Dolphin requirements. 

TL;DR: I'll close this. If I encounter this issue again in the future, I can reopen it.

By the way, thanks for all hard work, guys.

PR #2941, opened 2019-12-29
Retroarch 1.7.9.2

joolswills:
rebase of #2938 with asset changes

_This is a small modification on top of #2853, which was reverted because of a bug in RetroArch's build system._
Some changes since 1.7.6:
 - upstream dropped the `video_shader` parameter and replaced it instead with a CLI option (`--set-shader`), but in my tests this is not behaving the same - the shader from the command line overrides any game overrides, whereas before, the shader saved for the game would override the one in the config file. Also, for some reason the default for enabling the Video Shaders has been changed to _false_.    
I added a patch to bring back reading the `video_shader` parameter when every other auto-shader preset (game/core/content/global) is not present.  

- the default compile options enable support for OpenGL1, disable them on the Pi when the BRCM libs are used since it produces a linking error (missing `-lGL`).
- disabled `vulkan` and `wayland` for non-X11 platforms (fixes previous PR's build with RetroArch < 1.8.0 and SDL not having Wayland enabled - see https://github.com/libretro/RetroArch/issues/9583).

Things needed:


joolswills:
@cmitu just a heads up. I'm getting segfaults from lr-atari800 with this update of retroarch although it works when verbose logging is on (interesting bug). 

Disabling the shader config patches resolves it though so I believe it may be related to that. The core seems to init the screen exit and reinit so I wonder if it's related to calling an init routine twice to trigger the issue. I've not had a chance to build with debugging yet but I think a backtrace will probably clarify things.

cmitu:
I'll take a look.

PR #2942, opened 2019-12-30
frotz: remove module flags

cmitu:
Since `frotz` has only a text interface (`ncurses` based), enable it for all platforms.
I don't have an Odroid to check if it's available in the repositories, but if they're using also upstream repositories, then it should be installable.


hooperre:
I'm curious if the same can be done for `OpenTyrian`? By removing the module flags I was successfully able to build and run it on the Pi 4. I'd be curious to know why it isn't able to be removed. I'm still learning.

joolswills:
We are still working on enabling and testing modules and there's a lite.more too it (related to sdl1). Please use the forum for discussion.

PR #2943, opened 2020-01-01
Vero4k needs explicit video-mali build since the merger

hissingshark:
Now that the Mali branch has been merged into the SDL2 super branch #2809 we need to explicitly set `--enable-video-mali` as per other Mali platforms.

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -79,7 +79,7 @@ function build_sdl2() {
         conf_flags+=("--disable-video-x11")
     fi
     ! isPlatform "x11" && conf_flags+=("--disable-video-vulkan")
-    isPlatform "mali" && conf_flags+=("--enable-video-mali" "--disable-video-opengl")
+    isPlatform "mali" || isPlatform "vero4k" && conf_flags+=("--enable-video-mali" "--disable-video-opengl")
     isPlatform "rpi" && conf_flags+=("--enable-video-rpi")
     isPlatform "kms" || isPlatform "rpi" && conf_flags+=("--enable-video-kmsdrm")
 


scriptmodules/supplementary/sdl2.sh
```diff
@@ -79,7 +79,7 @@ function build_sdl2() {
         conf_flags+=("--disable-video-x11")
     fi
     ! isPlatform "x11" && conf_flags+=("--disable-video-vulkan")
-    isPlatform "mali" && conf_flags+=("--enable-video-mali" "--disable-video-opengl")
+    isPlatform "mali" || isPlatform "vero4k" && conf_flags+=("--enable-video-mali" "--disable-video-opengl")
     isPlatform "rpi" && conf_flags+=("--enable-video-rpi")
```
joolswills: Why doesn't the previous line handle this? Is the Mali flag not set for vero4k? If not that needs to be changed there not here.

hissingshark:
The mali flag usually leads to unwanted dependancy issues in various modules, so overall it has been cleaner using a separate flag for the platform.

joolswills:
I don't like this hack really. It needs to be implemented differently using platform flags if we are to keep vero4k support.

joolswills:
If there are dependency issues (which?) We should visit those instead.

hissingshark:
Fair enough.  I'll review everything and see what I can do differently.

joolswills:
Thanks.

hissingshark:
The most prevalent issue for adding `mali` to the Vero4k platform flags is the presence of `rp_module_flags="!mali"` in modules that at least work on the Vero4k (can't speak for other Mali platforms):

1. dosbox-sdl2
2. mupen64plus
3. reicast

Easily fixed by removing `!mali` for the affected modules, but with the potential consequences to incompatible Mali platforms.

---
I note here a glitch with dosbox-sdl2.  By borrowing the dependency list from dosbox it tries to pull in SDL1.2, even though it isn't needed for an SDL2 platform (and is unavailable here).  Would a PR to fix that be acceptable?

---
The only genuine dependancy issue using `mali` is that it tries to install `mali-fbdev`, which doesn't exist on this platform.  This affects:

1. retroarch
2. sdl2

Perhaps this could be handled by adding negation to the depends handler? So it would be used e.g. `isPlatform "vero4k" && depends+=(vero3-userland-dev-osmc !mali-fbdev)` 
Happy to write/PR this functionality.
Or is there already a platform based package blacklist implemented in RetroPie? (I've not found it)

psyke83:
@hissingshark 

It may be the case that some scriptmodules exclude the `mali` platform simply due to insufficient testers. For example, I wrote the `dosbox-sdl2` installation script, and probably disabled Mali support only because I have no hardware to verify it works. Although vero4k is apparently different to other Mali boards, I think it's still a better idea to submit fixes for the modules instead of removing the `!mali` flag from your board definition.

Regarding `mali-fbdev`, take a look at dependency handling in `helpers.sh`, specifically [https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/helpers.sh#L217-L220](this)
block. You could add a similar section that reads something like the following:

```
        if [[ "$required" == "mali-fbdev" ]]; then
            isPlatform "vero4k" && required="vero3-userland-dev-osmc"
        fi
```
With the above code committed (assuming it's correct), then you no longer need to worry about scriptmodules that mark `mali-fbdev` for installation; in fact, you could then remove or replace existing references to `vero3-userland-dev-osmc` in the individual scriptmodules with `mali-fbdev` and you should have no problems.

OT: I'm curious if the vero4k platform kernel/userland drivers supports KMS? If so, it might be worth experimenting with that so that you could leverage the generic KMS support that's been committed recently. But that's a discussion for another time.

hissingshark:
@psyke83 
Sounds like removing the `!mali` from tested modules could be the way to go then.

 Re: dependency handling, I knew I'd seen something about that before.  How frustrating.  You're absolutely right, a clause in `helper.sh` would nail that going forward.  A lot neater too.

If @joolswills agrees I'll close this PR and submit one that applies the above instead.

Yes, we have no SDL1.2 because there is no mali-fbdev driver support.  Indeed my entire project was only possible because "mihailescu2m" added that to SDL2 at v2.0.2.  I'll take a look at the Xorg idea.  I've been trying for years to find a way of getting at least some version of Minecraft running on the Vero4K, if not the official .jar.

I'll ask the devs about KMS.

Thank you for the feedback.

psyke83:
No worries. If you're interested in testing a proof of concept, you can try switching to this test branch and then install dosbox (the original, not sdl2 version): https://github.com/psyke83/RetroPie-Setup/tree/bad_sdl

If it works as expected, dependency resolution for `dosbox` should trigger installation of `xorg` and `matchbox-window-manager`, and the `dosbox` entry of emulators.cfg should have the `XINIT:` prefix. Finally, the "+Start DOSBox.sh" script has been modified to launch dosbox through runcommand (otherwise it will bypass the emulator.cfg and thus not run via Xorg properly). 

This is perhaps too hacky, but it would be valuable to know if a solution like this would be a viable means of getting more SDL1 software running on Mali. This could also be useful for KMS systems (including RPI4), as they have the same issue with SDL1 applications that require acceleration (and SDL2's kmsdrm driver has a tearing bug for RPI3/4 that doesn't manifest on the Xorg driver).

joolswills:
This should resolve the dosbox issue

https://github.com/RetroPie/RetroPie-Setup/pull/2969

Please see LINUX-HEADERS commit for an example of handling specific vero4k dependencies Vs mali

hissingshark:
Now addressed by PR: #3007

PR #2946, opened 2020-01-03
srb2: update to version 2.2

cmitu:
New version 2.2 was released last December, with a lot of updates and bugfixes.
The assets have also changed and the module doesn't install anymore because of that and because the (build) binary name changed.

The changes include:
* updated binary name to fix the install
* updated dependencies (`libgme` and `libopenmpt`). Even though the project's page lists `libupnp` also, I don't see any reference to it, though it bundles a copy of the `miniupnpc` library so it might use that.
* updated assets names and location as they're checked during the configuration from `cmake`.

NOTE: I tested with the assets from http://ppa.launchpad.net/stjr/srb2/ubuntu/pool/main/s/srb2-data/srb2-data_2.2.0-20191207191146.tar.xz (the SRB2 Ubuntu PPA). These would need to be updated on the RetroPie's server also in order for the build to succeed.

### DIFF
--- a/scriptmodules/ports/srb2.sh
+++ b/scriptmodules/ports/srb2.sh
@@ -15,7 +15,7 @@ rp_module_licence="GPL2 https://raw.githubusercontent.com/STJr/SRB2/master/LICEN
 rp_module_section="exp"
 
 function depends_srb2() {
-    getDepends cmake libsdl2-dev libsdl2-mixer-dev
+    getDepends cmake libsdl2-dev libsdl2-mixer-dev libgme-dev libopenmpt-dev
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

cmitu:
It currently doesn't build because of the old assets -  they're not updated on the RetroPie's site. We can use the URL I mentioned, from the PPA, what do you think ?

joolswills:
I'll update the assets on our site.

joolswills:
done.

cmitu:
Works fine, thank you !

PR #2948, opened 2020-01-06
lr-atari800 - fix for intermittent segfault on load

joolswills:
seems to be a lot of code that's called more than once - there's an odd timing issue that can cause a segfault in retroarch. When lauching, the emulator seems to trigger video initialisation twice, but removing this co_switch seems to resolve the crash but this core looks like it needs work

@cmitu


joolswills:
It sure this fixes all cases. Putting a sleep(1) at the end of the update_variables function seems to help but that's not really a fix. I need to read the libretro API as this looks like an issue caused by changing libretro stuff at the wrong time.

joolswills:
Crash is gone with retroarch 1.8.4 but there is corrupt osd text. Will need to revisit.

PR #2949, opened 2020-01-07
mesa-drm: switch to RetroPie repo & exclude extraneous files

psyke83:
Not sure if you'll be bothered with the wildcard to pull in the libkms library & symlinks (currently: `libkms.so  libkms.so.1  libkms.so.1.0.0`). I can change it to explicitly name the files if you want.

An earlier version of my modesetting PR was using the `modeprint` tool from mesa-drm, but it wasn't possible to install `libkms.so` in the same directory as `modeprint`, as the binary would not resolve the library in the same folder. This isn't a problem with `modetest`, so we can cut down on the extraneous files now.

If this is merged, please trigger a rebuild of the mesa-drm binaries so that runcommand doesn't break for users (due to the path change compared to the current binary package).

### DIFF
--- a/scriptmodules/supplementary/mesa-drm.sh
+++ b/scriptmodules/supplementary/mesa-drm.sh
@@ -22,7 +22,7 @@ function depends_mesa-drm() {
 }
 
 function sources_mesa-drm() {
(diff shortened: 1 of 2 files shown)


joolswills:
binaries built. Thanks.

PR #2950, opened 2020-01-07
emulationstation: fix installation on Raspbian Jessie

cmitu:
EmulationStation version `2.7.6`, installed on older Debian platforms, doesn't have the `resources` folder. Selectively add it just for newer releases of EmulationStation and Debian (Stretch or newer).

(reported in the forums in https://retropie.org.uk/forum/topic/24483/v/17)


joolswills:
I'll accept it but I've removed some other jessie compatibility stuff - Maybe we should just disable updating completely on Jessie. I thought it was made clear enough on the message :-)

joolswills:
The problem with disabling updating, is they don't get the code till they update. So they get at least one more update.

PR #2954, opened 2020-01-13
added VVVVVV to ports

joolswills:
A few sources take a long time to compile with -O2 btw. Limited testing but it works.


pjft:
Will test during the week.

joolswills:
Confirmed working by @pjft 

Added author to description to adhere with license.

PR #2957, opened 2020-01-15
retroarch: update to v1.8.4

cmitu:
Only change is the rebased patches, it should fix the crashes reported in the `lr-atari800` core.


joolswills:
Thanks.

PR #2958, opened 2020-01-17
runcommand.sh: assign space to button B

crcerror:
This change will make radiolists and checklists useable

```
#!/bin/bash
pizza=`dialog --checklist "Pizza with ..." 0 0 4 \
 cheese "" on\
 chicken "" off\
 ham "" off\
 tuna "" off 3>&1 1>&2 2>&3`
dialog --clear
clear
echo "Your order: Pizza with $pizza"
```

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -132,7 +132,7 @@ function start_joy2key() {
 
         # call joy2key.py: arguments are curses capability names or hex values starting with '0x'
         # see: http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html
-        "$ROOTDIR/supplementary/runcommand/joy2key.py" "$JOY2KEY_DEV" kcub1 kcuf1 kcuu1 kcud1 0x0a 0x09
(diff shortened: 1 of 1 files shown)


joolswills:
We don't include any radiolists / checklists. But we need tab - 227b65e5ef6c3dcf05ead85819e1053ae13c7a58

joolswills:
If you have some additional dialog, you can include joy2key in it and remap additional buttons.

crcerror:
I know how to utilize the joy2key script.
But checklists may be usefull if you want to install more than one theme/emulator/port

I can show you a good usecase of such dialogs (hopefully) soon. But I must admit I was not aware that the space key was utilzed in 2015 :)

crcerror:
> 
> 
> If you have some additional dialog, you can include joy2key in it and remap additional buttons.

[Midnight Joy --- Midnight Commander useable with joypad](https://retropie.org.uk/forum/topic/19267/)

joolswills:
How about we map button 3 as space?

crcerror:
honestly I think it is not neccesary. As you said we can source the code in own scripts and exchange the button. As long as there is no need because you always use `--menu` form for all dialogs we can left code as it is.
Maybe for RetroPie v4.5 if you decide to use radio/selection lists then it would make sense. Just my opinion.

crcerror:
@joolswills As promised usage of this feature can be found here
https://retropie.org.uk/forum/topic/15961
Thank you for maintaining RetroPie :)

PR #2959, opened 2020-01-17
Add .cue extension to 3DO

mikedahlgren:
Verified that .cue file extensions work correctly.  Already listed as supported in the wiki.


joolswills:
Thanks.

PR #2963, opened 2020-01-23
Added CPU min and max clockspeed to bashwelcometweak

parker-hemphill:
For the new Pi4 I've added tweaks to the bashwelcometweaks.sh to add clockspeed for CPU, Core, and V3D.  Doesn't extend the length of retropiewelcome display.

### DIFF
--- a/scriptmodules/supplementary/bashwelcometweak.sh
+++ b/scriptmodules/supplementary/bashwelcometweak.sh
@@ -27,6 +27,10 @@ function getIPAddress() {
     [[ -n "$ip_route" ]] && grep -oP "src \K[^\s]+" <<< "$ip_route"
 }
 
+function getClockSpeed() {
+    expr $(echo $(vcgencmd measure_clock $1|awk -F= '{print $2}') / 1000000 )
+}
+
 function retropie_welcome() {
     local upSeconds="$(/usr/bin/cut -d. -f1 /proc/uptime)"
     local secs=$((upSeconds%60))
@@ -126,6 +130,9 @@ function retropie_welcome() {
                 out+="Temperature........: CPU: $cpuTempC°C/$cpuTempF°F GPU: $gpuTempC°C/$gpuTempF°F"
                 ;;
             10)
+                out+="Clock Speed........: CPU: $(getClockSpeed arm)MHz Core: $(getClockSpeed core)MHz V3D: $(getClockSpeed v3d)MHz"
+                ;;
+            11)
                 out+="${fgwht}The RetroPie Project, https://retropie.org.uk"
                 ;;
         esac
--- a/scriptmodules/supplementary/bashwelcometweak.sh
+++ b/scriptmodules/supplementary/bashwelcometweak.sh
@@ -27,6 +27,10 @@ function getIPAddress() {
     [[ -n "$ip_route" ]] && grep -oP "src \K[^\s]+" <<< "$ip_route"
 }
 
+function getClockSpeed() {
+    expr $(echo $(vcgencmd measure_clock $1|awk -F= '{print $2}') / 1000000 )
+}
+
 function retropie_welcome() {
     local upSeconds="$(/usr/bin/cut -d. -f1 /proc/uptime)"
     local secs=$((upSeconds%60))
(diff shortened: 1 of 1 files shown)


hhromic:
Note that this bash snippet does not only execute on RPI devices, but also X86 and other supported platforms where `vcgencmd` doesn't apply.

parker-hemphill:
I’ll adjust it accordingly and resubmit.

parker-hemphill:
Adjusted to use `lscpu` for universal functionality.  Tested on Debian x86, Raspian, and Arch x86.

hhromic:
Thanks, is `lscpu` installed by default in stock RetroPie, i.e. stock Raspbian Lite?

hhromic:
Bgtw, imho the clock speeds report you had initially (CPU, Core, V3D) were far more useful than displaying the min/max clocks. To be clear, I didn't ask to remove usage of `vcgencmd` but to use it correctly such as in how the GPU temperature is collected currently.

Showing min/max CPU clocks is not of much value tbh.

parker-hemphill:
lscpu exist on Debian lite version and Raspian Lite.  I believe it even exist on DietPie which is another popular distro used on Raspberry Pie installs.  I agree my original pull is more useful but I believe the min and max CPU is useful for RPI users since there is a lot of tinkering with /boot/config.txt.  I'll do some digging and see if I can find a universal solution across Debian Buster and Raspian lite since those are the default choices for RetroPie.

hhromic:
> lscpu exist on Debian lite version and Raspian Lite. I believe it even exist on DietPie which is another popular distro used for RetroPie installs. 

I meant more to make sure it is installed *by default* in these distributions. For instance does it come pre-installed with Raspbian Lite (the distribution RetroPie is based on)?

You must make sure that your contribution works with what is pre-installed.

cmitu:
`lscpu` is part of `util-linux` package, which is marked as _essential_ by `apt`, so I think it's installed by default on a Debian (or derivative).

parker-hemphill:
I believe it is. I’m trying to find something by default that will provide the same type of information in x86 based RetroPie builds before I resubmit my pull. Worst case I’ll have it only display if vcgencmd exist.

joolswills:
Not sure tbh this extra info is needed. Not sure it's useful on a 8 core desktop also - less is more. Im also not convinced on adding more stuff to this already rather cluttered display. Sorry.

joolswills:
I was actually considering removing some stuff in the past :-)

parker-hemphill:
Totally get that. I was thinking it was useful for RPI4 devices as people tinker and try to do things to squeeze more performance. The latest firmware removing GPU overclock shows this is a valid need IMO. would adding it as a function and alias with local variables instead be appropriate?

joolswills:
Not sure I follow regarding removal of overclock? As it's gone (and probably won't return) I would think there is less need - but maybe I have misunderstood? It's also a static config and I think the display of variable info is more important.

parker-hemphill:
> Not sure I follow regarding removal of overclock?
Users are doing more tinkering with CPU clock and things like CPU_GOVERNOR.  My previous comment was wondering if this would be better as a function added to the bashwelcometweak, something like pi_info that displays clock speeds and other things useful for RPI users.  I'm going to start a post on Reddit under /s/retropie to get an idea of what other users would be interested in seeing and have a better answer for what I want to add and so I don't keep wasting your time having to reply to my comments for this pull :D

parker-hemphill:
According to a new entry https://www.raspberrypi.org/documentation/configuration/config-txt/overclocking.md there is still the option to overclock core and CPU,  These correlate on the latest Pi firmware and affect one another so I feel there is still a need for this information to be displayed in retropiebashwelcome.

parker-hemphill:
BUMP.  See latest commit/comment before this one please

timothybrown:
I like this, very useful when tinkering with overclock settings!

@parker-hemphill A slight correction: As of the most recently released bootloader, overclocking is supported for everything *except* `core_freq`. The `gpu_freq` option will only apply to `v3d`, `isp` and `h264` now and setting `core` will have no (or a detrimental) effect.

The Raspberry Pi Documentation is severely out of date in several areas, unfortunately.

joolswills:
Sorry - I won't be including this.

PR #2964, opened 2020-01-31
runcommand: fix rare default kms mode detection problem

psyke83:
Certain KMS systems may enumerate multiple CRTCs, some of which have
empty/invalid modes defined. If the first grepped entry happens to be such a
mode, modesetting will break. Fix by using a more accurate grep expression
to find an active CRTC with a valid mode that is marked with the "preferred"
flag.

Issue can be reproduced on Pi 3B with the vc4-kms-v3d overlay.


joolswills:
Thanks.

PR #2967, opened 2020-02-01
Updated bombermaaan script.

bjaraujo:


joolswills:
Thanks.

joolswills:
I assume you prefer it fixed to a specific version ? I'm happy with that anyway.

bjaraujo:
Yes, because although I try to have the master in a building state, it is not guaranteed. Thanks for merging.

PR #2972, opened 2020-02-02
Rpi4 gles3

joolswills:
Enabled gles3 for Ra / mupen64 plus libretro cores

Needs testing etc. Feedback welcome.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -55,6 +55,7 @@ function build_retroarch() {
         params+=(--disable-ffmpeg)
     fi
     isPlatform "gles" && params+=(--enable-opengles)
+    isPlatform "gles3" && params+=(--enable-opengles3)
     isPlatform "rpi" && isPlatform "mesa" && params+=(--disable-videocore)
     # Temporarily block dispmanx support for fkms until upstream support is fixed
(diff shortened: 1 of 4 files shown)


dankcushions:
Tested on RPI4 - works great! GLES3 is reported in both lr-m64p and lr-m64p-next, and my go-to test (the Jumbotron in Mario Kart 64, which I believe needs GLES3 to work) works once the right core options are enabled.

quicksilver7837:
JumboTron on Mario kart 64 works already in gliden64 on the pi 4. So I would assume that it's not a gles3 exclusive feature. Unless of course mupen64plus+gliden64 is already enabled for gles3?

dankcushions:
@quicksilver7837 I think standalone already compiles with GLES3 support on pi4. this is just for the lr- cores :) for me lr-m64p-next didn't show the jumbotron before GLES3, and did after (but you have to set `rdram colour copy = async`, or something like that, in the core options).

quicksilver7837:
Yes I knew this was for just for the libretro cores only. I was just making the observation that if the standalone mupen64plus was not enabled for gles3 but was displaying the jumbo tron then it may not be a gles3 feature. I tested on a pi 3 and the jumbo Tron doesn't display properly so as you say it makes sense that it is a gles3 only feature.

PR #2974, opened 2020-02-03
N64: Mesa: support high resolutions modes

dankcushions:
Because we only had one resolution for Mesa, we weren't able to traverse the resolutions loop properly to add the high resolution modes, so I did it another way.

~Also took the opportunity to retire every video plugin bar GLideN64 for pi4, since I doubt they offer any advantage now we have a GLES3 device with decent drivers. Kept them around for mesa on pi3 as its GLES2 pathway probably has some limitations.~

**Note**: only tested on pi4 (mesa). will test on pi3 for mesa and non-mesa once my new SD card arrives! non-mesa should be functionally identical to prev version, though.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -176,20 +176,27 @@ function configure_mupen64plus() {
     isPlatform "kms" && resolutions=("%XRES%x%YRES%")
 
     if isPlatform "rpi"; then
-        isPlatform "videocore" && addEmulator 1 "${md_id}-auto" "n64" "$md_inst/bin/mupen64plus.sh AUTO %ROM%"
-        for res in "${resolutions[@]}"; do
-            local name=""
-            local nativeResFactor=""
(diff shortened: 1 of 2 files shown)


joolswills:
I don't want to retire the other drivers yet. Maybe better to split the Mesa resolution stuff out and just deal with that here.

dankcushions:
done! retiring the old plugins is a seperate change and we're probably better to wait for the official release where we have much more people testing!

joolswills:
Thanks. Not had a chance to test yet but let me know when you have tried it on older RPI.

dankcushions:
There is an issue with this on pi3 + old drivers:
```/opt/retropie/emulators/mupen64plus/bin/mupen64plus.sh: line 15: [[: 640x480: value too great for base (error token is "640x480")```
will investigate!

dankcushions:
ok, fixed issue and ok on pi3 + old drivers :)

joolswills:
Thanks.

PR #2977, opened 2020-02-05
Enabling a variety of sdl1 applications for rpi4

joolswills:
This isn't everything but I have tested pretty much everything here - some of the sdl1 do not work with the dispmanx driver and have been left excluded.

for the sdl1 apps that work on dispmanx this will technically enable all the packages for other kms platforms (Only tinker board currently and this needs revisiting). The plan is to rework this so we can choose a backend based on platform and allow users to switch if needed - so for tinkerboard these would probably default to using X. 

In some cases I have enabled dispmanx by default where it wasn't before - but users on rpi1/2/3 can turn off via the dispmanx config. Some of them used X also and just needed the `!kms` removing.

### DIFF
--- a/scriptmodules/emulators/daphne.sh
+++ b/scriptmodules/emulators/daphne.sh
@@ -14,7 +14,7 @@ rp_module_desc="Daphne - Laserdisc Emulator"
 rp_module_help="ROM Extension: .daphne\n\nCopy your Daphne roms to $romdir/daphne"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/RetroPie/daphne-emu/master/COPYING"
 rp_module_section="opt"
-rp_module_flags="!x86 !mali !kms"
+rp_module_flags="dispmanx !x86 !mali"
 
 function depends_daphne() {
     getDepends libsdl1.2-dev libvorbis-dev libglew-dev zlib1g-dev
@@ -70,6 +70,8 @@ _EOF_
 
     chown -R $user:$user "$md_inst"
     chown -R $user:$user "$md_conf_root/daphne/dapinput.ini"
+
+    setDispmanx "$md_id" 1
 
     addEmulator 1 "$md_id" "daphne" "$md_inst/daphne.sh %ROM%"
     addSystem "daphne"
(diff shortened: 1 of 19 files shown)


mth75:
Can you include the TI99 emulator? Thx for the great progress:

https://github.com/RetroPie/RetroPie-Setup/wiki/TI-99

joolswills:
I didn't check that one. I will.

mth75:
Thx!

joolswills:
ti99sim should work now.

mth75:
Thx, indeed Ti99sim works now (fullscreen). LinApple however still isn't fullscreen (no change comparing to the last installscript)

joolswills:
Thanks. I need to adjust that in a similar way.

joolswills:
Linapple should be full screen now.

mth75:
and it is! Thank you.

mth75:
I think you are aware of the fact that selecting the resolution in runcommand (for video and/or rom) hasn't got any effect at the moment? I've tested this with LinApple and Ti99Sim and Frotz. Also, both LinApple and Ti99Sim are fully blown fullscreen at the moment (no borders so to say).

joolswills:
The resolution change on rpi4 is only for sdl2 applications. We have a custom sdl2 to handle this. It's not like on older RPI where an app can inherit the mode change and read the new resolution

mth75:
Thanks, so either emu's should be updated to sdl2 (not likely in some cases) or will become troublesome in the future.

joolswills:
Yeah. We may have an alternative solution in the future - this is a solution just to get things working. if you add XINIT: to the beginning of the launch CMD in emulators.cfg you can launch under X and then you can select a mode.

joolswills:
I plan to make this configurable in the future.

mth75:
that would be a nice. In the meantime I will try to reach out to the Ti99Sim developer, this emu does needs some love.

cmitu:
We could probably enable `dispmanx` for `basilisk` by default.
There's a SDL2 version for it, but I think that might be better as a separate scriptmodule.

joolswills:
Yeah. There's a few more emulators I've not checked..some don't work and will need X like sdltrs. 

Once this is merged feel free to add more :-)

joolswills:
I also need to add a change to runcommand to ignore dispmanx config if XINIT is used. People can customise it then in the meantime.

joolswills:
I've added that now.

joolswills:
Merging. As mentioned this is a stopover solution as I intend to add the facility to launch under X for sdl1 apps. This would also be the default for any other Kms targets.

It's also possible to create own emulator.cfg entries prefixing XINIT: to force launching under X11.

PR #2978, opened 2020-02-08
lr-xrick - added libretro core

joolswills:
* rework xrick to work with libretro core
 * remove data.zip from xrick due to licence uncertainty
 * various module cleanups


joolswills:
Going to merge this. May remove standalone xrick in the future once the libretro module is well tested.

PR #2979, opened 2020-02-08
eduke32: fix build on gcc 6.3.x

psyke83:
Recast klabs function to integer to prevent compiler error on older gcc versions.

--

Only verified that it fixes compilation via chroot. If you can, please verify the build works on stretch before merging. Thanks.


joolswills:
Confirm it builds. Can't test currently but stretch but I'm sure it's fine. If not someone can let us know. Cheers for the proper fix (my workaround was a temporary measure as I was rebuilding all packages. I'll rebuild stretch binaries now also).

PR #2980, opened 2020-02-09
Add langest/RetroStats to experimental

langest:
Adding my stats module to experimental packages.

One issue I am having is that that after installation, there is no uninstall option and it doesn't show up in the list as installed.
I have added whatever needs to happen to uninstall into the script in the remove function. How can I make this option appear?

### DIFF
--- a/scriptmodules/supplementary/retro-stats.sh
+++ b/scriptmodules/supplementary/retro-stats.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
Needs to be installed to "$md_inst" to show up as installed. However there are other issues currently also like hardcoded paths and such. Due to how busy I am currently, I won't be able to help with this until after the rpi4 release.

langest:
Great, thanks for the info.
That should be enough for me to fix it. I'll change build to install and will take care of the hardcoded paths.

langest:
It seems to install and uninstall properly now.
All hardcoded/absolute paths have been replaced by relative paths.

Please review when you feel like you have the time.

Thanks

Issue #2961, opened 2020-01-21
Review default emulators for Pi4


dankcushions:
**SNES**
~Current: [lr-snes9x-2010](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-snes9x2010.sh)
Proposed: [lr-snes9x](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-snes9x.sh)
Libretro fork of snes9x upstream. Most accurate/compatible, and runs fullspeed on pi4 (but not on a pi3). **Would suggest making the default for pi4**.~
**DONE! https://github.com/RetroPie/RetroPie-Setup/commit/31fe47492f7061924529e15c03583bf8d954817a**

**Megadrive/Genesis**
~Current: [lr-picodrive](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-picodrive.sh)
Proposed: [lr-genesis-plus-gx](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-genesis-plus-gx.sh)
Libretro fork of genesis-plus-gx upstream. Better audio, at least. Was previously agreed to be default for pi2 and up: https://github.com/RetroPie/RetroPie-Setup/issues/688 - might have been lost in a refactor.
**Would suggest making the default for pi2 and up.**~
**DONE! https://github.com/RetroPie/RetroPie-Setup/commit/c2c64b761f5264962c5d7b477e58b73a81d31f42**

**Mastersystem**
**Gamegear**
~As per the issue above, but I've not tested this yet.
**Would suggest making the default for pi2 and up.**~
**DONE! https://github.com/RetroPie/RetroPie-Setup/commit/c2c64b761f5264962c5d7b477e58b73a81d31f42**

**N64**
Currently [lr-mupen64plus-next](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-mupen64plus-next.sh) is in the experimental packages. At this point I think it's mature enough to be in main packages for pi4, with the abandoned [lr-mupen64plus](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-mupen64plus.sh) retiring to optional. lr-mupen64plus should probably stay in main for Pi3 and previous, but it does seem to work better for Pi3 + mesa.
[...]

PR #2984, opened 2020-02-10
Package moves

joolswills:
Some of these have been discussed in #2961 

Ive done it slightly differently but some of these are better defaults.

Im considering moving all n64 to optional also. Will make it easier for users to make a considered choice imho.

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -173,16 +173,14 @@ function install_mupen64plus() {
 function configure_mupen64plus() {
     local res
     local resolutions=("320x240" "640x480")
-    isPlatform "kms" && resolutions=("%XRES%x%YRES%")
+    isPlatform "kms" && res="%XRES%x%YRES%"
 
     if isPlatform "rpi"; then
-        # mesa needs to run at full screen as it doesn't benefit from our SDL scaling hint
-        if isPlatform "mesa"; then
-            res="${resolutions[0]}"
-            addEmulator 1 "${md_id}-GLideN64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM% $res 0 --set Video-GLideN64[UseNativeResolutionFactor]\=1"
+        # kms needs to run at full screen as it doesn't benefit from our SDL scaling hint
+        if isPlatform "kms"; then
+            addEmulator 0 "${md_id}-GLideN64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM% $res 0 --set Video-GLideN64[UseNativeResolutionFactor]\=1"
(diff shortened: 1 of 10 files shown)


scriptmodules/emulators/mupen64plus.sh
```diff
@@ -173,16 +173,14 @@ function install_mupen64plus() {
...
+    isPlatform "kms" && res="%XRES%x%YRES%"
 
     if isPlatform "rpi"; then
-        # mesa needs to run at full screen as it doesn't benefit from our SDL scaling hint
-        if isPlatform "mesa"; then
-            res="${resolutions[0]}"
-            addEmulator 1 "${md_id}-GLideN64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM% $res 0 --set Video-GLideN64[UseNativeResolutionFactor]\=1"
+        # kms needs to run at full screen as it doesn't benefit from our SDL scaling hint
+        if isPlatform "kms"; then
+            addEmulator 0 "${md_id}-GLideN64" "n64" "$md_inst/bin/mupen64plus.sh mupen64plus-video-GLideN64 %ROM% $res 0 --set Video-GLideN64[UseNativeResolutionFactor]\=1"
```
dankcushions: I think you want this one with `addEmulator 1` as it’s the best default for kms

joolswills:
It's disabled in favour of lr-mupen64plus-next as requested.

scriptmodules/emulators/mupen64plus.sh
joolswills: I just forgot to make that default also.

joolswills: I've done that now.

dankcushions: in https://github.com/RetroPie/RetroPie-Setup/issues/2961 i suggested mupen64plus standalone stays as default. it should be more performant/compatible as i think the libretro version isn't threaded yet, and may have some bugs that standalone does not.

scriptmodules/libretrocores/lr-snes9x.sh
```diff
@@ -41,6 +41,8 @@ function configure_lr-snes9x() {
     mkRomDir "snes"
     ensureSystemretroconfig "snes"
 
+    local def=0
+    ! isPlatform "armv6" && ! isPlatform "armv7" && def=1
     addEmulator 0 "$md_id" "snes" "$md_inst/snes9x_libretro.so"
```
dankcushions: does this want to be `addEmulator $def` rather than `addEmulator 0`?

joolswills: It does yep.

dankcushions:
thumbs up for lr-snes9x. just gave it a quick test and it seems to stand up to runahead = 1, and threaded video = disable, which are two big latency reducers.

scriptmodules/emulators/mupen64plus.sh
joolswills: I'm not convinced. I may switch it.

PR #2988, opened 2020-02-13
lr-flycast - only set video_shared_context for gl targets

joolswills:
this option was originally added before the core was enabled on arm. it may not even be needed on x11/x86 anymore

@gizmo98 do you remember why this was added. I think you added this code originally

If it's not needed on x11/GL we can remove completely


gizmo98:
It was not running without this option at that time. It can be removed if it is no longer needed.

joolswills:
I'll merge as is for now do some testing on pc later to see if it's still required.

PR #2989, opened 2020-02-13
srb2: fix building on Stretch, disable libOpenMPT

cmitu:
Since `libOpenMPT` is optional, disable it on Debian/Raspbian Stretch to prevent a build failure.
Ubuntu 18.04 includes a newer version and it builds fine with it.


joolswills:
Thanks.

PR #2990, opened 2020-02-14
lzdoom: update to 3.84 release

psyke83:
* Remove kmsdrm atexit patch (applied upstream)

---
Tested OK on buster rpi3 legacy and fkms.

OT: It seems that gzdoom (non-legacy) has recently brought back GL2/GLES support via the softpoly renderer, which allows acceleration via the SDL2 render driver. I gave it a quick test, but it runs far too slow, and requires Vulkan support to be built-in, or else the softpoly renderer will not initialize correctly. Might be worth keeping an eye on in the near future, though.


joolswills:
Thanks

PR #2992, opened 2020-02-16
reicast: change input mapping scripts and simplify launch script

cmitu:
Reicast input handling has been problematic for quite some time. It looks like it's the result of 2 commits:
 - https://github.com/reicast/reicast-emulator/commit/8e4e2c67f2f2d00da7567d779fc734c68e676562,  which removed the input configuration performed from the CLI 
- https://github.com/reicast/reicast-emulator/commit/8b5c2a3fac54bb74e8e79ff17fe0890af9044273, which added SDL input support, but also changed the way the input `.cfg` files are named.

Instead of relying on the `input:evdev_...` CLI parameters, the configuration is loaded from the `api`_`<controller_filtered_name>` files, where `<controller_filtered_name>` is produced by [`make_mapping_filename`](https://github.com/reicast/reicast-emulator/blob/11a02e571b37144d205eebb1c2974404c4b338f6/libswirl/input/gamepad_device.cpp#L203). 

The current configuration scripts (both from EmulationStation and Reicast) create the configuration files as `controller_<gamepad_name_without_spaces>`, which are not loaded by the _Reicast_'s current version.

This PR changes the names of the `.cfg` file produced by the Reicast (auto)configuration and removes the input mapping from the emulator's starting script (not needed anymore).

Credit goes to user *rejesterd*, which analyzed this in https://retropie.org.uk/forum/topic/24544.

EDIT: the auto-configuration script called from EmulationStation doesn't reliably create a correct auto-configuration file for `reicast`, however SDL doesn't export any API to get the `evdev` event codes so we can't resolve this at the moment without modifying EmulationStation and SDL.

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -150,7 +150,7 @@ function input_reicast() {
     ./reicast-joyconfig -f "$temp_file" >/dev/tty
     iniConfig " = " "" "$temp_file"
     iniGet "mapping_name"
-    local mapping_file="$configdir/dreamcast/mappings/controller_${ini_value// /}.cfg"
+    local mapping_file="$configdir/dreamcast/mappings/evdev_${ini_value//[:><?\"]/-}.cfg"
     mv "$temp_file" "$mapping_file"
     chown $user:$user "$mapping_file"
 }
--- a/scriptmodules/emulators/reicast/reicast.sh
+++ b/scriptmodules/emulators/reicast/reicast.sh
@@ -19,62 +19,6 @@ biosdir="$HOME/RetroPie/BIOS/dc"
 
 source "$rootdir/lib/inifuncs.sh"
 
-function mapInput() {
-    local js_device
-    local js_device_num
-    local ev_device
-    local ev_devices
-    local ev_device_num
-    local device_counter
-    local conf="$configdir/dreamcast/emu.cfg"
-    local params=""
-
-    # get a list of all present js device numbers and device names
-    # and device count
-    for js_device in /dev/input/js*; do
-        js_device_num=${js_device/\/dev\/input\/js/}
-        for ev_device in /dev/input/event*; do
-            ev_device_num=${ev_device/\/dev\/input\/event/}
-            if [[ -d "/sys/class/input/event${ev_device_num}/device/js${js_device_num}" ]]; then
-                file[$ev_device_num]=$(grep --exclude=*.bak -rl -m 1 "$configdir/dreamcast/mappings/" -e "= $(</sys/class/input/event${ev_device_num}/device/name)" | tail -n 1)
-                if [[ -f "${file[$ev_device_num]}" ]]; then
-                    #file[$ev_device_num]="${file[$ev_device_num]##*/}"
-                    ev_devices[$ev_device_num]=$(</sys/class/input/event${ev_device_num}/device/name)
-                    device_counter=$(($device_counter+1))
-                fi
-            fi
-        done
-    done
-
-    # emu.cfg: store up to four event devices and mapping files
-    if [[ "$device_counter" -gt "0" ]]; then
-        # reicast supports max 4 event devices
-        if [[ "$device_counter" -gt "4" ]]; then
-            device_counter="4"
-        fi
-        local counter=0
-        for ev_device_num in "${!ev_devices[@]}"; do
-            if [[ "$counter" -lt "$device_counter" ]]; then
-                counter=$(($counter+1))
-                params+="-config input:evdev_device_id_$counter=$ev_device_num "
-                params+="-config input:evdev_mapping_$counter=${file[$ev_device_num]} "
-            fi
-        done
-        while [[ "$counter" -lt "4" ]]; do
-            counter=$(($counter+1))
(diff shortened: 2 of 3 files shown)


joolswills:
Thanks!

joolswills:
Mapping seems to work great. Thanks.

Some unrelated problem to note.

Exiting (at least on rpi4) with start and select causes a crash. I will debug this. Theres a couple of minor issues I noticed but again unrelated to this. I'm happy to.merge. thanks!

joolswills:
The crash happens when exiting during a game. It's ok when hitting tab to access in the game menu.

Just mentioning here as I'm thinking maybe while what looks like a lot of changes have been happening, maybe we should revert to an older version. This was mentioned before by one of the RetroPie devs (maybe you)?

If we did would this controller change be affected? I didn't check the history yet to see what was changed when.

cmitu:
The changes in the controller handling (referenced in my commit message) were done 1 year ago and there have been quite a lot of changed until May/June when activity kind of stopped.

The mapping script from ES doesn't seem to work - there's quite a difference between what is generated by the joystick mapping script from Reicast and what is generated after ES auto-configuration.

joolswills:
It worked for my controller at least enough that I didn't notice. There's a bunch of problems outside of this - I'm inclined to move it to experimental. Lr-flycast and redream are also available now. 

Do you want me to merge this for now or hold on?

cmitu:
Let's hold on for a bit, I'll take a closer look at ES's mapping script for `reicast`.

cmitu:
I've taken a look at the `reicast.sh` auto-mapping script for EmulationStation and there are some cases which are not handled correctly anymore. 

Looking at how the buttons are mapped, it assumes the `evdev` button indexes are starting from 288L, while some gamepads have this starting at 304L (see https://gitlab.freedesktop.org/libevdev/libevdev/blob/libevdev-1.6.0/include/linux/input-event-codes.h#L364).

Here's 2 examples with the output from a `evtest` report: 
1. SNES clone (iBuffalo):
```
Input device name: "USB,2-axis 8-button gamepad  " 
 Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
    Event code 288 (BTN_TRIGGER)
    Event code 289 (BTN_THUMB)
    Event code 290 (BTN_THUMB2)
    Event code 291 (BTN_TOP)
    Event code 292 (BTN_TOP2)
    Event code 293 (BTN_PINKIE)
    Event code 294 (BTN_BASE)
    Event code 295 (BTN_BASE2)
  Event type 3 (EV_ABS)
    Event code 0 (ABS_X)
      Value    128
      Min        0
      Max      255
      Flat      15
    Event code 1 (ABS_Y)
      Value    128
      Min        0
      Max      255
      Flat      15
  Event type 4 (EV_MSC)
    Event code 4 (MSC_SCAN)
```
2. Sony Dualshock 4
```
nput device name: "Sony Interactive Entertainment Wireless Controller"
Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
    Event code 304 (BTN_SOUTH)
    Event code 305 (BTN_EAST)
    Event code 307 (BTN_NORTH)
    Event code 308 (BTN_WEST)
    Event code 310 (BTN_TL)
    Event code 311 (BTN_TR)
    Event code 312 (BTN_TL2)
    Event code 313 (BTN_TR2)
    Event code 314 (BTN_SELECT)
    Event code 315 (BTN_START)
    Event code 316 (BTN_MODE)
    Event code 317 (BTN_THUMBL)
    Event code 318 (BTN_THUMBR)
  Event type 3 (EV_ABS)
    Event code 0 (ABS_X)
      Value    130
      Min        0
      Max      255
      Fuzz      10
      Flat      15
    Event code 1 (ABS_Y)
      Value    123
      Min        0
      Max      255
      Fuzz      10
      Flat      15
    Event code 2 (ABS_Z)
      Value      0
      Min        0
      Max      255
      Fuzz      10
      Flat      15
    Event code 3 (ABS_RX)
      Value    122
      Min        0
      Max      255
      Fuzz      10
      Flat      15
    Event code 4 (ABS_RY)
      Value    128
      Min        0
      Max      255
      Fuzz      10
      Flat      15
    Event code 5 (ABS_RZ)
      Value      0
      Min        0
      Max      255
      Fuzz      10
      Flat      15
    Event code 16 (ABS_HAT0X)
      Value      0
      Min       -1
      Max        1
    Event code 17 (ABS_HAT0Y)
      Value      0
      Min       -1
      Max        1
```

For the 1st case, the script creates a correct mapping, but for the 2nd it doesn't. It also ignores the _hat_ input types, so in the 2nd case (the Sony controller) it doesn't map the D-Pad buttons at all (`btn_dpad1_left`, `btn_dpad1_right`, etc. in the `[dreamcast]` section).

I think the auto-mapping can be fixed, but it needs an `evdev` compatible call to retrieve the button scancodes correctly, similar to the mapping script from `reicast` (using the `python-evdev` package).

cmitu:
Update the status and also added the note about the limitations preventing - right now - a reliable auto-configuration method using the EmulationStation SDL scancodes or input indexes.

joolswills:
Thanks. We can revisit the mapping issues later.

PR #2993, opened 2020-02-16
lr-mupen64plus-next - temporary workaround for crash using videocore platform

hhromic:
* the core is unable to obtain pointers to the GL functions from RetroArch when using the legacy drivers.
* the issue is under investigation but for now build from the last known commit for videocore.


joolswills:
Thanks!

PR #2994, opened 2020-02-17
PSX: Use previous ari64 dynarec for ARM targets, rather than experimental lightrec one

dankcushions:
Lightrec crashes or is less performant at the moment.

Fixes https://retropie.org.uk/forum/topic/24895/pi4-lr-pcsx-rearmed-from-source-freezes-with-some-games


joolswills:
Thanks.

PR #2999, opened 2020-02-21
openbor: properly enable dispmanx

cmitu:


joolswills:
You will a launch script. See xrick for example.

cmitu:
That would work better.

joolswills:
There's logic to switch off dispmanx if launched from X. Also it can be switched off manually for videocore so it can't be hardcoded. Thanks

cmitu:
@joolswills yes, it felt like a hack, 'twas a hack. Should be better now.

joolswills:
Thanks.

PR #3000, opened 2020-02-22
Module flags

joolswills:
packages - expand platform flags to allow excluding all platforms and including certain ones

This should be backward compatible with the previous behaviour. By default all platforms are included.
Previously if you wanted a rpi+videocore only module you had to use lots of flags to exclude other platforms - eg

!x86 !mali !kms

which isn't ideal as new platforms are added.
Now this can be changed now by using a !all flag

!all videocore

which would first exclude all platforms, then enable the module just for videocore.

### DIFF
--- a/scriptmodules/emulators/gpsp.sh
+++ b/scriptmodules/emulators/gpsp.sh
@@ -14,7 +14,7 @@ rp_module_desc="GameBoy Advance emulator"
 rp_module_help="ROM Extensions: .gba .zip\n\nCopy your Game Boy Advance roms to $romdir/gba\n\nCopy the required BIOS file gba_bios.bin to $biosdir"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/gizmo98/gpsp/master/COPYING.DOC"
 rp_module_section="opt"
-rp_module_flags="noinstclean !x86 !mali !kms"
+rp_module_flags="noinstclean !all arm !mali !kms"
 
 function depends_gpsp() {
     getDepends libsdl1.2-dev libraspberrypi-dev
--- a/scriptmodules/supplementary/controlblock.sh
+++ b/scriptmodules/supplementary/controlblock.sh
@@ -14,7 +14,7 @@ rp_module_desc="ControlBlock Driver"
 rp_module_help="Please note that you need to manually enable or disable the ControlBlock Service in the Configuration section. IMPORTANT: If the service is enabled and the power switch functionality is enabled (which is the default setting) in the config file, you need to have a switch connected to the ControlBlock."
 rp_module_license="NONCOM https://raw.githubusercontent.com/petrockblog/ControlBlockService2/master/LICENSE.txt"
 rp_module_section="driver"
-rp_module_flags="noinstclean !x86 !mali"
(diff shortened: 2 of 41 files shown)


joolswills:
This needs some testing. I have changed the scope of modules in some cases also but it's also possible I've made changes that are not quite right.

scriptmodules/emulators/gpsp.sh
```diff
@@ -14,7 +14,7 @@ rp_module_desc="GameBoy Advance emulator"
 rp_module_help="ROM Extensions: .gba .zip\n\nCopy your Game Boy Advance roms to $romdir/gba\n\nCopy the required BIOS file gba_bios.bin to $biosdir"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/gizmo98/gpsp/master/COPYING.DOC"
 rp_module_section="opt"
-rp_module_flags="noinstclean !x86 !mali !kms"
+rp_module_flags="noinstclean !all arn !mali !kms"
```
cmitu: Typo here (`arn` should be `arm`)

scriptmodules/supplementary/controlblock.sh
```diff
@@ -14,7 +14,7 @@ rp_module_desc="ControlBlock Driver"
 rp_module_help="Please note that you need to manually enable or disable the ControlBlock Service in the Configuration section. IMPORTANT: If the service is enabled and the power switch functionality is enabled (which is the default setting) in the config file, you need to have a switch connected to the ControlBlock."
 rp_module_license="NONCOM https://raw.githubusercontent.com/petrockblog/ControlBlockService2/master/LICENSE.txt"
 rp_module_section="driver"
-rp_module_flags="noinstclean !x86 !mali"
+rp_module_flags="noinstclean !all rpi"
```
cmitu: `!all` should be first ?

scriptmodules/supplementary/splashscreen.sh
```diff
@@ -12,7 +12,7 @@
 rp_module_id="splashscreen"
 rp_module_desc="Configure Splashscreen"
 rp_module_section="main"
-rp_module_flags="noinstclean !x86 !osmc !xbian !mali"
+rp_module_flags="noinstclean !all rpi !osmc !xbian"
```
cmitu: `!all` should be first (2) ?

joolswills: It needs to be before the other platform flags. Should be ok?

Ideally I'd like to move the "is module supported" flags separate from the others but it may break 3rd party packages. I may reorg when I do the backend selection code.

joolswills: Unless I've misunderstood?

cmitu: No, I though it should be _first_ - it's my misunderstanding.

scriptmodules/emulators/gpsp.sh
joolswills: Thanks.

joolswills:
I'm going to merge this. If some flags are wrong I'm sure it will be spotted and we can adjust.

PR #3002, opened 2020-02-22
runcommand: override locale settings for printf

cmitu:
Reported in https://retropie.org.uk/forum/topic/24822/, affecting the X11 video modes calculation.

When a locale where `.` is not a decimal separator, `printf` throws an error when trying to round the refresh rate:
```
pi@retropie $ _number=60.02 && LC_ALL=ru_RU.UTF-8 printf '%.0f\n' ${_number::-2}
-bash: printf: 60.: invalid number
```

Overriding the `LC_NUMERIC` locale for the `printf` call seems ok.


hhromic:
Why not just remove the decimal symbol as well and print the number as a regular integer?

    _number=60.02 && printf "%d\n" ${_number::-3}

Is there something else to take into account?

PR #3003, opened 2020-02-23
lr-opera: replace lr-4do due to name and repository changes

cmitu:
4DO Libretro repository changed the location and the name of the core in https://github.com/libretro/opera-libretro/commit/8d62f9e8d3cf6c425cc1cbc488a7bc6f1f7d6a7b.

Fixed the installation with the new libretro core name.


joolswills:
Needs a rename hook to move configs etc. It's not enough to just change the name.

joolswills:
Sorry. I missed you did that

PR #3006, opened 2020-02-28
System reorg

joolswills:
Reorganisation of system splitting out some code into functions

__jobs var with a simple default based on memory and nproc
Set MAKEFLAGS -jX based on this
some memory var changes
Rework of all compiler flag logic

### DIFF
--- a/scriptmodules/admin/builder.sh
+++ b/scriptmodules/admin/builder.sh
@@ -109,7 +109,7 @@ function chroot_build_builder() {
 
             rp_callModule image chroot "$md_build/$dist" \
                 sudo \
-                MAKEFLAGS="$makeflags" \
+                __makeflags="$makeflags" \
                 DISTCC_HOSTS="$distcc_hosts" \
                 __platform="$platform" \
                 __has_binaries="$__chroot_has_binaries" \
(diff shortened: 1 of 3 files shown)


joolswills:
I'm leaning towards removing the auto vectorization flags. They can cause issues and
I have not managed to measure any performance apart from synthetic floating point benchmarks.

Better to enable them if and when is proven useful imho. The default flags should be reliable.

joolswills:
Will rebase this soon.

joolswills:
This is rebased. I'll do a few checks but unless anyone sees any issue will merge soon.

cmitu:
(not a bug) Because of `platform_all` , the unknown message platform shows `all` as one of the possibilities:

> Unknown platform - please manually set the __platform variable to one of the following: **all** armv7-mali generic-x11 imx6 odroid-c1 odroid-c2 odroid-xu rpi1 rpi2 rpi3 rpi4 tinker vero4k x86

joolswills:
Ah good spot. I'll rename the function.

joolswills:
I made a set_plaform_defaults for the default optimisation and put it before to simplify things (removes a check). Tbh it doesn't need to a function really, but I had some ideas for future defaults.

PR #3007, opened 2020-03-02
Advertise Vero4K as a Mali platform

hissingshark:
Addresses Issue #2943 more appropriately.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -441,5 +441,5 @@ function platform_vero4k() {
     __default_cflags="-I/opt/vero3/include -L/opt/vero3/lib -O2 -mcpu=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard -ftree-vectorize -funsafe-math-optimizations"
     __default_asflags=""
     __default_makeflags="-j4"
-    __platform_flags="arm armv7 neon vero4k gles"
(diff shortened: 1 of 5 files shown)


joolswills:
Thanks - Everything works ok ? there's still some vero4k logic in some modules - maybe some of those cases can be simplified also ?

scriptmodules/system.sh
```diff
@@ -441,5 +441,5 @@ function platform_vero4k() {
     __default_cflags="-I/opt/vero3/include -L/opt/vero3/lib -O2 -mcpu=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard -ftree-vectorize -funsafe-math-optimizations"
     __default_asflags=""
     __default_makeflags="-j4"
-    __platform_flags="arm armv7 neon vero4k gles"
+    __platform_flags="arm armv7 neon mali vero4k gles"
```
joolswills: You can remove vero4k from platform flags btw as it's automatically included.

hissingshark: Good point.

hissingshark:
Some of the vero4k logic is to trigger platform specific flags that were added upstream, but I'll see if there's anything else more generic I can handle in `helpers.sh`, to unclutter the individual modules.

joolswills:
Thanks

PR #3008, opened 2020-03-03
fbneo: add nes & fds subsystems

barbudreadmon:


cmitu:
Is any BIOS file needed for FDS support ? FCEUMM and Nestopia use `disksys.rom` as BIOS file (taken from the RetroArch's _system_ folder).

barbudreadmon:
Yes, fds requires the "fdsbios" romset, should contain 1 file with crc `5e607dcf`, this file is part of the No-Intro roms. Actually, most of this nes/fds driver is using No-Intro roms (we favored those over MESS romsets this time), the only exceptions should be homebrews/hacks/translations.

joolswills:
thanks.

cmitu:
I guess this will need a recent version of FBNeo, the new subsystems are not - yet - listed in https://github.com/libretro/FBNeo/blob/master/src/burner/libretro/README.md.

barbudreadmon:
Yes, the support was added only yesterday. Also, this new nes/fds driver is on the accurate side, probably the most accurate nes emulator nowaday with mesen (which is missing from your setup scripts ?). Actually there are a few known case of nes game that will only render properly on our emulator (one such case would be micro machines). The bad news is that it's barely faster than mesen, which probably makes it a bad option for pi3 (i didn't test tbh).

Issue #2998, opened 2020-02-20
removing reicast leaves behind the +START REICAST script in the roms directory


mulcahey:
When I remove reicast, the '+Start Reicast.sh' script still exists in the dreamcast roms directory. 
Edit: Forgot to include specs.  This is on an rpi4, and I updated the retropie script and uninstalled reicast about 15 minutes ago.  That installation of reicast was originally installed (from source) about 1 hour ago.

PR #3010, opened 2020-03-05
reicast: remove 'Start' link when emulator is removed

cmitu:
Should fix https://github.com/RetroPie/RetroPie-Setup/issues/2998.


joolswills:
Thanks

PR #3011, opened 2020-03-07
Unneeded vero4k/mali reference

hissingshark:
Sorry, you'd closed the previous PR before I'd finished.  Found this straggler that yields a "malimali", which obviously breaks the binary download path (we host our own now)...

The only other optimisations I can see are to move a few `isPlatform "vero4k" && depends+=(vero3-userland-dev-osmc)` lines - that don't involve any other module specific packages - to helpers.sh. 

Those are in ScummVM, Reicast and PPSSPP.  I know there are similar lines for the RPi, but it's an option anyway.  Thoughts?


joolswills:
Please use the draft mode or make it clear a PR is a wip if you don't want it merged.

hissingshark:
Apologies, hadn't seen that feature. Will do in future.

joolswills:
Thanksm

PR #3012, opened 2020-03-10
Get depends rework

joolswills:
@cmitu


joolswills:
This needs some testing. I've done some but it's not ready to merge yet.

joolswills:
Seems to be fine from my testing.

PR #3020, opened 2020-03-17
splashscreen: add a configuration file for user settings

cmitu:
This prevents resetting of the configuration when upgrading, since now the user configuration is kept in the scriptmodule itself.

Has been reported a few times in the forum - just didn't get around to fix it.

EDIT: Besides the randomization option, the splashscreen duration is now configurable (default: 12 sec, min: 10 sec, max 100 sec).

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -164,7 +164,7 @@ function choose_splashscreen() {
             options+=("$i" "$splashdir")
             ((i++))
         fi
-    done < <(find "$path" -type f ! -regex ".*/\..*" ! -regex ".*LICENSE" ! -regex ".*README.*" ! -regex ".*\.sh" | sort)
+    done < <(find "$path" -type f ! -regex ".*/\..*" ! -regex ".*LICENSE" ! -regex ".*README.*" ! -regex ".*\.sh"  ! -regex ".*\.pkg" | sort)
     if [[ "${#options[@]}" -eq 0 ]]; then
         printMsgs "dialog" "There are no splashscreens installed in $path"
         return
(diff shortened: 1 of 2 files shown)


joolswills:
Won't the separate config ini still get overwritten on update? Also maybe things like rootdir should still be hardcoded .. also regexpression? Could be easy to break but I'm generally for configurability of course.

joolswills:
Also maybe some failsafes. What if the config gets deleted etc.

cmitu:
@joolswills it shouldn't be removed on upgrades, it should at least keep the `RANDOMIZE` setting. You're right about the ROOTDIR/DATADIR fallbacks, I'll do a few tests and modify.

joolswills:
Not removed but updating should call install_bin which will reset some of it. Although the settings in question are probably best not configurable

joolswills:
Maybe best to just have randomize and also could add some splash delay / timing config in there. And avoid setting them if already set (eg for the timing).

cmitu:
@joolswills left only the `RANDOMIZE` option as configurable and fixed the permissions on the `.cfg` file. I'll look into a configurable timing, don't merge yet.

cmitu:
I made the splashscreen duration configurable. 
The new `dialog` configuration screen (using `--rangebox`) can use D-Pad Up/Down to change the duration value.

scriptmodules/supplementary/splashscreen.sh
```diff
@@ -183,7 +183,9 @@ function randomize_splashscreen() {
     )
     local cmd=(dialog --backtitle "$__backtitle" --menu "Choose an option." 22 86 16)
     local choice=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
-    iniConfig "=" '"' "$md_inst/asplashscreen.sh"
+    iniConfig "=" '"' "$configdir/all/$md_id.cfg"
+    chown $user:$user "$configdir/all/md_id.cfg"
```
joolswills: Missing a `$`

joolswills:
One minor typo. Once that's sorted are you happy for me to merge?

cmitu:
OK, I fixed the typo - ok to merge.

joolswills:
Thanks.

cmitu:
@mth75 Yes, it looks like a configuration is not created on update/upgrade. Thank you for reporting.

cmitu:
@mth75 should be fixed now, try updating your RetroPie-Setup script.

mth75:
@cmitu, confirmed! It works now. Thx!

PR #3023, opened 2020-03-17
packages: fix spelling error

cmitu:
Small spelling error - should be `md_id`. Noticed it during source installs, because the `build` phase tries to write the `.pkg` info file without the install folder being present.


joolswills:
Thanks.

PR #3026, opened 2020-03-18
stats: modify the pkgflags stats logic after the flags reorganization

cmitu:
After https://github.com/RetroPie/RetroPie-Setup/pull/3000, the logic for calculating the Video category for a package has changed with the introduction of the `!all` flag. The modification updates the `pkgflags` page to better match a package to the corresponding video section (Mali/DRM&KMD/VideoCore/X11).

I also added a title (which should be visible as a tool tip on a desktop browser) on each package row entry to show the script flags, just to make it easier to spot errors.


joolswills:
Awesome. Thanks!

PR #3029, opened 2020-03-21
lr-hatari: fix linkage on Ubuntu 18.04 & friends (gcc7)

cmitu:
(reported in https://retropie.org.uk/forum/topic/25181/) `lr-hatari` fails building with a linker error on gcc 7.x, due to not finding the `capsimage` functions.
Re-arranged the linking order so `libcapsimage.so.5` is picked up last and removed an extra include in the module `.diff`.


joolswills:
Thanks.

PR #3031, opened 2020-03-23
Packaging / setup tweaks

joolswills:
setup - show install origin in manage package sections for overview
 * useful to get an overview - used bold markup but shows as grey from framebuffer - but is easier to read when differently styled imho

Wanted to get an opinion on this before I commit.

Also added a dependency section to setup also as now packages remember how they were installed it is useful to be able to switch between binary / source.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -319,7 +319,7 @@ function section_gui_setup() {
 
         if [[ "$num_pkgs" -gt 0 ]]; then
             options+=(
-                U "Update all ${__sections[$section]} packages" "This will update any installed ${__sections[$section]} packages. The packages will be updated by the method used previously."
+                U "Update all installed ${__sections[$section]} packages" "This will update any installed ${__sections[$section]} packages. The packages will be updated by the method used previously."
             )
         fi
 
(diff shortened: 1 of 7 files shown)


joolswills:
Added the dependency section

scriptmodules/admin/setup.sh
```diff
@@ -262,7 +262,15 @@ function package_setup() {
                 ;;
             X)
                 local text="Are you sure you want to remove $md_id?"
-                [[ "${__mod_section[$idx]}" == "core" ]] && text+="\n\nWARNING - core packages are needed for RetroPie to function!"
+                case "${__mod_section[$idx]}" in
+                    core)
+                        text+="\n\nWARNING - core packages are needed for RetroPie to function!"
+                        ;;
+                    depends)
+                        text+="\n\nWARNING - this package is required by other retropie packages - removing may cause other packages to fail."
```
cmitu: maybe capitalize `retropie` (RetroPie).

joolswills: Yeah. Sloppy text again from me.

Apart from this are you ok with it as I'll merge if so.

cmitu: yes, it looks fine.

PR #3034, opened 2020-03-31
Update esthemes.sh

zanac:
Added a very cool theme made for 240p for old CRT Arcade Monitor

### DIFF
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -235,7 +235,7 @@ function gui_esthemes() {
         'arcadeforge pixel-heaven-text'
         'arcadeforge 240p_Bubblegum'
         'arcadeforge 240p-honey'
-        'AndreaMav' 'es-theme-arcade-crt'
+        'AndreaMav' 'arcade-crt'
         'dionmunk clean'
         'c64-dev epicnoir'        
     )
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -235,6 +235,7 @@ function gui_esthemes() {
         'arcadeforge pixel-heaven-text'
         'arcadeforge 240p_Bubblegum'
         'arcadeforge 240p-honey'
+        'AndreaMav' 'es-theme-arcade-crt'
     )
     while true; do
         local theme
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -236,6 +236,8 @@ function gui_esthemes() {
         'arcadeforge 240p_Bubblegum'
         'arcadeforge 240p-honey'
         'AndreaMav' 'es-theme-arcade-crt'
+        'dionmunk clean'
+        'c64-dev epicnoir'        
     )
     while true; do
         local theme


joolswills:
Did you test this? I don't think it will work as it shouldn't include the es-theme prefix. I would prefer permission from the author before including also.

zanac:
Yes, i tested it... it work very well!!! I just wrote to the author and he will write the permission here soon.....

AndreaMav:
Hi guys! I'm the author of the theme... of course i will grant the full permission to publish it on retropie: it's an honour for me!
thanks in advance!

scriptmodules/supplementary/esthemes.sh
```diff
@@ -236,6 +236,8 @@ function gui_esthemes() {
         'arcadeforge 240p_Bubblegum'
         'arcadeforge 240p-honey'
         'AndreaMav' 'es-theme-arcade-crt'
+        'dionmunk clean'
```
zanac, line 239: ok

scriptmodules/supplementary/esthemes.sh
```diff
@@ -236,6 +236,8 @@ function gui_esthemes() {
...
+        'c64-dev epicnoir'
```
zanac, line 240: ok

zanac:
If you adjust the pull i will be glad ;)

hhromic:
@zanac are you 100% sure this works?

As @joolswills mentioned the es-theme prefix should not be there:

    es-theme-arcade-crt  -> arcade-crt

AndreaMav:
should i change the name of the theme on the git from es-theme-arcade-crt to >> arcade-crt?

AndreaMav:
> @zanac are you 100% sure this works?
> 
> As @joolswills mentioned the es-theme prefix should not be there:
> 
> ```
> es-theme-arcade-crt  -> arcade-crt
> ```

i changed git in 
https://github.com/AndreaMav/arcade-crt

hhromic:
> should i change the name of the theme on the git from es-theme-arcade-crt to >> arcade-crt?

No you should not. Your repository should be named `es-theme-ABCD` and the entry in the scriptmodule should be `ABCD`. RetroPie [automatically adds the `es-theme-` prefix](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/supplementary/esthemes.sh#L35) under the hood to clone the theme. This is done to not pollute the menu with repetitive text.

Cheers and thanks for the contribution!

scriptmodules/supplementary/esthemes.sh
```diff
@@ -235,6 +235,9 @@ function gui_esthemes() {
         'arcadeforge pixel-heaven-text'
         'arcadeforge 240p_Bubblegum'
         'arcadeforge 240p-honey'
+        'AndreaMav' 'arcade-crt'
```
hhromic, line 238: You should not separate the strings with single quotes in the middle. Pay attention to how is done in the other entries. You quote the entiry string separated by space like this:

    'AndreaMav arcade-crt`

Otherwise the scriptmodule breaks. You are not testing these changes before submitting aren't you?

zanac: Sorry, i confirm... i just tested the theme, not the script :)

AndreaMav: thank you very much!

zanac:
@joolswills what you suggest? If you prefer i cancel the pull and repeat it in the right way..... we are not so great github users ;)

zanac:
i will send a new one

PR #3037, opened 2020-04-07
Added Supermodel emu, EasyRPG-Player, optional lr-* cores and updates

Mictjs:
Tested and working on Ubuntu 18.04


joolswills:
Many thanks for all this. We will need them submitted with one PR per core though for reviewing as there's too much to manage here in one PR. Some cores we exclude intentionally also but it will be easier if they are split up to get them included.

Thanks again.

Mictjs:
Ok. I will send them divided.

PR #3056, opened 2020-04-09
Add lr-fbalpha2012-cps1

Mictjs:
Tested and working on Ubuntu 18.04

### DIFF
--- a/scriptmodules/libretrocores/lr-fbalpha2012-cps1.sh
+++ b/scriptmodules/libretrocores/lr-fbalpha2012-cps1.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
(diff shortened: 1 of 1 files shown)


dankcushions:
appreciate the effort! but (sorry) what's the justification for adding this? what advantage does it bring to retropie? better performance? more compatibility? i don't think either of these are true for this particular emulator, and potentially others you've submitted.

as far as i know the two fba cps-only emulators exist purely for very low memory systems (eg wii) that struggle with the entire (old) FBA binary in memory. we don't support any such systems.

i don't think we should add cores for every libretro core/emulator, just because we can. it's extra maintenance work and it's clutter in the package manager. please could you include a use-case for each submitted emulator? thanks :)

yerebikko:
"...lorsque l'arrogance est supérieure à la connaissance du projet lui-même."

Fonctionne sur ubuntu comme indiqué ci-dessus

PR #3102, opened 2020-04-11
joy2key - convert to python3 and use python3-pyudev dependency

joolswills:
* Ubuntu 20.04 will no longer include python-pyudev
 * Use stderr for /dev/tty opening error msg
 * Use python3 in runcommand module for compilation
 * Open with 'append' which was added in f614804 fails with python3 on rpi4 console. Use lower level os.open and O_WRONLY to avoid possible buffer issues. Could not reproduce the ghost inputs the append was added for


joolswills:
This works on my Ubuntu 20.04 system and my rpi4 with buster. Also runs on stretch (but only tested in chroot) but no dependency issue with the python3 package or running the code. Going to merge.

If the ghost input issue that @psyke83 noticed arises I may need to revisit.

PR #3103, opened 2020-04-14
dosbox - allows for larger C drive

nemo93:
2 very small changes to improve a bit the Dosbox experience.

* larger C drive is required to install games like Fallout, M.A.X, KKND, etc. I went with 1GB to be absolutely sure but 768GB should be fine too.

* relying on the elegant method described [in the doc](http://dosonthepi.blogspot.com/2015/02/dosbox-configuration-for-individual.html) - which has you populate the **[autoexec]** section at the bottom of the .conf file (one .conf file = one game) - works great with Retropie. This allows for a single extension type (.conf) to be registered. Yet Dosbox default config file is also called anytime and along the user's own config file for the game. This comes with side effects like Mount points declared twice, sound card parameters mess and even preventing games to run. I've fixed that behavior by removing the **[-userconf](https://www.dosbox.com/DOSBoxManual.html#Parameters)** flag from the launch script (+Start....sh).

Note: it is my very first attempt at issuing a PR for Retropie (more to come for Dosbox as detailed in the [forum](https://retropie.org.uk/forum/topic/25041/dosbox-a-thread)). Therefore let me know what's wrong and I'll follow your guidance. Thanks for your time.

### DIFF
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -104,7 +104,7 @@ function midi_synth() {
 
 params=("\$@")
 if [[ -z "\${params[0]}" ]]; then
-    params=(-c "@MOUNT C $romdir/pc" -c "@C:")
+    params=(-c "@MOUNT C $romdir/pc -freesize 1024" -c "@C:")
 elif [[ "\${params[0]}" == *.sh ]]; then
     midi_synth start
     bash "\${params[@]}"
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -111,7 +111,7 @@ elif [[ "\${params[0]}" == *.sh ]]; then
     midi_synth stop
     exit
 elif [[ "\${params[0]}" == *.conf ]]; then
-    params=(-conf "\${params[@]}")
+    params=(-userconf -conf "\${params[@]}")
 else
     params+=(-exit)
 fi
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -111,7 +111,7 @@ elif [[ "\${params[0]}" == *.sh ]]; then
     midi_synth stop
     exit
 elif [[ "\${params[0]}" == *.conf ]]; then
-    params=(-userconf -conf "\${params[@]}")
+    params=(-conf "\${params[@]}")
 else
     params+=(-exit)
 fi


joolswills:
Thanks. My only worry regarding is if the user config breaks existing set ups? Eg. people with some options set but relying on master config file? I can't remember too well as not used dosbox for a while though.

nemo93:
According to the [doc](https://www.dosbox.com/DOSBoxManual.html#ConfigFile): DOSBox will load configuration files that are specified with -conf. If none were specified, it will try to load "dosbox.conf" from the local directory. If there is none, DOSBox will load the user configuration file. This file will be created if it doesn't exist.

Therefore the -userconf flag is redundant. Provided there's at least one dosbox.conf provided along with Retropie which is the case anyway. And by letting it you'll face the issue described in the PR in case you're trying to tidy your setup.

I'd really recommend users to go with one conf per game as it's less confusing and ease the debug process if needed. Also lot cleaner and less prone to errors (than with bash script).

cmitu:
I don't think the `-userconf` should be removed.

> Therefore the -userconf flag is redundant. Retropie comes along with a default dosbox.conf anyway. And by letting it you'll face the issue described in the PR in case you're trying to tidy your setup.

It's not redundant. If you don't specify `-userconf`, then the default configuration (which is produced and maintained by RetroPie installer) is not loaded when you use `-conf`.

From https://www.dosbox.com/wiki/Usage:

> **-userconf[1]**
Load the configuration from the user's profile or home directory. This is the default behavior, so this switch is useful when using `-conf`, below.
**-conf configfilelocation**
Load the config file specified in configfilelocation. Useful for specifying particular options for specific games. If used after -userconf, or if you use multiple -conf switches, options present in multiple configs will be overwritten by the last, except for autoexec entries, which will be appended in order.

> I'd really recommend users to go with one conf per game as it's less confusing and ease the debug process if needed. Also lot cleaner and less prone to errors (than with bash script).

That means each game's `.conf` will have to contain a full DosBox config - whereas having a per-game `.conf` file will allow general settings to be set once (via `-userconf`) and only game specific options reside in game's `.conf` file. 

I agree that using a `.conf` file is preferable to using a Bash script.

nemo93:
Thanks @cmitu. In that case would you like me to rework this PR to exclude the -userconf? I'm going to make a note somewhere in the wiki in case user faces issue I described in the PR above. 

I need to double check but the issue with the mount point might only occur whenever you'd like to run DOSbox' command-line (ie. to install a game). I do have a separate dosbox.conf stored under roms/pc along with other .conf to launch games. This allows me to run command-line from ES or Pegasus. I'm using a setup where all games are stored under a "games" folder under roms/pc (I also got a "utils" and "win311" dirs). The issue could be with the "+Start Dosbox" shell script which mounts roms/pc to C:. I understand it's a specific/non-standard setup.

cmitu:
> Thanks @cmitu. In that case would you like me to rework this PR to exclude the -userconf? I'm going to make a note somewhere in the wiki in case user faces issue I described in the PR above.

I think that would be the preferable, yes.

nemo93:
Done! Sorry for that.

joolswills:
Thanks. I squashed and merged it - changing the commit description a bit.

PR #3105, opened 2020-04-19
love: updated license and download links, updated 'mari0' game

cmitu:
* updated the license, since it's actually ZLIB and not GPL(3).
* the project has moved the source and development to Github (see https://love2d.org/wiki/Building_L%C3%96VE). 
* adjusted the dependencies to remove `mercurial` (since we're using `git` to download the sources) and `devil`, not used since [2014](https://github.com/love2d/love/commit/eca8c17a693e85635b043353e431dcb9d5012e75).
* updated the included Mari0 game to 1.6.2, compatible with the 11.2.


joolswills:
Thanks. I think it's worth mentioning all the changes in the commit also such as the mari0 game update.

cmitu:
@joolswills - ok, expanded the commit info, should cover all the info added in the PR.

PR #3112, opened 2020-04-26
retroarch - setting `video_smooth` is no longer necessary

hhromic:
* `video_smooth` defaults to `false` since RetroArch 1.8.5

Ref: https://github.com/libretro/RetroArch/pull/9710

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -157,7 +157,6 @@ function configure_retroarch() {
     iniSet "system_directory" "$biosdir"
     iniSet "config_save_on_exit" "false"
     iniSet "video_aspect_ratio_auto" "true"
-    iniSet "video_smooth" "false"
     iniSet "rgui_show_start_screen" "false"
     iniSet "rgui_browser_directory" "$romdir"
 


dankcushions:
i don't reaaaally like these kind of changes - we can't guarantee projects (especially libretro!) will keep the same defaults, so i think if we have gone to the effort of setting the ones we want (especially this one which had quite a bit of discussion!), they should stay - there's no harm!

that said, since the change came from a project leader, i think it's 100% safe in this case, just maybe not in others :)

hhromic:
That's understandable :)
If you prefer to not change this, you can simply close this PR. No hard feelings!
I just wanted to clean-up a bit the scriptmodule of redundant settings.

Btw, we should consider disabling the bilinear filtering for the menu as well. It currently looks quite blurry on modern high-def displays (Settings -> User Interface -> Appearance -> Menu Linear Filter).

joolswills:
I was going to merge this, but @dankcushions has a point that it could be changed. Maybe it's best to leave it. I'm kinda 50/50 on this one.

dankcushions:
i'm fine with merging on this occasion, just not making a habit of these kind of changes. if libretro change it (which i think is highly unlikely in this case), we will simply exact brutal revenge on @hhromic 🗡️ 

(or just get him to fix it 😄 )

joolswills:
ok. Merging :-)

PR #3114, opened 2020-04-28
New themes

zanac:


AndreaMav:
I tried to reload all the files today .. I hope now, using an app, that everything is ok ... the theme has been tested as per wiki present in the git, both with raspberry pi4 and with raspberry pi3.

cmitu:
@AndreaMav this PR is still not correct - has multiple commits that should be squashed into one and it includes extra lines (hence the conflict). I can create a simpler one and and you can close this request.

joolswills:
Theme is now included.

PR #3116, opened 2020-04-28
srb2: update to 2.2.2 and add extra asset file

cmitu:
Upstream project added an extra `.pk3` asset file in 2.2.2, we need to include it in the scriptmodule and added to the hosted assets archive.

Since the assets and releases are somehow dependent (the missing asset is a `patch.pk3`), I've locked the version to 2.2.2. We'll need to check bumping the version to include the extra/new assets.

The new asset files are in `assets/installer`, can be downloaded http://ppa.launchpad.net/stjr/srb2/ubuntu/pool/main/s/srb2-data/srb2-data_2.2.2-20200223174531.tar.xz and added to our current assets archive. It should be enough to fix the installation from binary right now and future source installs.

### DIFF
--- a/scriptmodules/ports/srb2.sh
+++ b/scriptmodules/ports/srb2.sh
@@ -21,7 +21,7 @@ function depends_srb2() {
 }
 
 function sources_srb2() {
-    gitPullOrClone "$md_build" https://github.com/STJr/SRB2.git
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Assets updated.

cmitu:
Actually, I was wrong about the binary installation - I don't know why, I though assets would be downloaded during install, but they're of course taken from the binary.
Source installation works fine.

PR #3121, opened 2020-05-01
drastic: workaround faulty input handling in a KMS context

cmitu:
_DraStic_ has input problems when running in a KMS context, reported a few times in the forums.
There is a workaround by running it a X11 session, with a minimal window manager (similar to Steam Link). 

Tested on a PI3 (Stretch) and PI4 (Buster), I don't know if the bug is present on Pi3 (Buster) with the legacy BRCM drivers

### DIFF
--- a/scriptmodules/emulators/drastic.sh
+++ b/scriptmodules/emulators/drastic.sh
@@ -17,7 +17,12 @@ rp_module_section="exp"
 rp_module_flags="!all arm !armv6 !mali"
 
 function depends_drastic() {
-    getDepends libasound2-dev libsdl2-dev zlib1g-dev
+    local depends=(libasound2-dev libsdl2-dev zlib1g-dev)
+    if isPlatform "kms" && ! isPlatform "x11"; then
+        depends+=(matchbox-window-manager xorg xserver-xorg-input-all)
+    fi
+
+    getDepends ${depends[@]}
 }
 
 function __binary_url_drastic() {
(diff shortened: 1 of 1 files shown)


psyke83:
Have you tried checking if `SDL_HINT_NO_SIGNAL_HANDLERS` or `SDL_HINT_VIDEO_DOUBLE_BUFFER` helps with this issue?

cmitu:
@psyke83 I tried a few hints, but the ones you mentioned, but it doesn't seem to make a difference.

joolswills:
Just to note drastic also crashes on exit here (launching with no parameters).

I'm not that keen on such a drastic workaround :-) but will consider options.

cmitu:
We don't need to merge it, we can look further for a solution.

What I notice it's different on X11 than on KMS/DRM is that my touchpad gets activated (the DS has a touchpad) on X11. On a RPI 3 system, where the RPI driver is used, the issue doesn't occur and the left analog joystick acts as a mouse.

psyke83:
I haven't looked at drastic yet, but in case it helps, the issue that affected steamlink and DualShock controllers was caused by a systemd bug: https://github.com/systemd/systemd/issues/5703

I'm not 100% sure if the buster version includes that fix, but I imagine it should?

cmitu:
@psyke83 if any, this bug would break the gamepad appearing correctly under X11, but I can check via `udevadm` what's reported for the controller. SDL2 seems to get the inputs alright (tested via the SDL_EVENT_LOGGING hint), but what happens inside `drastic` with the events it's probably the issue.

cmitu:
I checked, but the `udev`/`systemd` but is not present, instead the `ID_INPUT_JOYSTICK` is set correctly.

Bluebotlaboratories:
Is there a way for me to install this version of DraStic?

joolswills:
As we didn't find a solution I think your changes are the best bet @cmitu

Merging. Thanks!

PR #3123, opened 2020-05-02
system.sh: arithmetic expansion smallfix

meleu:
Noticed some messages right after calling retropie_setup.sh and before the dialog:
```
/home/meleu/RetroPie-Setup/scriptmodules/system.sh: linha 49: "8078788" / 1024 : erro de sintaxe: operando esperado (token com erro é ""8078788" / 1024 ")
```
It happens because the bash version on the computer I'm running it is 4.3, and it doesn't accept quotes inside the arithmetic expansion `$((...))` (it was [fixed in bash-4.4](http://git.savannah.gnu.org/cgit/bash.git/tree/CHANGES?h=bash-4.4-testing#n3433))

For consistency, I also removed the `$` prefix for variables inside the arithmetic expansion (same style used on line 56).

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -46,7 +46,7 @@ function test_chroot() {
 
 function conf_memory_vars() {
     __memory_total_kb=$(awk '/^MemTotal:/{print $2}' /proc/meminfo)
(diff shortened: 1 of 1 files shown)


joolswills:
thanks.

PR #3124, opened 2020-05-02
N64: Disabled hybrid upscaling filter (causes slowdown on RPI4)

dankcushions:
Currently we're getting slowdowns on mupen64plus-GLideN64 with default settings. See https://retropie.org.uk/forum/topic/25112/getting-the-best-n64-experience-on-a-pi-4/44


joolswills:
Thanks.

PR #3125, opened 2020-05-03
Sixaxis fixes for bare installation

psyke83:
Mark "libbluetooth3" for installation in sixaxis module (with service stack reload if necessary).

Just a thought, but perhaps we could consider adding "sixaxis" to the default package list for image builds? The package shouldn't really cause any issues for users that will never use a DualShock, and the messy third-party driver is split off (and not installed by default until the user selects via the gui option of the driver).

I would also like to rip out the "custombluez" module completely, but perhaps we can leave that until stretch support is fully dropped.

### DIFF
--- a/scriptmodules/supplementary/customhidsony.sh
+++ b/scriptmodules/supplementary/customhidsony.sh
@@ -10,7 +10,7 @@
 #
 
 rp_module_id="customhidsony"
-rp_module_desc="Custom hid-sony driver backported from kernel 4.15"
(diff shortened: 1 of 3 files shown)


joolswills:
Regarding the update all - I believe it was break to abort on an error. So adding continue may not be the right fix as it won't abort on an error. The logic was broken before though.

psyke83:
@joolswills 

OK, I've backed out the "update all" change from this PR and will leave you to solve that to your satisfaction.

joolswills:
Thanks.

PR #3126, opened 2020-05-03
lr-vice - added additional targets to lr-vice

joolswills:
* added x64sc, x128, xpet, xplus4 and xvic on top of the default x64

### DIFF
--- a/scriptmodules/libretrocores/lr-vice.sh
+++ b/scriptmodules/libretrocores/lr-vice.sh
@@ -10,37 +10,59 @@
 #
 
 rp_module_id="lr-vice"
-rp_module_desc="C64 emulator - port of VICE for libretro"
-rp_module_help="ROM Extensions: .cmd .crt .d64 .d71 .d80 .d81 .g64 .m3u .prg .t64 .tap .x64 .zip .vsf\n\nCopy your Commodore 64 games to $romdir/c64"
+rp_module_desc="C64 / C128 / PET / Plus4 / VIC20 emulator - port of VICE for libretro"
+rp_module_help="ROM Extensions: .cmd .crt .d64 .d71 .d80 .d81 .g64 .m3u .prg .t64 .tap .x64 .zip .vsf\n\nCopy your games to $romdir/c64"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/libretro/vice-libretro/master/vice/COPYING"
 rp_module_repo="git https://github.com/libretro/vice-libretro.git master"
 rp_module_section="opt"
 rp_module_flags=""
+
+function _get_targets_lr-vice() {
+    echo x64 x64dtv x64sc x128 xpet xplus4 xvic
+}
 
 function sources_lr-vice() {
     gitPullOrClone
 }
 
 function build_lr-vice() {
-    make clean
-    make
-    md_ret_require="$md_build/vice_x64_libretro.so"
+    mkdir -p "$md_build/cores"
+    local target
+    for target in $(_get_targets_lr-vice); do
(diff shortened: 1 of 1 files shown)


joolswills:
Did this the other day and forgot it wasn't quite complete. Need to fix up install still.

joolswills:
Probably should have done more testing. Unfortunately the libretro makefiles are not smart enough to work without cleaning in-between, so this may be best done as separate modules that share code. Or just include maybe xvic and plus4. Not sure a libretro core for c128 or PET is that useful Vs the standalone emulator.

joolswills:
Building xvic and xplus4 cores seems a reasonable compromise so I've gone with that for now vs having additional modules (which would have the overhead of downloading the source each time also).

joolswills:
Ive changed my mind and I think as it requires a full build for each system, they are best implemented as separate modules, but which can share code with lr-vice. I'll revisit.

joolswills:
I've looked into this again, and I don't like the duplication of module info etc with having lots of modules. Maybe I could implement some kind of sub-module function so a single scriptmodule can include additional modules whilst sharing the repository information etc.

I

joolswills:
Added additional vice targets to the module.

scriptmodules/libretrocores/lr-vice.sh
```diff
@@ -10,37 +10,57 @@
...
-    addSystem "c64"
+    local target
+    local name
+    for target in $(_get_targets_lr-vice); do
+        if [[ "$target" == "x64" ]]; then
+            name=""
+        else
+            name="-${target}"
+        fi
+        addEmulator 1 "$md_id${name}" "c64" "$md_inst/vice_${target}_libretro.so"
```
cmitu, line 62: Maybe move this into the `if` and set `x64` as default emulator
```` sh
if [[ "$target" == "x64" ]]; then
         addEmulator 1 "$md_id$" "c64" "$md_inst/vice_${target}_libretro.so"
 else
        addEmulator 0 "$md_id-${target}" "c64" "$md_inst/vice_${target}_libretro.so"
fi
````

File: scriptmodules/libretrocores/lr-vice.sh [hunk 2]
```diff
@@ -10,37 +10,57 @@
...
 }
 
 function build_lr-vice() {
-    make clean
-    make
-    md_ret_require="$md_build/vice_x64_libretro.so"
+    mkdir -p "$md_build/cores"
+    local target
+    for target in $(_get_targets_lr-vice); do
+        make clean
```
Gemba, line 32: Maybe it is advisable to put `make clean` outside the loop? Cuts compile time by roughly 2/3.

joolswills: I had that originally and it produced issues with the resulting cores. Maybe upstream libretro makefile has been improved, but feel free to test. Upstream mentions needing to clean before also afair.

joolswills: Actually just see my comments on the PR as I've already mentioned it.

joolswills: Here: https://github.com/RetroPie/RetroPie-Setup/pull/3126#issuecomment-623208782

But if anything has changed I can retest. Cheers.

Gemba: Compiled ok on my system but it seems sth is still lurking in the resulting builds: In their [README](https://github.com/libretro/vice-libretro/blame/master/README.md#L195) they note to run `make clean` before each building of an emulator.
So better keep this scriptmodule keep as-is.

PR #3127, opened 2020-05-04
added xpadneo - linux driver for xbox one wireless gamepads

joolswills:
Threw this together. I've seen this driver mentioned on the forums a few times. Untested.

### DIFF
--- a/scriptmodules/supplementary/xpadneo.sh
+++ b/scriptmodules/supplementary/xpadneo.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


cmitu:
Looks better than my initial PR.
 One thing you could take from is the module versioning, based on the repo `VERSION` [file](https://github.com/atar-axis/xpadneo/blob/master/VERSION). The file should be copied to `$md_inst` before build.

joolswills:
Will do. thanks. (Sorry for late reply - I meant to reply then got sidetracked).

joolswills:
Sorry for delay. Tweaked it to use the VERSION file. Maybe your description etc is better I will check that also.

Removed noinstclean and added nobin.

Checkout is done outside of repo and rsynced so it's safe to clean md_inst. But nobin is probably a good flag even though I don't build driver bins. Other drivers should probably have this too.

PR #3129, opened 2020-05-05
add lr-mesen: Mesen NES/Famicom emulator, libretro core

cmitu:
Runs ok on the PI4, but is too slow for a stock Pi 3B. I enabled it for the `xu4` , I don't think it would run well on a Tinkboard.

The upstream maintains the libretro port and is quite active.

### DIFF
--- a/scriptmodules/libretrocores/lr-mesen.sh
+++ b/scriptmodules/libretrocores/lr-mesen.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-mesen"
+rp_module_desc="High-accuracy NES and Famicom emulator"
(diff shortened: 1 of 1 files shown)


joolswills:
I don't like including specific targets in this case. Better to just exclude maybe armv6. Users can decide if they want to run it and performance can be documented.

joolswills:
The x86 flag should be used over x11 when targeting cpu btw.

cmitu:
Thanks, I updated the platform support.

scriptmodules/libretrocores/lr-mesen.sh
```diff
@@ -0,0 +1,43 @@
...
+rp_module_licence="GPL3 https://raw.githubusercontent.com/sourmesen/mesen/master/LICENSE"
+rp_module_section="exp"
+rp_module_flags="!armv6"
+
+function sources_lr-mesen() {
+    gitPullOrClone "$md_build" https://github.com/SourMesen/Mesen.git
+}
+
+function build_lr-mesen() {
+    make -C Libretro
```
joolswills, line 24: Should have a make clean also (for now - I may change this for all modules in the future).

scriptmodules/libretrocores/lr-mesen.sh
```diff
@@ -0,0 +1,43 @@
...
+
+function install_lr-mesen() {
+    md_ret_files=(
+        'Libretro/mesen_libretro.so'
+        'LICENSE'
+        'README.md'
+    )
+}
+
+function configure_lr-mesen() {
```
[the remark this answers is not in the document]
cmitu: shoule be fine now.

scriptmodules/libretrocores/lr-mesen.sh
cmitu: done

joolswills:
Thanks.

PR #3130, opened 2020-05-05
esthemes: new theme from @AndreaMav

cmitu:
Should replace #3114. 
Theme added with author's permission (https://retropie.org.uk/forum/topic/25313/).


joolswills:
thanks.

PR #3131, opened 2020-05-07
solarus - fix missing "false" case in wrapper script

hhromic:
* when not in videocore, the script fails due to missing "false" case

Shameful bug from my previous PR :disappointed:

Tested by a user running RetroPie on RPI4.


cmitu:
If the intent is to use `LD_PRELOAD` when running on `videocore` enabled platforms, wouldn't it be easier to just use:
```
if isPlatform "videocore"; then 
    export LD_PRELOAD="/opt/vc/lib/libbrcmGLESv2.so"
fi
```

The `depends` step already pulls `libraspberrypi-dev` (for `videocore`), which pulls `libraspberrypi0`, the package that actually contains `/opt/vc/lib/libbrcmGLESv2.so`.

hhromic:
The `isPlatform` function does not exist in the context of the wrapper script (runs outside of RetroPie's env). That's why I'm proxying its result via the `true`/`false` built-ins.

I agree that the check for the existence of the lib is technically not required, I just added it to be safe.

joolswills:
Thanks.

PR #3133, opened 2020-05-09
reicast - use python3 for reicast-joyconfig

joolswills:
I did a quick test and the reicast-joyconfig seemed to work with python3. A deprecated warning but I only did a quick test.

We may need to revisit using the other upstream branch - but last I checked some of the code paths were not wired in. But something to look into.

### DIFF
--- a/scriptmodules/emulators/reicast.sh
+++ b/scriptmodules/emulators/reicast.sh
@@ -17,12 +17,12 @@ rp_module_section="opt"
 rp_module_flags="!armv6"
 
 function depends_reicast() {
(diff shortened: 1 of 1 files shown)


cmitu:
Running this on a Buster system, I get a warning from the `evdev` module in several places, clobbering the output:
```
DeprecationWarning: Please use InputDevice.path instead of InputDevice.fn
  print("File: %s" % dev.fn)
```
Looks like it was added as part of `evdev` 1.0.0 (Jun 02, 2018), so it should be safe to update the `.py` script to use `InputDevice.path`.

joolswills:
Will merge. We can fix the deprecated msg later.

PR #3135, opened 2020-05-10
Install libnss3 with Kodi to support Widevine DRM

mediaminister:
The package libnss3 is needed to play Widevine DRM content in Kodi.

This fixes the following problem:
https://github.com/asciidisco/plugin.video.netflix/issues/489


joolswills:
I prefer to leave this to users to install I think. Sorry. Would need a rebase anyway.

PR #3136, opened 2020-05-12
retroarch - update to v1.8.6 and rebase/rework patches

joolswills:
Tested on RPI3 Buster (Thanks @hhromic), and PC with Ubuntu 20.04

Will test on RPI4 and merge if all good (and update assets)


joolswills:
Tested working on rpi4.

PR #3140, opened 2020-05-15
Add rpi platform to lr-mame2010

pjft:
Should address https://github.com/libretro/mame2010-libretro/issues/130
and be a (hopefully working) alternative to
https://github.com/RetroPie/RetroPie-Setup/pull/2773

### DIFF
--- a/scriptmodules/libretrocores/lr-mame2010.sh
+++ b/scriptmodules/libretrocores/lr-mame2010.sh
@@ -23,7 +23,7 @@ function build_lr-mame2010() {
     rpSwap on 750
     make clean
     local params=()
-    isPlatform "arm" && params+=("VRENDER=soft" "ARM_ENABLED=1")
+    isPlatform "arm" && params+=("VRENDER=soft" "ARM_ENABLED=1" "FORCE_DRC_C_BACKEND=1")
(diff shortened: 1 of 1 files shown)


joolswills:
Why "rpi" - isn't this a CPU issue ?

pjft:
Ah - your suggestion is effectively to add it to the previous "arm" check, is that it? That makes more sense indeed.

scriptmodules/libretrocores/lr-mame2010.sh
```diff
@@ -23,7 +23,7 @@ function build_lr-mame2010() {
     rpSwap on 750
     make clean
     local params=()
-    isPlatform "arm" && params+=("VRENDER=soft" "ARM_ENABLED=1")
+    isPlatform "arm" && params+=("VRENDER=soft" "ARM_ENABLED=1" "FORCE_DRC_C_BACKEND = 1")
```
joolswills, line 26: Should there be spaces around the = ?

pjft: Actually it worked with the spaces so I didn't notice it, but you're right - it doesn't need them so I removed them for compliance. Thanks for spotting them.

joolswills:
Thanks.

PR #3144, opened 2020-05-16
lr-theodore: switch to upstream repository

cmitu:
Fixes the repository source, replacing the libretro fork with the [upstream repo](https://github.com/Zlika/theodore).
Seems like the libretro fork is used only for PR submitted upstream, the buildbot uses the same upstream repository (https://retropie.org.uk/forum/topic/3049/thomson-computers-emulator/29).


joolswills:
Thanks.

PR #3148, opened 2020-05-19
Bluetooth updates

joolswills:
bluetooth - use bt-device from bluez-tools over old python code

 * added bluez-tools dependency
 * use bt-device for Trusting / Connecting / Removing of devices

 bluetooth - split out boot bluetooth connect script from main code

 * running RetroPie-Setup on boot isn't efficient / it should work without RetroPie-Setup present.
 * running it as root on boot causes /opt/retropie/configs/all to be chowned to root due to setupDirectories call

@psyke83 
This needs some testing - but works ok here. I did touch a bit of the Playstation pairing code with the first commit - perhaps I should have split out the second commit - I can do that if there's issues

the code relating to the platstation code was the following

```diff
 -       $(get_script_bluetooth bluez-test-device) disconnect "$mac_address" 2>&1
 -       $(get_script_bluetooth bluez-test-device) trusted "$mac_address" yes 2>&1
 -       local trusted=$($(get_script_bluetooth bluez-test-device) trusted "$mac_address" 2>&1)
 -       if [[ "$trusted" -eq 1 ]]; then
 +       bt-device --disconnect="$mac_address" 2>&1
 +       bt-device --set "$mac_address" Trusted 1 2>&1
 +       if [[ "$?" -eq 0 ]]; then
```

My plan is to use the bluez-tools commandline tools for more stuff - and get rid of the old python. This is a first step - we may not be able to replace everything of course.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -30,7 +30,7 @@ function _get_connect_mode() {
 }
 
 function depends_bluetooth() {
-    local depends=(bluetooth python-dbus python-gobject)
+    local depends=(bluetooth python-dbus python-gobject bluez-tools)
     if [[ "$__os_id" == "Raspbian" ]]; then
         depends+=(pi-bluetooth raspberrypi-sys-mods)
     fi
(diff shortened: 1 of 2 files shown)


psyke83:
Please don't merge this yet - it breaks DS3 pairing. Sorry for the delay in testing, will try to troubleshoot properly soon.

psyke83:
Update: I forgot to trigger installation of the new dependencies, so it was user error. It seems to work fine after `bluez-tools` is installed.

Small comment regarding these lines:
```
+        bt-device --disconnect="$mac_address" 2>&1
+        bt-device --set "$mac_address" Trusted 1 2>&1
```
Is it necessary to show output here? It's resulting in some message spam in the bluetooth GUI at the end of the pairing procedure that wasn't present before.

joolswills:
No. I'll sort that.

It would be good to use the bt-agent tool. I think it needs to be started and then a connection made. But I'm very reluctant to break your additions etc.

What do you think?

psyke83:
I'm not sure if `bt-agent` alone can pair those controllers given than non-standard steps are required for the USB authentication through the BlueZ siaxis plugin, but I can test any further changes that make incorporating it.

joolswills:
Revisited this. So looks like the old code had a `2>&1` to redir stderr to stdout. I guess I copied that blindly. I will redirect stdout to /dev/null instead.

I will then merge this. If I've broken anything we will find out soon enough.

joolswills:
I may need to redirect stderr to silence those lines, but I didn't test that code. If we do, maybe it's best to capture it (for logging) and display if there's an error.

Anyway. Merging this as it is. Further improvements to get rid of the last of the old python code would be good in the future.

PR #3150, opened 2020-05-22
usbromservice - fix mounting ext3/ext4 partitions

joolswills:
Mounting ext3/ext4 and other native filesystems was broken due to having uid=X,gid=X in the default mount options.
Now we just set these options on a per filesystem basis via FS_MOUNTOPTIONS

This is not yet tested.


joolswills:
Tested working.

PR #3151, opened 2020-05-24
audiosettings: updates for Pi4 and support for discrete internal ALSA devices

cmitu:
Modified to support the 2nd HDMI port on the Pi4 and for supporting the new upstream defaults, which use discrete ALSA cards. Handle the case when on-board audio is disabled.

On-board audio support is enabled via the `snd_bcm2835` module. Depending on the `enable_compat_alsa` parameter value (0/1), loading the module creates 1 or ~~3~~ 2/3 sound devices. 

The default so far has been `enable_compat_alsa=1`, but this will change in the new 5.4 LTS kernel (see https://github.com/raspberrypi/firmware/commit/7449540e3787cc14e9ee7f63690d6569943bce26 and the discussion in the [RPF Forums](https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=269769&start=200#p1659490)). `raspi-config` already handles this since [January](https://github.com/RPi-Distro/raspi-config/commit/629699ff9b2e73c92bafe45280515d816fbcaaf5), so I added a similar list of options to the `audiosettings` module.

### DIFF
--- a/scriptmodules/supplementary/audiosettings.sh
+++ b/scriptmodules/supplementary/audiosettings.sh
@@ -12,7 +12,7 @@
 rp_module_id="audiosettings"
 rp_module_desc="Configure audio settings"
 rp_module_section="config"
-rp_module_flags="!x86 !mali"
+rp_module_flags="!all rpi"
 
 function depends_audiosettings() {
     if [[ "$md_mode" == "install" ]]; then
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/audiosettings.sh
```diff
@@ -21,12 +21,38 @@ function depends_audiosettings() {
...
+    if aplay -l | grep -q "bcm2835 ALSA"; then
+        _bcm2835_alsa_compat_audiosettings
+    else
+        _bcm2835_alsa_internal_audiosettings
+    fi
+}
+
+function _bcm2835_alsa_compat_audiosettings() {
     local cmd=(dialog --backtitle "$__backtitle" --menu "Set audio output." 22 86 16)
+    local HDMI="HDMI"
```
joolswills, line 42: Please can you make the var lowercase.

scriptmodules/supplementary/audiosettings.sh
```diff
@@ -35,27 +61,101 @@ function gui_audiosettings() {
...
+
+function _bcm2835_alsa_internal_audiosettings() {
+    local cmd=(dialog --backtitle "$__backtitle" --menu "Set audio output." 22 86 16)
+    local options=()
+    local card_index, card_label
+
+    # Get the list of Pi internal cards
+    while read card_no card_label; do
+        options+=("$card_no" "$card_label")
+    done < <(aplay -ql |aplay -ql | sed -En 's/^card ([0-9]+).*\[bcm2835 ([^]]*)\].*/\1 \2/p')
```
joolswills, line 103: Minor whitespace issue

scriptmodules/supplementary/audiosettings.sh
cmitu: Done.

scriptmodules/supplementary/audiosettings.sh
cmitu: hm, not sure which one is it, but I removed the 2nd `aplay` pipe - obviously a mistake.

joolswills: I missed the duplicate. Just saw the missing space. Heh. Thanks.

joolswills:
Thanks. Looks good. Will merge after testing.

PR #3153, opened 2020-05-25
Updated to 2.2.4

Snoni:
This is untested because I'm lazy
And also stupid a little

### DIFF
--- a/scriptmodules/ports/srb2.sh
+++ b/scriptmodules/ports/srb2.sh
@@ -21,7 +21,7 @@ function depends_srb2() {
 }
 
 function sources_srb2() {
-    gitPullOrClone "$md_build" https://github.com/STJr/SRB2.git "SRB2_release_2.2.2"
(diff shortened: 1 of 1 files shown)


cmitu:
Seems to work, note that the new game data includes an additional `pk3` ([patch_music.pk3](https://github.com/STJr/SRB2/commit/4244480d6345148cd8d0e330e50494f9bf4114b8)).

joolswills:
I see from the source the assets are hosted on another GitHub repo and puleld in by various scripts for different targets (assets/Debian/rules).. I worry this is due to copyright or something but if they are just managed elsewhere and built separately we should handle this ourselves. We currently host an assets archive but I would prefer another solution than updating this.

Tbh I'm suspicious of the legality of this for inclusion currently and I am thinking about removing it. But please correct me if I am mistaken.

joolswills:
SRB2 is already updated past this version.

PR #3155, opened 2020-05-28
retroarch: version 1.8.8 update

cmitu:
Rebased RetroPie patched for 1.8.8 and added - for now - the option to keep the global core options in a single file.

NOTE: tested on Pi4 only at the moment.


joolswills:
Thanks.

PR #3156, opened 2020-05-29
image: update the URL for latest Buster image

cmitu:
Latest Buster Raspberry Pi OS image has changed location. Previous URL returns older (Feb 2020) release.


joolswills:
I'll rebase my changes over this on second thoughts.

joolswills:
Thanks.

PR #3158, opened 2020-06-01
attractmode: enable for KMS/DRM

cmitu:
Added 'attractmode' for the 'kms' platform, since it now includes support for DRM/KMS.
The 'sfml-pi' branch on https://github.com/mickelson/sfml-pi/ can be built with OpenGL or GLES support, by default it uses OpenGL.

~~Because the 'drmSetMaster'/'drmDropMaster' calls need 'root' privileges, a separate setuid helper is installed, which runs these privileged calls. It's needed by AttractMode to background itself when launching a game/emulator.~~

Added 2 corrections:
 - don't remove an empty system name (can happen when `delEmulator` is used for a Port)
 - really replace `/` in the system's full name (when adding/deleting a system)

 When using KMS, `attractmode` starts with 'preferred' video mode, as reported by the display's EDID. This could be different than the current video mode set via 'config.txt' or via 'cmdline.txt' by users.
  Modified the startup script to hint the video mode via 'SFML_DRM_MODE' and 'SFML_DRM_REFRESH' environment variables.

### DIFF
--- a/scriptmodules/supplementary/attractmode.sh
+++ b/scriptmodules/supplementary/attractmode.sh
@@ -13,7 +13,7 @@ rp_module_id="attractmode"
 rp_module_desc="Attract Mode emulator frontend"
 rp_module_licence="GPL3 https://raw.githubusercontent.com/mickelson/attract/master/License.txt"
 rp_module_section="exp"
-rp_module_flags="!mali !kms frontend"
+rp_module_flags="!mali frontend"
 
 function _get_configdir_attractmode() {
     echo "$configdir/all/attractmode"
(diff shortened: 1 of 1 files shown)


cmitu:
I have removed the setuid binary, it shouldn't be needed after the latest changes added upstream.

But I've noticed an issue with the `attract.cfg` configuration file generation, which leads to an incomplete system list. I haven't found the culprit yet, so don't merge this yet - I may be able to figure out why it's not working correctly.

cmitu:
I found the cause of the mis-generated configuration file. When `kodi` is installed, `configure_kodi` calls `delEmulator`, which then calls the `_del_system_attractmode` with an empty `$1`, which deletes almost all systems added (`display <system>` lines).
I added an extra check to make sure that `_del_system_attractmode` doesn't accept an empty system name.

joolswills:
Actually I think that part of delEmulator just needs removing.

joolswills:
I'll remind myself and rework it as needed. Fine to leave the safety in though.

cmitu:
Added the modified startup script, should be ok now.

joolswills:
Thanks. Kms binary is built also.

PR #3159, opened 2020-06-04
Add .mdf to Saturn extensions

Arnvidr:
As documented on the wiki, .mdf is a valid extension for the Saturn emulators. Tested that the change allows loading these files on my local installation.


joolswills:
Thanks.

PR #3162, opened 2020-06-10
lr-np2kai: update build instructions

cmitu:
Looks like upstream renamed the build folder (`sdl2` renamed to `sdl`), so the build instructions have been updated. 
Similar upstream change at https://github.com/libretro/libretro-super/commit/0c7dde6c4d3ffff3fe1fcb5af5a9813319254a67.


joolswills:
Thanks.

PR #3165, opened 2020-06-13
runcommand: improve 'xrandr' output parsing for mode id

cmitu:
The mode id is a hex, so it can contain (multiple) `[a-f]` apart from `[0-9]`.

Reported in the forums a couple of times ([here](https://retropie.org.uk/forum/topic/26683/) and [here](https://retropie.org.uk/forum/topic/26362/)). 

`xrandr` code for printing the mode id looks to be [here](https://gitlab.freedesktop.org/xorg/app/xrandr/-/blob/master/xrandr.c#L618).


joolswills:
Thanks!

PR #3166, opened 2020-06-14
Added performance fix for RPI darkplaces quake

madmodder123:
Without "gl_nopartialtextureupdates 1" the Raspberry Pi version of darkplaces-quake runs at ~5 FPS, with it enabled you can get 30fps without issue


cmitu:
Your `.sh` file has DOS line-endings, making the `.diff` to the previous module version basically 100%. You should correct that first.

madmodder123:
Fixed it :)

joolswills:
Formatting is still wrong - I will look into this as I didn't need this change when I last run this on the rpi4.

joolswills:
Unable to reproduce the issue on my set up. I have 30+ FPS without this option (@ Adding it made no difference. I'm going to close, but feel free to provide more info.about your config. Maybe it's another configuration related to your set up (you're not running from the desktop environment right?)

joolswills:
Sorry I had made an assumption that this was a rpi4 issue. But I just reread, and you don't specify. I'll re-open and test on rpi2/3

Details of your set up would be useful btw.

If it's needed for rpi1/2/3 videocore drivers, i will add it, but not via this PR as currently your code will add the line again and again. Also it.may he better to use the config.cfg and via our iniSet functions.

madmodder123:
Ah yeah I forgot my details:
Raspberry Pi 4 2gb
Retropie 4.6

When you load up the game the demo works fine, but when  you start a new game it goes to 5FPS without that option toggled.

joolswills:
Can't reproduce it - this is better dealt with on the forum first.

joolswills:
Ok I can reproduce the issue with the new information provided. Will implement the workaround or another fix.

madmodder123:
I also wanna add that the reason I put the option in "autoconfig.cfg" and not "config.cfg" is that if you put it in the latter, it will be erased if you set everything to default.

joolswills:
Thanks.

joolswills:
Please see #3179

PR #3168, opened 2020-06-17
Update cgenius  from 2.3.6 to 2.4.4.1

madmodder123:
2.3.6 had a bug where the controller configuration menu wouldn't show up, the latest fixes this and other bugs:
https://github.com/gerstrong/Commander-Genius/blob/master/changelog.txt


joolswills:
Thanks.

PR #3170, opened 2020-06-18
add lr-smsplus-gx and lr-gearsystem

protocultor:
SMSPlus-GX and Gearsystem are Sega 8 bit emulator cores, with better compatibility for Master System games than existing cores in RetroPie.
Both have been tested on a RPi3B+.

### DIFF
--- a/scriptmodules/libretrocores/lr-gearsystem.sh
+++ b/scriptmodules/libretrocores/lr-gearsystem.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks. Looks good code wise. Will test etc.

dankcushions:
Do we need to add BOTH?

dankcushions:
https://retropie.org.uk/forum/topic/26771/new-scriptmodule-for-master-system-game-gear-emulator has more details and i think justifies the inclusion of both. looks cool! thanks :)

protocultor:
I was following the format of previously accepted PRs; should've added the info.
Thanks to both for having a look a this.

joolswills:
Thanks. Also binaries are built for RPI.

PR #3171, opened 2020-06-19
Update fs-uae.sh

angrystar170:
Fix line 31 (link changed to Ubuntu bionic repo)


joolswills:
Thanks. Don't think it should use that repo for Debian though. But there's a Debian 10 repo - I'll add that with a check for Debian version.

joolswills:
I have added the Debian 10 repo. Closing this but please comment if it doesn't work for some reason.

PR #3172, opened 2020-06-19
Update pcsx2.sh

angrystar170:
Add dependency package "libportaudio2:i386" and "libsdl2-2.0-0:i386"

### DIFF
--- a/scriptmodules/emulators/pcsx2.sh
+++ b/scriptmodules/emulators/pcsx2.sh
@@ -47,6 +47,7 @@ function depends_pcsx2() {
 function install_bin_pcsx2() {
     local version
     [[ -n "${__os_ubuntu_ver}" ]] && version="-unstable"
+    dpkg --add-architecture i386 && apt update
     aptInstall "libportaudio2:i386"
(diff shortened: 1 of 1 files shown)


cmitu:
Aren't those packages pulled automatically - as dependencies - by installing the `pcsx2` package ?

angrystar170:
Yes. 

For reference, I am using Debian-based Linux (MX-19).

cmitu:
If those packages are already installed as dependencies, what's the point of installing them separately ?
Also, the `i386` architecture is added separately in the as part of the `depends` function, it doesn't need to be added again.

angrystar170:
I manually installed "libportaudio2:i386" and "libsdl2-2.0-0:i386" using the apt command.

cmitu:
> This is because the retropie installation script cannot automatically import dependency packages.

This is not the job of the installation script (to import dependency packages), but something that should be handled by `apt`. If the `pcsx2` package doesn't list the necessary dependencies, then there is a problem with the DEB package, not an problem of the install script.

joolswills:
Thanks. But closing based on comments etc.

PR #3174, opened 2020-06-21
lr-parallel-n64: Added aarch64 flags

wrigglevision:
These added lines will make for an effortless install on arm64 distros.


joolswills:
I have a branch with aarch64 changes so won't be accepting anything until I've got a PR for that open. But appreciated.

PR #3175, opened 2020-06-21
lr-parallel-n64: Added aarch64 flags

wrigglevision:
These new flags will allow ARM64 distros to install Parallel N64 through normal setup scripts.


joolswills:
I have a branch with aarch64 changes so won't be accepting anything until I've got a PR for that open. But appreciated.

Btw you don't need to close and ooeb a new PR if making changes after opening.

wrigglevision:
Got it. Many thanks.

PR #3176, opened 2020-06-23
gemrb: update to latest stable version and refresh deps

cmitu:
* updated the dependency list to include the Vorbis and VLC libraries.
* choose explicitly the SDL1 backend at build time.
* added the `dispmanx` flag to enable proper scaling on FKMS (Pi4) platforms (untested though).
* explicitly depend on `python2-dev` since python3 port is not complete.


joolswills:
Thanks.

PR #3179, opened 2020-06-26
darkplaces-quake - fix rpi4 slowdowns and rework logic

joolswills:
* enable gl_nopartialtextureupdates by default on rpi (fixes slowdown on rpi4 in game, doesn't seem to harm rpi1-3/videocore)
 * remove makefile patch and override instead with makefile vars for parameters - this should make it simpler for adding additional gles platforms
 * apply makefile.inc changes on all systems (which allows passing through of cpu parameters)
 * override makefile.inc OPTIM_RELEASE parameters so we use our default optimisations

Few additional notes:

On RPI4 we use opengl vs OpenGLES2 on RPI1-3

There's one line of glsl shader code that fails on Mesa GLES2 on rpi4.
Line 41 of shader_glsl.h - it doesn't like the use of invariant here. I found a related Mesa fix, but maybe that was only for opengl. I'm going to dig further. Commenting it out will produce a working GLES2 build.

With rpi4 and opengl `-benchmark demo1` does 23-24fps average on my system @ 1024x768
With rpi4 and gles2, it does 55-56fps. However there are some minor but noticeable glitches when playing.

So I'm leaving it with opengl for now. There may be some further opengl config optimisations also. The gles2 code doesn't look as it's had as much attention so the opengl renderer looks to be more accurate.

The latest upstream code seems to have broken gles2 support currently, and needs opengl 3.2 so we can't move to that on the RPI currently.


joolswills:
I have now added additional changes to build a GLES version also on the rpi4. It won't be set as the default on a first install but can be switched to with the runcommand menu.

Frame rate is a lot higher as mentioned, but there are some graphicall glitches with enemies/enemy explosions, which may be improved by tweaking some options.

joolswills:
Would appreciate if anyone has time to test this before it's merged.

PR #3180, opened 2020-06-27
Fix systemd-nspawn detection

tallen116:
I am building an image based on RPi 0/1 4.5.1 utilizing systemd-nspawn.  The existing logic to determine chroot environments doesn't seem to hold true when utilizing systemd-nspawn container instead of the traditional chroot command.

I also ran into an issue when updating/compiling the xpad kernel driver.  I believe there is no reason to modprobe the kernel driver in a chroot environment.


joolswills:
Looks good. Thanks.

PR #3181, opened 2020-06-28
gemrb: switch to SDL2 for rendering

cmitu:
GemRB supports both SDL1 and SDL2 for graphic rendering, SDL2 being now the preferred one.

This commit adds a `-sdl2` version for `gemrb` that builds the port with SDL2 and modifies the original scriptmodule so we can build both versions using the same functions.


cmitu:
I've removed the extra `-sdl2` module and instead set `gemrb` to build with SDL2.

joolswills:
Thanks.

PR #3184, opened 2020-07-03
Improved aarch64 support including RPI3/4

joolswills:
Various changes to improve compatibility with aarch64 targets and rpi4 (on 64bit raspberry pi os).


bluestang2006:
Pardon my intrusion but this PR caught my attention because of the aarch support. On a clean, full updated 64-bit Raspberry Pi OS, the setup script "successfully" installed the cores and main packages (basic install). Obviously, the splashscreens are a problem with omxplayer not supported on 64-bit RPiOS but that requires a little more of an effort that is beyond my skills at that moment.

One thing of note, which I'm not sure matters all that much, is you can pass the `"-march=armv8-a+crc -mtune=cortex-a72"` (RPi4) and the `"-march=armv8-a+crc -mtune=cortex-a53"` (RPi3) cpu flags in the system.sh script rather than `"-march=native"` and RetroPie will build successfully.

Either option results in the ASIMD instructions being reported as a capability in the RA verbose logs. Note: I've only tested a few roms with the MAME2003+, MAME2016, and FBNEO builds so I can't speak for the other emus.

I suppose the real question is if the `+crc` flag is needed because that is not a default option for armv8-a in gcc 8.3.0 or at least that is what I understood from reading the docs.

joolswills:
I haven't finished the cpu flags yet, but am aware thanks.

joolswills:
BTW `-march=native` is the equivalent of `-march=armv8-a+crc` (see gcc -march=native -Q --help=target)

psyke83:
Small note: given that you already fixed the ppsspp build, it's probably safe to remove the `!aarch64` flag from `lr-ppsspp.sh` as the build function is shared.

joolswills:
Thanks. Will do.

scriptmodules/supplementary/sdl2.sh
```diff
@@ -88,7 +88,7 @@ function build_sdl2() {
     sed -i 's/libgl1-mesa-dev,/libgl1-mesa-dev, '"${conf_depends[*]}"'/' ./debian/control
     sed -i 's/confflags =/confflags = '"${conf_flags[*]}"' \\\n/' ./debian/rules
 
-    if isPlatform "rpi"; then
+    if isPlatform "rpi" && [[ -d "/opt/vc/include" ]]; then
```
psyke83, line 91: Rather than applying a workaround for the header issue, perhaps it would be better/cleaner to stop building SDL2 with both drivers, and just remove these blocks entirely? You won't need to change anything in the SDL source, but on the next branch update/rebase, you can leave out this commit: https://github.com/RetroPie/SDL-mirror/commit/2db72c15114a190d18ebac343aafd3f53c4d849a

Just a suggestion, as I recall @hhromic recently mentioned some kind of issue with some SDL2 software not working correctly unless the KMS driver was forced via environment variable (IIRC).

bluestang2006: The latest source of SDL2 can be built with x11, egl, and kms/drm support simultaneously for a Pi4. If RA is also built with the same support (x11, egl, kms/drm) support as well, the user can use both the desktop environment and headless without any performance impact.

joolswills:
Enabled lr-ppspp on aarch64

camccar:
Just tested this and can confirm it fails on basic install due to omxplayer

joolswills:
@bluestang2006 @camccar 

If you want to test stuff that is in development, please provide logs. Just saying something fails is not enough information. I fixed the omxplayer dependencies already, but am aware of an issue with omxiv which is a separate tool. Perhaps you meant that.

I'm not really after user testing btw yet. And to note, there are no plans currently to switch RetroPie to 64bit on the RPI.

camccar:
> @bluestang2006 @camccar
> 
> If you want to test stuff that is in development, please provide logs. Just saying something fails is not enough information. I fixed the omxplayer dependencies already, but am aware of an issue with omxiv which is a separate tool. Perhaps you meant that.
> 
> I'm not really after user testing btw yet. And to note, there are no plans currently to switch RetroPie to 64bit on the RPI.

Hey I know you said you aren't looking for testing but i'm still getting the omx player error. I'm attaching a screenshot and logs.

omxplayer

[rps_2020-07-18_083115.log.gz](https://github.com/RetroPie/RetroPie-Setup/files/4941738/rps_2020-07-18_083115.log.gz)

joolswills:
Doesn't look like you are running the latest code from this branch.

joolswills:
I'm not sure you are running code from this branch at all actually.

camccar:
> I'm not sure you are running code from this branch at all actually.

Let me double check one minute

camccar:
Everything works great thank you. Much appreciated.

joolswills:
Fixed up some errors in mupen64 (wrong flag checks in a couple of places).

Im just tested the changes.

joolswills:
Mupen64plus seems ok now. I will be merging this soon.

joolswills:
It's typical but I need to fix up a lr-flycast issue o just noticed moments after merging. Will commit shortly.

akamming:
@joolswills  Great work, just tested the aarch64 installation on the pi4. All main packages seem to work fine!

I tried to run some Saturn games (lr-yabause) as well, only to notice the build failed.  i tried to raise an issue in the lr-yabause repository, but got feedback this is caused by the fact that the setup script calls just "make", where it should call "make HAVE_SSE=0" and this should be changed in the retropie setup script. I changed manually in the script and conformed it worked....

i know aarch64 on pi4 is not supported, so i don't expect anything, but if you do  want to support this in the future, you will now know how to make the lr-yabause core build as well on aarch64 on the pi4.  

Keep up the good work!

PR #3185, opened 2020-07-03
Make joystick's "B" button cancel console dialogs

c0d3h4x0r:
For consistency with EmulationStation.


c0d3h4x0r:
I'm realizing it would also be super nice to have the L/R shoulder buttons (buttons 5 & 6, presumably) map to PageUp and PageDown as well, so I'm working on an additional commit for that here, too.

c0d3h4x0r:
Got that working too.  Much nicer for operating the textual menus using just a gamepad.

Gemba:
Hi @c0d3h4x0r - looks we are on a similar path. Maybe you want to chime in at https://retropie.org.uk/forum/topic/27128/enhancements-for-dialog-navigation-controls which is good practice for Retropie before filing a PR.

c0d3h4x0r:
Honestly, this is a very trivial change, so I'm not sure why it even requires a discussion, but sure, I'll hope over there and take a look.

PR #3188, opened 2020-07-05
Totally rewritten new & improved Bluetooth input device configurator

c0d3h4x0r:
This fixes countless reliability and correctness problems with the existing Bluetooth configurator and back-end pairing script, and greatly enhances the user experience by way of clearer messaging, more responsive visual feedback, elimination of the entire step that prompted the user to manually select a bluez agent capability type (DisplayYesNo and the like), and scoping the entire configurator down to only working with *input* devices.

Because this is a massive high-churn rewrite, trying to diff the changes is pointless.  I encourage you to simply fetch and try out this version and run it through a bunch of test cases to see how it performs.  In all of my self-testing, which has been pretty thorough (I've paired a variety of keyboards and gamepads in a Bluetooth-dense environment with plenty of non-input devices crowding the scans, and I've also exercised all of the possible error cases I could possibly make happen), the new experience is lightyears better, faster, and more reliable than the old one, so I'd very much like to see this rewrite taken into mainline.


joolswills:
Thanks but it would be a good idea to open a discussion first before undertaking rewrites of stuff - work on the bluetooth code is ongoing (see recent changes).

>Because this is a massive high-churn rewrite, trying to diff the changes is pointless.

I disagree. Everything included in RetroPie is reviewed, and I can see issues with this - styleguide / functionnaming and so on.

Certainly more information will be needed regarding the changes for any change of some of this to get in.

joolswills:
Some comments:

All functions within the bluetooth module need to be named ending with _bluetooth
Functions should use local vars

But I'm not sure if the changes are wanted as mentioned, as my plan was to move to the bluez-tools utils, and away from python helper scripts.

joolswills:
Also even if it's a large rewrite, commit msgs need to reflect what is being changed Vs things like "massive fixes". I don't think this can be accepted as it is now sorry.

c0d3h4x0r:
This implementation is lightyears better (in terms of design correctness and user experience) than what’s in your `master` branch right now.  I’ve already sunk 20+ hours into it, so if all it needs is another couple hours to make it conform to your coding conventions, I’m willing to do that.  But simply saying you won’t take it just because it’s high churn is a pretty lousy reason.  There was no way to fix the multitude of problems with the existing code without doing a massive rewrite, because the existing code was designed around fundamental misunderstandings of how Bluetooth pairing is supposed to work.

I’m not sure what you mean by “the bluez-tools utils”, but if you’re referring to the bluez-simple-agent that I removed, that script was an absolute train wreck.  I replaced it from scratch using a modified version of the agent I recently (re)wrote for the MiSTer project.  It was actually amusing to see that the pair-agent script that had originally been used by MiSTer was a derivative of the same bluez-simple-agent and contained a number of the same bugs.

c0d3h4x0r:
I should also point out that I was very careful to preserve all the special-cased code for things like sixaxis controllers, to avoid regressing edge cases for hardware I don’t personally use.  If you do try to diff my changes, you’ll at least be able to see that, if the diffing algorithm is smart enough to track the moved preserved snippets.  (Sadly, my experience has been that GitHub’s built-in differ does a pretty sad job.)

c0d3h4x0r:
How about instead of closing this PR outright, you leave it open to give me a chance to revise it to meet your requirements?

c0d3h4x0r:
If I go to the trouble of incrementally rebuilding back to this proposed change through a series of tiny commits in a new branch, and fix the function locals and function naming concerns, will you consider taking it then?

c0d3h4x0r:
bluez-utils: were you referring to the deprecated package listed at https://www.bluez.org/download/?

joolswills:
I closed it as I felt from your wording in your reply there was going to be a conflict. I'm sorry you spent 20+ hours. It would be advisable to check first before spending so much time without any prior discussion.

I was referring to bluez-tools, a not completely finished but still useful toolset that I was hoping to replace the python code with https://github.com/khvzak/bluez-tools

I looked over your changes in some detail. I'm not sure I want to introduce new python code to maintain. It would be best to start with suggesting an improvement to our code and going from there but I do have some plans already.

Btw this module should not be restricted to input devices. It can be used to pair audio devices for example as is used as such in existing tutorials on the forum. 

In general large changesets are not just accepted and we need detailed commits for such things. Especially for a core part of retropie.

c0d3h4x0r:
I’m not looking for a conflict; I just want to help make this area of functionality better.  It has been a frustratingly broken area for a long time.

The Python code is just one part of the changes, and a separable one at that.  It’s also easy to remove the input devices check from what I’ve done.

Please tell me where I can find ongoing discussion/work regarding this particular area, and I’ll gladly join in and participate.  I’m not emotionally attached to the code I write, but I have some clear learnings and experience working in this area and would like to put it to good use.

joolswills:
What's frustratingly broken? Your language is very dramatic. Details are needed.

c0d3h4x0r:
1) “registered devices” is not a useful user concept.  What matters to the user is whether a device is paired or not, not whether it is known to `bt-device`.  Because of it being built around “registered” devices instead of paired devices, the logic in the “Pair a device” choice is badly broken, such that on all scans subsequent to the first, no devices are found or offered for pairing, because they are already “registered” but not yet paired.

2) bluez agent capabilities are not something that should ever need to be exposed to the user or that they should be requires to manually choose, as they are an implementation property of the agent, not an inherent property of the device being paired.  If the user is operating the menu, then they must have a display, so that already limits the sensible choices to DisplayOnly, DisplayYesNo, or KeyboardDisplay.  And from there, the right choice of the three depends on whether they have a keyboard attached, a joystick attached (since a joystick can be used to select yes/no buttons), or neither attached.  There’s no reason to present the user with a choice at all here, especially when it’s one they won’t even understand.

3) The “Connect all devices” option is completely broken (command-not-found errors) due to an unnecessary usage of a `$()` subshell.

4) The “Display devices” option produces extremely human-unfriendly output, and again is ‘t helpful because of showing “registered” rather than paired devices.

5) The reason I replaced the pairing agent wholesale is because the existing one does not pair reliably due to its misusage of the bluez APIs.

6) There are many spots in the bluetooth configurator where selecting a menu option givea no immediate visual feedback while the script is working, which is just a bad user experience.

c0d3h4x0r:
I can file separate issues for each of these major problem categories if you prefer, and split the fix out for each one separately.

joolswills:
1) actually that was a recent ish change. I found it more useful than displaying everything. The idea being that the devices are paired. If they don't pair they can be removed and re-added but I've not seen it to be a massive issue.

2) I like having the choice here for compatibility. It usually works automatically choosing first option.

3) this should be an easy fix. It's an old hack though for issues on old bluez stack

..

I agree with some of your points. This would have been a good first post imho.

I will go over the issues later and get back to you if I think some input would be useful. Thanks.

c0d3h4x0r:
1. It’s an extremely confusing and maddening user experience to have to go manually “remove” known-but-not-paired devices just to get them to show up again in another scan attempt, especially when the pairing operation only succeeds about 50% of the time, requiring a retry.

c0d3h4x0r:
2. I’m sorry you like having the choice there, because I don’t, and neither does any other RetroPie user I’ve personally talked about it with, because nobody understand what it does or why TeyroPie is asking them to make a choice there when no other device or operating system’s pairing wizard asks them to make such a choice.  Can you explain why you like it?  If it’s just a developer-only reason, it should at least be hidden behind some kind of debug flag or environment variable.  Making the user choose is entirely unnecessary and thus a bad UX.

c0d3h4x0r:
And again, not trying to be adversarial or argumentative — just trying to point out the user’s point of view.  I’m always a strong advocate for the user’s experience.

Thank you for taking these points into consideration, and let me know if there’s any other way I can help.  For example, if you want me to look into bluez-tools and prototype reworking the script to use it, I’m happy to take that on and report back with my findings.

PR #3189, opened 2020-07-05
Added new menu option to set disable_ertm=Y for Xbox One controller bluetooth support. 

jinglejangle:
…and /sys/module/bluetooth/parameters/disable_ertm=Y

It will survive reboots. It toggles it on and off. It will create bluetooth.conf if it isn't in /etc/modprobe.d/
It will toggle the parameter if it already exists in that file.

Tested on 3A+ with various xbox one bluetooth controllers. 

My xbox one controllers can pair properly once this is set. And stop working again once it is set to N.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -330,6 +330,47 @@ function connect_bluetooth() {
     done < <(list_registered_bluetooth)
 }
 
+function _get_ertm_mode() {
+	local ertm=$(cat /sys/module/bluetooth/parameters/disable_ertm)
+	echo $ertm; 
+}
+
+function toggle_bluetooth_ertm() {
+	# Xbox one controllers aren't paring properly until this is set to disable_ertm=Y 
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. I may integrate this differently, but did you try the xpad neo driver btw? It may work then without this.

jinglejangle:
Sure, I totally agree a different approach should be used really
Was just some sunday after noon fun for me, and got my pad working how I wanted. 
I did try xbox drivers and things but will systematically check again what's going on to ensure. 

I thought about maybe checking the name when bluetooth pairing with it for Xbox Wireless Controller or whatever it is and setting it automatically if detected. This approach doesn't scale at all, imagine if every pad needed a different toggle! and also it's a bit rude about making the system config changes, i'd really want those all handled properly too, not just as i've done here editing config files directly and calling out to service and things. 


but I wanted a quick menu option PoC and there it is, thought it would be rude not to share it, thanks for all the top work on the project. 




Thanks

joolswills:
I dont think this is needed now with the xpadneo driver. If I'm wrong let me know.

PR #3190, opened 2020-07-05
retroarch - updated to v1.9.3

joolswills:
* dropped hotkey hack patch as now integrated with config option input_hotkey_block_delay with default of 5 (as was our patch)
 * rebased disable search patch (code moved back to retroarch.c for now)
 * rebased / updated shader path config enable patch (retroarch_load_shader_preset_internal parameter changes etc)

It may be best to wait unless there is a reason we need to move to 1.8.9 due to ongoing reorganization.

Also we need to address at least the save state etc path defaults.

Do we go with the new defaults and just inform users they need to copy files, set to a per system folder or set to the ROMs folder as it is now. Will require changes to all libretro modules to do anything other than leaving it as defaults.

Opening this PR primary to keep track of changes and discuss best options. I can't remember if there was other changes we need to work around?


cmitu:
> Do we go with the new defaults and just inform users they need to copy files, set to a per system folder or set to the ROMs folder as it is now. Will require changes to all libretro modules to do anything other than leaving it as defaults.

IMHO we should change the save/state dirs to the ROM folders - or to a more sensible value. While the changes make sense, the defaults are set in `$HOME/.config/retroarch`, which is more of a configuration folder than a data folder. For states, there's no clear separation per-system or per-core, so it's possible to overwrite a save state if the game name is identical - for instance _Alladin_ exists in both SNES/MD.

For certain cores (i.e. arcades), the save folder is used for storing additional save data (NVRAM, Hiscore data, input/DIP switch configuration). Those would need to be moved if the save folder changes.

New version seems to work fine. 
One notable change is https://github.com/libretro/RetroArch/pull/10874/. Before that, RetroArch would exit if the content was incorrect - for instance incorrect arcade/MAME romsets. Now the menu is shown and an info message is displayed - the user would need to exit RetroArch from the menu.

joolswills:
We will have to revert the upstream commit then that changes the default location as I don't see a clean way of doing it otherwise.

cmitu:
> I don't see a clean way of doing it otherwise.

Can't we set the `savestate_directory`/`savefile_directory` with configuration-called function ?
It could analyze any system's `retroarch.cfg` and see if they're set (just in case it's already set), then set them if not found. Of course, modify `ensureSystemretroconfig` so that the new folders would be set by default when a new system is added.

Another option would be to modify `runcommand` and set the folders via the generated `/dev/shm/retroarch.cfg`, depending on the system. Since it's used as `--appendconfig`, it shouldn't overwrite any  configuration done by the user (via the system's `retroarch.cfg`).
This has the disadvantage of hiding the configuration, I think an explicit setting approach would be better.

bluestang2006:
The latest upstream commits to RetroArch enable users to use WIP Mesa Vulkan driver on a Pi 4. This needs more investigation.
https://pastebin.com/LtBmMDfj - 32-bit RPiOS.

joolswills:
@bluestang2006 What needs investigating ? How did you build RetroArch ? Not sure the relevance in regards to this PR - we build with vulkan disabled.

joolswills:
@cmitu

joolswills:
@cmitu 

>Can't we set the savestate_directory/savefile_directory with a configuration-called function ?
>It could analyze any system's retroarch.cfg and see if they're set (just in case they're already configured), then set them if not found. Of course, modify ensureSystemretroconfig so that the new folders would be set by default when a new system is added.

This won't help with roms in subfolders. 

>This has the disadvantage of hiding the configuration, I think an explicit setting approach would be better.

I'm still leaning towards a retroarch code change. I don't see how to do it nicely otherwise. It's a shame they didn't add some backward compatibility with a setting like "Use content folder". Maybe we can add something like this.

bluestang2006:
> @bluestang2006 What needs investigating? How did you build RetroArch ? Not sure the relevance in regards to this PR - we build with vulkan disabled on the RPI.

At this point, the simple answer is to enable X11 for the RPi 4 builds along with the --enable-vulkan flag. 

As for investigating, there is an obvious performance hit with X11 as best as I can tell, but I think it is related to ES since RetroPie uses both. A pure RetroArch build doesn't have this performance hit at all. At this stage of my tomfoolery, that's the only major difference I see.

Also, since the commits that enabled the Vulkan contexts came after the 1.8.9 release that means they will be a part of the 1.9.0 release. 

Obviously this project has other major issues to solve before it updates to 1.8.9 so it is a fair point that these observations are not related to this.

However, they may be useful for 1.9.0.

joolswills:
This isn't relevant to this PR though. Please use the forum.

hhromic:
> One notable change is libretro/RetroArch#10874. Before that, RetroArch would exit if the content was incorrect - for instance incorrect arcade/MAME romsets. Now the menu is shown and an info message is displayed - the user would need to exit RetroArch from the menu.

This is now fixed in this PR: https://github.com/libretro/RetroArch/pull/11063
We should probably wait until version 1.9.0 to get it.

joolswills:
@hhromic I agree.

I added a revert for the default saves/states/system path change.

I'm actually a slight fan of having saves/savestates in the content folder as it's handy with network shared romsets, without need for separate shares etc.

Maybe upstream could be persuaded for some option.

Another workaround btw would be to "migrate" these files on launch with runcommand. However it would need to be well documented so users are aware.

But tbh, it's easier to keep the old behaviour and it's not a difficult patch to maintain compared to others we had like the hotkey delay.

joolswills:
Updated to v1.9.0 and rebased patches. Untested currently.

cmitu:
It looks like https://github.com/libretro/RetroArch/pull/11171 is needed when enabling the VC driver (`--enable-dispmanx`), otherwise compilation fails.

joolswills:
I added the commit in question @cmitu

joolswills:
This will need sorting before we move to v1.9.0 - https://github.com/RetroPie/RetroPie-Setup/issues/3249

bluestang2006:
Re: saves and default paths - https://youtu.be/cNPtAOjaDJE

joolswills:
Just to note on this. I am waiting for v1.9.1. this PR needs to be rebased which I will do but we want the config code handling as referenced https://github.com/RetroPie/RetroPie-Setup/issues/3249 wh

joolswills:
Sorry for close/reopen. Accidentally clicked button while typing.

bluestang2006:
Looks like RetroArch v1.9.1 finally landed. Lots of new features, new gles configure flags, shader options, etc.

joolswills:
* dropped hotkey hack patch as now integrated with config option input_hotkey_block_delay with default of 5 (as was our patch)
 * rebased disable search patch (code moved back to retroarch.c for now)
 * rebased / updated shader path config enable patch (thanks @cmitu)
 * revert the default saves/states/sytem path change as we rely on the old behaviour
 * default to using ON/OFF text instead of rgui_switch_icons graphics

joolswills:
This is still a WIP. Configuration flags etc have not been reviewed / adjusted yet. But wanted to update on progress.

bluestang2006:
@joolswills I'm curious if you are getting a bunch of warnings like the one below trying to compile any version of RetroArch?
`usr/include/valgrind/config.h:336: warning: "PACKAGE_NAME" redefined #define PACKAGE_NAME "Valgrind"`
`input/drivers/../../config.h:4: note: this is the location of the previous definition #define PACKAGE_NAME "retroarch"`

I grabbed the latest [64-Bit RPiOS image](https://downloads.raspberrypi.org/raspios_arm64/images/raspios_arm64-2021-04-09/2021-03-04-raspios-buster-arm64.zip) and tried to install your branch in this PR and ran into this issue. RetroArch failed to install, mainly due to ffmpeg but there are all sorts of notes and warnings such as the ones above. If I add the configure option --disable-ffmpeg RA compiles and installs fine. However, the version reported is 3.14.0? 

This is a fully updated 64-bit RPiOS, which evidently had a minor version bump 10.3 to the base files via apt.

joolswills:
Not checked on aarch64 yet but will backport / patch the issue when I test. Thanks @bluestang2006 

Will push v1.9.2 changes shortly.

joolswills:
I will squash this and merge if there's no further issues. Thanks to @cmitu for reporting an older bug I noticed which was fixed for 1.9.3

https://github.com/libretro/RetroArch/issues/12387

I've not yet tested on anything but the rpi4 but am keen to merge if working fine on other rpis and x86.

bluestang2006:
> Not checked on aarch64 yet but will backport / patch the issue when I test. Thanks @bluestang2006
> 
> Will push v1.9.2 changes shortly.

v1.9.3 no longer requires the patch. I have just built v1.9.3 on aarch64 without any issues.

joolswills:
1.9.4 is out but I'll merge this and then test 1.9.4 or else I will never get this in.

joolswills:
I will squash and merge it shortly.

joolswills:
I have updated to v1.9.4 now after some testing. Binaries will be updated shortly (binaries are at v1.9.3)

PR #3191, opened 2020-07-06
hatari - switch to git repository (v2.2.1) and symlink bios (tos.img)

joolswills:
* update module help to include bios information

It's tested working but perhaps our default config/launchers need to be revisited as a fair amount may have changed.


joolswills:
Also added .zip extension. Tested working on latest hatari.

joolswills:
Binaries built for stretch and buster.

PR #3195, opened 2020-07-12
lzdoom: update to 3.86a

psyke83:
* This was announced as the final lzdoom release to support GL2 rendering.
* Update Freedoom to v0.12.1 with checksum check to update older versions.

---
Tested on RPI4.


joolswills:
Thanks!

PR #3197, opened 2020-07-13
ti99sim: update to version 0.16.0, switch to SDL2

cmitu:
Updated the emulator to the latest version, which is using SDL2, so `dispmanx` is not necessary mandatory.

Modified the installation to include some utilities (might be useful for disc/cartridge manipulation) and to include the docs and license.

### DIFF
--- a/scriptmodules/emulators/ti99sim.sh
+++ b/scriptmodules/emulators/ti99sim.sh
@@ -12,25 +12,35 @@
 rp_module_id="ti99sim"
 rp_module_desc="TI-99/SIM - Texas Instruments Home Computer Emulator"
 rp_module_help="ROM Extension: .ctg\n\nCopy your TI-99 games to $romdir/ti99\n\nCopy the required BIOS file TI-994A.ctg (case sensitive) to $biosdir"
-rp_module_licence="GPL2 http://www.mrousseau.org/programs/ti99sim/"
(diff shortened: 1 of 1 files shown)


joolswills:
If switching to sdl2, dispmanx flag isn't used. It's for sdl1 dispmanx driver.

joolswills:
I'll self host the archive as with previous version btw.

cmitu:
@joolswills ok, added the necessary changes.

joolswills:
If you want to switch to the xz archive, we will need the xz dependency. But it may be better to use gz. I can convert the archive.

joolswills:
(use gz. I'll make the archive available now on our server).

joolswills:
Just to note, we probably don't need xz-utils dependencies anymore in other packages, as it's required by dpkg-dev which is required by build-essentials, but this wasnt always the case on older debian afair.

joolswills:
Thanks

cmitu:
@joolswills just noticed the typo `libsd2` instead of `libsdl2` - sorry.

joolswills:
You had a typo in the dependencies btw. I fixed it after the merge.

joolswills:
Just to note, this version doesn't build on Stretch due to c++ version required (--std=c++2a parameter) not being recognised.

We will need to switch back to the older version for Stretch probably.

PR #3200, opened 2020-07-20
runcommand: validate the CrtcId/ModeId values for SDL2 env vars

cmitu:
Before setting the SDL2 env variables that configure the modesetting for DRM/KMS, validate the CrtcId/ModeId tuple against the current list reported by `modetest`. Changes in the kernel DRM driver could make the existing modesetting configuration (stored in `videomodes.cfg`) invalid because of different CrtcId and/or ModeId in new versions of the driver.


joolswills:
Awesome. Thanks!

PR #3202, opened 2020-07-21
joy2key - added repeat support for easier navigation of large menus

joolswills:
Remove unneeded loop from read_event.

Make process_event return the chars to output and split out code to write to terminal so it can
be called from the added repeat code in the main processing loop

Added a JS_DEL (delay before repeat) and use JS_REP (time betweeen repeats) for the key repeat speed

Added js_hold variable, where we store the last button press input received. If no other button press
or release is done, we do a repeat of the last output.

### DIFF
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -198,13 +198,12 @@ def process_event(event):
     if js_type & JS_EVENT_INIT:
         return False
 
-    hex_chars = ""
+    hex_chars = None
 
     if js_type == JS_EVENT_BUTTON:
         if js_number < len(button_codes) and js_value == 1:
             hex_chars = button_codes[js_number]
-
-    if js_type == JS_EVENT_AXIS and js_number <= 7:
+    elif js_type == JS_EVENT_AXIS and js_number <= 7:
         if js_number % 2 == 0:
             if js_value <= JS_MIN * JS_THRESH:
                 hex_chars = axis_codes[0]
@@ -215,13 +214,10 @@ def process_event(event):
                 hex_chars = axis_codes[2]
             if js_value >= JS_MAX * JS_THRESH:
                 hex_chars = axis_codes[3]
-
-    if hex_chars:
-        return hex_chars
     else:
-        return None
+        return False
 
-    return False
+    return hex_chars
 
 js_fds = []
 tty_fd = []
--- a/scriptmodules/supplementary/runcommand/joy2key.py
+++ b/scriptmodules/supplementary/runcommand/joy2key.py
@@ -23,9 +23,11 @@ from pyudev import Context
 
 JS_MIN = -32768
 JS_MAX = 32768
-JS_REP = 0.20
 
-JS_THRESH = 0.75
+JS_DEL = 0.5
+JS_REP = 0.07
+
+JS_THRESH = 0.9
 
 JS_EVENT_BUTTON = 0x01
 JS_EVENT_AXIS = 0x02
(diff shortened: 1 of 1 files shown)


joolswills:
Fixed the analogue stick logic issues. Was a simple check needed if the js_hold was set for that joystick. If so, we ignore any input other than release. This makes.the analogue sticks function like a digital pad in the menus. Put the ranges and thresholds back as they were.

Would appreciate if you can test this @cmitu

cmitu:
I gave it a try and the scrolling behavior via the analog joystick has some quirks:
- up/down scrolls, but skips about 1-2 entries, depending on how fast the stick is pushed. What's more annoying is that sometimes an extra scroll is generated when the stick is put back into position.
- keeping the joypad at max up/down doesn't always trigger a continuous scroll in the list, just like when D-Pad up/down are kept pressed.

I tested using the left joystick on a DuasShock 4 (v2) connected via USB. The right joystick has a different behavior - up/down switches between the buttons, while left/right scroll up/down in the list (but it behaves similarly w.r.t. scrolling).

I tried to get an event list via `evtest`, but the DS spams a continuous `EV_SYN` event that makes the events logging impractical.

joolswills:
Strange, didn't get that in my 8bitdo and the newest changes were to make the analogue stick behave better not worse - annoying. I don't have that joystick but I have a ds3. I'll recheck.

joolswills:
Does the joypad have an accelerometer that creates events when tilting etc?

cmitu:
Yes, it has, but it's created as a separate input device:
```
/dev/input/event3:	Sony Interactive Entertainment Wireless Controller Touchpad
/dev/input/event4:	Sony Interactive Entertainment Wireless Controller Motion Sensors
/dev/input/event5:	Sony Interactive Entertainment Wireless Controller
```

joolswills:
I wonder if it could still be interfering - can you test forcing to just the pad device? Also do you have any other analogue sticks to test with?

cmitu:
I'll give it a try with an Wii U Pro (no motion sensors) and re-check I'm using the latest `joy2key`.

cmitu:
I re-rested using a PS3 controller and it's behaving better - doesn't jump over lines during scrolling, but auto-scrolling when the joystick is at max up/down doesn't always triger, just like for the DS4.

madmodder123:
> 
> 
> I re-rested using a PS3 controller and it's behaving better - doesn't jump over lines during scrolling, but auto-scrolling when the joystick is at max up/down doesn't always triger, just like for the DS4.

Same behavior with wired PS2 controllers using the blue dual port USB adapters, the right stick is also mapped but the mapping is incorrect (Left goes up, Right goes down, Up goes left, Down goes right), not sure if it matters since most will use the left stick but I thought I should mention it.

Gemba:
Hmmm.
If this change relates to my [suggested changes](https://retropie.org.uk/forum/topic/27128/enhancements-for-dialog-navigation-controls) feel free to yank some parts of it for your change here.
I tested my proposed change and it works correctly with D-pad as well as with analog hat.

cmitu:
I re-tested using a Wii U Pro controller, which doesn't have motion sensors.
The behavior for scrolling is similar to the PS3 - it doesn't skip the entries on the list when operating the analog joysticks. What's different is that the right joystick works similar to the left one - i.e. up/down scrolls in the list, while left/right switches between the dialog buttons.

For comparison here is the auto-configuration for both controllers - https://gist.github.com/cmitu/df5fc1cb0d6441566593a6cd7aa7613e.

joolswills:
I didn't make any changes regarding the mapping btw. Just auto repeat behaviour, or are you saying mapping stuff has changed?

joolswills:
@Gemba I'll take a look. My changes did actually work fine with the controller's I tested on.

joolswills:
Closing as the sdl2 solution is better.

PR #3203, opened 2020-07-23
LZDOOM - Added joypad support for menus

madmodder123:
Added fixes from https://github.com/protocultor/gzdoom/commit/ce79b391f7f6fc8417b98a1caeae9b1affd55024
Also enabled joypads by default for raspberry pi (I set the default deadzones to 0.1 instead of 0 so the analog sticks don't spam the menu controls, it can still be adjusted lower than 0.1 if wanted).
You can also map the menu button to whatever you want.
Now you don't need a keyboard at all to set up or play the game.

Tested and working on RPI 4 with lzdoom 3.86a (running retropie 4.6)

### DIFF
--- a/scriptmodules/ports/lzdoom.sh
+++ b/scriptmodules/ports/lzdoom.sh
@@ -26,6 +26,9 @@ function depends_lzdoom() {
 
 function sources_lzdoom() {
     gitPullOrClone "$md_build" https://github.com/drfrag666/gzdoom "3.86a"
+	if isPlatform "rpi"; then
+		applyPatch "$md_data/01_rpi_fixes.diff" # Enables use of joypad to control menus so you don't need a keyboard!
+	fi
 }
 
 function build_lzdoom() {
--- a/scriptmodules/ports/lzdoom.sh
+++ b/scriptmodules/ports/lzdoom.sh
@@ -26,8 +26,10 @@ function depends_lzdoom() {
 
 function sources_lzdoom() {
     gitPullOrClone "$md_build" https://github.com/drfrag666/gzdoom "3.86a"
-	if isPlatform "rpi"; then
+	if isPlatform "rpi" && ! grep -q CNTRLMNU_OPEN_MAIN "$md_build/wadsrc/static/language.enu"; then
+		# Failsafe to apply joypad patch if it hasn't been added upstream yet https://github.com/coelckers/gzdoom/pull/1072
 		applyPatch "$md_data/01_rpi_fixes.diff" # Enables use of joypad to control menus so you don't need a keyboard!
+		# Safe to remove this patch if LZDOOM gets updated beyond 3.86a and includes joypad support for menus
 	fi
 }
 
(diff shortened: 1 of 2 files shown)


hhromic:
Hi, thanks for the contribution. I think you should try to get this merged upstream instead of here in RetroPie.
Once merged, we can bump the version in RetroPie that includes this change.

madmodder123:
> 
> 
> Hi, thanks for the contribution. I think you should try to get this merged upstream instead of here in RetroPie.
> Once merged, we can bump the version in RetroPie that includes this change.
The original author made a pull request on April 20th for the ORIGINAL GZDOOM repo (https://github.com/coelckers/gzdoom/pull/1072) that hasn't been merged yet.
LZDOOM was recently updated to 3.86a* (This was announced as the final lzdoom release to support GL2 rendering https://github.com/RetroPie/RetroPie-Setup/commit/754024ab67eef3885041336f539a7f6dd6800c02) so I wasn't sure if we would continue using newer versions.
Could I just edit the code to check if the source files have been updated to include the fix (patch if they haven't, ignore if they have been updated), as a temporary fix until they get the code merged (if at all)?

madmodder123:
Oh, it looks like he also made a [commit in April for LZDOOM](https://github.com/protocultor/gzdoom/commit/8ba7e49184c2be58c0ad03a7602dcd2c4b0ff6fd) as well as the original GZDOOM repo. It hasn't been merged yet either, I also noticed his defbinds.txt is out of date, so it would fail if someone tried to patch 3.86a with that commit (mine works fine though).

joolswills:
Thanks. Patch looks pretty clean but might be worth querying on the existing PR on upstream if this or an alternative would be accepted. Patches like this need to be maintained and it can be problematic due to the amount of software installable by retropie.

Would prefer to wait for a comment from upstream tbh.

madmodder123:
> 
> 
> Thanks. Patch looks pretty clean but might be worth querying on the existing PR on upstream if this or an alternative would be accepted. Patches like this need to be maintained and it can be problematic due to the amount of software installable by retropie.
> 
> Would prefer to wait for a comment from upstream tbh.

I'll ask around the upstream repo, thanks for the input!

protocultor:
It's done at GZDoom at least:
https://github.com/coelckers/gzdoom/pull/1163

drfrag666:
@protocultor You said "I have yet to see another controller that reaches a number that high of buttons, in any platform." What happens if there are more buttons? Nothing AFAIK. LZDoom supports up to 4 controllers at the same time with 32 buttons each for fake splitscreen.
BTW i've extended support for the old g3.3mgw branch a bit more and i'll release 3.87 in a few days. Due to my bad personal situation i'm forced to add paypal.me and ko-fi links to my GitHub i hope for a very limited time.

protocultor:
@drfrag666 I was trying to justify the inclusion of buttons 14 and 15 as "cancel" and "accept" for menus, respectively. As you say, even if there are more buttons, it just means there are more buttons to accept and cancel in menus.
Good luck with your personal situation; these are difficult times for many of us.

drfrag666:
Merged upstream but i haven't pushed it yet, it's in the legacy branch for now.

joolswills:
I'm going to close this as the needed changes will be included in the next release as mentioned.

Btw @drfrag666 would we be better off tracking a development branch rather than the last release for this?

protocultor:
Thanks to the author, but at this point in time, this is worth closing, since the "official" releases already include everything.
All of the backend changes were already included in version _3.87c_, and the only one missing, the "open menu with controller" button in the options, appears in _3.88b_.

joolswills:
Thank you!

PR #3207, opened 2020-08-04
sdltrs: switch to an active fork, build with SDL2

cmitu:
Switched SDLTRS to be built from https://gitlab.com/jengun/sdltrs, which is an active fork of the original SDLTRS emulator.

* use the SDL2 branch, which should be more portable (KMS/Odroid/etc.)
* set a minimum of configuration dir in the default configuration file ($HOME/.sdltrs.t8c)
* although not explicitely specified, the main license seems to be BSD-2-Clause (https://opensource.org/licenses/BSD-2-Clause)

### DIFF
--- a/scriptmodules/emulators/sdltrs.sh
+++ b/scriptmodules/emulators/sdltrs.sh
@@ -13,47 +13,60 @@ rp_module_id="sdltrs"
 rp_module_desc="Radio Shack TRS-80 Model I/III/4/4P emulator"
 rp_module_help="ROM Extension: .dsk\n\nCopy your TRS-80 games to $romdir/trs-80\n\nCopy the required BIOS file level2.rom, level3.rom, level4.rom or level4p.rom to $biosdir"
 rp_module_section="exp"
-rp_module_flags="dispmanx !mali !kms"
+rp_module_licence="BSD https://gitlab.com/jengun/sdltrs/-/raw/master/LICENSE"
+rp_module_flags=""
 
 function depends_sdltrs() {
-    getDepends libsdl1.2-dev libxt-dev
+    getDepends libsdl2-dev libreadline-dev
 }
 
 function sources_sdltrs() {
-    gitPullOrClone "$md_build" https://github.com/RetroPie/sdltrs.git
+    gitPullOrClone "$md_build" https://gitlab.com/jengun/sdltrs.git sdl2
 }
 
 function build_sdltrs() {
-    cd src/linux
     make clean
+    ./autogen.sh
+    ./configure --prefix="$md_inst"
     make
-    md_ret_require="$md_build/src/linux/sdltrs"
+    md_ret_require="$md_build/sdl2trs"
 }
 
 function install_sdltrs() {
-    chmod 644 README
     md_ret_files=(
-        'src/linux/sdltrs'
-        'README'
+        'sdl2trs'
+        'README.md'
+        'LICENSE'
     )
 }
 
 function configure_sdltrs() {
+    local common_args
     mkRomDir "trs-80"
 
-    addEmulator 1 "$md_id-model1" "trs-80" "$md_inst/sdltrs -model 1 -romfile $biosdir/level2.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
-    addEmulator 0 "$md_id-model3" "trs-80" "$md_inst/sdltrs -model 3 -romfile3 $biosdir/level3.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
-    addEmulator 0 "$md_id-model4" "trs-80" "$md_inst/sdltrs -model 4 -romfile3 $biosdir/level4.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
-    addEmulator 0 "$md_id-model4p" "trs-80" "$md_inst/sdltrs -model 4p -romfile4p $biosdir/level4p.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
+    common_args="-fullscreen -nomousepointer -showled"
+    addEmulator 1 "$md_id-model1" "trs-80" "$md_inst/sdl2trs $common_args -m1 -romfile $biosdir/level2.rom -disk0 %ROM%"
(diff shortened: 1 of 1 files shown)


scriptmodules/emulators/sdltrs.sh
```diff
@@ -10,46 +10,56 @@
 #
 
 rp_module_id="sdltrs"
-rp_module_desc="Radio Shack TRS-80 Model I/III/4/4P emulator"
+rp_module_desc="Radio Shack TRS-80 Model I/III/4/4P emulator (SDL2 version)"
```
joolswills, line 13: Don't think we need to add SDL2 version - if we are dropping the previous.

scriptmodules/emulators/sdltrs.sh
```diff
@@ -10,46 +10,56 @@
...
-    addEmulator 0 "$md_id-model3" "trs-80" "$md_inst/sdltrs -model 3 -romfile3 $biosdir/level3.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
-    addEmulator 0 "$md_id-model4" "trs-80" "$md_inst/sdltrs -model 4 -romfile3 $biosdir/level4.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
-    addEmulator 0 "$md_id-model4p" "trs-80" "$md_inst/sdltrs -model 4p -romfile4p $biosdir/level4p.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
+    addEmulator 1 "$md_id-model1" "trs-80" "$md_inst/sdl2trs -fullscreen -nomousepointer -model 1 -romfile $biosdir/level2.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
+    addEmulator 0 "$md_id-model3" "trs-80" "$md_inst/sdl2trs -fullscreen -nomousepointer -model 3 -romfile3 $biosdir/level3.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
+    addEmulator 0 "$md_id-model4" "trs-80" "$md_inst/sdl2trs -fullscreen -nomousepointer -model 4 -romfile3 $biosdir/level4.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
+    addEmulator 0 "$md_id-model4p" "trs-80" "$md_inst/sdl2trs -fullscreen -nomousepointer -model 4p -romfile4p $biosdir/level4p.rom -showled -diskdir $romdir/trs-80 -disk0 %ROM%"
```
joolswills, line 49: if adding more parameters, I think it might be a good time to dedupe this a bit - can use a variable or an array or so.

scriptmodules/emulators/sdltrs.sh
```diff
@@ -10,46 +10,56 @@
...
     addSystem "trs-80"
 
     [[ "$md_mode" == "remove" ]] && return
 
-    moveConfigFile "$home/sdltrs.t8c" "$md_conf_root/trs-80/sdltrs.t8c"
```
joolswills, line 52: The config location changed or the old one was incorrect ?

If it moved, and is the same format, it should be preserved/migrated.

scriptmodules/emulators/sdltrs.sh
```diff
@@ -10,46 +10,56 @@
...
+    iniConfig "=" "" "$home/.sdltrs.t8c"
+    iniSet "statedir"   "$romdir/trs-80"
+    iniSet "diskdir"    "$romdir/trs-80"
+    iniSet "cassdir"    "$romdir/trs-80"
+    iniSet "disksetdir" "$romdir/trs-80"
+    iniSet "harddir"    "$romdir/trs-80"
+
+    moveConfigFile "$home/.sdltrs.t8c" "$md_conf_root/trs-80/sdltrs.t8c"
+    chown "$user:$user" "$md_conf_root/trs-80/sdltrs.t8c"
```
joolswills, line 62: This will override any users custom configs for those values - so it should create a config in a temporary location and use "copyDefaultConfig" probably. 

Are all these needed as diskdir is also set on the commandline etc.

scriptmodules/emulators/sdltrs.sh
cmitu: Yes, that would clear things a bit.

scriptmodules/emulators/sdltrs.sh
cmitu: The `sdltrs.t8c` is default on Windows, on Linux is `.sdltrs.t8c (https://gitlab.com/jengun/sdltrs/-/blob/sdl2/src/trs_sdl_interface.c#L926). I assumed it was wrong, but in the RP repo - on which the current version is based - this is actually correct (https://github.com/RetroPie/sdltrs/blob/6c6935a0c730f65185c36b623f2c2f8f6d0ad3cf/src/trs_sdl_interface.c#L900). 
I'll redo this part.

scriptmodules/emulators/sdltrs.sh
cmitu: I don't know why I always think `iniSet` is not overwriting existing values. I'll adjust the configuration part and see if we can get rid of `diskdir` (might be un-necessary).

cmitu:
@joolswills thanks for the review, I'll adjust the scriptmodule.

scriptmodules/emulators/sdltrs.sh
joolswills: Or it might be better to include it in the config only rather than forcing on commandline - which would be more flexible perhaps for users.

cmitu:
- I added the changes for migrating the user settings - the new version kept the format and the options.
- emulator commandline - I removed the `diskdir` param, since it's not strictly needed for running the ROM,  `diskdir` is set in the default config to the same folder. I kept the `romfile` parameter, since it allows to specify a romfile for each of the 4 models, while the configuration file only knows about 3 romfiles (i.e. uses `romfile3` for both Model 3 and Model 4 emulation). 
- I removed the symlinks for the `romfile`s, since we pass the full path names to the emulator, they don't need to be in the binary install folder.

I reverted the description change, makes sense.

joolswills:
Awesome. Thanks.

PR #3210, opened 2020-08-08
amiberry - adjust amiberry launch script for argument changes

joolswills:
* use "--config " instead of "-config=" and "--autoload "
 * may adjust this further and split out amiberry launch to its own code rather than sharing the uae4arm launch code


joolswills:
Binaries updated.

PR #3211, opened 2020-08-09
runcommand - fix aspect ratio using dispmanx backend with sdl1 on fkms

joolswills:
Added sdl1 patch to allow overriding of graphics_get_display_size as it returns 0 on fkms for width and height,
which skips the scaling code and ability to adjust aspect ratio.

New ENV vars are SDL_DISPMANX_WIDTH and SDL_DISPMANX_HEIGHT which we now set manually in runcommand on fkms
when a dispmanx backend is used with sdl1.

This should resolve aspect ratio issues rpi4 users were having with some sdl1 emulators that worked with the dispmanx
backend, but always stretched to fill the entire screen.

This also enables the use of SDL_DISPMANX_RATIO to tweak the ratio when scaling - from 0.2 to 6 - currently
unused in any RetroPie modules by default, but used to be used for tweaking the vice scaling, to more accurately
produce a correct aspect ratio for the C64. But can be used by users manually via emulators.cfg if required.

bumped sdl1 version to force an update - including for Stretch, although this is primarily a buster/rpi4 issue

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -1019,7 +1019,12 @@ function config_dispmanx() {
     if [[ -f "$DISPMANX_CONF" ]]; then
         iniConfig " = " '"' "$DISPMANX_CONF"
         iniGet "$name"
-        [[ "$ini_value" == "1" ]] && COMMAND="SDL1_VIDEODRIVER=dispmanx $COMMAND"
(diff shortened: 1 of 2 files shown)


joolswills:
Seems to be ok from my testing.

Will require users to update RetroPie-Setup and then manually update sdl1 from Manage packages -> depends

Or update RetroPie-Setup and then re-install a sdl1 application which will also trigger sdl1 to update

joolswills:
Oops. Wrong button :-)

PR #3213, opened 2020-08-17
ti99sim: use an older version for Debian Stretch

cmitu:
Added back the `dispmanx` flag for the - older - SDL1 version.
Uses the[0.15.0](https://www.mrousseau.org/programs/ti99sim/archives/ti99sim-0.15.0.src.tar.xz) version when compiled with SDL1, which needs to be copied to the RetroPie archive web server.

### DIFF
--- a/scriptmodules/emulators/ti99sim-sdl1.sh
+++ b/scriptmodules/emulators/ti99sim-sdl1.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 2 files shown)


joolswills:
I'd prefer this to be split into two modules or else we get a non working dispmanx setting for the sdl2 version. Also it's better to use GCC for the comparison as it's a GCC version issue.

joolswills:
The sdl1 version can be called ti99sim-sdl1 as we do with scummvm.

cmitu:
@joolswills sure, we can split it. Should we disable the `-sdl1` version for KMS systems ?

joolswills:
Can leave it enabled if it works on kms.

joolswills:
Looks great. Thanks.

PR #3214, opened 2020-08-20
openmsx: make the system ROM folder available in $biosdir

cmitu:
Allows users to copy the necessary OpenMSX boot ROMs via file shares (`\\retropie\bios\openmsx`).

### DIFF
--- a/scriptmodules/emulators/openmsx.sh
+++ b/scriptmodules/emulators/openmsx.sh
@@ -54,4 +54,8 @@ function configure_openmsx() {
 
     addEmulator 0 "$md_id" "msx" "$md_inst/bin/openmsx %ROM%"
     addSystem "msx"
+
+    mkUserDir "$home/.openMSX/share/systemroms"
+    chown -R "$user" "$home/.openMSX"
+    moveConfigDir "$home/.openMSX/share/systemroms" "$biosdir/openmsx"
 }


scriptmodules/emulators/openmsx.sh
```diff
@@ -54,4 +54,8 @@ function configure_openmsx() {
 
     addEmulator 0 "$md_id" "msx" "$md_inst/bin/openmsx %ROM%"
     addSystem "msx"
+
+    mkUserDir "$home/.openMSX/share/systemroms"
```
joolswills, line 58: This could fail if the folder doesn't exist?

Probably best to symlink the entire config folder as we do elsewhere also.

joolswills (changes requested):

scriptmodules/emulators/openmsx.sh
cmitu: No, `mkUserDir` uses `mkdir -p`, which created the parent folders if it's not necessary.

https://github.com/RetroPie/RetroPie-Setup/blob/87839ed171dc3fa99adef80869fa152e19452091/scriptmodules/helpers.sh#L468-L475

joolswills: True - I should know this. However it will make them as root and your chosen isn't $user:$user

But I still prefer the main config symlinked to match elsewhere - and moveCondigDir is basically designed to do that, rather than a subfolder.

This is an old module as I guess it got overlooked.

cmitu:
I have a few small additions for `openmsx` - including a gamepad config generation script for EmulationStation.
I can squash this change into that since it might - indeed - be beneficial to have the entire `openmsx` config dir available in `$configdir` (so it can be inspected/edited etc.).

joolswills:
Sounds good. Cheers.

PR #3220, opened 2020-09-08
system / packages - added signing of binary packages

joolswills:
* default signing key can be overridden with __gpg_signing_key
 * added gpg dependency by default
 * download retropie key from ubuntu keyserver if it doesn't exist in keyring
 * added download helper function
 * rework logic for rp_installBin and rp_createBin
 * copy secret key to chroots for package building

All core/opt bins are now signed. Some experimental packages are still being built still but all bins will be signed soon.

After this is included I may switch away from wget to curl due to wget limitations with its use of  stderr (see download function).

But first will reduce number of modules handling downloads themselves to use the download function. Also the download and extract function will be reworked etc.

But the main premise behind this is to reduce the risk of installing prebuilt binaries, for example if our server was exploited and someone replaced a core binary with something malicious.

### DIFF
--- a/scriptmodules/admin/builder.sh
+++ b/scriptmodules/admin/builder.sh
@@ -96,7 +96,13 @@ function chroot_build_builder() {
         [[ ! -d "$md_build/$dist" ]] && rp_callModule image create_chroot "$dist" "$md_build/$dist"
         if [[ ! -d "$md_build/$dist/home/pi/RetroPie-Setup" ]]; then
             sudo -u $user git clone "$home/RetroPie-Setup" "$md_build/$dist/home/pi/RetroPie-Setup"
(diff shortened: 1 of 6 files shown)


joolswills:
This is ready for testing now.

joolswills:
Sorry. Actually I'm going to make some more changes.

joolswills:
Did a bunch more changes. Split out some stuff into functions and adjusted sdl1/2 modules accordingly so they use signed files too. 

(Technically could sign the debs directly but this is simpler, and I'm considering reworking sdl1/2 debs to maybe just be included in a standard retropie package and installed from there. Would be simpler.)

This should work, but my function help is very rushed so I am planning to go over that.

Probably bugs and typos. I will review but feel free to test in the meantime.

joolswills:
Merging.

PR #3221, opened 2020-09-12
Compatibility for Jetson Nano and Nintendo Switch

theofficialgman:
I'm aware that this change was asked for many months/years ago. We have a growing userbase using our fork of RetroPie on their nintendo switch's and would like this changed merged into master.

The reasoning for denying last time was the lack of implementing using a hardware string. The hardware string is a ARM kernel ONLY feature and is not officially supported on any other kernels (including ARM64) so this way of testing for hardware is not an option. Checking in /sys/firmware/devicetree/base/model is currently the best was of indentifying the device since all nintendo switch's with linux built for them have icosa as their default model as well as the jetsons using NVIDIA Jetson Nano Developer Kit.

If you have comments or suggestions, I have open ears.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -376,12 +376,19 @@ function get_platform() {
             "Allwinner sun8i Family")
                 __platform="armv7-mali"
                 ;;
-            *)
-                case $architecture in
-                    i686|x86_64|amd64)
-                        __platform="x86"
-                        ;;
-                esac
+            *)  #For now nvidia jetson nano no have Hardware name string by /proc/cpuinfo you can info by base/model 
+                if grep -q "NVIDIA Jetson Nano Developer Kit" /sys/firmware/devicetree/base/model 2>/dev/null; then
+                    __platform="jetson-nano" 
+
+                elif  grep -q "icosa" /sys/firmware/devicetree/base/model 2>/dev/null; then
+                    __platform="tegra-x1" 
+                else
+                    case $architecture in
+                        i686|x86_64|amd64)
+                            __platform="x86"
+                            ;;
+                    esac
+                fi
                 ;;
         esac
     fi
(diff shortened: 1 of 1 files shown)


joolswills:
Actually the reasoning was that it had formatting issues and needed cleaning up. 

This PR will need some changes for the same reason I'm afraid, but if prefer to do it myself. I'll look into adding it soon.

joolswills:
The way the platform flags are set etc has changed also.

theofficialgman:
I see. You are probably looking for something more like this for the flags at least. Let me know if there is anything I can help with

scriptmodules/system.sh
```diff
@@ -473,6 +480,20 @@ function platform_odroid-xu() {
...
+    __default_cpu_flags="-march=armv8-a+crc -mcpu=cortex-a57 -mtune=cortex-a57"
+    __platform_flags+=(aarch64 x11 gl)
+    __default_cflags+=" -ftree-vectorize -funsafe-math-optimizations"
+    __default_asflags=""
+}
+
+function platform_tegra-x1() {
+    __default_cpu_flags="-march=armv8-a+crc -mcpu=cortex-a57 -mtune=cortex-a57"
+    __platform_flags+=(aarch64 x11 gl)
+    __default_cflags+=" -ftree-vectorize -funsafe-math-optimizations -ffast-math -fPIE"
```
joolswills, line 493: why -fPIE - I think that probably should be left out.

Also the other cflags should be excluded from here also (See other platforms - we don't include them by default). fast-math could cause incompatibility issues also.

joolswills:
I did notice though that I missed the correct platform_flags for odroid-xu so I will fix that.

scriptmodules/system.sh
theofficialgman: I see. I was looking at the odroid and tinker for reference which do have cflags defined, which is why I left them in from the previous users config

theofficialgman: maybe it would be best to leave out those math based optimizations though and leave enabling those to the cores themselves if they can handle it

joolswills: Yeah. It's safer I think.

joolswills:
Please try our #3222

joolswills:
It's based on the changes provided and the previous PR. But with some logic tweaks. And simplified GCC args

No need for mtune if setting mcpu

This will fail if there's any 32bit OS Linux that works on them. Will need additional logic then. Also as I own neither hardware I should be considered unsupported.

joolswills:
Thanks for your PR. It was useful to implement the changes even if I did it slightly differently. Appreciated.

theofficialgman:
we can't run 32bit os anyway, nvidia doesn't give the required support for the tegra x1 gpu

theofficialgman:
no problem, I just wanted it implemented. I don't really care how it is done or who ends up doing it

joolswills:
Cheers. Appreciated.

joolswills:
Thanks for bringing this back to my attention. Closing as implemented. Feel free to contribute additional fixes.

PR #3222, opened 2020-09-12
system - added recognition for Jetson Nano and Tegra X1

joolswills:
* by user request - not officially supported

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -377,11 +377,26 @@ function get_platform() {
                 __platform="armv7-mali"
                 ;;
             *)
-                case $architecture in
-                    i686|x86_64|amd64)
-                        __platform="x86"
-                        ;;
-                esac
+                # jetson nano and tegra x1 can be identified via /sys/firmware/devicetree/base/model
+                local model_path="/sys/firmware/devicetree/base/model"
+                if [[ -f "$model_path" ]]; then
+                    # ignore end null to avoid bash warning
+                    local model=$(tr -d '\0' <$model_path)
+                    case "$model" in
+                        "NVIDIA Jetson Nano Developer Kit")
+                            __platform="jetson-nano"
+                            ;;
+                        icosa)
+                            __platform="tegra-x1"
+                            ;;
+                    esac
+                else
+                    case $architecture in
+                        i686|x86_64|amd64)
+                            __platform="x86"
+                            ;;
+                    esac
+                fi
                 ;;
         esac
     fi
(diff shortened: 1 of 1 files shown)


scriptmodules/system.sh
```diff
@@ -377,11 +377,26 @@ function get_platform() {
...
+                # jetson nano and tegra x1 can be identified via /sys/firmware/devicetree/base/model
+                local model_path="/sys/firmware/devicetree/base/model"
+                if [[ -f "$model_path" ]]; then
+                    # ignore end null to avoid bash warning
+                    local model=$(tr -d '\0' <$model_path)
+                    case "$model" in
+                        "NVIDIA Jetson Nano Developer Kit")
+                            __platform="jetson-nano"
+                            ;;
+                        *icosa*)
```
joolswills, line 389: Wasn't sure if this was requiring a partial match or not.

theofficialgman: it should be icosa and that's it

joolswills:
This is hardcoded for only aarch64 for these boards btw. I don't have anything to test with. If there's more arm x11 + GL boards I may rework it somewhat. I have plans to split out the systems configs after more refactoring anyway.

scriptmodules/system.sh
joolswills: I'll remove the wildcards

joolswills: Done.

joolswills: Thanks.

theofficialgman:
nvidia renames the model if its not a tegra x1 so there shouldn't be any work of it not being aarch64

joolswills:
Thanks. Please let me know if this works and I'll merge it.

theofficialgman:
Just finished testing your latest changes (37c14a2) and built a system that it compatible (lr-parallel-n64). Seems to work as expected.

joolswills:
I just made one further change. Since they both share the same platform specs as far as we are concerned, I made the jetson nano share the tegra-x1 platform config. Just a bit of deduplication

joolswills:
Thanks. My last change shouldn't affect anything so I'll merge in a bit.

theofficialgman:
that's fine. jetson nano really only differs in number of gpu cuda cores which isn't affected here

joolswills:
Interesting. I know very little about these systems. Maybe I should get a second hand switch to have a play.

theofficialgman:
you have to get an original switch model (not the new switch) since you can't hack those

PR #3223, opened 2020-09-13
Default Core Options for lr-mupen64plus-next

dankcushions:
Parity with the defaults for mupen64plus-GLideN64 standalone.


joolswills (approved):
Thanks. Looks good.

PR #3224, opened 2020-09-15
openmsx: added joystick autoconfiguration, config tweaks and enabled for 'mali'

cmitu:
* added a joystick auto-configuration for Emulationstation. Joypad mappings:
  * A: MSX controller A; OSD menu: Back/Cancel
  * B: MSX controller B; OSD menu: Action (toggle/execute menu item)
  * D-Pad/Left Analog: MSX controller joystick; OSD Menu: Up/Down/Left/Right
  * Start - Toggle the openMSX OSD menu
  * Select - Show on-screen keyboard
  * Y - MSX F1
  * X - MSX F2
  * Left Trigger - MSX F3
  * Right Trigger - MSX F4

  The joystick's A/B/D-Pad are functional in the openMSX menu, however B is action and A is cancel/back for some reason.

  The joystick is loaded using the `retropie-init.tcl` script, automatically loaded and executed during startup. Per-game overrides are possible, as long as the correct configuration script for a gamepad is found (needs some documentation).

* created a minimal configuration, with some performance optimization for Pi0/Pi1 and configuring a default machine.
  Auto-saving is disabled, because it will save the joystick configuration and it might pose problems when using different joysticks.

* symlinked the configuration folder into `$configdir/msx/openmsx`.
  This is useful for adding new machines/configurations and editing the gamepad auto-configurations.

* symlinked the user's _systemroms_ folder to `$biosdir/openmsx`.
  openMSX uses file hashes to identify firmware files, so the filenames/folder structure is not important.

* added one emulator variation for each major MSX model (MSX2/MSX2+/MSX TurboR)
  They need the appropriate BIOS/firmware files, otherwise they won't run.
  NOTE: there are variations of machines using the open source C-BIOS for each major model, but C-BIOS only supports cartridges, so no tape/disk/floppy images will work.

### DIFF
--- a/scriptmodules/emulators/openmsx.sh
+++ b/scriptmodules/emulators/openmsx.sh
@@ -14,7 +14,7 @@ rp_module_desc="MSX emulator OpenMSX"
 rp_module_help="ROM Extensions: .rom .mx1 .mx2 .col .dsk .zip\n\nCopy your MSX/MSX2 games to $romdir/msx"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/openMSX/openMSX/master/doc/GPL.txt"
 rp_module_section="opt"
-rp_module_flags="!mali"
+rp_module_flags=""
 
 function depends_openmsx() {
     local depends=(libsdl2-dev libsdl2-ttf-dev libao-dev libogg-dev libtheora-dev libxml2-dev libvorbis-dev tcl-dev libasound2-dev)
--- a/scriptmodules/emulators/openmsx.sh
+++ b/scriptmodules/emulators/openmsx.sh
@@ -11,7 +11,7 @@
 
 rp_module_id="openmsx"
 rp_module_desc="MSX emulator OpenMSX"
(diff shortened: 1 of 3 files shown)


cmitu:
I'll try to find a way to make 'A'/'B' behave similarly to the RGUI behavior, but other than that it should be fine.

joolswills:
Lots to review. Looks good. Extra points for coding anything in TCL :-)

cmitu:
I added a small modification to the bindings for _Start_ to - really - work as a toggle for the OSD Menu.

While it's possible to swap A/B in the OSD menu (to behave similarly to RGUI), it looks like the button configuration is reset each time the OSD menu is re-created - i.e. the _Exit_ prompt re-creates an OSD menu dialog and A/B behaviour is swapped back to the defaults. There might be other similar pop-up menus, so for consistency I'll leave it as is.

joolswills:
I never got around to testing this sorry. However, it looks decent, it's one emulator and I know a lot of work was put in, so will merge. Sorry for the delay.

Although I am impressed seeing tcl used, I'm not sure it aligns with the ease of use¹ our existing code demonstrates, but we will adjust!

 ¹This is sarcasm btw.

PR #3232, opened 2020-09-26
Update cpu flags for tegra x1 devices.

mrcmunir:
Add cpu specific flags for tegra tx1 devices  (crc+fp+simd)

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -494,7 +494,7 @@ function platform_odroid-xu() {
 }
 
 function platform_tegra-x1() {
-    __default_cpu_flags="-mcpu=cortex-a57"
+    __default_cpu_flags="-march=armv8-a+crc -mtune=cortex-a57 -mcpu=cortex-a57+crc+fp+simd"
     __platform_flags+=(aarch64 x11 gl)
 }
 


joolswills:
Please can you give some information as to why you think this is needed?

https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html

>-mcpu=name[+extension…]

>**This specifies the name of the target ARM processor. GCC uses this name to derive the name of the target ARM architecture (as if specified by -march) and the ARM processor type for which to tune for performance (as if specified by -mtune)**. Where this option is used in conjunction with -march or -mtune, those options take precedence over the appropriate part of this option.

>Many of the supported CPUs implement optional architectural extensions. Where this is so the architectural extensions are normally enabled by default. If implementations that lack the extension exist, then the extension syntax can be used to disable those extensions that have been omitted. For floating-point and Advanced SIMD (Neon) instructions, the settings of the options -mfloat-abi and -mfpu must also be considered: floating-point and Advanced SIMD instructions will only be used if -mfloat-abi is not set to ‘soft’; and any setting of -mfpu other than ‘auto’ will override the available floating-point and SIMD extension instructions.

>For example, ‘cortex-a9’ can be found in three major configurations: integer only, with just a floating-point unit or with floating-point and Advanced SIMD. **The default is to enable all the instructions, but the extensions ‘+nosimd’ and ‘+nofp’ can be used to disable just the SIMD or both the SIMD and floating-point instructions respectively.**

mrcmunir:
@joolswills oh i see I added this due In my tests I saw that you will gain some performance stability if I specify it (not to much but 2-3fps).. but it may be that it is something very specific case to gcc7 version .
It seems perfect to me if it is not specified by default this flags (in good use safe anyway)
On the other hand, do you have any discord RetroPie or maybe I Open thread in the forums?  
Some recommendations for it I seen in very test diferences compilations and reports in ms perfomance with some options specially in Retroarch.
It's very useful in tegra x1 devices:
-Disable Vysnc Option on Retroarch gain perfomance in some cores n64, dc , psx prevent stuterring in few hard games in big resolution.
 I don't see if some option for control Vsync ON or off in the init the script?
-Enable Thread video Retroarch option big up perfomance helped a lot on Yabasanshiro , n64 , psx , dc , psp so video_threaded = true will be needed specific isPlatform "jetson-nano" too (maybe others boards aarch64 useful too for big fps specially in cores multithread). I see in the Retroarch.sh is only x86 enabled for this.

dankcushions:
can you show the benchmarks?

> I don't see if some option for control Vsync ON or off in the init the script?

it's not set by retropie so it uses retroarch defaults. you could turn it on or off in /all/retroarch.cfg

> -Enable Thread video Retroarch option big up perfomance helped a lot on Yabasanshiro , n64 , psx , dc , psp so video_threaded = true will be needed specific isPlatform "jetson-nano" too (maybe others boards aarch64 useful too for big fps specially in cores multithread). I see in the Retroarch.sh is only x86 enabled for this.

it's on my default for all _except_ x86: https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/retroarch.sh#L163-L165

mrcmunir:
> can you show the benchmarks?

In 64/dc cores helped light better gives the feeling of holding up 2-3fps faster stability (according to the retroarch counter maybe false positive but.. not sure how precise counter Retroarch it for fps number)

> it's not set by retropie so it uses retroarch defaults. you could turn it on or off in /all/retroarch.cfg

It is what I was doing for now manually (but I did not know if it could be done better) is what I tell users with performance stuterring problems in some games from 720p it makes strange behavior with vsync especially in psx/n64 cores stuterring issue when vsync Retroarch enable.
 
> it's on my default for all _except_ x86: https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/retroarch.sh#L163-L165

Ops I miss "!" x86 platform so see that now ok correct. Thanks for it :)

PR #3238, opened 2020-10-11
Recent change in cannonball cmake files makes RetroPie build/installation process no longer function

hotspoons:
This commit: https://github.com/djyt/cannonball/commit/ffac08c25c2937d046e376ed59a5f0154c38d969 broke the build and installation process for the cannonball engine as designed in RetroPie-Setup. By updating the arguments to gitPullOrClone to lock the version at a previously-buildable version, cannonball is now able to build and install correctly.

A better solution would be to update the build process here to work with the new cmake files in the master version of their repository, but this fix will at least allow the engine to function.

### DIFF
--- a/scriptmodules/ports/cannonball.sh
+++ b/scriptmodules/ports/cannonball.sh
@@ -23,7 +23,7 @@ function depends_cannonball() {
 }
 
 function sources_cannonball() {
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Yeah it also seems to want a very new cmake. Whether it's actually required I'd have to check. I'll merge this and we can review updating the retropie code later. Cheers.

PR #3239, opened 2020-10-13
lr-hatari: update local patch to fix build

cmitu:
Build broke because of upstream commit 05c7f4e9f399c2 (Makefile update).


joolswills:
Thanks.

PR #3240, opened 2020-10-17
lr-vecx: update build for GPU rendering support

cmitu:
The core now uses GPU rendering and it needs additional parameters for compilation.

Not sure  how it works on other GLES platforms (Odroid/Tinker/etc), it assumes that `libGLESv2` is the name of the GLES lib. Would be useful to test at least on Odroid.

### DIFF
--- a/scriptmodules/libretrocores/lr-vecx.sh
+++ b/scriptmodules/libretrocores/lr-vecx.sh
@@ -20,8 +20,12 @@ function sources_lr-vecx() {
 }
 
 function build_lr-vecx() {
+    local params
+    isPlatform "videocore" && params+="platform=rpi"
+    isPlatform "gles" && params+=" HAS_GLES=1"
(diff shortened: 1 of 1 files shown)


scriptmodules/libretrocores/lr-vecx.sh
```diff
@@ -20,8 +20,12 @@ function sources_lr-vecx() {
 }
 
 function build_lr-vecx() {
+    local __flags
```
joolswills, line 23: Please remove __ prefix. We reserve that for globals.

joolswills (changes requested):

scriptmodules/libretrocores/lr-vecx.sh
cmitu: Ok, replaced.

PR #3244, opened 2020-10-23
Add GLES3.x flags for Odroid XU3 and XU4.

dankcushions:
XU3 supports GLES 3.0: https://www.hardkernel.com/shop/odroid-xu3/
XU4 supports GLES 3.1: https://www.hardkernel.com/shop/odroid-xu4-special-price/

However, XU only supports GLES 2.0: https://www.hardkernel.com/shop/odroid-xu/, so kept that as a separate platform.

May fix https://retropie.org.uk/forum/topic/28256/odroid-xu4-lr-flycast-input-causing-graphical-bug

(untested!)

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -369,8 +369,11 @@ function get_platform() {
             "Freescale i.MX6 Quad/DualLite (Device Tree)")
                 __platform="imx6"
                 ;;
-            ODROID-XU[34])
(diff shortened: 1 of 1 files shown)


joolswills:
Sorry but this isn't too useful - enabling gles3 isn't the issue. It needs testing first imho. I have been fixing up odroid issues recently and will visit this when I get a chance.

PR #3245, opened 2020-10-28
runcommand.sh multiple connected devices, pi4 bugfix and speedups

Nomarian:
As stated in [forum](https://retropie.org.uk/forum/topic/28237/runcommand-sh-line-1045-n-a-division-by-0) in raspbianOS on a pi4 you get a bash error whenever you launch a new game. This is because xrandr outputs 2 connected devices and MODE_CUR[5] array gets assigned as aspect ratio instead of resolution. This only looks for one output.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -243,14 +243,14 @@ function get_all_kms_modes() {
     declare -Ag MODE
     local default_mode="$(echo "$KMS_BUFFER" | grep -Em1 "^Mode:.*(driver|userdef).*crtc")"
     local crtc="$(echo "$default_mode" | awk '{ print $(NF-1) }')"
-    local crtc_encoder="$(echo "$KMS_BUFFER" | grep "Encoder map:" | awk -v crtc="$crtc" '$5 == crtc { print $3 }')"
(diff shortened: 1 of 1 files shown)


Nomarian:
of course, you could also have multiple monitors and get the same error message, but at least this fixes it for single outputs

cmitu:
The formatting makes the diff much larger than it should be, so it's hard to see just the changes you've added.  It also has unrelated changes - the `get_all_tvs_modes` function is not used in a X11 environment or on the Pi4.

Can you post the output of `xrandr --verbose` for your setup on pastebin.com ?

Nomarian:
Yeah, sorry about that. I can't even test it, OCD took over.


[xrandr output](https://tpaste.us/MQ51)


I have no clue if multiple displays are supported. if it does, this will definitely break that.

cmitu:
> ... which this doesn't really address for multiple primary displays.

I don't think the runcomand mode switching supports multiple displays - especially when they could have different resolutions. It doesn't know which one is used for display and emulators (RetroArch for instance) could have configurations to choose the display which can override the screen where ES (and `runcommand`) will run.

Nomarian:
And you can have multiple resolutions with different displays. which at least for me seems like a headache to handle.


Anyway, if you think the changes are too large. I think you should reject this PR, I will redo my repo and use small commits little by little this time.

PR #3247, opened 2020-10-30
runcommand.sh: Fixes x11 multiple output/display error

Nomarian:
Ok, small change of which you can actually see.

using {exit} on awk is equivalent to a | head -n1. So it exits at the first /connected/. thus filling the array with only one output.

Also why does grep -o even have {1,}, that's what + is for

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -382,9 +382,9 @@ function get_x11_mode_info() {
 
     if [[ -z "$mode_id" ]]; then
         # determine current output
-        mode_id[0]="$($XRANDR --verbose | grep " connected" | awk '{ print $1 }')"
+        mode_id[0]="$($XRANDR --verbose | awk '/ connected/ { print $1;exit }')"
         # determine current mode id & strip brackets
-        mode_id[1]="$($XRANDR --verbose | grep " connected" | grep -o "(0x[a-f0-9]\{1,\})")"
+        mode_id[1]="$($XRANDR --verbose | awk '/ connected/ {print;exit}' | grep -o "(0x[a-f0-9]\{1,\})")"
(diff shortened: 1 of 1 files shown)


cmitu:
> using {exit} on awk is equivalent to a | head -n1. So it exits at the first /connected/. thus filling the array with only one output.

Doesn't `grep -m1` give the same result ?

Nomarian:
Probably but it's not used in here.

cmitu:
Look ok as is, but I prefer `grep` to `awk` for readability.

I had a chance to test with a 2 screen setup - though not on a Pi. I don't get the error reported in the forums, but the video mode listing is broken/incorrect. Besides the modifications you added, adding a similar change to 
https://github.com/RetroPie/RetroPie-Setup/blob/2fd5e223f57277118916dc211bf0a1eeaf9525e6/scriptmodules/supplementary/runcommand/runcommand.sh#L284

```
        local output="$($XRANDR --verbose | grep -m1 " connected" | awk '{ print $1 }')"
```
will at least show the correct video modes for the 1st display's. What's the result on your setup ? If it's ok, then you can add this and I'd be happy to merge the changes.

Nomarian:
Really? even with xrandr reporting 2 connected devices? anyway, it launched, no errors, everything seems fine.

cmitu:
I was asking about the video mode listing, when you try to change the video mode from the runcommand launch menu. Do you get a correct listing with just this patch ?

Nomarian:
Yes. Before it just showed the output and the id, no resolution or aspect ratio listed at all!


Also, I didn't even know what you were talking about because the output was always incorrect until now.

cmitu:
> Yes. Before it just showed the output and the id, no resolution or aspect ratio listed at all!

Ok, thank you for the confirmation and the PR. 
I'll run an additional test on my setup see if needs an additional change for video mode listing.

PR #3251, opened 2020-11-02
lr-mame, mame: use python3 to build recent mame versions

cmitu:
The change impacts `lr-mame`,`lr-mess`,`lr-mame2016`,`lr-mess2016` and `mame`.
It should fix building on new Ubuntu 20.04+ installations, which don't have a default `python` executable.
Since `python3` is automatically needed by the installation (via `python3-pyudev` for `joy2key`), no explicit dependency for `python3` was added.

### DIFF
--- a/scriptmodules/emulators/mame.sh
+++ b/scriptmodules/emulators/mame.sh
@@ -53,10 +53,7 @@ function build_mame() {
     fi
 
     # Compile MAME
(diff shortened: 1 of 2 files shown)


joolswills (approved):

joolswills:
Thanks.

PR #3253, opened 2020-11-04
get_all_x11_modes speedup, translates the logic to a single awk call

Nomarian:
The function of get_all_x11_modes does around 6 forks per line. this adds up to around 5 seconds before the selection dialog is launched. this translates the logic of the entire function into a single awk call. making it take 1 second.

### DIFF
--- a/retropie_packages.sh
+++ b/retropie_packages.sh
@@ -9,7 +9,7 @@
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
 #
 
-__version="4.7"
+__version="4.7.1"
 
 [[ "$__debug" -eq 1 ]] && set -x
 
--- a/scriptmodules/admin/image.sh
+++ b/scriptmodules/admin/image.sh
@@ -211,8 +211,6 @@ function create_image() {
     local chroot="$2"
     [[ -z "$chroot" ]] && chroot="$md_build/chroot"
 
-    image+=".img"
-
     # make image size 300mb larger than contents of chroot
     local mb_size=$(du -s --block-size 1048576 "$chroot" 2>/dev/null | cut -f1)
     ((mb_size+=492))
@@ -275,9 +273,6 @@ function create_image() {
     kpartx -d "$image_name"
 
     trap INT
-
-    printMsgs "console" "Compressing $image ..."
-    gzip -f "$image"
 }
 
 # generate berryboot squashfs from filesystem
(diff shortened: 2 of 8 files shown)


joolswills:
Please can you adjust it using our styleguide - we use 4 spaces indent.

Nomarian:
Fixed?


This takes into account the other [PR](https://github.com/RetroPie/RetroPie-Setup/pull/3252) fix, so only one can be accepted.

joolswills:
Actually it looks like the original has some wrong indent. That should be fixed first. I can do that and you can rebase your changes or it can be done before adding your changes in a separate commit in this PR

joolswills:
Sorry about that - I fixed the indent so this will need to be rebased, but it will be a cleaner diff then.

Nomarian:
Never had to do a rebase, so its a learning experience for me. Hope its ok now.

Nomarian:
OK now?

cmitu:
> Any thoughts?

Looks ok, though it's not as easy to read as the previous version.
In fairness, without knowing the input format expected, it's not going go be easy to grok the code in either versions.

Can you try to group the last `awk` block commands so it clear which line is processed and how it affects the output - something like:
```
...
    # Print CRTC identifier (CONNECTOR:0xID)
    print output ":" substr($2,2,length($2)-2) # id
    $1="";$2=""
    sub(/^[ \t]+/,"") # trim spaces
    info=$0
    
    # Get width from the 2nd video mode line
    getline; width=$3; 
    
    # Read height & vrefresh from 3d video mode line
    getline; height=$3; vrefresh=$NF
    
    # Print video mode details
    printf("%sx%s @ %s (%s)\n", width, height, vrefresh, info)
```

Nomarian:
You can just copy paste the awk call entirely and see the output yourself. So its much easier to see what the output is going to be.


However, one key aspect I dislike, is that the output is structured, its always 2 lines. but I suppose you can print a command, and have a special line coming in the code.

cmitu:
> Also, I think its easier to test given that you can just copy paste the awk call entirely and see what the output looks like

I did the test before, that's why I said it looks ok. I tested with an X11 env again a few days ago and it's fine - we can merge this if anyone else doesn't have any objections.

Can you squash your commits into a single one ?

Nomarian:
Apparently not. (I still don't quite understand git)

cmitu:
Use `git rebase -i <commitHash>`, picking the 1st commit in the series, then squash the subsequent commits in the editor.

Nomarian:
Can I just close and make another PR?

cmitu:
Sure, you can make a new PR if the rebasing hasn't worked out.

PR #3257, opened 2020-11-11
Use https instead of http for retroachievements communication

snewell:
Communication with retro achievements happens over http, and this switches it to https (relevant pull request in retroarch: https://github.com/libretro/RetroArch/pull/11551).  Opening this since retropie is using retroarch-1.8.8, and the retroarch pull request is to get the change into master.

Verified this builds on my setup (raspberry pi4) and I can login to retro achievements like normal.

Probably a good idea to make sure it's accepted upstream before accepting this patch, but I figured eyes on the RetroPie script can't hurt.


cmitu:
Upstream RetroArch is the better place for this. Once it's integrated there, it will be picked up in a future version of the RetroPie's scriptmodule, with the version that includes it.

joolswills:
Yeah - unlikely to backport this tbh.

PR #3259, opened 2020-11-14
runcommand.sh: get_all_x11_modes speedup.

Nomarian:
Translate bash logic to one awk call

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -278,33 +278,53 @@ function get_all_x11_modes()
 {
     declare -Ag MODE
     local id
-    local info
     local line
-    local verbose_info=()
-    local output="$($XRANDR --verbose | grep " connected" | awk '{ print $1 }')"
+    while read -r id; do
+        # populate CONNECTOR:0xID into an array
+        MODE_ID+=($id) # output:id as in (hdmi:0x44)
+        
+        read -r line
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -278,33 +278,54 @@ function get_all_x11_modes()
...
+        MODE[$id]="$line"
+    done < <( $XRANDR --verbose | awk '
+        # defines the type of line
+        # true is the "header" (output and id)
+        # false is the "description" (Mode: and everything that begins with a space)
+        { type = /^[^ \t]+/ }
+        
+        # Exit after the first output
+        type && output {exit} # New header and output set means new output
+        #type { output="" } # New header, Reset output
```
joolswills, line 297: Is this commented out line supposed to be left?

Nomarian: Yes. Everything works fine by not exiting. but because get_x11_mode_info exits after the first output found, I thought it best to also do the same for now. if you remove the 2 lines above and uncomment you would really get all the x11 modes. but it might break with x11_mode_info. So I want to keep that commented until that ever gets fixed.

Nomarian:
Alright, removed the comment. if someone has multiple connected devices, the line can be added later I guess.

joolswills:
Thanks!

Nomarian:
_sigh_... I did it wrong, The log looks bad now.

PR #3265, opened 2020-11-30
drastic: disable cursor under KMS

cmitu:
After the changes in #3121, starting `drastic` on a Pi4 shows the X11 cursor.

Since there's no way to remove it for users that don't have a mouse/touchpad available and `drastic` has its own input stylus handling for those input devices, disable the cursor via the WM.


joolswills:
Thanks.

PR #3268, opened 2020-12-08
Move to associative arrays for module data

joolswills:
This reworks a lot of internal package logic and simplifies a lot of functions.

It was something I had in mind a long time ago, and the longer I left it, the messier it got.

Previously all the module data was in indexed arrays with matching offsets. Each scriptmodule folder had an initial index and it was incremented per module.

Many package functions expected an index. Indexes were never unique and could change depending on added modules etc. We also had to convert between them and module IDs.

This caused a lot of issues to extend package organisation. Eg. The idea to allow external repos with additional packages would have been messy to try and arrange the order.

Now it's all done by module id. There is a numbered index of all modules found, which increments which is used in the GUI setup menus. But it's not needed by any packaging functions.

md_idx is dropped and the few modules that used it for install checks etc now use md_id.

All packaging functions use the id.

Retropie_packages now uses the module id only.

I may have forgotten to mention some things and this isn't complete and fully tested yet. Doing an early draft PR while I finalise it before review.

But comments on the changes welcome.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -286,6 +286,7 @@ function package_setup() {
                 rps_logInit
                 {
                     rps_logStart
+                    clear
                     rp_callModule "$id" remove
                     rps_logEnd
                 } &> >(_setup_gzip_log "$logfilename")
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -153,6 +153,7 @@ function post_update_setup() {
 
 function package_setup() {
     local id="$1"
+    local default=""
 
     # associative array so we can pull out the messages later for the confirmation requester
     declare -A option_msgs=(
@@ -238,9 +239,9 @@ function package_setup() {
             options+=(H "Package Help")
         fi
 
-        cmd=(dialog --backtitle "$__backtitle" --cancel-label "Back" --menu "Choose an option for $id\n$status" 22 76 16)
+        cmd=(dialog --backtitle "$__backtitle" --cancel-label "Back" --default-item "$default" --menu "Choose an option for $id\n$status" 22 76 16)
         choice=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
-
+        default="$choice"
         local logfilename
 
         case "$choice" in
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -456,7 +456,7 @@ function update_packages_setup() {
     local id
     for id in ${__mod_id[@]}; do
         if rp_isInstalled "$id" && [[ "${__mod_section[$id]}" != "depends" ]]; then
-            rp_installModule "$id" "_update_" || return 1
+            rp_installModule "$id" "_update_"
         fi
     done
 }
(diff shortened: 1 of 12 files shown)


joolswills:
It also isn't a change that can really be done incrementally hence it's a single commit. It will also need a lot of testing when I think it's ready as it affects all parts of code involving package/module installation.

joolswills:
There is also the opportunity later to use a single associative array for the module data (combining module id and module data as a single key). But I haven't decided whether this is really an advantage apart from making it easier to extend if adding new module parameters which would require additional code anyway. 

It would mean additional logic to iterate for example through all licenses vs looping through a single array. But I thought I'd mention it.

joolswills:
Another thing I wanted to add after this, is a status array. Right now, if a module is not supported by a platform, it's just skipped. It would be useful to have all indexed even if not shown to the user. A lot of users ask why something doesn't show up etc. Potentially we could list disabled modules with info why they are not installable (assuming it wouldn't be too confusing via our dialog menus)

And one more thing which isnt specific to this PR but I think has been confusing is how we order items. Again if it can be done without being too confusing, it would be useful to have a heading above emulators, libretrocores, and ports.

But also I was considering to add some additional menus  such as:

* Modules currently installed (all sections)
* Installing via scriptmodule category rather than module section. Eg it shows all emulators, libretro cores and so on in different menus
* Also a section to multiselect various packages for installation as RetroPie had in the early days (it was the main way to install then)

I'm going off scope on this a bit, but I think there could be some ways of improving how package data is handled after this which wasn't possible before, once we are free of the hardcoded index orientated code.

joolswills:
So. I still think there is potential for more bugs due to lots of manual replacement of variables and such. But it's ready for testing I guess to locate the remaining issues. I think there potentially things I've missed in setup, and I managed to delete my install by updating with this code (fixed now).  Be warned :-)

joolswills:
Oh and I need to update the first commit with info from this PR. Put back to draft until I sort that sorry.

joolswills:
There are also other issues with this code I need to resolve sorry. Going to need to spend more time on it.

scriptmodules/admin/setup.sh
```diff
@@ -453,10 +453,10 @@ function config_gui_setup() {
...
 function update_packages_setup() {
     clear
-    local idx
-    for idx in ${__mod_idx[@]}; do
-        if rp_isInstalled "$idx" && [[ "${__mod_section[$idx]}" != "depends" ]]; then
-            rp_installModule "$idx" "_update_" || return 1
+    local id
+    for id in ${__mod_id[@]}; do
+        if rp_isInstalled "$id" && [[ "${__mod_section[$id]}" != "depends" ]]; then
+            rp_installModule "$id" "_update_" || return 1
```
cmitu, line 459: Maybe this should be handled similar to https://github.com/RetroPie/RetroPie-Setup/commit/0e63f091bd20569ab68f9f6bb259ababb245b176 - don't fail on a single package installation.

joolswills: I agree. I'll adjust that in a separate commit from the reworking so it's clear. Thanks.

joolswills:
This is ready to be merged now - I updated the first commit with more information as mentioned.

I will give it some more testing and merge in the next few days - if anyone else has some time to test it would be useful. Thanks.

@cmitu thanks for your help with this.

PR #3270, opened 2020-12-13
audiosettings: add PulseAudio configuration support.

cmitu:
When PulseAudio is installed and active, it takes over as the default ALSA device.
Remove any `.asoundrc` when it's enabled and configure the PulseAudio default output sink instead.

Added the option to enable/disable PulseAudio, depending on the running configuration.

### DIFF
--- a/scriptmodules/supplementary/audiosettings.sh
+++ b/scriptmodules/supplementary/audiosettings.sh
@@ -30,7 +30,10 @@ function gui_audiosettings() {
     # The list of ALSA cards/devices depends on the 'snd-bcm2385' module parameter 'enable_compat_alsa'
     # * enable_compat_alsa: true  - single soundcard, output is routed based on the `numid` control
     # * enable_compat_alsa: false - one soundcard per output type (HDMI/Headphones)
-    if aplay -l | grep -q "bcm2835 ALSA"; then
+    # If PulseAudio is enabled, then try to configure it and leave ALSA alone
+    if _pa_cmd_audiosettings systemctl -q --user is-enabled pulseaudio.socket; then
+        _pulseaudio_audiosettings
+    elif aplay -l | grep -q "bcm2835 ALSA"; then
         _bcm2835_alsa_compat_audiosettings
     else
         _bcm2835_alsa_internal_audiosettings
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/audiosettings.sh
```diff
@@ -30,15 +30,18 @@ function gui_audiosettings() {
     # The list of ALSA cards/devices depends on the 'snd-bcm2385' module parameter 'enable_compat_alsa'
     # * enable_compat_alsa: true  - single soundcard, output is routed based on the `numid` control
     # * enable_compat_alsa: false - one soundcard per output type (HDMI/Headphones)
-    if aplay -l | grep -q "bcm2835 ALSA"; then
+    # If PulseAudio is enabled, then try to configure it and leave ALSA alone
+    if sudo -u "$user" XDG_RUNTIME_DIR=/run/user/$SUDO_UID systemctl -q --user is-enabled pulseaudio.socket; then
```
joolswills, line 34: I believe this will output an error line if pulseaudio isn't installed - probably don't want to show that

joolswills (commented):
Rather than "sudo -u "$user" "$sudo_env" everywhere, could maybe de-duplicate this further. eg you could de-duplicate the entire command not just one part of it, use a private function, and so on.

scriptmodules/supplementary/audiosettings.sh
```diff
@@ -159,3 +178,65 @@ EOF
...
+    local sink_label
+
+    # Check if PulseAudio is running, otherwise 'pacmd' will not work
+    if ! sudo -u "$user" "$sudo_env" pacmd stat>/dev/null; then
+        printMsgs "dialog" "PulseAudio is enabled, but not running\nAudio settings cannot be set right now"
+        return
+    fi
+    while read sink_index sink_label; do
+        options+=("$sink_index" "$sink_label")
+    done < <(sudo -u "$user" "$sudo_env" pacmd list-sinks | grep -e index -e alsa.name | awk -F [:=] '/index/ { idx=$2; getline; print idx,gensub(/"|bcm2835\s+/,"","g", $2) }')
```
joolswills, line 196: This line could be maybe split / laid out differently for readabilithy- if you're using awk you could probably lose grep also ?

scriptmodules/supplementary/audiosettings.sh
joolswills: Yeah it dumps a "Failed to get unit state for pulseaudio.socket" if not installed. Can be redirected.

cmitu: Yes, it errors out. Should be handled by the helper function.

scriptmodules/supplementary/audiosettings.sh
cmitu: Re-arranged it a bit and made it entirely in `awk`. It's more readable, but not shorter.

cmitu:
I added a helper function for running the commands in the user's context and used it instead of adding the long-ish `sudo` command for PA helpers calls.

PR #3271, opened 2020-12-14
xpadneo: bump version to v0.9

cmitu:
Update to v0.9


joolswills:
Thanks, but I think I will prefer to keep it locked at a tag and manually incremented for now.

joolswills:
BTW note that I am planning on removing of wget usage in favour of curl in the future.

cmitu:
Ok, so then just bump the version, v0.9 has some new controller support, plus some bug fixes and rumble improvements.

joolswills:
Thanks!

PR #3272, opened 2020-12-15
ppsspp / ppsspp-1.5.4 - added ppsspp 1.5.4 using parent ppsspp module

joolswills:
Users reported that this version runs faster on videocore on the rpi1/2/3

Adjusted ppsspp sources to allow branch parameter and fixed build subfolder to ppsspp
so that paths are the same for md_ret_files, including adjusting lr-ppsspp in the same way.

Older code builds ok with our current ppsspp module script on rpi1/2/3 + videocore gles2,
but not rpi4, which it isn't needed on anyway. This may need changing if we have to update
ppsspp further in the future.

### DIFF
--- a/scriptmodules/emulators/ppsspp-1.5.4.sh
+++ b/scriptmodules/emulators/ppsspp-1.5.4.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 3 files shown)


Aryetis:
Here's some footage comparing the currently available binary and the 1.5.4 one on a Raspberry pi 3 (non b+). Same Iso, same settings, same SD card, same Retropie version, etc.

https://www.youtube.com/watch?v=OwykKZMc3nc&feature=youtu.be

joolswills:
Thanks for the footage. I will build binaries and merge this once that's done.

PR #3273, opened 2020-12-17
bluetooth - registered / paired device logic rework

joolswills:
I made some changes before to hide previously registered devices from the connect menu, to make it
easier when dealing with multiple devices of the same name. However due to the way bluetoothctl works,
previously "seen" devices are added to the system, meaning devices that are not paired are registered and cached,
and then won't be shown again.

This changeset addresses the following:

Only actually paired devices will be hidden from the register and connect menu

New functions list_paired and list_connected have been added

Switched to use bt-device to check for connected devices for consistency. Private function handles
grepping for Paired or Connected via bt-device --info

display_active_and_registered function renamed to simpler status and reworked to use the new functions.
It outputs to console - the GUI menu just calls this and then outputs to a dialog menu - making this now
a more useful commandline function

remove_device shows all registered devices, even ones seen before. These previously seen devices show
up when scanning so we still want to be able to remove them. I have adjusted it though so paired devices are
shown first, followed by any known devices.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -165,12 +165,12 @@ function status_bluetooth() {
     local name
 
     while read mac; read name; do
-        paired+="$mac: $name\n"
+        paired+="$mac - $name\n"
     done < <(list_paired_bluetooth)
     [[ -z "$paired" ]] && paired="There are no paired devices"
 
     while read mac; read name; do
-        connected+="$mac: $name\n"
+        connected+="$mac - $name\n"
     done < <(list_connected_bluetooth)
     [[ -z "$connected" ]] && connected="There are no connected devices"
 
@@ -215,7 +215,7 @@ function remove_device_bluetooth() {
     fi
 }
 
-function register_bluetooth() {
+function pair_bluetooth() {
     declare -A devices=()
     local mac
     local name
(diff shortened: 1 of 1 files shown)


joolswills:
Added some additional commits. Some variable renaming and a fix for the connect function which I broke in a15d58a8

The boot/background and connect to all devices gui functionality shouldn't be needed. I don't think the udev rules are either. I'm inclined to remove some of this stuff from the menu perhaps, but maybe leave it for commandline only. Needs some further looking into but I'm keen to get this PR included asap so would be something for later.

Background connect is a nasty hack which was only required due to issues with my 8bitdo devices on Jessie bluez afair as there were issues with devices not automatically connecting.

joolswills:
Going to merge this but will note:

This will caused paired PS3 controllers to show up when scanning, which @cmitu

joolswills:
Going to merge this but will note:

This will cause paired PS3 controllers to show up when scanning, which @cmitu noted - this was also the case before my changes in 411b4eb1. However my changes didn't show devices known previously by bluez, which is a problem as our current scanning using bluetoothctl will register lots of devices it sees, meaning they may not be paired but won't show when scanning.

Part of this PR adjusts this to filter out only paired devices. Which as mentioned won't include PS3 - once we have a pure bluez solution for pairing PS3 etc this should be resolved. But I believe the adjusted behaviour is better, and should resolve issues some users have had, and I saw myself in testing.

Other changes are needed anyway, including maybe removal of some of the workarounds for older bluez. 

This is my defence if I break anything :-)

PR #3278, opened 2020-12-28
esthemes: enable branch support for git

cmitu:
This would enable installation of themes where the default branch is named `main`, instead of `master`.
Github changed this a while back, see the [renaming repo](https://github.com/github/renaming).

One such theme is @lipebello's new theme (https://github.com/lipebello/es-theme-retrorama-turbo).


cmitu:
Actually, this may be better handled in `gitPullOrClone`.

PR #3279, opened 2021-01-01
own_sdl2=0 was being ignored

nhand42:
The inivar own_sdl2 in /opt/retropie/configs/all/retropie.cfg was ignored for value "0".

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -240,6 +240,7 @@ function _mapPackage() {
                 iniConfig " = " '"' "$configdir/all/retropie.cfg"
                 iniGet "own_sdl2"
                 [[ "$ini_value" == "1" ]] && own_sdl2=1
+                [[ "$ini_value" == "0" ]] && own_sdl2=0
                 [[ "$own_sdl2" -eq 1 ]] && pkg="RP sdl2 $pkg"
             fi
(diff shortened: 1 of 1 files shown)


hhromic (changes requested):
Given that previously we have:
```bash
local own_sdl2=1
```
The actual fix is to just replace:
```
[[ "$ini_value" == "1" ]] && own_sdl2=1
```
into:
```bash
[[ "$ini_value" == "0" ]] && own_sdl2=0
```

not add a new conditional.

nhand42:
> Given that previously we have:
> 
> ```shell
> local own_sdl2=1
> ```
> 
> The actual fix is to just replace:
> 
> ```
> [[ "$ini_value" == "1" ]] && own_sdl2=1
> ```
> 
> into:
> 
> ```shell
> [[ "$ini_value" == "0" ]] && own_sdl2=0
> ```
> 
> not add a new conditional.

Unless x11 platform is set, in which case the test for 1 is still required.

hhromic:
Ah that's a good point indeed. I don't know what is the actual desired behaviour.
Let's wait for another dev to chime in then. Thanks for noticing this bug anyway!

joolswills:
I'll sort this. I may change the logic a bit, but I agree it's broken.

PR #3280, opened 2021-01-02
esthemes: enable default branch detection for themes repositories.

cmitu:
This would enable installation of themes where the default branch is named `main`, instead of `master`.
Github changed this a while back, see the [renaming repo](https://github.com/github/renaming).

One such theme is @lipebello's new theme (https://github.com/lipebello/es-theme-retrorama-turbo).

**Note**: this is a re-worked version of #3278. While `gitPullOrClone` could be modified to detect the default main branch of a repo, its main usage is in the `_source` scriptmodules function - where the branch can be specified, so default branch detection is not strictly necessary.
`esthemes` doesn't have the ability to specify a branch when cloning a theme. Since a theme repository isn't usually multi-branched (so there's no need to add branch name support in the selection list), I added the default branch detection logic just for this scriptmodule, when installing a theme.

Not sure if the default branch name should be `main` though (instead of `master`), given the new Github defaults.

### DIFF
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -24,6 +24,7 @@ function depends_esthemes() {
 function install_theme_esthemes() {
     local theme="$1"
     local repo="$2"
+    local default_branch
     if [[ -z "$repo" ]]; then
         repo="RetroPie"
     fi
@@ -31,8 +32,11 @@ function install_theme_esthemes() {
         theme="carbon"
         repo="RetroPie"
     fi
+    # Get the name of the default branch, fallback to 'master' if not found
+    default_branch=$(runCmd git ls-remote --symref --exit-code "https://github.com/$repo/es-theme-$theme.git" HEAD | grep -oP ".*/\K[^\t]+")
+    [[ -z "$default_branch" ]] && default_branch="master"
     mkdir -p "/etc/emulationstation/themes"
-    gitPullOrClone "/etc/emulationstation/themes/$theme" "https://github.com/$repo/es-theme-$theme.git"
+    gitPullOrClone "/etc/emulationstation/themes/$theme" "https://github.com/$repo/es-theme-$theme.git" "$default_branch"
 }
 
 function uninstall_theme_esthemes() {


scriptmodules/supplementary/esthemes.sh
```diff
@@ -24,15 +24,19 @@ function depends_esthemes() {
...
+    local default_branch
     if [[ -z "$repo" ]]; then
         repo="RetroPie"
     fi
     if [[ -z "$theme" ]]; then
         theme="carbon"
         repo="RetroPie"
     fi
+    # Get the name of the default branch, fallback to 'master' if not found
+    default_branch=$(runCmd git ls-remote --symref "$repo" HEAD | cut -s -d'/' -f3 | cut -d$'\t' -f1)
```
joolswills, line 36: I think the two cuts in this case are maybe less readable than a regexp via grep or sed. Eg

```
 | grep -oP ".*/\K\w+"
```

Although I feel there should be something simpler still maybe.

joolswills:
Looks good in general. I left a comment regarding extracting branch name. Also it may be worth checking this against a couple of non github repos if you didn't already to make sure it defaults ok.

I'm happy to leave default as master for now. We can change if needed. Did you see a case where it fails and the default is needed? It may be better to return a failure for example.

scriptmodules/supplementary/esthemes.sh
cmitu: Does `| awk -F'[\t/]' '/ref/ { print $(NF-1); exit}' ` look better ? It's not shorter, but easier to read.

joolswills: Think I still prefer the grep in this case. Actually could use bash regular expressions too. I guess it's not important as it's not a performance thing but I would guess the grep is most efficient. (despite using perl regexp and so not being portable). Awk is readable if you know awk - I like it but it's long and seems a bit basic for an awk task :-)

cmitu: OK, added the `grep` filter for the default branch name.

cmitu:
I've tested with `git ls-remote` and it works with other repositories (gitlab or self-hosted).
About errors - `ls-remote` doesn't show any error message if the remote reference is not found (though _HEAD_ should always be available). If the repo URL is incorrect, Github and Gitlab will prompt for user/password, but that happens with a normal clone also, not only with `git ls-remote`).

For now I've left the fallback, since the functions in `gui_esthemes` don't check for errors after installation and errors are shown only on the console. If we want to show an error when the theme download fails, I can change it to fail when `git ls-remote` errors out and add a check for `md_ret_errors` in the `gui_esthemes` function.

joolswills:
Thanks.

PR #3281, opened 2021-01-03
Lr dosbox pure

joolswills:
Have done limited testing. Put together quickly so need to do some more work (module based off lr-dosbox which was pretty basic).


joolswills:
Oops. Need to rebase.

hlad:
there should be more valid extensions, not just .zip [https://github.com/schellingb/dosbox-pure/blob/main/dosbox_pure_libretro.cpp#L1765](https://github.com/schellingb/dosbox-pure/blob/main/dosbox_pure_libretro.cpp#L1765)

joolswills:
Thanks - I'll add them.

kashkash18:
When will be add to experimental packets in retropie?

PR #3284, opened 2021-01-08
moonlight: add support for open-source GFE servers

psyke83:
There are open-source implementations of NVIDIA's GFE server which
expands Moonlight compatibility to other GPU manufacturers on both
Linux and Windows:

1. https://github.com/loki-47-6F-64/sunshine
(original project with CPU and NVIDIA HW encoder support)

2. https://github.com/LS3solutions/openstream-server
(forked project focused on Windows support with partial AMD HW encoding)

Unfortunately, moonlight-embedded does not work out of the box, so
provide a new host menu to assist in configuring the options needed
to attain full compatibility with these servers.

### DIFF
--- a/scriptmodules/supplementary/moonlight.sh
+++ b/scriptmodules/supplementary/moonlight.sh
@@ -185,6 +185,35 @@ function get_resolution_moonlight() {
     fi
 }
 
+function get_host_moonlight() {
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. Looks good :-)

PR #3285, opened 2021-01-08
Box86, Wine, and Space Cadet 3D Pinball

GeorgeMcMullen:
Thank you in advance for reviewing my pull request. Attached are four scripts with the intention of being able to run x86 and Windows games through RetroPie on a Raspberry Pi 4. Other Raspberry Pi's may be compatible, but will require further testing.

For further details, see: https://retropie.org.uk/forum/topic/28528/box86-and-wine-on-rpi4/

The scripts must be run in order, and they will check to make sure the preceding dependencies have been met.

- mesa.sh - This is an updated Mesa driver, which is required by Box86. The version that is installed with the RetroPie distribution is not yet compatible.
- box86.sh - This is set as an emulator because it is.
- wine.sh - This is the x86 version of Wine and requires box86 in order to run. It is installed as a Port, because as the name says, Wine Is Not an Emulator.
- spacecadet3dpinball.sh - This is the original Windows XP game, hosted on the Internet Archive. It is also installed as a Port as it is more like a rom and not an emulator. It is included mostly as a reference to installing other games. One note - you must not hit the Escape key to quit or you will be in a hung state in the game. Use Alt-F4 instead.

These could be split into separate pull requests, but I haven't because of the dependency chain. 

Looking forward to any questions or feedback you might have!

Thanks again,

\- George

### DIFF
--- a/scriptmodules/emulators/box86.sh
+++ b/scriptmodules/emulators/box86.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="box86"
+rp_module_desc="Box86 emulator"
+rp_module_help="Place your x86 binaries $romdir/box86"
+rp_module_licence="MIT https://github.com/ptitSeb/box86/blob/master/LICENSE"
+rp_module_section="exp"
+rp_module_flags="rpi4 x11"
+
+function _latest_ver_box86() {
+    # This defines the Git tag / branch which will be used. Main repository is at:
+    # https://github.com/ptitSeb/box86
+    echo v0.1.8
+    # The following is not working yet. Releases must be non-prerelease and non-draft.
+    # wget -qO- https://api.github.com/repos/ptitSeb/box86/releases/latest | grep -m 1 tag_name | cut -d\" -f4
+}
+
+function depends_box86() {
+    if compareVersions $__gcc_version lt 7; then
+        md_ret_errors+=("Sorry, you need an OS with gcc 7 or newer to compile $md_id")
+        return 1
+    fi
+
+    local dep_idx="$(rp_getIdxFromId "mesa")"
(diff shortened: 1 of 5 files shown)


joolswills:
Thanks. Unfortunately I don't want to include a custom compiled mesa into retropie currently, so any addition will need to wait until mesa is updated in Raspberry Pi OS.

I also wouldn't want to include any games - this game looks to be copyrighted still also. 

I think for now this would be best managed yourself on the forum as currently - but please remove any references to the copyrighted game.

GeorgeMcMullen:
Thanks @joolswills. Good points on both counts. 

- I'll keep an eye on the Mesa driver being updated in Raspberry Pi OS. Hopefully it is soon!
- I'll remove Space Cadet from this branch. I think it might be good to have some script as a reference for others to create their own. Do you have any suggestions?
- Do you think Wine belongs in Ports or as an Emulator/System? Box86 is in emulators, but that is really an emulator. But my script does not create a ROMs directory for it. As a comparison, DosBox is in emulators, but that is also a true emulator.

Is it ok to keep this pull request open for the time being?

Thanks again,

\- George

timfel:
@GeorgeMcMullen FYI, I've just tried installing only box86 from git and downloading the Wine 6.0 release version, and they work fine on RPi 4 with the buster based retropie. I have not removed `xserver-xorg-video-fbturbo` as per your PR, nor have I updated mesa manually. So afaict, just the bits for cloning and building box and getting wine should be enough.

mth75:
@timfel A lot of games work without the latest Mesa drivers ... a lot won't (think vulkan). You probably will be referred to the forum: https://retropie.org.uk/forum/topic/28528/box86-and-wine-on-rpi4/

timfel:
@mth75 Sure, but what does it matter if we're going to wait for the bundled mesa to get updated anyway? I would think just having the option to have wine and then it improving over time as the packages get updated is not much different than for other experimental emulators

mth75:
@timfel The point is George is getting feedback for the install scripts , in the meantime I'm building my library of Window games + there is no saying when the Mesa-drm version will be on par for usage with the Box86/Wine combination. One example: Broken Sword 3 needs the 20.x Mesa versions to play.

https://retropie.org.uk/forum/topic/29241/the-which-games-works-with-box86-wine-topic-on-the-retropie-rpi4-400/41?_=1613228544075

GeorgeMcMullen:
Thanks @mth75 and @timfel. It might be worth it to find out which, if any, games don't actually require the new Mesa drivers. The potential issue is that people may get confused as to what they will actually be able to run. I see this happen a lot on the forums about which MAME roms go with which MAME emulator.

PR #3286, opened 2021-01-11
lr-hatari: fix building with the capsimage libs

cmitu:
Upstream `hatari` has dropped support for CAPS4, so it's not looking for `caps5`, but the libretro core still looks for [`caps5`](https://github.com/libretro/hatari/blob/c05015c395a17af12e1eb7e9e4fb876f74b00122/src/uae-cpu/newcpu.c#L166).


joolswills:
Thanks. I had forgotten about the libretro core when I updated this.

PR #3287, opened 2021-01-11
Docsview

joolswills:
wikiview was no longer working due to docs moving - so renamed and adjusted the old wikiview module to docsview.

The module could do with some more work really, but at least it's functional now.


joolswills:
Made some minor functional changes as it was rescanning the entire folder each time and didn't remember dialog menu position.

PR #3290, opened 2021-01-15
Wget to curl

joolswills:
Move RetroPie to use curl rather than wget.

This changeset switches the helpers download function to use curl (including some function enhancements such as stdout support) and makes downloadAndExtract also use the core download function.

In addition almost all module code using wget has been switched to use the download or downloadAndExtract function excluding one case where wget was used in an external script which has been switched to curl.

Packaging and setup functions have been adjusted to curl also.

This is left as draft as there's a few things I want to look at still, such as the stderr output capturing and some additional tweaks here and there perhaps. But feel free to test.

### DIFF
--- a/scriptmodules/packages.sh
+++ b/scriptmodules/packages.sh
@@ -92,9 +92,9 @@ function rp_callModule() {
             rp_hasBinary "$md_id"
             local ret="$?"
 
-            # check if we had a network failure from wget
-            if [[ "$ret" -eq 4 ]]; then
-                __ERRMSGS+=("Unable to connect to the internet")
+            # check if we get a couldn't resolve host or failed to connect to host from curl
+            if [[ "$ret" -eq 6 || "$ret" -eq 7 ]]; then
+                __ERRMSGS+=("Unable to connect to the internet - curl returned $ret")
                 return 1
             fi
 
@@ -370,7 +370,7 @@ function rp_hasBinary() {
     [[ -z "$url" ]] && return 1
 
     if rp_hasBinaries; then
-        wget --spider -q "$url"
+        curl -o /dev/null -sfI "$url"
         return $?
     fi
     return 1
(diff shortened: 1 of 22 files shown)


scriptmodules/packages.sh
```diff
@@ -92,9 +92,9 @@ function rp_callModule() {
             rp_hasBinary "$md_id"
             local ret="$?"
 
-            # check if we had a network failure from wget
-            if [[ "$ret" -eq 4 ]]; then
-                __ERRMSGS+=("Unable to connect to the internet")
+            # check if we get a couldn't resolve host or failed to connec to host from curl
```
cmitu, line 95: Small typo here (_connec_).

scriptmodules/supplementary/raspbiantools.sh
```diff
@@ -77,7 +77,7 @@ function install_firmware_raspbiantools() {
     for pkg in "${pkgs[@]}"; do
         if hasPackage "$pkg" "$ver" ne; then
             deb="${pkg}_${ver}_armhf.deb"
-            if ! wget -O"$deb" "$url/$deb"; then
+            if download "$url/$deb"; then
```
cmitu, line 80: shouldn't this be `! download "$url/$deb"` ?

joolswills: Yes. Thanks.

scriptmodules/packages.sh
joolswills: Fixed

scriptmodules/supplementary/raspbiantools.sh
joolswills: Fixed thanks.

joolswills:
I have added timeout defaults which can be overridden and have reworked the error capturing.

This is ready now for review/testing.

joolswills:
Actually I need to rebase it first due to a conflict with the spelling fixes.

joolswills:
Apologies - not ready. Issue related to my stdout/stderr changes when piping output of download

joolswills:
Reverted to the previous working redirection code. So ready for testing and review. Thanks.

joolswills:
Going to merge. I have tested this a fair amount but will keep an eye if there are any related issues reported.

joolswills:
I will run a full binary rebuild now it's merged also which should pick up any issues I have missed

PR #3291, opened 2021-01-17
dkmsManager fixes

psyke83:
First commit is safe, as it fixes the module error mechanism when dkms fails to build against the running kernel. This can happen if there was a kernel/header upgrade but the system hasn't rebooted to the new kernel yet.

Second commit is dodgy (hence the DNM). It tries to build for all kernels found via a search of installed Debian packages, and marks installation as successful if it detects the module marked as installed for any one of the kernel versions found... but, this is not really a guarantee that it has built/installed the module for the kernel that will be active on the next reboot.

Would appreciate feedback. The problem with only fixing the error detection is that a routine update of RetroPie's packages in combination with an OS upgrade that included a new kernel will still leave xpad in a broken state (although it will at least report the error to the user).

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1445,19 +1445,24 @@ function dkmsManager() {
     local mode="$1"
     local module_name="$2"
     local module_ver="$3"
-    local kernel="$(uname -r)"
+    local kernel
+    local kernels=($(for kernel in $(dpkg -S /lib/modules/*/ | cut -d ":" -f2); do basename "$kernel" | grep -v "$(uname -r)"; done))
     local ver
+
+    # make sure to include running kernel if not in chroot
+    if [[ "$__chroot" -eq 0 ]]; then
+        kernels+=("$(uname -r)")
+    fi
 
     case "$mode" in
         install)
             if dkms status | grep -q "^$module_name"; then
                 dkmsManager remove "$module_name" "$module_ver"
             fi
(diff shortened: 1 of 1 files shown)


scriptmodules/helpers.sh
```diff
@@ -1445,19 +1445,24 @@ function dkmsManager() {
...
+        kernels+=("$(uname -r)")
+    fi
+
     case "$mode" in
         install)
             if dkms status | grep -q "^$module_name"; then
                 dkmsManager remove "$module_name" "$module_ver"
             fi
-            if [[ "$__chroot" -eq 1 ]]; then
-                kernel="$(ls -1 /lib/modules | tail -n -1)"
```
psyke83, line 1457: In case it helps explain why I didn't (re)use this to guess the pending kernel... it doesn't seem to be a reliable method to find the target kernel, in a chroot or otherwise. After re-installing the `raspberrypi-headers` and `raspberrypi-kernel` packages, I tried invoking this on a real rpi3, and it erroneously selected the `v8+` folder.

joolswills:
Hope you don't mind. Switched to draft instead of donotmerge. Looks good btw, but maybe
It's worth comparing the kernel version code Vs mitus etc. Cheers.

psyke83:
I wasn't aware of the draft feature - good to know! As for the PR itself, I prefer @cmitu's more distro-agnostic solution, so I'll close this and leave Cristi to submit when he's ready. Thanks.

PR #3292, opened 2021-01-19
helpers: improve DKMS module installation

cmitu:
Modified the `dkmsManager` function to account for cases when the running
kernel is replaced before the module installation, so `dkms` cannot build the module.
If the module cannot be built for the running kernel version, then instruct `dkms` to build
for the installed kernels that have a valid `build` folder/symlink present.

Installations performed under a 'chroot' are also handled with the new additions,
so there is no need for separate handling.

This covers situations when the user does an upgrade of the distro's packages and
RetroPie's drivers at the same time. Raspberry Pi OS replaces the current kernel
and the kernel headers for the running version are no longer found, so the driver(s) installed
are not updated from their scriptmodule by `dkmsManager`.

As a side note, Raspberry Pi OS has 4 kernels shipped with the same package:
 - kernel.img ($ver+) for Pi 0/1
 - kernel7.img ($ver-v7+) for Pi 2/3
 - kernel7l.img ($ver-v7l+) for Pi4
 - kernel8.img ($ver-v8+) for Pi4 with the `aarch64` architecture (enabled by `arm_64bit` option in `configtxt`)
For each kernel, there's a corresponding folder under `/lib/modules`. Explained in [1].

The 'chroot' installation didn't install the module for all the versions on the RPI image, it chose
the last sub-folder (which would be the `$ver-v8+`, for `aarch64`), so the latest images don't have the `xpad` driver installed correctly.

Ref:
 [1] the `kernel` option from https://www.raspberrypi.org/documentation/configuration/config-txt/boot.md

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1453,13 +1453,19 @@ function dkmsManager() {
             if dkms status | grep -q "^$module_name"; then
                 dkmsManager remove "$module_name" "$module_ver"
             fi
-            if [[ "$__chroot" -eq 1 ]]; then
-                kernel="$(ls -1 /lib/modules | tail -n -1)"
-            fi
             ln -sf "$md_inst" "/usr/src/${module_name}-${module_ver}"
-            dkms install --force -m "$module_name" -v "$module_ver" -k "$kernel"
-            if dkms status | grep -q "^$module_name"; then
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks very much for this. Well researched and appreciate all the info. A good reason to rebuild images too. Thanks again - great job (thanks to @psyke83 also - team effort!) :-)

PR #3295, opened 2021-01-23
packages - allow adding of 3rd party module repos

joolswills:
The structure should be such that you have

ext/REPOSITORY/scriptmodules

modules are only read from the following subfolders of scriptmodules

emulators
libretrocores
ports
supplementary
admin

Tagging: @zerojay and @hiulit

This should make things easier and keep your repos separate etc. Needs testing.


hiulit:
Hey! I didn't see that! I'll take a look at it. Thanks!

PR #3296, opened 2021-01-23
Merge module arrays and include all modules in setup

joolswills:
rather than multiple arrays for each module info we now use a single array with

__mod_info(ID/KEY) where key is type/desc/help/licence/flags etc

we still keep __mod_id() - a standard indexed array with every module id (used for setup menus etc) and
__mod_idx() which references the index # from the id

this will make it easier to extend in the future

### DIFF
--- a/scriptmodules/admin/builder.sh
+++ b/scriptmodules/admin/builder.sh
@@ -29,7 +29,7 @@ function module_builder() {
         ! fnExists "install_${id}" && continue
 
         # skip already built archives, so we can retry failed modules
-        [[ -f "$__tmpdir/archives/$__binary_path/${__mod_type[$id]}/$id.tar.gz" ]] && continue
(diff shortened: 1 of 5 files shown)


joolswills:
I have also added a change here so all modules are shown even if not enabled for the current system. Details in the commit msg.

joolswills:
I made some styling tweaks to manage packages. Now installed modules are white to make it clearer to see which are installed at a quick glance. I tried out various styles including disabled modules being red, but that made it less clear and a bit too busy. Blue also worked ok for installed modules - so please let me know.

This PR had a bit of feature creep but I wanted an example of how the module info is easier to extend now etc.

joolswills:
Did some more testing. Seems ok. Styling changes can be tweaked if not to people's liking.

PR #3298, opened 2021-01-26
packages - allow adding of 3rd party module repos

joolswills:
3rd party repositories can be checked out to RetroPie-Setup/ext

Repositories should be structured so you have:

ext/REPOSITORY/scriptmodules

Modules are only read from the following subfolders of scriptmodules
 * emulators
 * libretrocores
 * ports
 * supplementary
 * admin

### DIFF
--- a/ext/README.txt
+++ b/ext/README.txt
@@ -1,0 +1,13 @@
+You can checkout additional repositories here to add 3rd party modules.
+
+Repositories should be structured so you have:
+
+ext/REPOSITORY_NAME/scriptmodules
+
+Modules are only read from the following subfolders of scriptmodules:
+ * emulators
+ * libretrocores
+ * ports
+ * supplementary
+ * admin
+
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -325,7 +325,14 @@ function section_gui_setup() {
         local pkg_origin
         local num_pkgs=0
         local info
+        local type
         for id in $(rp_getSectionIds $section); do
+            # do a heading for each origin and module type
+            if [[ "$type" != "${__mod_info[$id/type]}" ]]; then
+                info="${__mod_info[$id/vendor]} - ${__mod_info[$id/type]}"
+                pkgs+=("----" "$info ----" "Packages from $info")
+                type="${__mod_info[$id/type]}"
+            fi
             if ! rp_isEnabled "$id"; then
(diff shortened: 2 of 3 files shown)


joolswills:
This replaces #3295 which was broken. Currently testing this but seems ok.

joolswills:
I think I will change this around some more actually. Want it to be clear in the menus if something is a 3rd party repo and maybe display them separately. Sorry for prematurely merging previous commit as I hadn't thought it through completely.

joolswills:
I have worked around the issue by adding a vendor record to the modules. It's also displayed in the setup menus along with the module type which should help with confusion of the ordering which has happened before. But it also makes it clear which are 3rd party modules and is a simpler solution that a whole new menu section set.

From the commit msg:

Used to separate 3rd party packages from RetroPie ones. Added some extra menu items in setup showing
above each type like a heading with

---- VENDOR - TYPE ----

Where:
 * VENDOR is RetroPie or the name of the folder the 3rd party repo is installed in
 * TYPE is the module/package type such as emulator, libretrocore, port etc

Some users have found the ordering of the menus confusing and this should help with that also.

3rd party packages always appear at the end of a section

joolswills:
This is ready for testing if anyone has some time. Thanks.

joolswills:
Added a check to make sure we don't have duplicate `rp_module_id` names when scanning. Skip and display a message if found.

This helps avoiding duplicate IDs but also 3rd party repos should not override our modules but have their own unique names, otherwise debugging issues could be confusing and it could break things if a RetroPie module is replaced by another.

joolswills:
In the case of RetroPie-Extra it has two modules openbor and lr-bsnes that conflict (@zerojay - when this is included it's worth noting you willn eed to adjust your repo but it will make it easier to add your repo without mixing files with ours and making it easier for users to manage).

joolswills:
This does increase module loading time a bit but I don't think it's a big issue as it's on setup/packages initial launch but I had some ideas to improve module scanning in general in the future.

joolswills:
I'm going to merge. Hopefully no issues this time.

arnaldovalente:
How to access to $md_data in ext folder. It point to vendor one

joolswills:
Looks like I have a few issues to fix.

joolswills:
The md_data issue should now be fixed. Thanks for the info.

arnaldovalente:
Tank you for fix it

PR #3300, opened 2021-01-28
Update lr-snes9x.sh

solskogen:
Add back support for compiling it with distcc


solskogen:
Closing this. It still fails on 32bit.

PR #3302, opened 2021-02-02
Repo version check

joolswills:
These changes allow storing of source repo information in retropie.pkg as well as the ability
to check for a new version. To do this, repository information needs to be moved to the variable
md_module_repo in the format

md_module_repo="TYPE URL BRANCH COMMIT"

Currently types of file, git, or svn are suppported.

If using file the BRANCH and COMMIT are not used.
If using svn the BRANCH should be - and the COMMIT is the SVN revision.

This information is split up and provided to the module in the form of

md_repo_type
md_repo_url
md_repo_branch
md_repo_commit

In addition I have added some other related changes including some caching to help with the additional overhead the extra package information adds when navigating menus.

### DIFF
--- a/scriptmodules/emulators/hatari.sh
+++ b/scriptmodules/emulators/hatari.sh
@@ -12,7 +12,8 @@
 rp_module_id="hatari"
 rp_module_desc="Atari emulator Hatari"
 rp_module_help="ROM Extensions: .st .stx .img .rom .raw .ipf .ctr .zip\n\nCopy your Atari ST games to $romdir/atarist\n\nCopy Atari ST BIOS (tos.img) to $biosdir"
-rp_module_licence="GPL2 https://hg.tuxfamily.org/mercurialroot/hatari/hatari/file/9ee1235233e9/gpl.txt"
+rp_module_licence="GPL2 https://git.tuxfamily.org/hatari/hatari.git/plain/gpl.txt"
+rp_module_repo="git git://git.tuxfamily.org/gitroot/hatari/hatari.git v2.3.1"
 rp_module_section="opt"
 rp_module_flags=""
 
@@ -28,7 +29,7 @@ function _sources_libcapsimage_hatari() {
 
 function sources_hatari() {
     # shallow clone isn't supported via https:// on this repo
-    gitPullOrClone "$md_build" "git://git.tuxfamily.org/gitroot/hatari/hatari.git" "v2.3.1"
+    gitPullOrClone
     _sources_libcapsimage_hatari
 }
 
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -13,6 +13,7 @@ rp_module_id="ppsspp"
 rp_module_desc="PlayStation Portable emulator PPSSPP"
 rp_module_help="ROM Extensions: .iso .pbp .cso\n\nCopy your PlayStation Portable roms to $romdir/psp"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/hrydgard/ppsspp/master/LICENSE.TXT"
+rp_module_repo="git https://github.com/hrydgard/ppsspp.git master"
 rp_module_section="opt"
 rp_module_flags=""
 
@@ -27,7 +28,7 @@ function depends_ppsspp() {
 function sources_ppsspp() {
     local branch="$1"
     [[ -z "$branch" ]] && branch="master"
-    gitPullOrClone "$md_build/ppsspp" https://github.com/hrydgard/ppsspp.git "$branch"
+    gitPullOrClone "$md_build" "$md_repo_url" "$branch"
     cd "ppsspp"
 
     # remove the lines that trigger the ffmpeg build script functions - we will just use the variables from it
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -26,9 +26,7 @@ function depends_ppsspp() {
 }
 
 function sources_ppsspp() {
-    local branch="$1"
-    [[ -z "$branch" ]] && branch="master"
-    gitPullOrClone "$md_build" "$md_repo_url" "$branch"
+    gitPullOrClone "$md_build/ppsspp"
     cd "ppsspp"
 
     # remove the lines that trigger the ffmpeg build script functions - we will just use the variables from it
--- a/scriptmodules/emulators/simcoupe.sh
+++ b/scriptmodules/emulators/simcoupe.sh
@@ -13,6 +13,7 @@ rp_module_id="simcoupe"
 rp_module_desc="SimCoupe SAM Coupe emulator"
 rp_module_help="ROM Extensions: .dsk .mgt .sbt .sad\n\nCopy your SAM Coupe games to $romdir/samcoupe."
 rp_module_licence="GPL2 https://raw.githubusercontent.com/simonowen/simcoupe/master/License.txt"
+rp_module_repo="git https://github.com/simonowen/simcoupe.git master"
 rp_module_section="opt"
 rp_module_flags=""
 
@@ -26,7 +27,7 @@ function sources_simcoupe() {
     # dialect support but actually seem to build ok. Lock systems with older cmake to 20200711 tag,
     # which builds ok on Raspbian Stretch and hopefully Ubuntu 18.04.
     hasPackage cmake 3.8.2 lt && branch="20200711"
-    gitPullOrClone "$md_build" https://github.com/simonowen/simcoupe.git "$branch"
+    gitPullOrClone "$md_build" "$md_repo_url" "$branch"
 }
 
 function build_simcoupe() {
--- a/scriptmodules/emulators/simcoupe.sh
+++ b/scriptmodules/emulators/simcoupe.sh
@@ -13,21 +13,25 @@ rp_module_id="simcoupe"
 rp_module_desc="SimCoupe SAM Coupe emulator"
 rp_module_help="ROM Extensions: .dsk .mgt .sbt .sad\n\nCopy your SAM Coupe games to $romdir/samcoupe."
 rp_module_licence="GPL2 https://raw.githubusercontent.com/simonowen/simcoupe/master/License.txt"
-rp_module_repo="git https://github.com/simonowen/simcoupe.git master"
+rp_module_repo="git https://github.com/simonowen/simcoupe.git :_get_branch_simcoupe"
 rp_module_section="opt"
 rp_module_flags=""
+
+function _get_branch_simcoupe() {
+    local branch="master"
+    # latest simcoupe requires cmake 3.8.2 - on Stretch older versions throw a cmake error about CXX17
+    # dialect support but actually seem to build ok. Lock systems with older cmake to 20200711 tag,
+    # which builds ok on Raspbian Stretch and hopefully Ubuntu 18.04.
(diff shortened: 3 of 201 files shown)


joolswills:
This is a large changeset and I'm still doing tests so it's not quite ready for a review as I need to review it all again myself. Probably some bugs and typos as will happen with so many changes.

joolswills:
But feel free to try of course.

joolswills:
I have noticed some logic issues in some cases in the setup menus and _auto_ / _update_ parameters that I need to address. So be aware that there are still bugs in this.

joolswills:
I also need to convert the ports modules as I forgot about them :-) maybe the PR was a little premature.

joolswills:
Thanks to @cmitu who let me know about some logic issues. I reworked some of the changes, primarily the function to load package data.

This should resolve some of the issues where a package that had no pkg info data from an old install wasnt handled correctly. The caching is reworked. From the main commit changed (there were minor changes elsewhere too which I merged into relevant commits)

Previously we eval's the return of getPackageInfo so the package variables could be created
locally to the function. However, due to the use of a subshell we couldn't cache anything in a variable
for use later.

The functions is now renamed to rp_loadPackageInfo and loads the package information into __mod_info[ID/FIELD]

eg. __mod_info[ID/pkg_origin] will contain the retropie.pkg origin field for the installed module

It also uses just grep to load the data rather than our existing ini functions for performance. As it's not
considered a user editable file, performance is key here, as we need this data when traversing the
RetroPie-Setup menus.

If a full load of a retropie.pkg is done, __mod_info[ID/pkg_info] is set to 1, as a cache flag so the data can be
reused without reloading on an additional call. rp_SetPackageInfo will clear this flag

Optional parameters for rp_loadPackageInfo can choose individual fields but if used the data won't be
cached, but it can be a lot faster - we use this to grab pkg_origin for the section menus in RetroPie-Setup.

joolswills:
I have converted all ports that will work with the new functionality. I am going to add some changes for some modules that won't currently work next.

joolswills:
Have added rp_module_repo for packages that use md_inst and also adjusted sixaxis to use md_build (as it didn't seem to need to use a subfolder) although it would have worked as we now record the first source folder in gitPullOrClone.

However dkms packages on raspberry pi os won't get run if a kernel is updated so we need to force this. Ideally on a kernel update dkms packages would be rebuilt as they are with dkms packages on Ubuntu - I assume this is due to the way the raspberry pi packages are created, unless we are doing something incorrectly. Would be nice to hook into a package update to rebuild our dkms drivers outside of RetroPie-Setup so I will look into this, or just fall back on always updating them as currently which only works when updating the OS packages within RetroPie-Setup.

cmitu:
> However dkms packages on raspberry pi os won't get run if a kernel is updated so we need to force this.

The kernel update runs `dkms_autoinstaller` from `/etc/kernel/postint.d/dkms`, which re-runs the installation for installed modules. I think the kernel modules (setup via `dkms`) that we install are upgraded automatically even on normal (i.e. out of RetroPie-Setup) updates.

joolswills:
That's what I thought but I updated via apt-get with xpad installed (to new 5.10 kernel although I was on a rpi-update kernel previously) and it wasn't rebuilt, which is why I wanted to check it. I know you've spent some time on this stuff @cmitu so happy if you want to look into it - but no obligations of course :-)

cmitu:
I did the same update yesterday (no `rpi-update` usage though) and the modules I had installed were upgraded.
```
 dkms status
hid-xpadneo, v0.9, 5.10.11+, armv7l: installed
hid-xpadneo, v0.9, 5.10.11-v7+, armv7l: installed
hid-xpadneo, v0.9, 5.10.11-v7l+, armv7l: installed
xpad, 0.4, 5.10.11+, armv7l: installed (original_module exists)
xpad, 0.4, 5.10.11-v7+, armv7l: installed (original_module exists)
xpad, 0.4, 5.10.11-v7l+, armv7l: installed (original_module exists)
```

I can certainly run more tests, I wonder if using `rpi-update` didn't actually broke the module (`xpad`) since it doesn't install headers.

joolswills:
May have been that. I'm testing on another system also just to check if it wasn't my error.

joolswills:
Sorry. Seems ok. It was fine on my rpi-update system - brain glitch :-)

scriptmodules/supplementary/xpadneo.sh
```diff
@@ -12,6 +12,7 @@
 rp_module_id="xpadneo"
 rp_module_desc="Advanced Linux driver for Xbox One wireless gamepads"
 rp_module_licence="GPL3 https://raw.githubusercontent.com/atar-axis/xpadneo/master/LICENSE"
+rp_module_repo="git https://github.com/atar-axis/xpadneo.git 0.9"
```
cmitu, line 15: Small typo - the release tag is `v0.9`, not `0.9`.

scriptmodules/libretrocores/lr-mesen.sh
```diff
@@ -13,11 +13,12 @@ rp_module_id="lr-mesen"
 rp_module_desc="High-accuracy NES and Famicom emulator"
 rp_module_help="ROM Extensions: .nes .fds .unf .unif .zip\n\nCopy your NES roms to $romdir/nes\nFamicom roms to $romdir/fds\nCopy the recommended BIOS file disksys.rom to $biosdir"
 rp_module_licence="GPL3 https://raw.githubusercontent.com/sourmesen/mesen/master/LICENSE"
+rp_module_repo="git https://github.com/SourMesen/Mesen.git master"
```
cmitu, line 16: (Unrelated to this PR) Maybe switch to https://github.com/libretro/Mesen.
Original repository has been archived (see the author's [announcement](https://old.reddit.com/r/emulation/comments/j58hfu/mesen_and_mesens_github_repositories_are_now/g7sti7q/)).

scriptmodules/emulators/simcoupe.sh
```diff
@@ -13,20 +13,25 @@ rp_module_id="simcoupe"
 rp_module_desc="SimCoupe SAM Coupe emulator"
 rp_module_help="ROM Extensions: .dsk .mgt .sbt .sad\n\nCopy your SAM Coupe games to $romdir/samcoupe."
 rp_module_licence="GPL2 https://raw.githubusercontent.com/simonowen/simcoupe/master/License.txt"
+rp_module_repo="git https://github.com/simonowen/simcoupe.git master :_get_branch_simcoupe"
```
cmitu, line 16: I think there's an extra `master` here.

scriptmodules/emulators/ppsspp.sh
```diff
@@ -25,9 +26,7 @@ function depends_ppsspp() {
 }
 
 function sources_ppsspp() {
-    local branch="$1"
-    [[ -z "$branch" ]] && branch="master"
-    gitPullOrClone "$md_build/ppsspp" https://github.com/hrydgard/ppsspp.git "$branch"
+    gitPullOrClone
     cd "ppsspp"
```
cmitu, line 30: Is the `cd` still needed if cloning goes directly to `$md_build` ?

scriptmodules/emulators/hatari.sh
```diff
@@ -13,6 +13,7 @@ rp_module_id="hatari"
 rp_module_desc="Atari emulator Hatari"
 rp_module_help="ROM Extensions: .st .stx .img .rom .raw .ipf .ctr .zip\n\nCopy your Atari ST games to $romdir/atarist\n\nCopy Atari ST BIOS (tos.img) to $biosdir"
 rp_module_licence="GPL2 https://hg.tuxfamily.org/mercurialroot/hatari/hatari/file/9ee1235233e9/gpl.txt"
```
cmitu, line 15: (Unrelated to the PR's functionality) License link is broken, https://git.tuxfamily.org/hatari/hatari.git/plain/gpl.txt should be ok.

cmitu:
I gave it a test and left a few notes.
Source version seems to work fine, built also few packages that have dynamic branch/commit info and they're reported correctly.
Startup does take longer - maybe add a progress info message while the package info is read and cached ?

scriptmodules/emulators/ppsspp.sh
joolswills: Actually this needs to be put back to ppsspp subfolder.

joolswills:
Thanks. I'll fix the issues mentioned. The slowdown is caused by the additional check to avoid module name conflicts with 3rd party repos introduced by another PR. I don't think this is any slower - it was in the menus but I resolved that in later changes in this PR.

I'm reluctant to introduce a progress for 6 seconds load time on rpi3/4 but I may limit the check in a future PR to only check 3rd party repos and assume our IDs don't conflict which will help.

scriptmodules/supplementary/xpadneo.sh
joolswills: I should have run the check_repos after last changes :-) but it does pick up on this which is good. Thanks.

scriptmodules/libretrocores/lr-mesen.sh
joolswills: Have merged this into the original commit but have added it as an addition commit msg note.

scriptmodules/emulators/simcoupe.sh
joolswills: The additional master was wrong. But initializing `$branch` to master in the function seems correct unless that's not the default branch?  The function hook is only needed for stretch support so will remove in the future I guess but let me know if I've missed something. For now have removed master from md_module_repo variable.

joolswills:
Have rebased the changes into relevant commits including fixing rebasing over master with the colour changes.

I have a little more work to do, then I will take it out of draft for a final going over.

joolswills:
Thanks for the preliminary testing @cmitu . Let me know if I missed anything and if I misunderstood the simcoupe issue. Thanks.

joolswills:
Just to note I committed a fix for the performance issue in https://github.com/RetroPie/RetroPie-Setup/commit/4fc2e4ee9e7bc4c7c2bcdfe7a2f9f47dcf78646e

Menu navigation in this PR should be quicker than master, due to the faster package info loading and caching so I'm happy with the speed now.

joolswills:
Mupen64plus now supports checking for updates, and it works if any of the plugins repos have been updated. Was a lot of extra code, but the function hook support could be leveraged in the future if we need it. I put lots of details in the last few commits.

joolswills:
I found some cases where no internet connection wasn't handled well with the various checks. I have some code to improve it, but it needs some cleanups and comments. So this is still not quite ready yet.

scriptmodules/packages.sh
```diff
@@ -386,45 +440,197 @@ function rp_hasBinary() {
...
+                *)
+                    # fallback on forcing an update
+                    ret=2
+                    ;;
+            esac
+
+            # if we are currently not going to update - check the date of the module code
+            # if it's newer than the install date of the module we force an update
+            if [[ "$ret" -eq 1 && "$__ignore_module_date" -ne 1 ]]; then
+                local module_date="$(date -Iseconds -r "${__mod_info[$id/path]}")"
```
cmitu, line 616: Is this correct ? `module_date` shouldn't be using `$__mod_info[$id/date]` ?

joolswills: It compares against the last change date of the script/module file. So if a scriptmodule is changed an update is forced.

joolswills:
I have done a few additional tweaks including some extra caching and some fixes for some cases I missed (when no network connection is available). Also some extra checks and some output when checking for updates.

Will do a last check and then merge.

PR #3307, opened 2021-02-06
setup - change colours around to work with other terminals

joolswills:
On the Raspberry Pi console, bold text results in a lighter colour, rather than actually being bold.

One user (it may not be a common issue) reported on their terminal that they couldn't see installed packages. This was due to our "bold white" for installed applications which ended up being bold grey on a grey background - so they couldn't see what was installed without hovering over it.

I have switched the colours around here, so the section heading is just "bold" which will be a light
grey on the Raspberry Pi, and blue for installed apps, and red for unavailable packages.

I'm not completely happy with this, as it's more colourful than before, and perhaps too busy. However, this should work on terminals that decode it as 7 colours + bold rather than 15 colours etc.

--

Thoughts welcome. I was hoping to primarily highlight installed packages, and didn't want it to get busy with colour. However if this is not to people's liking it may be easiest to just switch back to plain text in the menus.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -335,15 +335,15 @@ function section_gui_setup() {
             # do a heading for each origin and module type
             if [[ "$last_type" != "$type" ]]; then
                 info="$type"
-                pkgs+=("----" "\Z4$info ----\Zn" "Packages from $info")
+                pkgs+=("----" "\Z4$info ----" "Packages from $info")
                 last_type="$type"
             fi
             if ! rp_isEnabled "$id"; then
-                info="\Zb*$id - Not available for your system\Zn"
+                info="\Z1$id\Zn - Not available for your system"
             else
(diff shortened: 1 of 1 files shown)


cmitu:
Yes, a bit too busy with the blue color for installed packages - which is kind of redundant since there is already an indication that the package is installed.
The rest looks ok (bold repository markers and red highlight for unavailable packages).

joolswills:
I agree with the blue. I quite liked the white we had, a si thought it made it easier at a quick glance to see but it's not needed with the unavailable packages in red as they don't confuse the listing as much. I'll adjust this to remove the blue, but may put the heading back as blue again as it matched with the numeric colours. I'll tweak it and commit and we can revisit if there's more issues reported by users.

joolswills:
I've adjusted it now, so the only changes in this PR are to remove the white for installed packages and àdd the red for unavailable. Also removed setting the colours back at end of each line as it doesn't seem required.

PR #3308, opened 2021-02-13
customhidsony updates to reflect currently supported base Debian version

psyke83:
* Remove custombluez, as this was only useful for Jessie-based distributions, which we no longer support.
* Switch to updated GASIA patch authored by @mwelchuk

@mwelchuk - can you let me know if there's any problem with me switching to your patch in RetroPie? I still don't own any GASIA pads to test this myself, but I'm confident that your quirk-based solution is better than my prior patch.


joolswills:
Hi @psyke83 - I didn't comment on this as it was in draft but I'm happy to remove the bluez module.

PR looks good to me btw, if you're happy for me to merge?

joolswills:
Thanks.

PR #3311, opened 2021-02-17
openmsx: fix building on Ubuntu 18.04/GCC 7

cmitu:
Alternatively, we could install `gcc-8` on Ubuntu and compile with it - seems to work.


joolswills:
Thanks. Will merge. If you prefer to change it feel free, but this seems a safe workaround for now without having to install a newer GCC. Thanks.

PR #3312, opened 2021-02-18
lzdoom: update to 3.87c

cmitu:
Looks like GL2 compatibility has been kept for LZDoom versions post 3.86, so we can update to the latest 3.x release tag.

Other:
* Removed the CFLAGS manipulation, since the math optimizations have been removed in c3751a4f0ee37451991b775e0bea8ba3df424a20, but left the warning for building with `-funsafe-math-optimizations`
* Use the direct link to the license page


joolswills:
Thanks.

PR #3314, opened 2021-02-21
dolphin - correct build from source script

rusher:
After recent modification of dolphin.sh, build from source failed, because 2 branches are set in git command.


joolswills:
Thanks.

PR #3315, opened 2021-02-21
packages - handle git repositories with two tag references

joolswills:
When the remote repository uses an annotated Git tag, the real commit is found
by looking for the "tag^{}" reference, since the the tag ref will point to the
tag object itself, instead of the tagged commit. See gitrevisions(7).

### DIFF
--- a/scriptmodules/packages.sh
+++ b/scriptmodules/packages.sh
@@ -528,10 +528,15 @@ function rp_getRemoteRepoHash() {
     set -o pipefail
     case "$type" in
         git)
-            cmd=(git ls-remote "$url" "$branch")
-            # grep to make sure we only return refs/heads/BRANCH and refs/tags/BRANCH in case there are
-            # additional references to the branch/tag eg refs/heads/SOMETHINGELSE/master which can be the case
-            hash=$("${cmd[@]}" 2>/dev/null | grep -P "\trefs/(heads|tags)/$branch" | cut -f1)
+            # when the remote repository uses an annotated git tag, the real commit is found by looking for the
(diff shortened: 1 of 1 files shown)


cmitu:
Looks fine, but for some reason `advmame` still shows that it needs an update. Need to test more

The description could be shorter:

> When the remote repository used an annotated Git tag, the real commit is found by looking for the "tag^{}" reference, since the the tag ref will point to the tag object itself, instead of the tagged commit. See gitrevisions(7).

cmitu:
I'm thinking that it fails here
https://github.com/RetroPie/RetroPie-Setup/blob/fb5a22cfb75ea977484a0631949c80a263ce39c7/scriptmodules/packages.sh#L628-L635

because `$__ignore_module_date` is un-intialized.

cmitu:
OK, re-tested with `advmame` (where I noticed the issue) and looks fine.

cmitu:
I've found out why the date diff occurs: `${__mod_info[$id/path]}` isnewer than the `pkg_date` because the scriptmodule was updated after building from source happened. Looks reasonable enough - scripmodule changes could produce different results.

cmitu (approved):

joolswills:
Updated the commit msg / PR description and code comment as suggested.

cmitu:
Mis-spelled the 'the the', copied from my description.

joolswills:
I completely missed that. I changed one word but my brain skipped that. I'll do a cosmetic commit for the comment.

joolswills:
Not the first time "the the" has made trouble hehe

PR #3316, opened 2021-02-23
cgenius: update to the latest stable version

cmitu:
* Bug fixes
* GUI, sound and controller improvements.
* Boost is no longer required, replaced with TinyXML2.
* Python scripting engine replaced with Lua (included).


joolswills:
Thanks.

PR #3318, opened 2021-03-01
Rework the checks and error messages displayed on network error

joolswills:
Create a new function isConnected in helpers.sh which calls getIPAddress, but also sets a new global __NET_ERRMSG if not connected. This is used in setup.sh and packages.sh so we don't duplicate the same message between files.

Split out a new function runCurl from the download function, to do lower level curl calls, but capture any errors into __NET_ERRMSG for display in setup.sh or via the packaging functions. Use a connect timeout of 10 in the download function rather than 60 which is quite long, but keep the 60 second limit on no data transfer.

Adjust rp_remoteFileExists to use runCurl, so we can capture any error when checking a remote file, and use this value for display in setup and in logs etc. Switch to using long versions of curl options for readability, and add --show-error so we still output errors when being silent.

Also switch rp_getFileDate to use new runCurl function, so that any curl argument overrides are handled, and use long format parameters also. No error handling is done here though.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -185,6 +185,7 @@ function package_setup() {
         hasFlag "${__mod_info[$id/flags]}" "nonet" && has_net=1
 
         if [[ "$has_net" -eq 1 ]]; then
+            dialog --backtitle "$__backtitle" --infobox "Checking for updates for $id ..." 3 60 >/dev/tty
             rp_hasBinary "$id"
             local ret="$?"
             [[ "$ret" -eq 0 ]] && has_binary=1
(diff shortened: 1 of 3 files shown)


joolswills:
@cmitu @hhromic 

This should be clearer in regards to error. May not be too useful for end users since we display the curl error, but good for debugging.

I'm still checking this over but please test. I may tweak the doxygen comments as they may not be as good as they could be but let me know what you think. Thanks.

joolswills:
I also added a small commit to add a dialog when checking for updates when going into a package in package management. I'm the case of slow connections its more useful than looking like nothing is happening etc.

PR #3319, opened 2021-03-02
openmsx: fix the wrong emulator parameter for model

cmitu:
Turns out `-m` is not a shorthand for `-machine`.
Changed the names emulators shown in runcommand selection, `+` may be problematic when storing the entry in emulators.cfg


joolswills:
Thanks.

PR #3322, opened 2021-03-05
add dosbox staging - an enhanced modernized dosbox fork

nemo93:
PR to include DOSBox Staging to Retropie. This fork is an attempt to revitalize DOSBox's development process. It's not a rewrite, but a continuation and improvement on the existing DOSBox codebase while leveraging modern development tools and practices. This fork is in very active development. 

Besides code modernization and refinements, Staging brings the following features - documented on the [Wiki](https://github.com/dosbox-staging/dosbox-staging/wiki):
* SDL 2.0
* General MIDI via Fluidsynth (2.x built-in statically)
* MT-32 via Munt (2.x built-in statically)
* Major [improvements](https://github.com/dosbox-staging/dosbox-staging/wiki/Gravis-UltraSound-Enhancements) to Gravis Ultrasound playback
* Support FLAC, Opus, Vorbis, and MP3 CD-DA tracks
* [Pixel-perfect](https://dosbox-staging.github.io/v0-75-0/#pp) scaling mode
* [AUTOTYPE ](https://github.com/dosbox-staging/dosbox-staging/wiki/AUTOTYPE-Candidates) command
* Audio pop and click prevention
* Support of GLSL shaders
* Improve Telnet-mode handling for Modem emulation
* and [many others](https://github.com/dosbox-staging/dosbox-staging/wiki/features-differences)

Repo: https://github.com/dosbox-staging/dosbox-staging
Official website: https://dosbox-staging.github.io/

All credits to @dreamer and @kcgen for the incredible work done on that fork.

### DIFF
--- a/scriptmodules/emulators/dosbox-staging.sh
+++ b/scriptmodules/emulators/dosbox-staging.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="dosbox-staging"
+rp_module_desc="modern DOS/x86 emulator focusing on ease of use"
+rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
+rp_module_licence="GPL2 https://raw.githubusercontent.com/dosbox-staging/dosbox-staging/master/COPYING"
+rp_module_repo="git https://github.com/dosbox-staging/dosbox-staging.git master"
+rp_module_section="exp"
+rp_module_flags="sdl2"
+
+function depends_dosbox-staging() {
+    getDepends cmake libasound2-dev libglib2.0-dev libopusfile-dev libpng-dev libsdl2-dev libsdl2-net-dev meson ninja-build
+}
+
+function sources_dosbox-staging() {
+    gitPullOrClone
+}
+
+function build_dosbox-staging() {
+    local params=(-Dbuildtype=release -Ddefault_library=static --prefix="$md_inst")
+
+    # Fluidsynth (static)
+    cd "$md_build/contrib/static-fluidsynth"
+    make
+    export PKG_CONFIG_PATH="${md_build}/contrib/static-fluidsynth/fluidsynth/build"
+
+    cd "$md_build"
+    meson setup "${params[@]}" build
+    ninja -C build
+
+    md_ret_require=(
+        "$md_build/build/dosbox"
+    )
+}
+
+function install_dosbox-staging() {
+    cd "$md_build/build"
+    meson install
+}
+
+function configure_dosbox-staging() {
+    configure_dosbox
+
+    [[ "$md_id" == "remove" ]] && return
+
+    local config_path=$(su "$user" -c "\"$md_inst/bin/dosbox\" -printconf")
+    if [[ -f "$config_path" ]]; then
+        iniConfig " = " "" "$config_path"
+        if isPlatform "rpi"; then
+            iniSet "fullscreen" "true"
+            iniSet "fullresolution" "desktop"
+            iniSet "output" "texturenb"
+            iniSet "core" "dynamic"
+            iniSet "cycles" "25000"
+        fi
+    fi
+}
--- a/scriptmodules/emulators/dosbox.sh
+++ b/scriptmodules/emulators/dosbox.sh
@@ -65,6 +65,7 @@ function configure_dosbox() {
     local def=0
     local launcher_name="+Start DOSBox.sh"
     local needs_synth=0
+    local config_dir="$home/.$md_id"
     case "$md_id" in
         dosbox-sdl2)
             launcher_name="+Start DOSBox-SDL2.sh"
@@ -76,6 +77,10 @@ function configure_dosbox() {
             # set dispmanx by default on rpi with fkms
             isPlatform "dispmanx" && ! isPlatform "videocore" && setBackend "$md_id" "dispmanx"
             ;;
+        dosbox-staging)
+            launcher_name="+Start DOSBox-Staging.sh"
+            config_dir="$home/.config/dosbox"
+            ;;
         *)
             return 1
             ;;
@@ -83,7 +88,7 @@ function configure_dosbox() {
 
     mkRomDir "pc"
 
-    moveConfigDir "$home/.$md_id" "$md_conf_root/pc"
+    moveConfigDir "$config_dir" "$md_conf_root/pc"
 
     addEmulator "$def" "$md_id" "pc" "bash $romdir/pc/${launcher_name// /\\ } %ROM%"
     addSystem "pc"
@@ -140,16 +145,18 @@ _EOF_
     chmod +x "$romdir/pc/$launcher_name"
     chown $user:$user "$romdir/pc/$launcher_name"
 
-    local config_path=$(su "$user" -c "\"$md_inst/bin/dosbox\" -printconf")
-    if [[ -f "$config_path" ]]; then
-        iniConfig " = " "" "$config_path"
-        iniSet "usescancodes" "false"
-        iniSet "core" "dynamic"
-        iniSet "cycles" "max"
-        iniSet "scaler" "none"
-        if isPlatform "rpi" || [[ -n "$(aconnect -o | grep -e TiMidity -e FluidSynth)" ]]; then
-            iniSet "mididevice" "alsa"
-            iniSet "midiconfig" "128:0"
+    if [[ "$md_id" == "dosbox" || "$md_id" == "dosbox-sdl2" ]]; then
+        local config_path=$(su "$user" -c "\"$md_inst/bin/dosbox\" -printconf")
+        if [[ -f "$config_path" ]]; then
+            iniConfig " = " "" "$config_path"
+            iniSet "usescancodes" "false"
+            iniSet "core" "dynamic"
+            iniSet "cycles" "max"
+            iniSet "scaler" "none"
+            if isPlatform "rpi" || [[ -n "$(aconnect -o | grep -e TiMidity -e FluidSynth)" ]]; then
+                iniSet "mididevice" "alsa"
+                iniSet "midiconfig" "128:0"
+            fi
         fi
     fi
 }


scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+
+rp_module_id="dosbox-staging"
+rp_module_desc="DOS emulator (enhanced and modern DOSBox fork)"
+rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
+rp_module_licence="GPL2 https://github.com/dosbox-staging/dosbox-staging/blob/master/COPYING"
+rp_module_repo="git https://github.com/dosbox-staging/dosbox-staging.git master"
+rp_module_section="exp"
+
+function depends_dosbox-staging() {
+    getDepends ccache build-essential meson ninja-build libasound2-dev libpng-dev libsdl2-dev libsdl2-net-dev libopusfile-dev libfluidsynth-dev
```
cmitu, line 20: `ccache` and `libfluidsynth-dev` shouldn't be listed. 

Since `libfluidsynth` is statically linked, I guess we should list the dependencies for `libfluidsynth`. `libglib2.0-dev` seems to be needed, other libraries may be optional since the `fluidsynth` build doesn't need Jack/ALSA/PulseAudio/etc. support.

`cmake` is also needed by the `fluidsynth` build. While technically it's installed when adding `emulationstation`, it's better to specifically add it.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+function sources_dosbox-staging() {
+    gitPullOrClone
+}
+
+function build_dosbox-staging() {
+    local params=(-Dbuildtype=release -Ddefault_library=static)
+
+    # Fluidsynth
+    cd $md_build/contrib/static-fluidsynth
+    make -j $(nproc)
```
cmitu, line 32: No need to add `-j $(nproc)`, the setup script will automatically adjust the concurrency based on the platfom/detected CPUs/available memory.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+function install_dosbox-staging() {
+    md_ret_files=(
+        'build/dosbox'
+        'COPYING'
+        'README'
+    )
+}
+
+function configure_dosbox-staging() {
+    local launcher_name="+Start DOSBox-staging.sh"
```
cmitu, line 53: Use `DOSBox-Staging` (capitalized `staging`), closer to the project's name.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="dosbox-staging"
+rp_module_desc="DOS emulator (enhanced and modern DOSBox fork)"
```
cmitu, line 13: Wouldn't it be better to use the website descriotion (_DOS/x86 emulator focusing on ease of use_) ? `fork` is also fine, just a suggestion.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+    bash "\${params[@]}"
+    exit
+elif [[ "\${params[0]}" == *.conf ]]; then
+    params=(-userconf -conf "\${params[@]}")
+else
+    params+=(-exit)
+fi
+
+# MT-32 Munt library
+export LD_LIBRARY_PATH="\$LD_LIBRARY_PATH:$md_inst"
```
cmitu, line 79: Is this still needed if `MT-32` is statically linked ?

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+
+function configure_dosbox-staging() {
+    local launcher_name="+Start DOSBox-staging.sh"
+    local def="0"
+
+    mkRomDir "pc"
+    rm -f "$romdir/pc/$launcher_name"
+    if [[ "$md_mode" == "install" ]]; then
+        cat > "$romdir/pc/$launcher_name" << _EOF_
+#!/bin/bash
```
cmitu, line 60: Better use
```
#!/usr/bin/env bash

File: scriptmodules/emulators/dosbox-staging.sh [hunk 7]
```diff
@@ -0,0 +1,98 @@
...
+#
+# if present
+#   /home/pi/.config/dosbox/dosbox-staging-git.conf
```
cmitu, line 63: Use `$user` instead of `pi`. It will be expanded in the script at runtime.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+    md_ret_files=(
+        'build/dosbox'
+        'COPYING'
+        'README'
+    )
+}
+
+function configure_dosbox-staging() {
+    local launcher_name="+Start DOSBox-staging.sh"
+    local def="0"
```
cmitu, line 54: Remove `def` (also see below)

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+        if [[ -f "$config_path" ]]; then
+	    iniConfig " = " "" "$config_path"
+	    iniSet "fullscreen" "true"
+	    iniSet "fullresolution" "desktop"
+	    iniSet "output" "texturenb"
+	    iniSet "cycles" "25000"
+        fi
+    fi
+
+    addEmulator "$def" "$md_id" "pc" "bash $romdir/pc/${launcher_name// /\\ } %ROM%"
```
cmitu, line 96: `$def` is unchanged, so just replace it with `0`. In time, we may decide (based on platform) if we make it default, though it would need to be in the main/optional section by then :).

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="dosbox-staging"
+rp_module_desc="DOS emulator (enhanced and modern DOSBox fork)"
+rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
+rp_module_licence="GPL2 https://github.com/dosbox-staging/dosbox-staging/blob/master/COPYING"
```
cmitu, line 15: Use the 'raw' variant - https://raw.githubusercontent.com/dosbox-staging/dosbox-staging/master/COPYING.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+        chmod +x "$romdir/pc/$launcher_name"
+        chown $user:$user "$romdir/pc/$launcher_name"
+
+        local config_path=$(su "$user" -c "\"$md_inst/dosbox\" -printconf")
+        if [[ -f "$config_path" ]]; then
+	    iniConfig " = " "" "$config_path"
+	    iniSet "fullscreen" "true"
+	    iniSet "fullresolution" "desktop"
+	    iniSet "output" "texturenb"
+	    iniSet "cycles" "25000"
```
cmitu, line 92: Is this value ok also for x86/PC platforms or it should be set to `auto` ?
You can test for the platform and preset another value if needed (see [here](https://github.com/RetroPie/RetroPie-Setup/blob/90173a8181354fa6bf99da85414b25e63635c241/scriptmodules/emulators/dosbox.sh#L134)).

scriptmodules/emulators/dosbox-staging.sh
joolswills, line 32: Please leave off the -j parameter. This is handled by Retropie.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+
+function sources_dosbox-staging() {
+    gitPullOrClone
+}
+
+function build_dosbox-staging() {
+    local params=(-Dbuildtype=release -Ddefault_library=static)
+
+    # Fluidsynth
+    cd $md_build/contrib/static-fluidsynth
```
joolswills, line 31: Please put quotes around paths

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,98 @@
...
+    make -j $(nproc)
+    export PKG_CONFIG_PATH="${md_build}/contrib/static-fluidsynth/fluidsynth/build"
+
+    cd $md_build
```
joolswills, line 35: Quotes

scriptmodules/emulators/dosbox-staging.sh
joolswills, line 20: Is ccache a requirement? May be better to leave this out.

joolswills:
If the launch script is basically the same or close to that of the dosbox launch script it may be worth looking if the code can be shared.

joolswills:
Thanks for this.

cmitu (commented):
Thank you for the PR !

I've left a few notes sprinkled here and there, looks fine so far. 1st version didn't properly  use `LD_LIBRARY_PATH`, but I see you've pushed some changes and removed the need for that.

One thing that I think needs to be added is a link to the configuration file to `/opt/retropie/configs/pc`, so the users will be able to modify the configuration without SSH access. You can do so by using the `moveConfigFile` function in the `configure` function.
```
moveConfigFile "$home/.config/dosbox/dosbox-staging.conf" "$md_conf_root/pc/dosbox-staging-git.conf"
```

Regarding the config file name, I recall seeing a discussion about `master` using the `-git` suffix, while releases being named just `dosbox-staging.conf`. Is there a parameter for the build where we can choose just `dosbox-staging.conf` (i.e. no `-git`) ?

I'll give a few more tests and see how it works.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Got it. Parameter left off.

scriptmodules/emulators/dosbox-staging.sh
nemo93: sorry for that.

scriptmodules/emulators/dosbox-staging.sh
nemo93: sure. Fixed.

scriptmodules/emulators/dosbox-staging.sh
nemo93: -j left off.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Nice one! thanks.

scriptmodules/emulators/dosbox-staging.sh
nemo93: I propose the following compromise: modern DOS/x86 emulator focusing on ease of use (modern is important given the massive work done to modernize the codebase).

scriptmodules/emulators/dosbox-staging.sh
nemo93: My mistake. Sorry. Removed.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Done.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Removed.

scriptmodules/emulators/dosbox-staging.sh
nemo93: "0" has been added as proposed.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Done.

scriptmodules/emulators/dosbox-staging.sh
nemo93: You're right. The above values are for "rpi". I forgot Retropie works on many platforms not only "rpi"!! I've added a condition for "rpi" platform. On x86 it's fine leaving on auto.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Done. Though it's not consistent with the other scriptmodule ([here](https://github.com/RetroPie/RetroPie-Setup/blob/90173a8181354fa6bf99da85414b25e63635c241/scriptmodules/emulators/dosbox.sh#L82)).

scriptmodules/emulators/dosbox-staging.sh
nemo93: `ccache` and `libfluidsynth-dev` both removed from list. I've added `cmake` and `libglib2.0-dev` instead + sorted list. For further reference Fluidsynth requirements are being listed [here](https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake#requirements).

nemo93: It really helps during my many testing sessions where I had to compile several times in a raw the emu. I guess under normal conditions this one not needed. Dropped.

scriptmodules/emulators/dosbox-staging.sh
cmitu: Looks fine.

nemo93:
Thanks @cmitu @joolswills for your very helpful feedback. I just pushed another commit to address your points. With respect to the 2 remaining questions:
1. launcher/wrapper code sharing. I did hesitate indeed to share code with the `dosbox.sh` [scriptmodule](https://github.com/RetroPie/RetroPie-Setup/blob/90173a8181354fa6bf99da85414b25e63635c241/scriptmodules/emulators/dosbox.sh#L106). I was unsure to add more conditions to that module all around just to keep the following bit. I decided to not do that in the end.
```
params=("\$@")
if [[ -z "\${params[0]}" ]]; then
    params=(-c "@MOUNT C $romdir/pc -freesize 1024" -c "@C:")
elif [[ "\${params[0]}" == *.sh ]]; then
    bash "\${params[@]}"
    exit
elif [[ "\${params[0]}" == *.conf ]]; then
    params=(-userconf -conf "\${params[@]}")
else
    params+=(-exit)
fi
```
2. config file name. Ah! That's a [long-running discussion](https://github.com/dosbox-staging/dosbox-staging/issues/73). It's not that easy to change version number given it is coded somewhere deep in the code. In short, you'll end up with `dosbox-staging-git.conf` if you compile from `git/master` and with `dosbox-staging.conf` whenever you download a stable release [binary](https://dosbox-staging.github.io/downloads/windows/). Given we'll be relying on `master` here the file will be named after `dosbox-staging-git.conf`. I've added a local variable for easy change if that name should change again.

dreamer:
@cmitu @nemo93 About the `-git` suffix for default .conf file when building from `master` branch.

The root cause for having this distinction was our old buildsystem - for development purposes me and @kcgen wanted to have separe .conf file for debug builds (as we retouch available features, translations, descriptions a lot during development) - Autoconf did not give us ability to easily distinguish between release and debug builds, but Meson removed this problem for us.

We can modify the buildsystem to name the default .conf file `dosbox-staging.conf` for release builds and keep `dosbox-staging-git.conf` for our debug builds, no matter the branch used. I guess that would solve the problem for you?

cmitu:
> We can modify the buildsystem to name the default .conf file `dosbox-staging.conf` for release builds and keep `dosbox-staging-git.conf` for our debug builds, no matter the branch used. I guess that would solve the problem for you?

Thank you for chiming in. 
Yes, it would be better for us if we can provide a consistent configuration filename when installing, but I don't want to impose new restrictions on your build/release process. I remember that when using `autoconf`, the naming was handled via `CONF_BRAND`, so I guess if you intent to implement something similar for the `meson` build system, the change would be transparent for the RetroPie script (i.e. not set a release branding and always use the `-git` suffix).

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -0,0 +1,100 @@
...
+                iniSet "fullscreen" "true"
+                iniSet "fullresolution" "desktop"
+                iniSet "output" "texturenb"
+                iniSet "core" "dynamic"
+                iniSet "cycles" "25000"
+            fi
+        fi
+    fi
+
+    moveConfigFile "$home/.config/dosbox/$config_filename" "$md_conf_root/pc"
```
cmitu, line 96: This should be 
```
moveConfigFile "$home/.config/dosbox/$config_filename" "$md_conf_root/pc/$config_filename"
```
otherwise the symlink is created to the `pc` folder instead.

nemo93: Great example of copy/pasting without double-checking :-( Silly me. It has been fixed and confirmed proper symlink.

joolswills:
Thanks for looking into / dealing with the changes. I will give it a good test and come back to you.

nemo93:
@joolswills @cmitu the default .conf file is now being named after `dosbox-staging.conf`, for consistency's sake.

dreamer:
@nemo93 I think it was a good decision to go with master branch for this PR :)

I'm not sure how much longer it will take until we'll have 0.77.0 stable release (probably another month of development and then another month of testing and bugfixing, maybe more). Once we'll release 0.77.0 version though, I advise to switch to release branch and stick to the stable version going forward. That's because master branch will probably destabilize pretty quickly as new features will be merged in.

joolswills:
I would like to squash all these commits before merge myself so I can include details from all the changes, but feel free to make further modifications as needed. Thanks.

nemo93:
thanks a lot @joolswills to me it looks very nice and can be merged as-is. For now and as @dreamer mentioned let's stick to master branch for now and we may switch to a tag in the future.

Thanks again to @kcgen and @dreamer for the amazing work done on that fork and to everyone involved in testing.

nemo93:
@joolswills Commits squashed. I just left your own commits msgs as I was unsure.

joolswills (commented):
Looks good. I removed the build-essentials dependency (handled by RetroPie), rebased over latest master and force pushed.

Btw I noticed you're using master branch on your fork. If you want to contribute further changes it may be easier to create a new branch for each PR.

joolswills:
Thanks.

PR #3324, opened 2021-03-11
lr-fbneo.sh: be less specific about version

barbudreadmon:
remove version from description, i think the previous 0.2.97.44 description is still causing misunderstandings for ppl who didn't update their retropie-setup script yet, and we'll probably change version more regularily from now on.


joolswills:
Thanks.

PR #3326, opened 2021-03-15
Checkinstall replaced by make install

luizoti:
This change allows installations on Debian, tested on: Debian 10

Linux HP 5.10.0-0.bpo.3-amd64  SMP Debian 5.10.13-1~bpo10+1 (2021-02-11) x86_64 GNU/Linux

### DIFF
--- a/scriptmodules/supplementary/ps3controller.sh
+++ b/scriptmodules/supplementary/ps3controller.sh
@@ -74,9 +74,9 @@ function remove_ps3controller() {
     )
     for paths in ${remove_paths[@]}; do
         if [[ -d ${paths} ]]; then
-            sudo rm -rfv ${paths}
-        elif [[ condition ]]; then
-            sudo rm -fv ${paths}
+            rm -rfv ${paths}
+        elif [[ -f ${paths} ]]; then
+            rm -fv ${paths}
         fi
     done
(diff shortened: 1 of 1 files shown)


cmitu:
That passes the installation, but the removal is not handled - there is no Debian package installed to remove it.

luizoti:
I update the PR with a method to remove compiled pkgs.

joolswills:
Sorry for delay on this.

I'm not keen on maintaining a list of files for removal like this, but you can get checkinstall from buster-backports and it looks like it's included with Debian Testing and Unstable so isn't being dropped. Maybe the package was unmaintained for a while (I've not looked into it further)

I'm inclined to leave this as is.

joolswills:
Thanks for the contribution though.

luizoti:
I remember that at the time I didn't have the pro debian 10 package and even if there is pro debian unstable, it is unstable, using the repositories in stable will probably break the distro, I wouldn't do it, so much so that at the time I didn't do it for other reasons. The list of files 

I'm pretty sure came from the sixad installation file itself. 

Anyway, I'm already in another distro and it doesn't affect me anymore.

joolswills:
Not sure what you mean sorry. checkinstall will be available on Debian 11 and is available for Debian 10 in the buster backports.

Closing - thanks for the feedback.

PR #3327, opened 2021-03-15
Detect Odroid platforms with expanded Hardware name in cpuinfo output

jwyse:
Fix platform detection for Odroid platforms, where /proc/cpuinfo Hardware field may end with (but not exactly match) "ODROID-XU[34]", e.g. "**Hardkernel** ODROID-XU4" on Ubuntu 20.04.2 LTS.


joolswills:
Thanks. Actually I have this change on a branch (for XU - the other boards didn't seem to need it when I tested with 20.04 but maybe different kernel version). However there are other issues with 20.04 on odroid with some driver dependency stuff afair so I may hold off until I've had a chance to look into it again.

PR #3328, opened 2021-03-21
audiosettings: fix PA options display when using mawk

cmitu:
`mawk` (default AWK on Debian) doesn't know about `gensub`, so use a Posix ERE instead.
Fix also the options index when choosing the PulseAudio sink so the correct audio output is chosen.


joolswills:
Thanks.

PR #3333, opened 2021-05-05
mame: remove binary name, not needed anymore

cmitu:
The suffix was removed in 0.229 (https://github.com/mamedev/mame/commit/adc23f3f74f64afa3a1258539a4f8ae2f7ae76c9)


joolswills:
Thanks.

PR #3334, opened 2021-05-08
xpadneo: update to 0.9.1

cmitu:
0.9.1 is a maintenance release:

"The highlights of this release are better compatibility with current Proton versions, fixes to Xbox One S|X controller support, removal of old dead configuration options, updated and improved documentation, better integration with DKMS, improved logging to aid issue resolving, future proofing the systemd and udev integration, and deprecating support for directional rumble which never properly worked."

The install script has slightly changed, so the build step was updated to accommodate that.


joolswills:
Thanks

PR #3335, opened 2021-05-08
lr-quicknes in main packages only for armv6 (pi1/0)

dankcushions:
As per https://docs.libretro.com/guides/retroachievements/#nes, lr-fceumm (default NES emulator for >armv6) now supports RetroAchievements, so I can't think of a reason to keep lr-quicknes in main packages for anything other than pi1/0 (where it's currently the default as it's the fastest).


joolswills:
It's not a big package but yeah, I agree. Thanks.

PR #3337, opened 2021-05-09
MAME Pulse dependency

dankcushions:
installing `lib-pulsedev` doesn't appear to have any effect on general sound, nor install `pulseaudio` (the purging of which is sometimes recommended to fix the ALSA vs Pulse audio issues)


joolswills:
Thanks

PR #3339, opened 2021-05-14
lr-theodore: Update LICENSE file path

Zlika:
Hello.
The current LICENSE path points to the old libretro clone of the upstream repository. The libretro clone does not exist anymore, the upstream repository must be used instead.


joolswills:
thanks.

PR #3340, opened 2021-05-15
lr-gpsp: remove the build workaround added in #3456036

cmitu:
The dynarec crash on ARM with GCC7+ have been solved in 6254bbb1d, building with GCC6 is no longer necessary.


joolswills:
Thanks.

PR #3342, opened 2021-05-20
move skyscraper output path to /.media

XenuIsWatching:
This will prevent emulation station from entering the directory attempting to parse every file within this directory for games.

In my experimentation, with 10000s of games each with media (screenshots, marquees, etc.) this reduced the startup time for emulation station from 8 minutes to 5 minutes.

Relevant discussions found here: https://github.com/muldjord/skyscraper/issues/307 ; https://github.com/RetroPie/EmulationStation/issues/752

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -1211,7 +1211,7 @@ function show_launch() {
             "$HOME/RetroPie/roms/$SYSTEM/images/${ROM_BN}-image"
             "$HOME/.emulationstation/downloaded_images/$SYSTEM/${ROM_BN}-image"
             "$HOME/.emulationstation/downloaded_media/$SYSTEM/screenshots/${ROM_BN}"
-            "$HOME/RetroPie/roms/$SYSTEM/media/screenshots/${ROM_BN}"
+            "$HOME/RetroPie/roms/$SYSTEM/.media/screenshots/${ROM_BN}"
         )
     fi
 
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -280,7 +280,7 @@ function _scrape_skyscraper() {
 
     if [[ "$use_rom_folder" -eq 1 ]]; then
         params+=(-g "$romdir/$system")
-        params+=(-o "$romdir/$system/media")
+        params+=(-o "$romdir/$system/.media")
         # If we're saving to the ROM folder, then use relative paths in the gamelist
         flags+="relative,"
     else


cmitu:
I don't agree with changing the default options because of a non-standard/advanced setup.

The scraping script is meant as an easy front-end option for users, for more advanced features, the command line and/or the configuration file can be used. You can modify the default output folders in the configuration for Skyscraper and then re-generate the gamelist(s).

Now, given @muldjord's modification for the new version (changing of the default output path from `media` to `.media`), I think a 3rd option could be added for the output folder (_use default_ or _use config file_) and this would accommodate your use case or the situations where users wish to have a different media storage folder.

XenuIsWatching:
> I don't agree with changing the default options because of a non-standard/advanced setup.

I wouldn't call this a non-standard/advanced setup. What's happened is that the default option for the easy front end creates problems with emulation station (and an annoyance for the user) when the media is place in the `rom` directory.

cmitu:
> I wouldn't call this a non-standard/advanced setup. 

Having 10k scraped ROMs on a NAS share is not a prevalent setup. Most users use an external USB disk/stick for large ROM collections and running with a ROM folder is specifically an advanced setup documented in the docs.

> What's happening is that when emulations station starts up. It begins searching it's rom folders for games. However, with the media folder in it, it will then begin checking each and every file if it is a diectory, and then if it has the extension (ex. .gba, .cue, etc...) of a game. Any read to a filesystem is an expensive process by a program, and will cause performance issues if not keep to a minimum.

Yes, I'm familiar with this kind of performance due to large number of media/roms, especially on a networked system. That's one the reasons the _Parse Gamelists_ option exists.

> What's happened is that the default option for the easy front end creates problems with emulation station (and an annoyance for the user) when the media is place in the `rom` directory.

That's not the _default_ option when scraping with Skyscraper - the default is to use the gamelists folder under `$HOME/.emulationstation`, just like the EmulationStation's internal scraper.

> Now while I'm _okay_ with adding an option to use the config file output folder. I still believe that default options are meant to give users the least annoyance as possible, and right now, the default path (with use rom folder) creates a much longer unnecessary load time.

As above, the default is not to use the ROM folders.

muldjord:
I've also decided against this change in Skyscraper. It will instead be a config option to hide it. Reasoning in the issue ticket. https://github.com/muldjord/skyscraper/issues/307

XenuIsWatching:
> I've also decided against this change in Skyscraper. It will instead be a config option to hide it. Reasoning in the issue ticket. muldjord/skyscraper#307

@muldjord I'd have to agree with your change

> That's not the default option when scraping with Skyscraper

> I don't agree with changing the default options

@cmitu I'm not sure what you're calling *default* now. Your first post calls it *default*, now you're not calling it *default*...?

I'm not quite sure I've been clearly communicating the issue clearly. This change was *never changing* any defaults (such as enabling the "use rom folder" to be enabled). This was just changing the hardcoded folder name..

Anyways, I'm just going to use the term "*hardcoded* location with rom folders" selected from now on for clarity

> Having 10k scraped ROMs on a NAS share is not a prevalent setup. Most users use an external USB disk/stick for large ROM collections and running with a network mounted ROM folder is specifically an advanced setup documented in the docs.

Although I have a NAS in my setup, those who have a USB disk/stick for large ROM collections would have the same issue (if they have use rom folders selected)...

>That's not the default option when scraping with Skyscraper - the default is to use the gamelists folder under $HOME/.emulationstation, just like the EmulationStation's internal scraper.

Unfortunately, due to how some systems in my setup don't have a gamelist.xml. This wouldn't be an option for me  :( 

What I had to do after these scripts ran was to just do a `mv media/ .media/` and a `sed -i s#/media/#/.media/#d gamelists.xml` to make emulation station happier similar to what tomaz did.

cmitu:
> @cmitu I'm not sure what you're calling _default_ now. Your first post calls it _default_, now you're not calling it _default_...?
> 
> I'm not quite sure I've been clearly communicating the issue clearly. This change was _never changing_ any defaults (such as enabling the "use rom folder" to be enabled). This was just changing the hardcoded folder name..

Let me explain better then. Right now the minimal UI used for invoking Skyscraper from RetroPie-Setup offers 2 options for  gamelist location. The options value  is stored in the configuration file (`skyscraper.cfg`) under the `use_rom_folder` key:
- (default, _use_rom_folder=0_) use the EmulationStation's gamelists folder (`$HOME/.emulationstation/gamelists`)
- (_use_rom_folder=1_) use the game ROMs folder

See also https://retropie.org.uk/docs/Scraper/#where-are-my-scraped-media-and-metadata-saved

Users will have to change the option in order to get the `gamelist.xml` generated in the ROM folder and stumble upon the performance issue.

Now, with your change, if the user had the configuration option _use_rom_folder=1_, on the next gamelist generation, all the media will be re-generated from the cache under the `.media` folder, and as @muldjord mentioned, will duplicate the existing `media` folder. 

That's why I don't agree changing the location when `use_rom_folder=1`, but otherwise adding a 3rd option (which would still be a user choice and not the default) would be fine.

XenuIsWatching:
> Now, with your change, if the user had the configuration option _use_rom_folder=1_, on the next gamelist generation, all the media will be re-generated from the cache under the `.media` folder, and as @muldjord mentioned, will duplicate the existing `media` folder.

I suppose one option would be to detect to if user has a media folder in the directory and to then prompt the user that this folder will be moved and the gamelists.xml will be updated with similar commands that I have mentioned earlier... 

> adding a 3rd option (which would still be a user choice and not the default) would be fine.

I could work on this (when I can find the time) as this simple enough, and update the pull request

cmitu:
> I suppose one option would be to detect to if user has a media folder in the directory and to then prompt the user that this folder will be moved and the gamelists.xml will be updated with similar commands that I have mentioned earlier...

That might work, but it would need the option to generate gamelist to the `.media` folders to be available (and selected) and run the gamelist translation when that option is enabled. This way the `media` folder doesn't get re-generated again on the next scraping session.

joolswills:
I believe this can be closed now?

XenuIsWatching:
Yes, I have started work on a separate branch that will make the folder name configurable... still have some more testing to do, mostly with moving the existing media folder and updating gamelists.xml
https://github.com/XenuIsWatching/RetroPie-Setup/tree/skyscraper_media_folder

I  will close this and submit a separate pull request for this when ready.

PR #3344, opened 2021-05-25
Updated deprecated '--unattend' option to use the '--flags unattend' o…

muldjord:
The `--unattend` option has been deprecated for a year and was recently removed from Skyscraper (it existed, but was hidden). This updated the script to use the correct `--flags unattend` option instead. Tested on Ubuntu 21.04, works as expected.


joolswills:
Thanks.

PR #3345, opened 2021-05-27
ppsspp: add a workaround for assets folder detection issues

cmitu:
Reported initially in https://retropie.org.uk/forum/topic/30620/, it looks like recent versions don't detect correctly the binary folder at runtime and they're unable to find the `assets` folder correctly.
Running from the emulator's folder fixes it, since `cwd` is added to the search path by the emulator.
Needs further investigation to find the cause and - possibly - a proper fix.


joolswills:
Thanks.

PR #3346, opened 2021-06-03
Added experimental PSX emulator lr-duckstation.

dankcushions:
Tested on Pi 4, arm32, gles 3.1. Should be compatible with aarch64 and x64, gles2 (ie pi2/3 - not sure about performance) and vulkan.

Libretro core is not released as source - binary is hosted by dev, under NC-type license. see https://www.duckstation.org/libretro/

### DIFF
--- a/scriptmodules/libretrocores/lr-duckstation.sh
+++ b/scriptmodules/libretrocores/lr-duckstation.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-duckstation"
+rp_module_desc="PlayStation emulator - Duckstation for libretro"
+rp_module_help="ROM Extensions: .exe .cue .bin .chd .psf .m3u .pbp\n\nCopy your PlayStation roms to $romdir/psx\n\nCopy compatble BIOS files to $biosdir"
+rp_module_licence="PROP"
+rp_module_section="exp"
+rp_module_flags="!all arm !armv6 aarch64 64bit"
+
+function __binary_url_lr-duckstation() {
+    local url="https://www.duckstation.org/libretro/duckstation_libretro_linux_"
+    isPlatform "aarch64" && echo "${url}aarch64.zip"
+    isPlatform "arm" && echo "${url}armv7.zip"
+    isPlatform "x86" && isPlatform "64bit" && echo "${url}x64.zip"
+}
+
+function install_bin_lr-duckstation() {
+    downloadAndExtract "$(__binary_url_lr-duckstation)" "$md_inst"
+}
+
+function configure_lr-duckstation() {
+    mkRomDir "psx"
+    ensureSystemretroconfig "psx"
+
+    # dynarec segfaults without redirecting stdin from </dev/null
+    addEmulator 0 "$md_id" "psx" "$md_inst/duckstation_libretro.so </dev/null"
+    addSystem "psx"
+}


scriptmodules/libretrocores/lr-duckstation.sh
```diff
@@ -0,0 +1,36 @@
...
+rp_module_desc="PlayStation emulator - Duckstation for libretro"
+rp_module_help="ROM Extensions: .exe .cue .bin .chd .psf .m3u .pbp\n\nCopy your PlayStation roms to $romdir/psx\n\nCopy compatble BIOS files to $biosdir"
+rp_module_licence="PROP"
+rp_module_section="exp"
+rp_module_flags="!all !armv6 aarch64 arm 64bit"
+
+function __binary_url_lr-duckstation() {
+    isPlatform "aarch64" && echo "https://www.duckstation.org/libretro/duckstation_libretro_linux_aarch64.zip"
+    isPlatform "arm" && echo "https://www.duckstation.org/libretro/duckstation_libretro_linux_armv7.zip"
+    isPlatform "x86" && isPlatform "64bit" && echo "https://www.duckstation.org/libretro/duckstation_libretro_linux_x64.zip"
```
joolswills, line 22: I'd prefer this without the url duplication.

scriptmodules/libretrocores/lr-duckstation.sh
```diff
@@ -0,0 +1,36 @@
...
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-duckstation"
+rp_module_desc="PlayStation emulator - Duckstation for libretro"
+rp_module_help="ROM Extensions: .exe .cue .bin .chd .psf .m3u .pbp\n\nCopy your PlayStation roms to $romdir/psx\n\nCopy compatble BIOS files to $biosdir"
+rp_module_licence="PROP"
+rp_module_section="exp"
+rp_module_flags="!all !armv6 aarch64 arm 64bit"
```
joolswills, line 17: The arm flag will override the !armv6 for armv6.

scriptmodules/libretrocores/lr-duckstation.sh
dankcushions: agree! how about now?

scriptmodules/libretrocores/lr-duckstation.sh
dankcushions: yup :( how about now? if >armv8 becomes supported I guess we would have to change the script but i can't think of a better way of excluding <armv7

scriptmodules/libretrocores/lr-duckstation.sh
```diff
@@ -0,0 +1,37 @@
...
+rp_module_help="ROM Extensions: .exe .cue .bin .chd .psf .m3u .pbp\n\nCopy your PlayStation roms to $romdir/psx\n\nCopy compatble BIOS files to $biosdir"
+rp_module_licence="PROP"
+rp_module_section="exp"
+rp_module_flags="!all armv7 armv8 aarch64 64bit"
+
+function __binary_url_lr-duckstation() {
+    local url="https://www.duckstation.org/libretro/duckstation_libretro_linux_"
+    isPlatform "aarch64" && echo "${url}aarch64.zip"
+    isPlatform "arm" && echo "${url}armv7.zip"
+    isPlatform "x86" && isPlatform "64bit" && echo "{url}x64.zip"
```
joolswills, line 23: missing `$`

scriptmodules/libretrocores/lr-duckstation.sh
joolswills: Might have been simpler to just put the `!armv6` at the end.

Also could use "neon" as only armv7+ has it to avoid having armv7 / armv8

scriptmodules/libretrocores/lr-duckstation.sh
dankcushions: 🤦

scriptmodules/libretrocores/lr-duckstation.sh
dankcushions: oh I didn't realise the conditions were applied incrementally 👍  i guess `!all arm !armv6 aarch64 64bit` makes sense then

joolswills: Should be fine yeah.

dankcushions: done!

joolswills:
Thanks. Please can you squash it and I'll merge.

joolswills:
Thanks

theofficialgman:
Isn't a change of licensing against the GPLv3 liscense of duckstation itself? meaning that this new lr-duckstation fork is not allowed to change to anything else and must release the source code

dankcushions:
i am not a lawyer. how libretro and duckstation upstream license their code is up to them. this is not an appropriate place to debate it.

PR #3348, opened 2021-06-05
dreamcast: Added .m3u support for lr-flycast

dankcushions:
See https://github.com/libretro/flycast/blob/master/core/libretro/libretro.cpp#L2257


joolswills:
Thanks.

PR #3350, opened 2021-06-06
runcommand.sh: updated doc for %XRES%, %YRES%

Gemba:
As discussed here initially: https://retropie.org.uk/forum/topic/30662/intellivision-jzintv-for-sdl2-proposed-scriptmodule/3

Added some escape sequences for doxygen to keep `%` char in HTML output.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -8,6 +8,8 @@
 # See the LICENSE.md file at the top-level directory of this distribution and
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
 #
+# Editor's note: You may notice "\c \%" escape sequences, these are needed
+# to avoid the initial % char to be swallowed by doxygen.
 
 ## @file supplementary/runcommand/runcommand.sh
 ## @brief runcommand launching script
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -2,12 +2,15 @@
 
 # This file is part of The RetroPie Project
 #
-# The RetroPie Project is the legal property of its developers, whose names are
-# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+# The RetroPie Project is the legal property of its developers, whose names
+# are too numerous to list here. Please refer to the COPYRIGHT.md file
+# distributed with this source.
```
joolswills, line 7: This may be best left so it's the same as other sources

joolswills:
Looks ok to me. It's not ideal with the quoting but I had missed that `%` was being dropped. Maybe we can work around this via the bash doxygen code we use so we don't need to quote it. I'll see if there's any alternative solutions but appreciate the work on this.

Gemba:
Whatever will be done to remediate the % situation, at least the documentation is there. :)

joolswills:
Think the quoting is the best way for now so merging. Thanks.

joolswills:
I rebased / squashed the commits and adjusted commit msg slightly. Thanks for the contribution.

PR #3351, opened 2021-06-06
lr-scummvm.sh: Recognize *.svm files in libretro.cpp

Gemba:
Making it full compliant to the documentation:
https://retropie.org.uk/docs/ScummVM/#example

Note: It was working before but only used the `%ROM%` file contents (=game short title/id) iff the extension was `.scummvm` in other cases it fell back to auto detection within that game subfolder.


joolswills:
I'd like to hear the thoughts from @hhromic as he included this module originally and I think worked on the docs too.

PR #3354, opened 2021-06-09
retroarch: fix the patch enabling `video_shader`

cmitu:
Looks like the v1.9.1 original patch didn't apply cleanly to v1.9.4 and the patch was incomplete.


joolswills:
Thanks.

PR #3355, opened 2021-06-11
Update splashscreen.sh

DanielLester83:
Removed dependencies in service for earlier launch. needs testing on the supported platforms.

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -42,8 +42,9 @@ function install_bin_splashscreen() {
 [Unit]
 Description=Show custom splashscreen
 DefaultDependencies=no
-After=console-setup.service
-Wants=console-setup.service
+#Disabled as unneeded (needs testing)
+#After=console-setup.service
+#Wants=console-setup.service
 ConditionPathExists=$md_inst/asplashscreen.sh
 
 [Service]


DanielLester83:
It works on my "zero w" in a "retroflag gpi case" which uses an overlay in the config.txt, that is all I can test so I have no idea if other setups have problems.

cmitu:
How much time is gained by removing the dependency ? Have you tested with image/video (any/both) ?

DanielLester83:
It saves roughly about 4 seconds with an image and the video I tried takes so long to load anyways that I couldn't tell a difference.

cmitu:
Removing the dependency does indeed remove about 4 sec during start-up (tested on Pi3/Pi4). There are a few drawbacks to this change:

* it will break the module on Pi4 (see #2852), since the DRM/KMS modesetting will take place after the splashscreen starts and it will blank the screen when it does.
* if the user has splashscreens hosted on an USB or external drive mounted, it will run before the `local-fs.target` and the splashscreen might not have the right files available. This is a minor thing - since most users will use a local sdcard - but might break existing configurations.

The change will not speed up the splashscreen when using videos, since `omxplayer` needs the `dbus` service and will wait for it to start (even if `splashscreen.service` is started earlier).

All in all, I'm not keen on merging this change. While it provides some speed up for Pi0/Pi3 users when using images only for the splashscreen, it will break on Pi4 and could affect other users.

joolswills:
Yeah. I'm for leaving this as is.

DanielLester83:
> Removing the dependency does indeed remove about 4 sec during start-up (tested on Pi3/Pi4). There are a few drawbacks to this change:
> 
> * it will break the module on Pi4 (see #2852), since the DRM/KMS modesetting will take place after the splashscreen starts and it will blank the screen when it does.
> * if the user has splashscreens hosted on an USB or external drive mounted, it will run before the `local-fs.target` and the splashscreen might not have the right files available. This is a minor thing - since most users will use a local sdcard - but might break existing configurations.
> 
> The change will not speed up the splashscreen when using videos, since `omxplayer` needs the `dbus` service and will wait for it to start (even if `splashscreen.service` is started earlier).
> 
> All in all, I'm not keen on merging this change. While it provides some speed up for Pi0/Pi3 users when using images only for the splashscreen, it will break on Pi4 and could affect other users.

Thanks for the feedback.

Maybe someone else could automate the different files based on these conditions or automate the install of plymouth which seems way tooo hard.

joolswills:
I'm closing this for now. I don't think it's worth tweaking as we can't know if users have splash screens on another disk or not. Of course you can disable the module and use a custom script as you like.

Regarding Plymouth, I don't see any advantage to using it. On Raspberry Pi OS, there is no Initrd, so not much benefit. 

If someone submits a PR with a chance so a user can choose for an earlier splash start, I may accept it. So long as it's not default and easy to switch back etc.

PR #3356, opened 2021-06-12
joy2key: new version using PySDL2

cmitu:
Added a new `joy2key` implementation, using PySDL2 for joystick event handling.
PySDL2 is a python module that wraps SDL2 (and other SDL libraries) using the built-in `ctypes` module.

Pros:
 * event handling is simplified a bit, using SDL's event loop.
 * (subjective) the code is a bit more structured and easy to follow.
 * joystick handling is rewritten based on EmulationStation code, movement is smoother and scrolling is improved.
 * support for input repeat to improve scrolling, just keep the input pressed and scrolling continues

Cons:
 * module is a bit larger (296 LOC vs 224 in current joy2key.py)
 * needs PySDL2 (which might not be packaged, see the notes below) and SDL2
 * arguably, device conf. to keyboard event mapping is more complex (abstracted as InputDev)

Notes:
 * availability of PySDL2 as a system package is good, but it's not standard in Debian 10 (stable) at the moment.
   The module is present though (as a backport) in Ubuntu 18.04 (and later) and Raspberry Pi OS (Buster).
   If not found at runtime, then the current `joy2key.py` is used instead and for this the `joy2keyStart` helper was modified.

 ~~* in the current version of `joy2keyStart`, `joy2key` is started from `$scriptdir`, the changes will make it run from `runcommand`'s installation folder. Not sure if this was an oversight or on purpose.~~

 * added PgUp/PgDown to the default parameters for `joy2key`, they are mapped in a similar fashion to EmulationStation to the shoulder buttons.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1213,7 +1213,7 @@ _EOF_
 ## @param but2 mapping for button 2
 ## @param but3 mapping for button 3
 ## @param butX mapping for button X ...
-## @brief Start joy2key.py process in background to map joystick presses to keyboard
+## @brief Start joy2key process in background to map joystick presses to keyboard
 ## @details Arguments are curses capability names or hex values starting with '0x'
 ## see: http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html
 function joy2keyStart() {
@@ -1223,18 +1223,30 @@ function joy2keyStart() {
 
     local params=("$@")
     if [[ "${#params[@]}" -eq 0 ]]; then
-        params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20 0x1b)
+        # Default button-to-keyboard mappings:
+        # * cursor keys for axis/dpad
+        # * enter, space and esc for buttons 'a', 'b' and 'x'
+        # * page up/page down for buttons 5,6 (shoulder buttons)
+        params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20 0x1b 0x00 kpp knp)
+    fi
+
+    # Choose the joy2key implementation here, since `runcommand` may not be installed
+    local joy2key="joy2key.py"
+    if hasPackage "python3-sdl2"; then
+        iniConfig " =" '"' "$configdir/all/runcommand.cfg"
+        iniGet "joy2key_version"
+        [[ $ini_value != "0" ]] && joy2key="joy2key_sdl.py"
     fi
 
     # get the first joystick device (if not already set)
     [[ -c "$__joy2key_dev" ]] || __joy2key_dev="/dev/input/jsX"
 
     # if no joystick device, or joy2key is already running exit
-    [[ -z "$__joy2key_dev" ]] || pgrep -f joy2key.py >/dev/null && return 1
+    [[ -z "$__joy2key_dev" ]] || pgrep -f "$joy2key" >/dev/null && return 1
 
-    # if joy2key.py is installed run it with cursor keys for axis/dpad, and enter + space for buttons 0 and 1
-    if "$scriptdir/scriptmodules/supplementary/runcommand/joy2key.py" "$__joy2key_dev" "${params[@]}" 2>/dev/null; then
-        __joy2key_pid=$(pgrep -f joy2key.py)
+    # if joy2key is installed, run it
+    if "$scriptdir/scriptmodules/supplementary/runcommand/$joy2key" "$__joy2key_dev" "${params[@]}" 2>/dev/null; then
+        __joy2key_pid=$(pgrep -f "$joy2key")
         return 0
     fi
 
--- a/scriptmodules/supplementary/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand.sh
@@ -28,11 +28,11 @@ function depends_runcommand() {
 }
 
 function install_bin_runcommand() {
-    cp "$md_data/runcommand.sh" "$md_inst/"
-    cp "$md_data/joy2key.py" "$md_inst/"
-    chmod a+x "$md_inst/runcommand.sh"
-    chmod a+x "$md_inst/joy2key.py"
-    python3 -m compileall "$md_inst/joy2key.py"
+    for file in "runcommand.sh" "joy2key.py" "joy2key_sdl.py"; do
+        cp "$md_data/$file" "$md_inst/"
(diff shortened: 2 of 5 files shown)


joolswills:
Thanks very much for this. Will test. Much appreciated!

joolswills:
It's started from scriptdir from RetroPie-Setup as it needs to work when nothing is installed. So run command should use the installed version, but RetroPie-Setup should use the local copy. But the installed version should always be in sync as it's installed after every update.

scriptmodules/helpers.sh
```diff
@@ -1223,18 +1223,24 @@ function joy2keyStart() {
...
     if [[ "${#params[@]}" -eq 0 ]]; then
-        params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20 0x1b)
+        # Default button-to-keyboard mappings:
+        # * cursor keys for axis/dpad
+        # * enter, space and esc for buttons 'a', 'b' and 'x'
+        # * page up/page down for buttons 5,6 (shoulder buttons)
+        params=(kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20 0x1b 0x00 kpp knp)
     fi
 
+    local joy2key="$(_joy2key_runcommand)"
```
joolswills, line 1233: Although joy2key is included as part of runcommand, and maybe this should be changed
A core helper function shouldn't rely on the runcommand module.
So this function would be better integrated into the core imho

joolswills:
Might be more logical if the code was handled like inifuncs.sh and archivefuncs.sh and shared between runcommand and RetroPie (but different copies). But we can do that later if needed.

But it's useful to be able to run it on a start of RetroPie-Setup even if runcommand isn't installed.

joolswills:
Please disregard the function location moving. Only need it to be standalone so it works without install. I will rework it later as I want to move some of our existing libraries/scripts we install such as inifuncs.sh and archivefuncs.sh and it would be logical to reorganize this at the same time.

cmitu:
OK, I added a few changes:
* modified `joy2keyStart` in `helpers.sh` so that it works without relying on the `runcommand` being installed.
* added the option to switch between the joy2key implementations (udev/sdl). I know this would ideally be part of `retropie.cfg`, but currently there's no UI/module for the options there. I've added it to `runcommand`'s config, since the option should also be available in the main `runcommand` script.
* modified the main `runcommand.sh` to use the new version of `joy2key`, since it doesn't rely on the implementation in `helpers.sh` (might be moved later to a helper `lib.sh`, similar to `inifuncs.sh`).
* in the new `.py` script, I added the handling of A/B being switched (via `retroarch.cfg` option of `menu_swap_ok_cancel_buttons`).

joolswills:
Thanks for this. Will give it a test.

joolswills:
The SDL python code fails on Raspberry Pi OS buster due to the python library lacking SDL_HINT_JOYSTICK_HIDAPI and SDL_HINT_NO_SIGNAL_HANDLERS

commenting the related code out makes it work. I guess it shouldn't be a big workaround if this is missing in the older module.

cmitu:
> The SDL python code fails on Raspberry Pi OS buster due to the python library lacking SDL_HINT_JOYSTICK_HIDAPI and SDL_HINT_NO_SIGNAL_HANDLERS
> 

Those can be guarded with a SDL version check - I assume this is with the default `libSDL2` from the distro, before the RetroPie's SDL2 version is added.

joolswills:
This was with our RetroPie SDL 2.0.10.

I like the changes. Works well so far. I think I'd prefer a longer delay when holding before repeating and faster repeat speed if that's doable. I've not really looked at the code yet. Basically so it better matches how keyboard input works.

I will have a look at the code when I have more time but thanks again for all your work on this. It's definitely a big improvement.

cmitu:
> I like the changes. Works well so far. I think I'd prefer a longer delay when holding before repeating and faster repeat speed if that's doable. I've not really looked at the code yet. Basically so it better matches how keyboard input works.

This can be tweaked at the top of the script by modifying `JS_POLL_DELAY_DEFAULT` and `JS_POLL_DELAY_DEBOUNCE` . The latter is used between input repeats.

joolswills:
Thanks. I'll have a play.

cmitu:
> The SDL python code fails on Raspberry Pi OS buster due to the python library lacking SDL_HINT_JOYSTICK_HIDAPI and SDL_HINT_NO_SIGNAL_HANDLERS

This should be fixed now. I've opted for using env variables, since the SDL library might support the hint, but the wrapper might not:
* `SDL_HINT_JOYSTICK_HIDAPI` was added in SDL 2.0.9 and supported by PySDL2 0.9.7
* `SDL_HINT_NO_SIGNAL_HANDLERS` was added in SDL 2.0.4 and supported by PySDL2 0.9.5

joolswills:
The poll_delay setting to use JS_POLL_DELAY_DEBOUNCE is missing for SDL_JOYAXISMOTION - I added it to the PR but please rebase or adjust as needed.

I think I prefer DEBOUNCE set to 250 btw.

joolswills:
Sorry I rebased my change as had the wrong wording in comment but then included the 250 increase change by accident. I'll leave it as it will be squashed but adjust the code as needed

joolswills:
I wonder if the `if mapped_events is not None:` blocks could be de-deplicated maybe also. Minor things but it's a large loop with lots of nesting so maybe could be split out more. The old joy2key was also guilty.

joolswills:
Also could avoid the `and event.jdevice.which in active_devices` checks for each event type by putting them all under an `elif event.jdevice.which in active_devices:` in the loop etc.

cmitu:
I added a couple of modifications, based on the comments and testing performed
 - simplified event handling with a helper method
 - moved the event repeat handling in a separate filtering method based on the event's last fire time and the # of times it was repeated. This allows the event loop to have a constant and higher poll cycle - instead of using `SDL_Delay` globally - and detect 'immediately' when an input is added or removed (i.e. fast button press which would be masked by a larger `SDL_Delay` )

joolswills:
Looks good to me. I'm going to merge if you're happy? Thanks for this.

cmitu:
I will do a test with a new install just to check the configuration is ok.

cmitu:
Looks ok on a new installation (tested on a Pi4 RPI OS Lite).
I squashed the branch commits, should be ready to merge.

joolswills:
Awesome work. Thanks.

PR #3358, opened 2021-06-13
redream: not available on aarch64 yet

cmitu:
Fix `redream` being installable on `aarch64`, now that #3357 fixed the issue of packages with custom binary url set.


dankcushions:
FYI this issue tracks aarch64 support for pi: https://gitlab.com/inolen/redream/-/issues/953

From comments on discord, it's in the works :+1:

PR #3361, opened 2021-06-20
add * to catch icosa_emmc and icosa both

theofficialgman:
Nintendo Switch users who launch from emmc instead of SD storage get the model name icosa_emmc

I wasn't aware of this myself until and switch user notified me of this

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -397,7 +397,7 @@ function get_platform() {
                         "NVIDIA Jetson Nano Developer Kit")
                             __platform="jetson-nano"
                             ;;
-                        icosa)
+                        icosa*)
                             __platform="tegra-x1"
                             ;;
                     esac


joolswills:
If there's no other different strings it might be worth using `icosa|icosa_emmc` as it's clearer then the IDs it matches.

theofficialgman:
those are the only two (icosa and icosa_emmc). We have an emmc dtb overlay that changes the model name among other things on switch.

Feel free to do either what is in this PR or just commit it yourself (I'll leave the PR as is). I'll close this if you decide to do icosa | icosa_emmc

joolswills:
Thanks. I added it with a thanks also. Btw I just merged a PR with some reorganisation of the platforms config. Basically split out cpu configuration but please feel free to test it and make sure it's ok on this system.

theofficialgman:
yup all good. Just tested the latest commit at the time of posting. the new platforms config seems to be working as expected.
Thank you

joolswills:
Thanks for your help.

PR #3362, opened 2021-06-20
Use Libretro's RetroArch joypad autoconfig presets to generate button labels

dankcushions:
Whilst using Libretro's [RetroArch joypad autoconfigs](https://github.com/libretro/retroarch-joypad-autoconfig/) has proven problematic, we can still leverage this database in some interesting ways. This PR uses it to apply meaningful button labels across the Retroarch menus, rather than the retropad abstract/button numbers.

Eg, for a DualShock 4:
IMG_5905

Only tested with UDEV devices - one with labels in the matched preset in the repo (pictured), one with none (but will have after https://github.com/libretro/retroarch-joypad-autoconfig/pull/767). When there's no labels, they appear as previously:
IMG_5906

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
@@ -239,6 +239,9 @@ function map_retroarch_joystick() {
     local value
     local type
     declare -A hat_map=([1]="up" [2]="right" [4]="down" [8]="left")
+    iniGet "input_driver"
+    local input_driver="$ini_value"
+    local autoconfig_preset=$(grep -rwl "$rootdir/emulators/retroarch/autoconfig-presets/$input_driver" -e "$DEVICE_NAME" | head -1)
     for key in "${keys[@]}"; do
         case "$input_type" in
             hat)
@@ -260,8 +263,7 @@ function map_retroarch_joystick() {
                 value="$input_id"
 
                 # workaround for mismatched controller mappings
-                iniGet "input_driver"
-                if [[ "$ini_value" == "udev" ]]; then
+                if [[ "$input_driver" == "udev" ]]; then
                     case "$DEVICE_NAME" in
                         "8Bitdo FC30"*|"8Bitdo NES30"*|"8Bitdo SFC30"*|"8Bitdo SNES30"*|"8Bitdo Zero"*)
                             if [[ "$_atebitdo_hack" -eq 1 ]]; then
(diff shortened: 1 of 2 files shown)


scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
```diff
@@ -238,6 +238,9 @@ function map_retroarch_joystick() {
     local key
     local value
     local type
+    iniGet "input_driver"
+    local input_driver="$ini_value"
+    local autoconfig_preset=$(grep -rwl "/opt/retropie/configs/all/retroarch/autoconfig-presets/$input_driver" -e "$DEVICE_NAME" | head -1)
```
joolswills, line 243: Can use `$configdir` here?

joolswills:
Apologies for the review as I missed that it's a WIP. Looks good though

scriptmodules/supplementary/emulationstation/configscripts/retroarch.sh
dankcushions: good spot, done!

dankcushions:
@joolswills no problem! if you're happy with it in principle then I am happy putting it for review 👍

joolswills:
Sorry for the delay on this. Will test

joolswills (changes requested):
Thanks. Works well from my testing. I would however like to move the autoconfig file location. As we use it as a read only reference, it could be checked out to the build folder and the configs copied to a subfolder of the retroarch `$md_inst`

dankcushions:
@joolswills thanks! was this what you had in mind?

joolswills:
This branch has conflicts btw so best to fix them before I review. Thanks

dankcushions:
Done a rebase! I believe it's good to review now.

joolswills:
Thanks. Looks good but I wonder if while moving it we should just do this at the source stage and copy it in install so it's part of the prebuilt bins. As `$md_inst` is cleaned on update it might make more sense than doing a new checkout over each time, although this would mean the version would be matched to when the binaries were built. It depends how important it is to have the latest version. As it's just for labels this may be enough.

dankcushions:
they probably don't change that much, but I suppose the same argument could be made for assets, overlays, shaders, etc. maybe the whole lot should be moved?

joolswills:
We install minimal assets so that's a bit different. Since we already handle the other stuff differently it's not a good idea to change that. It's also outside of this PR.

joolswills:
It's a small repo so I guess it's not a big issue to check it out at the configure stage so leave it. I'll merge this when I'm next working on stuff. Thanks

joolswills (approved):

joolswills:
Thanks.

PR #3365, opened 2021-06-25
Add the missing Nvidia Jetsons

theofficialgman:
I realize that the user could just force a model but this allows for easy use on the rest of the nvidia jetsons>

Some of the jetsons (the TX2 notibly) can't be determined by the model identifier so instead we use the dts filename
This idea (and most of the code) was pulled from https://github.com/rbonghi/jetson_stats/blob/master/jtop/jetson_variables

This adds the jetson nano 2GB (previously missing), the TX2, and jetson xavier systems.
I decided to do icosa* this time again just for the future incase we decide to specify something else


theofficialgman:
sorry about the force pushes.. I realized I have my username/email set wrong on my local git... can't figure out why the they still aren't matching up

theofficialgman:
alright there we go.. figured it out, its still one commit and there is no history of  the mistakes so its still PR worthy

scriptmodules/system.sh
```diff
@@ -388,18 +388,17 @@ function get_platform() {
...
-                        "NVIDIA Jetson Nano Developer Kit")
-                            __platform="jetson-nano"
-                            ;;
-                        icosa|icosa_emmc)
-                            __platform="tegra-x1"
-                            ;;
+                # jetsons can be identified by their dtsfilename
+                # see https://github.com/rbonghi/jetson_stats/blob/master/jtop/jetson_variables for similar implementation
+                if [ -f /proc/device-tree/nvidia,dtsfilename ]; then
+                    local JETSON_CODENAME=$(tr -d '\0' < /proc/device-tree/nvidia,dtsfilename)
```
joolswills, line 394: Please can you use lowercase variable names

scriptmodules/system.sh
```diff
@@ -388,18 +388,17 @@ function get_platform() {
...
+                    JETSON_CODENAME=$(echo ${JETSON_CODENAME#*"/hardware/nvidia/platform/"} | tr '/' '\n' | head -2 | tail -1 )
+                    case $JETSON_CODENAME in
```
joolswills, line 396: Please quote the variable.

scriptmodules/system.sh
```diff
@@ -516,13 +515,27 @@ function platform_odroid-xu() {
     __platform_flags+=(mali gles)
 }
 
-function platform_tegra-x1() {
+function platform_TX1() {
```
joolswills, line 518: Please keep the function names lowercase here.

joolswills (changes requested):

scriptmodules/system.sh
theofficialgman: I'm going to make all of these as resolved since I've updated the files

scriptmodules/system.sh
```diff
@@ -388,18 +388,17 @@ function get_platform() {
...
-                    case "$model" in
-                        "NVIDIA Jetson Nano Developer Kit")
-                            __platform="jetson-nano"
-                            ;;
-                        icosa|icosa_emmc)
-                            __platform="tegra-x1"
-                            ;;
+                # jetsons can be identified by their dtsfilename
+                # see https://github.com/rbonghi/jetson_stats/blob/master/jtop/jetson_variables for similar implementation
+                if [ -f /proc/device-tree/nvidia,dtsfilename ]; then
```
joolswills, line 393: Please use double `[[`

theofficialgman:
I've decided to change entirely how the system type is detected to match how nvidia does so in nv.sh
it should be immune to dtb name changes
will update soon

theofficialgman:
ok, so now we only check for the SOCFAMILY, instead of the actual devices friendly name or dtb filename.
this should be all the identification needed for retropie
changing the SOC name is not something that can be done easily in dts and will remain constant across all devices and boards that use that SOC

there are two methods used to get potential SOCFAMILY used, one uses the dts 4.9.X layout and the other uses 3.X dts layout that the earlier tegras used (tegra210 could be found with either version of the kernel and is checked both ways because of this)

theofficialgman:
oh my thats not what I wanted to do... lol

PR #3366, opened 2021-06-27
Runcommand menu rework

joolswills:
Switch to letters for items rather than numerical values to avoid gaps and to make it easier to add items to the
menu without having to renumber items etc, which has happened previously (hence most menus excluding
generated lists don't use number ids in RetroPie).

Remove the "Remove emulator/video mode choice" items and replace with an option in the sub-menu for clearing / removing.
Previously when choosing a default emulator for a rom, (or a default video mode for emulator, emulator + rom etc),
an additional menu item was added below to remove the setting. Although this enabled quick access to remove options
from the main menu, it wasn't that intuitive and doesn't match the way menus work elsewhere in RetroPie.

Now when choosing a default emulator / video mode, there is an option to clear it at the top of the list.

The framebuffer menu now uses the resolution as a key rather than a numbered list (This also fixes the broken
default-item not working). The video and framebuffer dialogs were merged into one function with some additional
descriptions and menu title information also.

While making these changes a couple of bugs were fixed with default-item functionality - Framebuffer resolution
default-item support was broken as mentioned, but also the main menu lacked this, so the position wouldn't be
remebered after entering a video menu selection etc.

PR also includes a cosmetic commit for trailing whitespace

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -296,7 +296,7 @@ function get_all_x11_modes()
     while read -r id; do
         # populate CONNECTOR:0xID into an array
         MODE_ID+=($id) # output:id as in (hdmi:0x44)
-        
+
         read -r line
         # array is x/y resolution @ vertical refresh rate ( details )
         MODE[$id]="$line"
@@ -305,36 +305,36 @@ function get_all_x11_modes()
         # true is the "header" (output and id)
         # false is the "description" (Mode: and everything that begins with a space)
         { type = /^[^ \t]+/ }
-        
+
         # Exit after the first output
         type && output {exit} # New header and output set means new output
-        
+
         # many outputs can be connected, but only the ones with the id are in use.
         # output must be connected and have an (id)
         type && / connected/ && /\(0x[0-9a-f]+\)/ {
             output=$1; next
         }
-        
+
         # parse mode and lines
         # If we are in a "description", and output is set (output being what we want)
         # And if $2 is an id, we are in a video mode description line
         !type && output && $2 ~ /^\(0x[0-9a-f]+\)$/ {
             # Print CRTC identifier (CONNECTOR:0xID)
             print output ":" substr($2,2,length($2)-2) # id
-            
+
             # get rid of what we printed
             $1="";$2=""
             sub(/^[ \t]+/,"") # trim spaces
-            
+
             # save rest of the line
             info=$0
 
             # Save width from the 2nd line of the video mode
             getline; width=$3
-            
+
             # Save height & vrefresh from the 3rd line of video mode
             getline; height=$3; vrefresh=$NF
-            
+
             # Print video mode details
             print width "x" height " @ " vrefresh " (" info ")"
         }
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -778,6 +784,13 @@ function choose_emulator() {
     local default="$2"
     local cancel="$3"
 
+    local mode_desc="default emulator for "
+    if [[ "$mode" == "emu_sys" ]]; then
+        mode_desc+="$SYSTEM"
+    else
+        mode_desc+="ROM ($ROM)"
+    fi
+
```
cmitu, line 793: Might want to remove `default` re-declaration and move `default_id` at the top of the function.

scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -778,6 +784,13 @@ function choose_emulator() {
     local default="$2"
     local cancel="$3"
 
+    local mode_desc="default emulator for "
```
cmitu, line 787: Maybe use a more suggestive var name instead of `mode_desc` (copy & paste from `choose_mode` ?)

cmitu (commented):
Looks and works fine. Left a couple of small notes.

One thing that you may consider adding is replacing the empty `()` - when there's no choice for video/emulator (+ROM) - with `not set` or 'none', it would be similar to the title of the selection screen (i.e. _Currently: (not set)_ when no choice exists).

scriptmodules/supplementary/runcommand/runcommand.sh
joolswills: Yeah must be. I'll sort. Thanks.

scriptmodules/supplementary/runcommand/runcommand.sh
joolswills: Will check this. It's not really a redeclaration in bash but will clean up if needed.

scriptmodules/supplementary/runcommand/runcommand.sh
joolswills: Actually this isn't a copy and paste error. The mode variable in both functions is unrelated to video but how the function behaves depending on the first parameter. So when I added a description var I named it appropriately.

It could be argued in the choose_mode function this variable name is confusing but that can be another PR if needed but I'm not too fussed.

joolswills: And more importantly to note the mode variable predates this PR if

> Looks and works fine. Left a couple of small notes.
> 
> One thing that you may consider adding is replacing the empty `()` - when there's no choice for video/emulator (+ROM) - with `not set` or 'none', it would be similar to the title of the selection screen (i.e. _Currently: (not set)_ when no choice exists).

I will do this. Hopefully there is no emulator ever called "not set" :-)

joolswills: Actually I've changed my mind. At least for now as this adds a heap of extra variables and logic for the menu as we have to handle

```
$emu_sys
$emu_rom
$vid_emu
$vid_rom
$fb_emu
$fb_rom
```

My main motivation with this PR was the variable menu options. I don't disagree this may be better but I think that may be better off in a future change.

scriptmodules/supplementary/runcommand/runcommand.sh
joolswills: Ah thought it was something in the PR. Yeah I'll remove the second declaration. Not moving the default_id declaration as it's not c89 ;-)

scriptmodules/supplementary/runcommand/runcommand.sh
cmitu: >  I don't disagree this may be better but I think that may be better off in a future change.

It's mostly cosmetic, so yes, it may be handled better in a future change.

joolswills:
I changed the submenus for emulator and ROM settings to use the ROM basename as with the main menu for consistency and readability.

Will merge.

PR #3369, opened 2021-06-30
joy2key: simplify debug message; fix python warning

cmitu:
`python3.8(+)` throws a warning for using 'is/is not' with with certain literal types [1].
Simplifying the debug message that shows the `python3-sdl2` wrapper version gets rid of this warning, the end result is the same.

[1] https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior

Reported in https://retropie.org.uk/forum/topic/30876.


joolswills:
Thanks.

mth75:
Is there a relation with: https://retropie.org.uk/forum/topic/30882/warning-parsing-error-for-controller ?

cmitu:
@mth75 not with this PR, but it's probably related to the recent changes in `joy2key`. Please add the info asked in the forum post and let's continue there.

PR #3370, opened 2021-07-02
joy2key: guard against malformed hat values

cmitu:
Added 2 fixes:

* reject malformed hat values for an input in the joystick auto-config file(s).
  If it's not one of `h<X>up`, `h<X>right`, `h<X>down`, `h<X>left`, reject the input value.

* fix `SDL_GetError` not being properly shown because of the wrong f-string.

1st fix should fix the error reported by @mth75 in https://retropie.org.uk/forum/topic/30882/.


joolswills:
Thanks.

PR #3372, opened 2021-07-07
Dosbox Staging - switch to tag from now on instead of master/main

nemo93:
Starting with 0.77.


scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -13,7 +13,7 @@ rp_module_id="dosbox-staging"
 rp_module_desc="modern DOS/x86 emulator focusing on ease of use"
 rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/dosbox-staging/dosbox-staging/master/COPYING"
-rp_module_repo="git https://github.com/dosbox-staging/dosbox-staging.git master"
+rp_module_repo="git https://github.com/dosbox-staging/dosbox-staging v0.77.0"
```
joolswills, line 16: Better without the removal of the .git extension (which is the advertised clone url by github - https://github.com/dosbox-staging/dosbox-staging.git)

joolswills (changes requested):

PR #3374, opened 2021-07-14
wolf4sdl: fix game launching

cmitu:
Looks like the new version doesn't have a default `DATADIR` to search for files.
Modified the launcher to start from the `ports/wolf3d` port folder so data files are found.


joolswills:
Sorry for the delay. Will merge thanks.

joolswills:
Binaries should be updated now also.

PR #3375, opened 2021-07-18
Add the missing Nvidia Jetsons (update)

theofficialgman:
sorry about the mess I made here.
https://github.com/RetroPie/RetroPie-Setup/pull/3365

As I said in one of my messages there, the previous implementation is not stable as model can be configured quite easily in the dts (and it often done so on custom carrier board).
Checking the dts filename is also not stable as this will change whenever a new board layout is made.

the SOCFamily is only changed when a new chip is created (all tegrax1 chips use tegra210 as an example but the tegrax2 has its own socfamily)

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -388,16 +388,38 @@ function get_platform() {
                 __platform="armv7-mali"
                 ;;
             *)
-                # jetson nano and tegra x1 can be identified via /sys/firmware/devicetree/base/model
-                local model_path="/sys/firmware/devicetree/base/model"
-                if [[ -f "$model_path" ]]; then
-                    # ignore end null to avoid bash warning
-                    local model=$(tr -d '\0' <$model_path)
-                    case "$model" in
-                        "NVIDIA Jetson Nano Developer Kit")
-                            __platform="jetson-nano"
+                # jetsons can be identified by device tree or soc0/family (depending on the L4T version used)
+                # refer to the nv.sh script in the L4T DTS for a similar implementation
+                if [[ -e "/proc/device-tree/compatible" ]]; then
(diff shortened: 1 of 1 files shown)


scriptmodules/system.sh
```diff
@@ -388,19 +388,32 @@ function get_platform() {
...
+                        __platform="tegra-3"
+                    elif [[ ${CHIP} =~ "tegra114" ]]; then
+                        __platform="tegra-4"
+                    elif [[ ${CHIP} =~ "tegra124" ]]; then
+                        __platform="tegra-k1-32"
+                    elif [[ ${CHIP} =~ "tegra132" ]]; then
+                        __platform="tegra-k1-64"
+                    elif [[ ${CHIP} =~ "tegra210" ]]; then
+                        __platform="tegra-x1"
+                    fi
```
joolswills, line 416: Could this be done with a case statement (and above) using wildcards instead.
Also using bash RE is overkill even without imho. Can use `[[ "$var" == *string* ]]` btw which is 
Faster but prefer a case statement with wildcards.

Please use lower case `chip` and make it a local variable.

joolswills (changes requested):
Btw, just for future reference this could have been force pushed to the original PR. Could reset the old branch to master, or even push this branch back to the old remote one. But it's ok, just for future reference.
Thanks for the changes.

scriptmodules/system.sh
theofficialgman: honestly I knew you were going to suggest using a case statement... I was just lazy and this is how nvidia implemented it in their scripts
I'll make the changes

joolswills (approved):

joolswills:
Thanks.

PR #3378, opened 2021-07-21
Update esthemes.sh

DTEAM-1:
Addition of  'DTEAM-1 cygnus-blue-flames'

see https://github.com/DTEAM-1/cygnus-blue-flames

Hi , I propose, if you agree, to add  cygnus-blue-flames theme as an official theme. It contains 168 systems for Retropie. It's a branch of the original cygnus.

You can see systems with a .gif in the readme file and also some screenshots. 

Thanks again


joolswills:
Thanks.

PR #3381, opened 2021-07-31
setup - add additional packaging menus

joolswills:
Add inst menu, for managing of all installed packages in a single list
Add all menu, for managing of all packages in a single list

Wrote this a month ago and need to go over it. I was planning on some other menus too, but I'll leave this here - interested if additional ways to manage packages like this is useful.

### DIFF
--- a/scriptmodules/admin/setup.sh
+++ b/scriptmodules/admin/setup.sh
@@ -364,6 +364,23 @@ _EOF_
 
 function section_gui_setup() {
     local section="$1"
+    local ids=()
+    case "$section" in
+        all|inst)
+            name="packages"
+            local id
+            for id in "${__mod_id[@]}"; do
+                # if we are showing installed packaged, skip those that are not installed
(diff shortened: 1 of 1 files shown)


cmitu:
Just gave it a quick test and it looks and works fine. 
Just one thing I noticed - the 'all' listing has the `supplementary` section added, with `wifi`, `audiosettings`, `consolefont` , `configedit` etc. Maybe the exclusion based on the section should include `config`, i.e.
```
! [[ "${__mod_info[$id/section]}" =~ ^(depends|config|)$ ]] && ids+=("$id")
```

joolswills:
Went for just excluding config. I may rework this later. I rebased it too. I'll undraft it.

joolswills:
> Looks ok now. One minor nitpick: for the `installed` packages, maybe show only _Update all installed packages_ and remove _Install all installed packages_.

Good point. I've fixed that now. Going to merge. Can improve later if needed

PR #3382, opened 2021-08-03
Split out Joy2key to it's own module

joolswills:
Move much of the helpers.sh start/stop logic and default parameters to a joy2key wrapper script.

Switch runcommand.sh to use new wrapper script

Add tab button to "y" key for use in edit dialogs

Remove runcommand $md_inst on update. If old joy2key is present in runcommand install, trigger joy2key module install.

Also removed the ini override which I requested originally. The new code works well, and the wrapper handles which script to call based on python libs installed.


cmitu:
Looks fine after a few testing rounds, just a couple of notes:
* it looks like there's no direct dependency between `runcommand` and the new `joy2key` module - I assume this is intentional ? If `joy2key` is removed, then it would need to be manually installed.
* the `depends` will try to install `python3-sdl2`, but if  it fails, wouldn't the installation of `joy2key` also fail ? At runtime, the `sdl2` module is probed, but if it's missing entirely, then there wouldn't be any `joy2key` to run. I think this is mostly pedantic, since all platforms supported should have the package.

joolswills:
Yeah the joy2key can be removed manually, but will be installed by default as runcommand is.

I'll fix the dependency logic. Since it installs both versions it should do a check. I overlooked that.

Thanks for the review.

joolswills:
Nov the dependency logic from system.sh to joy2key_depends

joolswills:
Add a missing `local file` definition.

Going to merge. Hopefully I've not missed anything.

PR #3384, opened 2021-08-09
Added I.K.E.M.E.N GO scriptmodule

SuperFromND:
This PR adds I.K.E.M.E.N GO as a scriptmodule.  See [the posts on this thread on the RetroPie forums](https://retropie.org.uk/forum/topic/28055/i-k-e-m-e-n-go-on-raspberry-pi-4) for more technical details.

[I.K.E.M.E.N GO](https://github.com/Windblade-GR01/Ikemen-GO) is an open-source clone of the [M.U.G.E.N](https://en.wikipedia.org/wiki/Mugen_(game_engine)) fighting game engine written in Golang.  M.U.G.E.N has been used to create a number of full games as well as lots of content for general use, and requests to add it to RetroPie [have been](https://retropie.org.uk/forum/topic/2917/ideas-so-mugen-diablo-i-n64-emulator-improve) [around for](https://retropie.org.uk/forum/topic/12776/m-u-g-e-n-fighting-games-for-retropie) [a while.](https://retropie.org.uk/forum/topic/8038/m-u-g-e-n-support)

In addition to just installing I.K.E.M.E.N GO, this module also sets up the engine's content folders in `/home/pi/RetroPie/roms/ports/ikemen-go` to allow for easy installation of content, as well as one folder in `/opt/retropie/configs/ports/ikemen-go/save` so that users can edit the config.json file (I.K.E.M.E.N GO's equivalent of M.U.G.E.N's `mugen.cfg`) if need be.  This is why there are two `chown` commands in the module, with the third (`chown -R $user:$user "$md_inst"`) being there to ensure I.K.E.M.E.N GO doesn't crash due to being unable to write its `Ikemen.log` file.

This scriptmodule has been tested and confirmed to work at least on the Pi 4, but due to the huge amount of dependencies required to compile it on the Pi, as well as some other issues (particularly when concerning composite video output and, at the moment, inputs), I've placed it under experimental.


scriptmodules/ports/ikemen-go.sh
```diff
@@ -0,0 +1,77 @@
...
+#
+
+rp_module_id="ikemen-go"
+rp_module_desc="I.K.E.M.E.N GO - Clone of M.U.G.E.N to the Go programming language"
+rp_module_licence="MIT https://raw.githubusercontent.com/Windblade-GR01/Ikemen-GO/master/License.txt"
+rp_module_help="Copy characters, stages, screenpacks, etc. to $romdir/ports/ikemen-go\n\nConfig files can be found at $configdir/ports/ikemen-go/save"
+rp_module_section="exp"
+
+function depends_ikemen-go() {
+    getDepends golang-go libgl1-mesa-dev xinit xorg xorg-dev libopenal1 libopenal-dev libgtk-3-dev libasound2-dev freeglut3-dev
```
joolswills, line 19: The dependencies should only include the development packages. As they depend on the binary libraries they can be excluded. Eg `libopenal1`. Also xorg package depends on xinit so that's not needed.
Is xorg-dev needed? Maybe can tweak this a bit to minimise dependencies

scriptmodules/ports/ikemen-go.sh
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function sources_ikemen-go() {
+    gitPullOrClone "$md_build" https://github.com/Windblade-GR01/Ikemen-GO.git
```
joolswills, line 23: This should be updated to use the `rp_module_repo` variable
 See other modules for examples.

scriptmodules/ports/ikemen-go.sh
```diff
@@ -0,0 +1,77 @@
...
+    go mod tidy
+    make TAGS='-tags al_cmpt' linux
+    # grabs default screenpack and content required for the game to run; note that the screenpack has a CC-BY-NC 3.0 license
+    make elecbyte
+    md_ret_require="$md_build/Ikemen_GO"
+}
+
+function install_ikemen-go() {
+    cp 'elecbyte/LICENCE.txt' 'ScreenpackLicense.txt'
+    mkdir save sound
```
joolswills, line 37: Is the save folder for writing? If so it is better to symlink it in the configure stage to `$md_conf_root/system` etc?

scriptmodules/ports/ikemen-go.sh
```diff
@@ -0,0 +1,77 @@
...
+    addPort "$md_id" "ikemen-go" "I.K.E.M.E.N GO" "XINIT:$md_inst/ikemen-go.sh"
+
+    moveConfigDir "$md_inst/chars" "$romdir/ports/ikemen-go/chars"
+    moveConfigDir "$md_inst/stages" "$romdir/ports/ikemen-go/stages"
+    moveConfigDir "$md_inst/data" "$romdir/ports/ikemen-go/data"
+    moveConfigDir "$md_inst/external" "$romdir/ports/ikemen-go/external"
+    moveConfigDir "$md_inst/font" "$romdir/ports/ikemen-go/font"
+    moveConfigDir "$md_inst/sound" "$romdir/ports/ikemen-go/sound"
+    moveConfigDir "$md_inst/save" "$configdir/ports/ikemen-go/save"
+
```
joolswills, line 66: These folders can't contain anything and are better if they don't exist prior to this stage.

joolswills (changes requested):
Thanks. Not tested yet but made a few comments that may be useful. Will give it a test though when I have a chance. Thanks.

scriptmodules/ports/ikemen-go.sh
SuperFromND: `xorg-dev` is [apparently needed for GLFW](https://www.glfw.org/docs/latest/compile.html#compile_deps) on Debian distributions, but the other non-`dev` packages could probably be removed without issue.

SuperFromND: Okay, I just tested: removing `xorg` is okay, but removing `xorg-dev` and replacing it with `libglfw3-dev` causes IKEMEN GO to not be able to start, with the following runcommand.log:
```
Parameters: 
Executing: xinit /dev/shm/retropie_xinitrc -- vt1 -keeptty

xinit: unable to run server "X": No such file or directory
Use the -- option, or make sure that /usr/bin is in your path and
that "X" is a program or a link to the right type of server
for your display.  Possible server names include:

    Xorg        Common X server for most displays
    Xvfb        Virtual frame buffer
    Xfake       kdrive-based virtual frame buffer
    Xnest       X server nested in a window on another X server
    Xephyr      kdrive-based nested X server
    Xvnc        X server accessed over VNC's RFB protocol
    Xdmx        Distributed Multi-head X server

xinit: giving up
xinit: unable to connect to X server: Connection refused
xinit: server error
```
So it does appear that  `xorg-dev` is necessary to get xinit to start.
On a related note, I tested and found that `freeglut3-dev` is seemingly unnecessary.

SuperFromND: Okay, further testing seems to indicate that `xorg` will suffice and that `xorg-dev` is overkill.

cmitu: `xterm` would also be needed to be installed - it's not included in `xorg` and it's used by the launching script.

I've compiled the new module ran it briefly from ES (Pi4) and the video it's noticeably slow (like it's using software rendering). The lack of fullscreen may be remedied through configuration, I haven't fiddled with it.
Maybe I need to re-build again, but I doubt is the dependencies that may be causing this. Any method to check whether the engine uses HW acceleration ? The log file from the install folder is empty.

SuperFromND: I believe that IKEMEN GO doesn't have any form of software rendering; i.e. it only has an OpenGL-based renderer; one not optimized specifically for the Pi, which is likely why it's slow.

Fullscreen can indeed be enabled in-game via the video settings (or alternatively editing `save/config.json` and manually setting fullscreen to true); the only reason I haven't set it to on by default (aside from it crashing composite-out mode) is that IKEMEN doesn't include a default config.json, generating it on first run instead.

From what I can tell, the log file only gets written to in the event of a crash.  I don't think IKEMEN GO has any form of `--verbose` option either, though I could be wrong on that.

cmitu:
(continued from the review)
I have tested again with the initial dependencies installed (i.e. `xorg-dev`) and rebuilding the binary. The video still doesn't get over than 22fps (in demo mode) with a modest resolution (800x600). Fullscreen is also not really fullscreen, the image is scaled in the bottom left corner of the screen. In arcade play mode it barely reaches 15 fps and the CPU is clearly the bottlenect (consuming 3 full cores or going over).

ikemen

@SuperFromND is this the same experience you have or there is rooms for improvement ?

SuperFromND:
That's odd; while I was getting those sorts of issues initially, eventually I was able to get IKEMEN to display centered by default, not in the bottom-left corner (the cause was _something_ involving `xorg` and `xorg-dev` being installed simultaneously; removing `xorg` and/or installing from a fresh RetroPie image seemed to fix it.)

I have noticed the bottom-left-corner thing crop up at certain resolutions, since it relies on changing video modes to achieve the fullscreen effect it's possible that the resolution of the Pi is changed to something higher than the resolution of the game.

Admittedly I've never tried setting a native resolution above 640x480, since that's what most characters and screenpacks are built around.

cmitu:
> @cmitu That's odd; while I was getting those sorts of issues initially, eventually I was able to get IKEMEN to display centered by default,

It does display centered by default on the 1st run and I suppose if the resolution is left untouched will stay this way. But it also doesn't fill up the entire screen, even if I configure in the video settings to be fullscreen, that's why I tried switching the resolution. My display supports both resolutions (800x600 and 640x480), but runs in a higher resolution by default and having the game in a tiny window (centered or not) is not ideal.

SuperFromND:
> It does display centered by default on the 1st run and I suppose if the resolution is left untouched will stay this way.

The way this was worded implied that something makes it default to the bottom-left corner on subsequent runs, and it made me want to see if perhaps setting a game resolution was causing a window's starting coordinate to "save" an offset, but I just tried switching the resolution to 960x720, restarted, switched it back to 640x480, restarted again, and the screen still starts in the center as I expected.

> But it also doesn't fill up the entire screen, even if I configure in the video settings to be fullscreen, that's why I tried switching the resolution.

Odd, changing the setting should at the very least *try* to change the video resolution to match that of the game's resolution, or at least the nearest one that the Pi likes.  I tried 800x600, both windowed and in fullscreen, and that seemed to work fine for me.  It's worth noting though that [fullscreen on IKEMEN has been known to be a bit wonky in general for a while](https://github.com/Windblade-GR01/Ikemen-GO/issues/324#issuecomment-894717432), so it's likely that something will address whatever problem you're having at some point in the future.

scriptmodules/ports/ikemen-go.sh
SuperFromND: `rp_module_repo` is used now; I didn't realize it was introduced until this comment, so thanks for letting me know!

scriptmodules/ports/ikemen-go.sh
SuperFromND: The `save` folder is indeed written to by the engine; it's used for saving a few different things, like `config.json`, replay files, and a few other miscellaneous files in certain modes like `randomtest`.

The script now symlinks it to the `$configdir/ports/ikemen-go/save` folder, created with user-ownership via `mkUserDir` so that IKEMEN can write the files successfully.

scriptmodules/ports/ikemen-go.sh
SuperFromND: I decided to turn `sound` and `save` into user-owned folders that are created and symlinked in the `configure` stage:
```sh
    mkUserDir "$romdir/ports/ikemen-go/sound"
    mkUserDir "$configdir/ports/ikemen-go/save"
    ln -sf "$romdir/ports/ikemen-go/sound" "$md_inst/sound"
    ln -sf "$configdir/ports/ikemen-go/save" "$md_inst/save"
```

SuperFromND:
Okay, I just updated the module with a notably large change: this commit now adds a secondary module, `golang-1.17`, as a dependency.  This is required for IKEMEN GO to compile as one of its dependencies requires Go 1.14 or later, and current Debian Buster only supplies up to 1.11 via `apt-get`.  I chose 1.17 as it's the latest version and should be future-proofed with future IKEMEN GO dependency updates for the forseeable future.

I didn't modify the other `golang` script that's included in supplementary, though it probably would be a smart idea to modify that one to reflect its version number (1.8).

SuperFromND:
> Fullscreen is also not really fullscreen, the image is scaled in the bottom left corner of the screen.
> 
> ikemen

Okay, I finally figured out why this happens!  It turns out if you have any specific video mode set in /boot/config.txt (whether manually or with `raspi-config`), IKEMEN will not scale itself fullscreen.  The reason is that IKEMEN relies on changing the video mode of the monitor to handle fullscreen (you can't just maximize it like with most other programs, you either get a window or a video mode-based fullscreen).

The workaround is, on top of leaving the video mode in /boot/config.txt as default (in other words, if there's any active `hdmi_group` or `hdmi_mode` parameters, comment them out), to use RetroPie's runcommand video mode settings to force 640x480 resolution (or whatever resolution you set IKEMEN to run in).  Not ideal, but at least it works?

No idea why you're getting such bad performance though; I'm getting fullspeed with only one core getting substantial usage (though quite a lot of memory usage, as to be expected since I have a decent amount of custom content loaded):
image

PR #3385, opened 2021-08-14
remove tegra 2 and change tegra 3/4 to gles

theofficialgman:
tegra 2 is armv7 and doesn't support NEON and is unlikely to ever be used
tegra 3 (and maybe 4?) only support gles (up to 2.0) not gl (in L4T linux at least)


joolswills:
Thanks.

PR #3386, opened 2021-08-20
lr-vice: preset a couple of SID options on ARMv6

cmitu:
The options changed are similar to the configuration used for the standalone VICE:

* Set the SID implementation to FastSID
* Decrease the sound sample rate to 22050

### DIFF
--- a/scriptmodules/libretrocores/lr-vice.sh
+++ b/scriptmodules/libretrocores/lr-vice.sh
@@ -66,6 +66,11 @@ function configure_lr-vice() {
 
     [[ "$md_mode" == "remove" ]] && return
 
+    if isPlatform "arm"; then
+        setRetroArchCoreOption "vice_sid_engine" "FastSID"
+        isPlatform "armv6" && setRetroArchCoreOption "vice_sound_sample_rate" "22050"
+    fi
+
     cp -R "$md_inst/data" "$biosdir"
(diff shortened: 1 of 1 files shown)


cmitu:
For standalone `vice`, we preset the options for any non `x11` platform, not sure if we should do the same for the libretro core (hence the draft).

joolswills:
Actually Vice should be updated. It's old code. I'll take a look

joolswills:
This looks good btw. Thanks.

joolswills:
So looking at the vice code, the original old logic was intended as: If not x86 then use fastsid. If armv6 then half sample rate.

With the RPI3/RPI4 the floating point Sid emulation is fine but it probably is still a big overhead for RPI2. I actually think it may still be worth defaulting to fastsid on all arm. As we.do in the standalone core.

I'm adjusting the logic in the standalone now to update it taking into account new flags etc.

joolswills:
https://github.com/RetroPie/RetroPie-Setup/commit/a39f05d72f2fe2d6604ee646a761d2ac0689a1b7

cmitu:
OK, I'll sync the changes to a39f05d.

cmitu:
OK, the options should be similar to the `vice` changes.

joolswills:
Thanks.

PR #3387, opened 2021-08-21
Add Hypseus Singe (Daphne) SDL2 Laserdisc emulator.

DirtBagXon:
This is to automate: https://github.com/DirtBagXon/hypseus-singe/blob/master/src/3rdparty/retropie/RETROPIE.md

This plugin is not restricted in arcitecture like the previous SDL1 version.
Supports Pi (inc. aarch64), ODroid & PC (i386/AMD64).

Standing Forum discussions:
https://retropie.org.uk/forum/topic/18505/new-hypseus-and-lr-daphne-to-add-on-retropie-setup
https://retropie.org.uk/forum/topic/27426/request-add-daphne-singe-emulator
https://retropie.org.uk/forum/topic/31155/integrate-daphne-emulator-back-into-the-retropie-script-for-ubuntu

This setup follows the existing Daphne plugin installation as closely as possible.
I have mirrored the included config file to the existing plugin where appropriate.

This is my first pull request for RetroPie, please let me know if anything else is required.

I have tagged as an experimental package, application should be stable.

~Would like to shift 'RetroPie' branch from:~

~https://github.com/DirtBagXon/hypseus-singe~
~to~
~https://github.com/RetroPie/hypseus-singe~

~If someone could provide details on how to submit on a new repo.~

### DIFF
--- a/scriptmodules/emulators/hypseus.sh
+++ b/scriptmodules/emulators/hypseus.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="hypseus"
+rp_module_desc="Hypseus Singe - Laserdisc Emulator"
+rp_module_help="ROM Extension: .daphne\n\nCopy your Daphne roms to $romdir/daphne"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/DirtBagXon/hypseus-singe/master/LICENSE"
+rp_module_repo="git https://github.com/DirtBagXon/hypseus-singe.git RetroPie"
+rp_module_section="exp"
+rp_module_flags="sdl2"
+
+function depends_hypseus() {
+    getDepends libvorbis-dev libogg-dev zlib1g-dev libmpeg2-4-dev libsdl2-dev libsdl2-image-dev libsdl2-ttf-dev cmake
+}
+
+function sources_hypseus() {
+    gitPullOrClone
+}
+
+function build_hypseus() {
+    rm -rf build
+    mkdir build
+    cd build
+    rpSwap on 1024
+    cmake ../src
+    make
+    rpSwap off
+    cp hypseus ../hypseus.bin
+    md_ret_require="hypseus"
+}
+
+function install_hypseus() {
+    md_ret_files=(
+        'sound'
+        'pics'
+        'fonts'
+        'hypseus.bin'
+        'LICENSE'
+    )
+}
+
+function configure_hypseus() {
+    mkRomDir "daphne"
+    mkRomDir "daphne/roms"
+
+    addEmulator 0 "$md_id" "daphne" "$md_inst/hypseus.sh %ROM%"
+    addSystem "daphne"
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    mkUserDir "$md_conf_root/daphne"
+
+    local dir
+    for dir in ram logs screenshots; do
+        mkUserDir "$md_conf_root/daphne/$dir"
+        ln -snf "$md_conf_root/daphne/$dir" "$md_inst/$dir"
+    done
+
+    copyDefaultConfig "$md_data/hypinput.ini" "$md_conf_root/daphne/hypinput.ini"
+
+    ln -snf "$romdir/daphne/roms" "$md_inst/roms"
+    ln -snf "$romdir/daphne/roms" "$md_inst/singe"
+
+    ln -sf "$md_conf_root/daphne/hypinput.ini" "$md_inst/hypinput.ini"
+
+    local common_args="-framefile \"\$dir/\$name.txt\" -homedir \"$md_inst\" -fullscreen \$params"
+
+    cat >"$md_inst/hypseus.sh" <<_EOF_
+#!/bin/bash
+dir="\$1"
+name="\${dir##*/}"
+name="\${name%.*}"
+
+if [[ -f "\$dir/\$name.commands" ]]; then
+    params=\$(<"\$dir/\$name.commands")
+fi
+
+if [[ -f "\$dir/\$name.singe" ]]; then
+    "$md_inst/hypseus.bin" singe vldp -retropath -manymouse -script "\$dir/\$name.singe" $common_args
+else
+    "$md_inst/hypseus.bin" "\$name" vldp $common_args
+fi
+_EOF_
+    chmod +x "$md_inst/hypseus.sh"
+    mkdir -p "$md_inst/framefile"
+}
--- a/scriptmodules/emulators/hypseus/hypinput.ini
+++ b/scriptmodules/emulators/hypseus/hypinput.ini
@@ -1,0 +1,58 @@
+# Sample hypinput.ini
+# All key options listed
+# Hypseus uses SDL2 Keycodes
+#
+# The first two entries are SDL2 keyboard codes or names (0 for "none")
+#
+# Find SDL2 keyboard code information here:
+# https://github.com/DirtBagXon/hypseus-singe/blob/master/doc/keylist.txt
+#
+# Hypseus Singe supports configuration on multiple joysticks
+# First joystick is defined as 0, second joystick as 1 etc.
+#
+# IMPORTANT: Find the joystick button and axis by running:
+# jstest /dev/input/js0 || jstest /dev/input/js1
+#
+# The third number in config is a joystick button code (or 0 for "none")
+# Since 0 is reserved for special meaning, joystick button 0 is
+# identified as 1. Button 1 is identified as 2, and so on.
+#
+# Defining 001 (or 1) identifies first joystick(0) button 0
+# Defining 111 identifies second joystick(1) button 10
+#
+# The fourth number in config (if specified) is the joystick axis
+# configuration (or 0 for "none"). Since 0 is reserved for
+# special meaning, joystick axis 0 is identified as 1.
+# Axis 1 is identified as 2, and so on.
+#
+# Only the first four switches are defined (SWITCH_UP->SWITCH_RIGHT) for axis
+#
+# Defining -001 (or -1) identifies first joystick(0) axis 0 in negative direction
+# Defining +102 identifies second joystick(1) axis 1 in positive direction
+
+# KEY_BUTTON3 Turns scoreboard on/off in lair/ace
+
+[KEYBOARD]
+KEY_UP = SDLK_UP SDLK_r 5 -002
+KEY_DOWN = SDLK_DOWN SDLK_f 7 +002
+KEY_LEFT = SDLK_LEFT SDLK_d 8 -001
+KEY_RIGHT = SDLK_RIGHT SDLK_g 6 +001
+KEY_COIN1 = SDLK_5 0 1
+KEY_COIN2 = SDLK_6 0 0
+KEY_START1 = SDLK_1 0 4
+KEY_START2 = SDLK_2 0 0
+KEY_BUTTON1 = SDLK_LCTRL SDLK_a 14
+KEY_BUTTON2 = SDLK_LALT SDLK_s 15
+KEY_BUTTON3 = SDLK_SPACE SDLK_d 16
+KEY_SKILL1 = SDLK_LSHIFT SDLK_w 0
+KEY_SKILL2 = SDLK_z SDLK_i 0
+KEY_SKILL3 = SDLK_x SDLK_k 0
+KEY_SERVICE = SDLK_9 0 0
+KEY_TEST = SDLK_F2 0 0
+KEY_RESET = SDLK_0 0 0
+KEY_SCREENSHOT = SDLK_F12 0 0
+KEY_QUIT = SDLK_ESCAPE SDLK_q 17
+KEY_PAUSE = SDLK_p 0 0
+KEY_CONSOLE = SDLK_BACKSLASH 0 0
+KEY_TILT = SDLK_t 0 0
+END


DirtBagXon:
In order to test locally, I had to hardcode the `function sources_hypseus()` url and branch.

```
function sources_hypseus() {
    gitPullOrClone "$md_build" "https://github.com/DirtBagXon/hypseus-singe.git" "RetroPie" "" 0
}
```

The macro for `$md_repo_url` expansion does not appear to be contained in this repo, so I assume it's parsed elsewhere, I didn't explore.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+rp_module_id="hypseus"
+rp_module_desc="Hypseus Singe - Laserdisc Emulator"
+rp_module_help="ROM Extension: .daphne\n\nCopy your Daphne roms to $romdir/daphne"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/DirtBagXon/hypseus-singe/master/LICENSE"
+rp_module_repo="git https://github.com/DirtBagXon/hypseus-singe.git RetroPie"
+rp_module_section="exp"
+rp_module_flags="sdl2"
+
+function depends_hypseus() {
+    getDepends libvorbis-dev libogg-dev zlib1g-dev libmpeg2-4-dev libsdl2-image-dev libsdl2-ttf-dev cmake
```
cmitu, line 21: It's missing the `libsdl2-dev` dependency. While it's installed automatically by any core package, it's better to explicitly add it.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function sources_hypseus() {
+    gitPullOrClone "$md_build" "$md_repo_url" "$md_repo_branch" "" 0
```
cmitu, line 25: This should be just
```
gitPullOrClone
```
Make sure you're using the latest RetroPie-Setup script for this to work.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+        'hypseus.bin'
+        'LICENSE'
+    )
+}
+
+function configure_hypseus() {
+    mkRomDir "daphne"
+    mkRomDir "daphne/roms"
+
+    addEmulator 1 "$md_id" "hypseus" "$md_inst/hypseus.sh %ROM%"
```
cmitu, line 52: This should be changed to
```
addEmulator 0 "$md_id" "daphne" "$md_inst/hypseus.sh %ROM%"
```
to ensure that `hypseus` is added as an alternative (and non-default) emulator for the `daphne` system.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+    addSystem "hypseus"
```
cmitu, line 53: This should be changed to 
```
addSystem "daphne"
```

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    mkUserDir "$md_conf_root/hypseus"
+
+    if [[ ! -f "$md_conf_root/hypseus/hypinput.ini" ]]; then
```
cmitu, line 59: `hypseus` should be `daphne`, so that the file is copied to `/opt/retropie/configs/daphne`. Same for the `cp` command below.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+        cp -v "$md_data/hypinput.ini" "$md_conf_root/hypseus/hypinput.ini"
+    fi
+    ln -snf "$romdir/daphne/roms" "$md_inst/roms"
+    ln -snf "$romdir/daphne/roms" "$md_inst/singe"
+    ln -sf "$md_conf_root/$md_id/hypinput.ini" "$md_inst/hypinput.ini"
```
cmitu, line 64: The symlink should be
```
ln -sf "$md_conf_root/daphne/hypinput.ini" "$md_inst/hypinput.ini"
```

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+
+if [[ -f "\$dir/\$name.singe" ]]; then
+    "$md_inst/hypseus.bin" singe vldp -retropath -framefile "\$dir/\$name.txt" -script "\$dir/\$name.singe" -homedir "$md_inst" -fullscreen \$params
+else
+    "$md_inst/hypseus.bin" "\$name" vldp -framefile "\$dir/\$name.txt" -homedir "$md_inst" -fullscreen \$params
+fi
+_EOF_
+    chmod +x "$md_inst/hypseus.sh"
+
+    chown -R $user:$user "$md_inst"
```
cmitu, line 84: Is regular user _write_ access really needed for the installation folder ?

scriptmodules/emulators/hypseus/hypinput.ini
```diff
@@ -0,0 +1,58 @@
+# Sample hypinput.ini
+# All key options listed
+# Hypseus uses SDL2 Keycodes
+#
+# The first two entries are SDL2 keyboard codes or names (0 for "none")
+#
+# Find SDL2 keyboard code information here:
+# https://gitlab.com/DirtBagXon/hypseus-singe/-/blob/main/doc/keylist.txt
```
cmitu, line 8: Should maybe change the link to <https://github.com/DirtBagXon/hypseus-singe/blob/master/doc/keylist.txt> or directly to <https://wiki.libsdl.org/SDLKeycodeLookup>.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+function configure_hypseus() {
+    mkRomDir "daphne"
+    mkRomDir "daphne/roms"
+
+    addEmulator 1 "$md_id" "hypseus" "$md_inst/hypseus.sh %ROM%"
+    addSystem "hypseus"
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    mkUserDir "$md_conf_root/hypseus"
```
cmitu, line 57: This is not needed, it's handled by the `addSystem` function.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,87 @@
...
+if [[ -f "\$dir/\$name.singe" ]]; then
+    "$md_inst/hypseus.bin" singe vldp -retropath -framefile "\$dir/\$name.txt" -script "\$dir/\$name.singe" -homedir "$md_inst" -fullscreen \$params
+else
+    "$md_inst/hypseus.bin" "\$name" vldp -framefile "\$dir/\$name.txt" -homedir "$md_inst" -fullscreen \$params
+fi
+_EOF_
+    chmod +x "$md_inst/hypseus.sh"
+
+    chown -R $user:$user "$md_inst"
+    chown -R $user:$user "$md_conf_root/hypseus/hypinput.ini"
```
cmitu, line 85: Replace `hypseus` with `daphne`.

cmitu:
Hi, I've left a few comments with the changes that I think are necessary.
Since `hypseus` is intended to be using the same `daphne` folder as the _Daphne_ emulator, `hypseus` shouldn't be added as a system or having a separate conf dir - I marked the places with the changes in the source.

I'm not sure why the `sources` function doesn't work - are you using a recent RetroPie-Setup version ?

Regarding the repository, there's no need to switch to one under the `retropie` account, we prefer to use the upstream repository when available. The `daphne` emulator is an exception. I'm guessing that, at the time it was added, the source was not available as a Git/TGZ/other repository so that's why it was added manually to the `retropie` org's Github repositories.
Are there any differenced between the master and the RetroPie branches in you repo ? Ideally we should always be using the main/stable branch so that the upstream developer(s) (i.e. you) don't have to do anything special for RetroPie installations.

scriptmodules/emulators/hypseus.sh
DirtBagXon: Yes user access is required for screenshots and logs, same as Daphne.

I can be more restrictive in hypseus though, via:

```
    mkUserDir "$md_inst/logs"
    mkUserDir "$md_inst/screenshots"
    chown -R $user:$user "$md_inst/logs"
    chown -R $user:$user "$md_inst/screenshots"
```

If that works ?

scriptmodules/emulators/hypseus/hypinput.ini
DirtBagXon: Will do, although that https://wiki.libsdl.org link keeps disappearing which is why I host locally :)

DirtBagXon:
OK, suggested changes pushed and tested.

The `sources` function worked with plain `gitPullOrClone` after a git pull - my mistake, thanks.

I am happy to keep the repository, I was just trying to follow the existing package format. But no issues there.

Regarding the `branch`, there is lots of, MacOSX and Windows, bloat in the master branch history, so thought it leaner to keep the Pi branch clean. Also there a change that uses the packaged `libmpeg2` of the distribution in this branch, as the 'optimised' version, linked statically in master, causes issues on the Pi.

I have been, and am happy to continue, maintaining as a branch.

scriptmodules/emulators/hypseus.sh
DirtBagXon: Also 'ram' folder where certain Daphne games store highscore tables etc.

Looked at this, but can't really see how to limit further without creating user writable directories in some other location...

Seems most obvious to leave in the main `-homedir` folder, like Daphne always has.

cmitu: A solution would be to use the _configdir_ for writable data and point `-homedir` to it. One - more - advantage is being able to copy the config folder between systems and keep the user generated content (e.g. hiscore tables). _configdir_ is already user writeable.

cmitu:
@DirtBagXon thanks for the changes, they look good.
I'll give it another and see how it works when installed from scratch (with and without Daphne installed beforehand).

scriptmodules/emulators/hypseus.sh
DirtBagXon: It might become a bit of a symlinking hell, but would involve creating a few more dirs (in  _$md_inst_ and _$md_conf_root/daphne/_) on install and symlinking would look something like this:

```
total 1872
drwxr-xr-x 9 pi   pi      4096 Aug 22 13:57 .
drwxr-xr-x 5 root root    4096 Aug 22 13:51 ..
drwxr-xr-x 2 pi   pi      4096 Aug 22 13:51 fonts
drwxr-xr-x 2 pi   pi      4096 Aug 22 13:57 framefile
lrwxrwxrwx 1 pi   pi        41 Aug 22 13:51 hypinput.ini -> /opt/retropie/configs/daphne/hypinput.ini
-rwxr-xr-x 1 pi   pi   1836348 Aug 22 21:14 hypseus.bin
-rwxr-xr-x 1 pi   pi       528 Aug 22 13:51 hypseus.sh
-rw-r--r-- 1 pi   pi     32472 Aug 22 13:51 LICENSE
drwxr-xr-x 2 pi   pi      4096 Aug 22 13:57 logs -> /opt/retropie/configs/daphne/logs
drwxr-xr-x 3 pi   pi      4096 Aug 22 13:51 pics
drwxr-xr-x 2 pi   pi      4096 Aug 22 19:26 ram -> /opt/retropie/configs/daphne/ram
-rw-r--r-- 1 pi   pi       286 Aug 22 13:51 retropie.pkg
lrwxrwxrwx 1 pi   pi        34 Aug 22 13:51 roms -> /home/pi/RetroPie/roms/daphne/roms
drwxr-xr-x 2 pi   pi      4096 Aug 22 13:57 screenshots -> /opt/retropie/configs/daphne/screenshots
lrwxrwxrwx 1 pi   pi        34 Aug 22 13:51 singe -> /home/pi/RetroPie/roms/daphne/roms
drwxr-xr-x 2 pi   pi      4096 Aug 22 13:51 sound
```

I can look at that, if it's something you would want to pursue ?

DirtBagXon: Moving _-homedir_ would create a lot more issues, as Daphne is notoriously fussy about its paths.

I think symlinking, as per above, would be simpler.

cmitu: Well, given the constraints, I guess using `$md_inst` for `homedir` is ok and there's no need to complicate the installation layout in this case.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,85 @@
...
+function depends_hypseus() {
+    getDepends libvorbis-dev libogg-dev zlib1g-dev libmpeg2-4-dev libsdl2-dev libsdl2-image-dev libsdl2-ttf-dev cmake
+}
+
+function sources_hypseus() {
+    gitPullOrClone
+}
+
+function build_hypseus() {
+    local params=()
```
joolswills, line 29: This doesn't look used.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,85 @@
...
+    mkRomDir "daphne/roms"
+
+    addEmulator 0 "$md_id" "daphne" "$md_inst/hypseus.sh %ROM%"
+    addSystem "daphne"
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    if [[ ! -f "$md_conf_root/daphne/hypinput.ini" ]]; then
+        cp -v "$md_data/hypinput.ini" "$md_conf_root/daphne/hypinput.ini"
+    fi
```
joolswills, line 59: This looks like a job for `copyDefaultConfig` 

https://retropie.org.uk/api/helpers_8sh.html#a49d0ead30005ce0d98d91e09e8a3cd2c

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,85 @@
...
+
+if [[ -f "\$dir/\$name.commands" ]]; then
+    params=\$(<"\$dir/\$name.commands")
+fi
+
+if [[ -f "\$dir/\$name.singe" ]]; then
+    "$md_inst/hypseus.bin" singe vldp -retropath -framefile "\$dir/\$name.txt" -script "\$dir/\$name.singe" -homedir "$md_inst" -fullscreen \$params
+else
+    "$md_inst/hypseus.bin" "\$name" vldp -framefile "\$dir/\$name.txt" -homedir "$md_inst" -fullscreen \$params
+fi
```
joolswills, line 78: There's a lot of shared params here. Maybe best handled in an array with logic to add needed params to avoid duplication?

scriptmodules/emulators/hypseus.sh
joolswills: I would still be keen to avoid a writable md_inst. Unless it needs to make random named temporary files
It should be avoidable. Just to note `$md_inst` is wiped on update
So anything user writable needs to be symlinked etc.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,85 @@
...
+_EOF_
+    chmod +x "$md_inst/hypseus.sh"
+
+    chown -R $user:$user "$md_inst"
+    chown -R $user:$user "$md_conf_root/daphne/hypinput.ini"
```
joolswills, line 83: `-R` doesn't make sense on a single file. But this line won't be needed if using copyDefaultConfig

joolswills:
I've made a few comments. Thanks for your work on this.

scriptmodules/emulators/hypseus.sh
joolswills: Another option to consider would be to symlink the other way around. Depending what static data is needed relative to `-homedir` it could be symlinked from the config area to `$md_inst`. Another option would be to adjust the source since this is a fork anyway with changes. But whichever seems most reasonable.

If we really can't avoid a writable md_inst, I may be persuaded but I'd like to avoid it. Thanks :-)

DirtBagXon: Rather than use the _$md_conf_root_ for _non-conf_ data, looking at some other packages would **$home** be appropriate for savegame data, logs and screenshots ?

e.g. from _fs-uae.sh_

    mkUserDir "$home/Documents/FS-UAE"
    mkUserDir "$home/Documents/FS-UAE/Configurations"

_i.e._

    mkUserDir "$home/hypseus/ram"
    mkUserDir "$home/hypseus/logs"
    mkUserDir "$home/hypseus/screenshots"

And then create appropriate symlinks?

That would leave only '_framefile_' which I can create directory on install. Nothing in RetroPie should be writing to that. Looking at source, this is just used as a secondary check for missing framefiles by the VLDP. So has no need to be writable by the user, but will try to created if it doesn't exist.

DirtBagXon:
Ok, so I have pushed my proposed changes on writable directories.

This aligns with the standard hypseus/daphne layout behaviour in _~/.hypseus_ when run as a standalone application, and also if hypseus.bin is run outside the RetroPie script arguments. So makes good sense.

I couldn't see a function to mkdir in _$md_inst_ - So fudged in _build_hypseus_ and _install_hypseus_ - It looks ok, but please let me know if there is a better way.

The bash array on arguments, I need to be convinced on. Firstly bash arrays are fugly :grin:. Also I don't believe the arguments are so long to make it viable. It is also far easier to debug in the current layout. Hopefully you agree.

Many thanks for the feedback. It is turning out far from the original daphne installer I tried to copy, but all changes are for the better and address some of the idiosyncrasies of the daphne install that bugged me already. I am happy if you are.

The layout after install will now look like this:

```
drwxr-xr-x 6 root root    4096 Aug 24 10:36 .
drwxr-xr-x 5 root root    4096 Aug 24 10:36 ..
drwxr-xr-x 2 root root    4096 Aug 24 10:36 fonts
drwxr-xr-x 2 root root    4096 Aug 24 10:36 framefile
lrwxrwxrwx 1 root root      41 Aug 24 10:36 hypinput.ini -> /opt/retropie/configs/daphne/hypinput.ini
-rwxr-xr-x 1 root root 1796160 Aug 24 10:36 hypseus.bin
-rwxr-xr-x 1 root root     528 Aug 24 10:36 hypseus.sh
-rw-r--r-- 1 root root   32472 Aug 24 10:36 LICENSE
lrwxrwxrwx 1 root root      22 Aug 24 10:36 logs -> /home/pi/.hypseus/logs
drwxr-xr-x 3 root root    4096 Aug 24 10:36 pics
lrwxrwxrwx 1 root root      21 Aug 24 10:36 ram -> /home/pi/.hypseus/ram
-rw-r--r-- 1 root root     286 Aug 24 10:36 retropie.pkg
lrwxrwxrwx 1 root root      34 Aug 24 10:36 roms -> /home/pi/RetroPie/roms/daphne/roms
lrwxrwxrwx 1 root root      29 Aug 24 10:36 screenshots -> /home/pi/.hypseus/screenshots
lrwxrwxrwx 1 root root      34 Aug 24 10:36 singe -> /home/pi/RetroPie/roms/daphne/roms
drwxr-xr-x 2 root root    4096 Aug 24 10:36 sound
```

scriptmodules/emulators/hypseus.sh
DirtBagXon: Made comments below on this, but bash arrays are fugly :grin:

scriptmodules/emulators/hypseus.sh
DirtBagXon: See https://github.com/RetroPie/RetroPie-Setup/pull/3387/commits/cfcf2a1f274cac397e89f964a44e8f80807f31a8

DirtBagXon:
Please let me know if there are any outstanding changes you would like to see.

Many thanks guys.

joolswills:
Sorry for the delay on this.

The directories and symlink layout could be improved I think. 

Unless I've misunderstood I think it would be better if `$home/.$md_id` was symlinked to `$md_conf_root` (via moveConfigDir if there will be existing files in this location).

And then the subfolders (if they need to be premade) can be made in `$md_conf_root`. This allows us to have all the configurations in one place.

With this change `"$md_inst/ram` `$md_inst/logs` and `$md_inst/screenshots` would symlink to $md_conf_root also - not to the `$home/.$md_id/` folder.

I'm happy to make these changes if it makes sense to you and I've not overlooked something.

joolswills:
Sorry I meant `$md_conf_root/daphe/hypseus` for the configuration location I guess - not `$md_conf_root/$md_id` - or even `$md_conf_root/daphne` if there will be no collisions with daphne configs.

joolswills:
Here's an example configure function with my changes in mind (thrown together quickly). Used a loop to de-duplicate some of the directory and symlink creation (doing it all together)

```bash
function configure_hypseus() {
    local common_args
    mkRomDir "daphne"
    mkRomDir "daphne/roms"

    addEmulator 0 "$md_id" "daphne" "$md_inst/hypseus.sh %ROM%"
    addSystem "daphne"

    mkUserDir "$md_conf_root/daphne"

    moveConfigDir "$home/.$md_id" "$md_conf_root/daphne"
    local dir
    for dir in ram logs screenshots; do
        mkUserDir "$md_conf_root/daphne/$dir"
        ln -snf "$md_conf_root/daphne/$dir" "$md_inst/$dir"
    done

    copyDefaultConfig "$md_data/hypinput.ini" "$md_conf_root/daphne/hypinput.ini"

    ln -snf "$romdir/daphne/roms" "$md_inst/roms"
    ln -snf "$romdir/daphne/roms" "$md_inst/singe"

    ln -sf "$md_conf_root/daphne/hypinput.ini" "$md_inst/hypinput.ini"

    common_args="-framefile \"\$dir/\$name.txt\" -homedir \"$md_inst\" -fullscreen \$params"

    cat >"$md_inst/hypseus.sh" <<_EOF_
#!/bin/bash
dir="\$1"
name="\${dir##*/}"
name="\${name%.*}"

if [[ -f "\$dir/\$name.commands" ]]; then
    params=\$(<"\$dir/\$name.commands")
fi

if [[ -f "\$dir/\$name.singe" ]]; then
    "$md_inst/hypseus.bin" singe vldp -retropath -script "\$dir/\$name.singe" $common_args
else
    "$md_inst/hypseus.bin" "\$name" vldp $common_args
fi
_EOF_
    chmod +x "$md_inst/hypseus.sh"
    mkdir -p "$md_inst/framefile"
}
```

DirtBagXon:
The main reason I ran the layout previously was because the files linked in _$home/.$md_id_ were not really '_config_'.

However, if you are happy to have the `logs`, `screenshots` and `ram` (_save games_) in _$md_conf_root/daphne/_ then all is good.

Your'e the boss, happy for you to make the changes. I am a little behind on the _helpers.sh_ functions now.

DirtBagXon:
> Sorry I meant `$md_conf_root/daphe/hypseus` for the configuration location I guess - not `$md_conf_root/$md_id` - or even `$md_conf_root/daphne` if there will be no collisions with daphne configs.

I don't foresee any conflict with existing `daphne` configs in doing the above, as the config file name is different, and the other directories, discussed above, are all written locally in the "user writable" _$md_inst_ of the old package. _i.e._ What we tried to move away from here...

Thanks for picking this up again.

joolswills:
Please feel free, or if you prefer, I can commit my changes and if they are ok, I can squash the commits.

joolswills:
> However, if you are happy to have the `logs`, `screenshots` and `ram` (_save games_) in _$md_conf_root/daphne/_ then all is good.

Yeah this should be fine. Thanks.

DirtBagXon:
> if you prefer, I can commit my changes and if they are ok, I can squash the commits.

Go for it.

scriptmodules/emulators/hypseus.sh
```diff
@@ -0,0 +1,95 @@
...
+function configure_hypseus() {
+    mkRomDir "daphne"
+    mkRomDir "daphne/roms"
+
+    addEmulator 0 "$md_id" "daphne" "$md_inst/hypseus.sh %ROM%"
+    addSystem "daphne"
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    mkUserDir "$md_conf_root/daphne"
```
joolswills, line 59: Reviewing my own code, but I think I can lose this line as the moveConfigDir should take care of it.

joolswills:
Sorry if I'm missing something, but I was wondering why the emulator needs the ~/.hypseus folder for logs / ram / screenshots as well as having the `$md_inst` folders with this structure ? I would have thought it would use one or the other ?

joolswills:
I think I misunderstood, and I need to change my patch. The `$home/.hypseus` folder was just to store the configs outside of the install folder ? I was under the impression it was needed to be there for some reason but now I'm thinking it isn't.

DirtBagXon:
> I think I misunderstood, and I need to change my patch. The `$home/.hypseus` folder was just to store the configs outside of the install folder ? I was under the impression it was needed to be there for some reason but now I'm thinking it isn't.

`$home/.hypseus` is the generic 'home' folder for all sub-folders if the `-homedir` argument isn't given:  
https://github.com/DirtBagXon/hypseus-singe#altering-hypseus-or-singe-rom-locations-in-bash-scripts

So provided _scriptmodules/emulators/hypseus.sh_ is run with `-homedir`, technically it isn't needed.

However, as I stated above:
> This aligns with the standard hypseus/daphne layout behaviour in ~/.hypseus when run as a standalone application, and also if hypseus.bin is run outside the RetroPie script arguments. So makes good sense.

So `$home/.hypseus`  would be created if the user ran the the binary within the cli without any arguments, or lacking `-homedir`.

Having a symlink there may then be considered 'tidy' to catch such invocations, but isn't necessary for operations called by an unaltered _scriptmodules/emulators/hypseus.sh_.

joolswills:
Sorry for late reply.

It may be better to just drop -homedir if we are redirecting the default homedir - simplifies it.

Did you test if my changes worked ok - I wasn't sure.

I will give this a test again with my mentioned change, and squash and merge. Sorry again for the delays.

DirtBagXon:
I think in an attempt to not over engineer and simplify this, we just ignore `~/.hypseus` - It's irrelevant really - just my OCD

Removing `-homedir` would just take us back into a symlinking knightmare I think.

I dropped the `moveConfigDir` and just let your changes create the needed writable dirs in `$md_conf_root/daphne`

```
--- hypseus.sh.pr     2021-12-23 09:00:39.704887197 +0000
+++ hypseus.sh  2021-12-23 09:01:20.074802393 +0000
@@ -58,7 +58,6 @@
 
     mkUserDir "$md_conf_root/daphne"
 
-    moveConfigDir "$home/.$md_id" "$md_conf_root/daphne"
     local dir
     for dir in ram logs screenshots; do
         mkUserDir "$md_conf_root/daphne/$dir"
```

Tested this, and all good, with `/opt/retropie/emulators/hypseus/` looking like this:

```
drwxr-xr-x 6 root root    4096 Dec 23 09:20 .
drwxr-xr-x 5 root root    4096 Dec 23 09:20 ..
drwxr-xr-x 2 root root    4096 Dec 23 09:20 fonts
drwxr-xr-x 2 root root    4096 Dec 23 09:20 framefile
lrwxrwxrwx 1 root root      41 Dec 23 09:20 hypinput.ini -> /opt/retropie/configs/daphne/hypinput.ini
-rwxr-xr-x 1 root root 1801072 Dec 23 09:20 hypseus.bin
-rwxr-xr-x 1 root root     528 Dec 23 09:20 hypseus.sh
-rw-r--r-- 1 root root   32472 Dec 23 09:20 LICENSE
lrwxrwxrwx 1 root root      33 Dec 23 09:20 logs -> /opt/retropie/configs/daphne/logs
drwxr-xr-x 3 root root    4096 Dec 23 09:20 pics
lrwxrwxrwx 1 root root      32 Dec 23 09:20 ram -> /opt/retropie/configs/daphne/ram
-rw-r--r-- 1 root root     286 Dec 23 09:20 retropie.pkg
lrwxrwxrwx 1 root root      34 Dec 23 09:20 roms -> /home/pi/RetroPie/roms/daphne/roms
lrwxrwxrwx 1 root root      40 Dec 23 09:20 screenshots -> /opt/retropie/configs/daphne/screenshots
lrwxrwxrwx 1 root root      34 Dec 23 09:20 singe -> /home/pi/RetroPie/roms/daphne/roms
drwxr-xr-x 2 root root    4096 Dec 23 09:20 sound
```

`/opt/retropie/configs/daphne` looking like this:

```
drwxr-xr-x  5 pi pi 4096 Dec 23 09:23 .
drwxr-xr-x 38 pi pi 4096 Dec 23 09:20 ..
-rw-r--r--  1 pi pi   81 Dec 23 09:23 emulators.cfg
-rw-r--r--  1 pi pi 2018 Dec 23 09:20 hypinput.ini
drwxr-xr-x  2 pi pi 4096 Dec 23 09:23 logs
drwxr-xr-x  2 pi pi 4096 Dec 23 09:20 ram
drwxr-xr-x  2 pi pi 4096 Dec 23 09:20 screenshots
```

Everything portable, and writables in `$md_conf_root/daphne` as suggested.

scriptmodules/emulators/hypseus.sh
DirtBagXon: See: https://github.com/RetroPie/RetroPie-Setup/pull/3387#issuecomment-1000170416

DirtBagXon:
Also added `-manymouse` as default Singe argument. This enables ABS mouse input by default which massively improves mouse accuracy (and gives lightgun [DolphinBar/Sinden] support out of the box).

There are no issues in enabling this on non-mouse games, as mouse input is disabled within SDL input in such games.

This seems to be a preferred default option for Retro distros.....

joolswills:
Thanks for your patience with this. I have squash the commits on my local system. I noticed `scriptmodules/emulators/hypseus/hypinput.ini` contains windows CRLF line endings. Is this required by the emulator ?

DirtBagXon:
No, sorry, I try to clean them up when I catch them.

They come over from the multiple systems it's used and developed on. Hypseus will cope with either CRLF or LF.

joolswills:
Thanks. I have converted the file and squashed the commits.

joolswills (approved):

joolswills:
Thanks for the contribution. Binaries are also available for the Raspberry Pi (on Raspberry Pi OS Buster).

PR #3388, opened 2021-08-22
Update system.sh

cameron98:
When installing on Elementary OS Odin, was unable to pass OS check due to missing capitalisation of Elementary. When corrected, install worked fine. I have corrected this in here, please can this change be added.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -275,7 +275,7 @@ function get_os_version() {
             fi
             __os_debian_ver="9"
             ;;
-        elementary)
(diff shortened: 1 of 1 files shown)


cmitu:
A similar test like for the _Neon_ Ubuntu version might be better, since it accomodates both versions - test for `[Ee]lementary` .

joolswills:
This seems familiar. I may have had this change in a branch somewhere (I think I did as @cmitu  did). I'll add it to support both as @cmitu suggests.

PR #3389, opened 2021-08-23
helpers: teach `aptUpdate` to hop releases

cmitu:
When the release of the remote repository is changed, `apt-get` will not update its packages without
`--allow-releaseinfo-change`, so fix that.

This has become problematic due to last week's Debian Bullseye release, which changed the <raspbian.raspberrypi.org> suite from 'stable' to 'oldstable' and prevents `apt` to refresh its index. Upgrading from 4.7.1 (image) to the current version is no longer possible because of this, since `subversion` is not installed on 4.7.1, but it's required now and due to it being updated in the repo, it's not installable.

In the future, we might switch to just using `apt update` (which doesn't have this problem), but there are several other places where `apt-get` is still used, so this change is meant just as a quick fix for the recent suite changes.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -175,7 +175,7 @@ function hasPackage() {
 ## @brief Calls apt-get update (if it has not been called before).
 function aptUpdate() {
     if [[ "$__apt_update" != "1" ]]; then
-        apt-get update
+        apt-get update --allow-releaseinfo-change
         __apt_update="1"
     fi
 }


joolswills:
Thanks. Looks good. I don't want to switch to apt though for reasons apt says in its man.

```
apt provides a high-level commandline interface for the package management system. It is
       intended as an end user interface and enables some options better suited for interactive
       usage by default compared to more specialized APT tools like apt-get(8) and apt-cache(8).
```

I think I read it also wouldn't guarantee parameter changes so best to stick with apt-get etc imho.

PR #3390, opened 2021-08-24
sdl2 - updated to retropie-2.0.20 branch

joolswills:
Added libsamplerate0-dev dependency to module.

Branch includes reworking the patch for kms modesetting via env variables, some minor changes to our Mali and RPI patches,
and some removal of no longer needed changes. It also includes removal of some debian/control build dependencies which were
added to upstream to align with Debian's packaging. The addition causes dependencies like libwayland-dev and libpulse-dev to
be required for all platforms. As we handle adding dependencies in our sdl module code, these have been removed in our
retropie-2.0.16+ sdl2 branch.

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -16,11 +16,11 @@ rp_module_section="depends"
 rp_module_flags=""
 
 function get_ver_sdl2() {
-    echo "2.0.10"
+    echo "2.0.20"
 }
 
 function get_pkg_ver_sdl2() {
-    local ver="$(get_ver_sdl2)+5"
+    local ver="$(get_ver_sdl2)+1"
     isPlatform "rpi" && ver+="rpi"
     isPlatform "mali" && ver+="mali"
     echo "$ver"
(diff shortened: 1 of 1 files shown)


bluestang2006:
@joolswills [`--enable-video-rpi`](https://github.com/RetroPie/SDL/blob/retropie-2.0.16/configure#L1652) and [`--enable-video-kmsdrm`](https://github.com/RetroPie/SDL/blob/retropie-2.0.16/configure#L1679) are now enabled by default in `2.0.16` if you want to save a few lines of code.

cmitu:
For RPI OS 11 (`bullseye`), we'll need to explicitely add `--disable-video-rpi` on non-`dispmanx`/`videocore`, otherwise compilation will fail. Detection for RPI tests just for the `bcm_host` library (which is included), but the `brcm[GLES|EGL]` libraries no longer exist - probably something that should be remedied in SDL's configure test.

bluestang2006:
> For RPI OS 11 (`bullseye`), we'll need to explicitely add `--disable-video-rpi` on non-`dispmanx`/`videocore`, otherwise compilation will fail. Detection for RPI tests just for the `bcm_host` library (which is included), but the `brcm[GLES|EGL]` libraries no longer exist - probably something that should be remedied in SDL's configure script.

Looks like there are some upstream commits that resolve this, but it will require bumping this PR to SDL 2.0.18.

joolswills:
Updated to use our sdl 2.0.18 branch.

joolswills:
Updated to use our sdl 2.0.20 branch.

arturi:
@joolswills thanks for your work! do you have an estimate when this would be released? Trying to install RetroPie on Bullseye and failing with sdl2 compilation errors, `EGL_DISPMANX_WINDOW_T`.

cmitu:
To fix the RPI driver detection on `videocore` platforms, an extra `sed` to fix `configure` works:

``` diff
     if isPlatform "rpi" && [[ -d "/opt/vc/include" ]]; then
         # move proprietary videocore headers
         sed -i -e 's/\"EGL/\"brcmEGL/g' -e 's/\"GLES/\"brcmGLES/g' ./src/video/raspberry/SDL_rpivideo.h
+        sed -i -e 's#<EGL/eglplatform#<brcmEGL/eglplatform#g' configure
         mv /opt/vc/include/EGL /opt/vc/include/brcmEGL

```

joolswills:
I have made the suggested changes. thanks.

joolswills:
Tested working on RPI3/4/Odroid XU4

joolswills:
I am currently building the binaries/debs. I will test updating via binary on a system and then if you are happy also @cmitu I will merge.

joolswills:
Binary install works fine also.

cmitu:
> I will test updating via binary on a system and then if you are happy also @cmitu I will merge.

Yes, should ok to merge.

PR #3395, opened 2021-08-29
Retroarch: Add libretro glsl and slang shaders if platform is not rpi

gizmo98:
The current RetroPie/common-shaders master branch is outdated (last update 2016). Libretro has an own glsl repo with glsl now. Common-shaders has no slang shaders for video drivers glcore and vulkan.

This PR adds three folders in the shader directory if platform is not rpi.
shaders/slang: Contains libretro  slang shaders for drivers glcore/vulkan
shaders/glsl: Contains libretro glsl shaders for driver gl
shaders/retropie: Contains common-shaders branch rpi shaders for driver gl

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -101,6 +101,10 @@ function update_shaders_retroarch() {
             gitPullOrClone "$dir/$shadersystem" "https://github.com/libretro/$shadersystem-shaders.git" "$branch"
             chown -R $user:$user "$dir/$shadersystem"
         done
+        # remove if not git repository for fresh checkout
+        [[ ! -d "$dir/retropie/.git" ]] && rm -rf "$dir/retropie"
+        gitPullOrClone "$dir/retropie" https://github.com/RetroPie/common-shaders.git "rpi"
+        chown -R $user:$user "$dir/retropie"
     fi
 }
 
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -87,11 +87,21 @@ function install_retroarch() {
 function update_shaders_retroarch() {
     local dir="$configdir/all/retroarch/shaders"
     local branch=""
-    isPlatform "rpi" && branch="rpi"
     # remove if not git repository for fresh checkout
     [[ ! -d "$dir/.git" ]] && rm -rf "$dir"
-    gitPullOrClone "$dir" https://github.com/RetroPie/common-shaders.git "$branch"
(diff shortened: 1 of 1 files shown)


joolswills:
I assume this will cause issues though for existing configs - maybe we should put a notice somewhere or at least have a topic on the forum ?

gizmo98:
It will certainly interfere with current configs. So it’s a good idea to add a notice. Our master branch is neglected. I cannot load shaders of folder retropie because all  precision fixes of branch ‚rpi‘ are missing. The other shaders can be loaded but all upstream changes and fixes are missing. Non of the current shaders can be used if video driver is glcore or vulkan.

joolswills:
Yeah - I agree with it in general. Perhaps we should make it so we have a similar folder structure for the rpi also - but obviously this is more config breakage unless we put it elsewhere and symlink.

gizmo98:
Really ugly solution:
Keep current shader folder and add folders glsl and slang. Remove content of shaders/retropie folder (these shaders are not usefull without fixes of branch common-shaders: rpi ). Just fill the three folders and keep the rest as it is.

barbudreadmon:
What about pi4 ? I think this PR doesn't allow usage of slang shaders on pi4, while it should also support vulkan & glcore (note that glcore won't be forced on gles platforms).

Note that glcore should probably always be prefered over gl on platforms compatible with both : from the tests i did last year it was slightly faster due to less retroarch overhead.

cmitu:
> What about pi4 ? I think this PR doesn't allow usage of slang shaders on pi4, while it should also support vulkan & glcore (note that glcore won't be forced on gles platforms).

Isn't `glcore` for OpenGL 3+ devices ? The Pi4 has only OpenGL 2.1 support, so it wouldn't work with it. 

For `vulkan`, this will probably be available on the next RPI OS release (based on Debian 11 _bullseye_) and we can probably add a (new) ` vulkan` flag to pull in the `slang` shaders for the platforms supporting it.

joolswills:
I'm not going to implement the PR as is anyway most likely. Just was referencing it. I've not yet decided how I will implement it.

barbudreadmon:
@cmitu `glcore` is kinda misnamed, it should have been `moderngl` or something, in the case of gles platforms i think RA's `glcore` video driver is associated with gles3 capabilities.

gizmo98:
@barbudreadmon does it make sense to enable glcore for Pi4? Rpi optimized glsl shaders will not work by default.

Should we test if glcore + rpi optimized ported slang shaders run better than gl + rpi optimized shaders?

barbudreadmon:
@gizmo98 you should probably give it a try

PR #3397, opened 2021-08-31
Skyscraper - fix critical dependency error (Debian 11+, Ubuntu 21.04+)

cobalt2727:
qt5 has been removed in all modern versions of Debian and Ubuntu, and the skyscraper script will not work at all without these changes - this dependency change should not break older versions of these operating systems and the package should install fine.
https://askubuntu.com/a/1335187

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -28,7 +28,7 @@ function sources_skyscraper() {
 }
 
 function build_skyscraper() {
-    qmake
+    QT_SELECT=5 qmake
     make
     md_ret_require="$md_build/Skyscraper"
 }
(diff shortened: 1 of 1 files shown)


cmitu:
The dependencies look ok, but the building might fail because `qt5` is not the default version and `qmake` doesn't know which version to choose:

```
~/RetroPie-Setup/tmp/build/skyscraper ~/RetroPie-Setup

Building 'skyscraper' : Scraper for EmulationStation by Lars Muldjord

qmake: could not exec '/usr/lib/x86_64-linux-gnu/qt4/bin/qmake': No such file or directory
make: *** No targets specified and no makefile found.  Stop.
```

This can be fixed in the build function with
``` sh

function build_skyscraper() {
    QT_SELECT=5 qmake
    make
    md_ret_require="$md_build/Skyscraper"
}
```
Can you add these changes to your PR ?

cobalt2727:
Done

joolswills:
Thanks.

PR #3399, opened 2021-09-08
uqm - rework module to build from source (and update to 0.8.0)

joolswills:
This avoids the messy logic for missing Raspberry Pi Buster packages, and simplifies future updates.


joolswills:
I'm going to merge this. It's an experimental module anyway but the way it operated with building debs on buster was a hack and made for issues with the package versioning code.

Works fine on my rpi4 and uses SDL2 now so should be more compatible.

PR #3401, opened 2021-09-21
Add libretro module script for TIC-80

LNRC:
This script was developed with assistance from mods and forum users on Retropie, full forum thread at https://retropie.org.uk/forum/topic/31366/tic-80-retroarch-lr-tic80-core-installation-script?_=1632144270795
The script is easy to install, the emulator runs well, and .tic/.zip files work without issues.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -220,6 +220,9 @@ solarus_fullname="Solarus Engine"
 ti99_exts=".ctg"
 ti99_fullname="TI99"
 
+tic80_exts=".tic .zip"
+tic80_fullname="TIC-80"
+
 trs-80_exts=".dsk"
 trs-80_fullname="trs-80"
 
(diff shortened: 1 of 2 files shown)


joolswills:
Thanks. Will test, but looks good.

joolswills:
Works well thanks. Interesting project. Might be nice to have a module for the standalone binary and the tools too etc.

I made a couple of minor changes. 

Wording in desc, moved extensions to `platforms.cfg` and removal of the build folder to force a clean build.

joolswills:
Will merge after testing in my chroot build environment. Thanks.

joolswills:
Removal of the build folder caused an issue, as I see the project has a build folder included that is needed. Would be better to build outside of existing folders which I am testing.

joolswills:
Seems to be ok using a different folder to build in. Merging. Binaries for RPI on Raspberry Pi OS Buster (base RetroPie image) will be available shortly.

joolswills:
Thanks.

PR #3402, opened 2021-09-21
Add uae4arm core support

Chips-fr:
uae4arm is now available as a core with quite features (save states, virtual keyboard, cd support, hd support, whdload slave support, floppy disk switching ...)
I tested adf, ipf (not sure where the lib was coming from but it was working), cd32 (cdtv should be supported too), whdload (lha) and all are ok.
I didn't delete the old uae4arm.sh but could be a good idea.
I modified system.sh since it was strange that asflags didn't include cpu flags. cpu flags are needed on raspbian for asm since neon support is not in the default gcc target.
I tried only on my pi 4. Haven't try others configuration (like aarch64) on retropie.
I was not sure about  rp_module* .


joolswills:
Thanks. If you would prefer the standalone version gone I can sort that after we include this. Currently there is code shared between uae4arm and amiberry as we included amiberry later and there is shared functionality.

Will give this a look and review. 

Regarding flags. Previously we didn't separate CPU flags from cflags and asflags was unset. We can now add them and your change seems reasonable but is it required for your libretro core or you just noticed it was missing?

Most makefiles handle the assembler opts for things like neon specific ASM so it hasn't been an issue. But it makes sense to include it but will need to build everything after that change to make sure.

Chips-fr:
You can keep the standalone, i just feel like that cores deliver a better user experience than standalone for gaming so keeping standalone seems not interesting for the long term.
Without the asflags modification in system.sh this core will not compile on raspbian due to the raspbian default target  (ARM11 so without neon).

joolswills:
Thanks for the info. I think most of the time libretro cores with neon asm pass the cflags to AS so this wasn't needed. But it makes sense now we have a separate variable for just the CPU flags to include it here.

joolswills:
Thanks. Merging. Would be good if you can add a LICENSE file if it's GPL2/3 etc.

PR #3403, opened 2021-09-24
Add Meritous v1.2 game

TurBoss:
Builds meritous from source
includes fix to handle buttons in gamepad

Captura de pantalla de 2021-09-24 16-29-19


joolswills:
Thanks. Will take a look.

TurBoss:
Hello, I was taking a look at the port but I was unable to figure how can I make it run in fullscreen

its 640x480 and its begin shown in a corner


how can I make it run at full screen by default, I tried setting the `fullscreen` arg but nothing..


Thanks

PR #3404, opened 2021-09-27
Add defaultRAConfig to replace ensureSystemretroconfig

joolswills:
Removed deprecated and unused shader parameter support from ensureSystemretroconfig.

defaultRAConfig takes a single parameter for the system name, but also allows additional
parameter pairs to be added to set default ini configuration key/values

eg. defaultRAConfig "dreamcast" "fps_show" "true"

It also uses $md_conf_root as a base, which ensureSystemretroconfig didn't handle, so we save and reset this for backward compatibility for anything still using ensureSystemretroconfig


joolswills:
Have renamed the function to `defaultRAConfig`.

PR #3406, opened 2021-09-29
duckstation: preset an extra option, added license link

cmitu:
Set the Memcard 1 save as "Save Ram (libretro .srm)" instead of the native "<game>_1.mcd" file. Should make easier consistent with other libretro cores behavior.

License link updated, taken from https://github.com/stenzek/duckstation#libretro-core.

### DIFF
--- a/scriptmodules/libretrocores/lr-duckstation.sh
+++ b/scriptmodules/libretrocores/lr-duckstation.sh
@@ -11,8 +11,8 @@
 
 rp_module_id="lr-duckstation"
 rp_module_desc="PlayStation emulator - Duckstation for libretro"
-rp_module_help="ROM Extensions: .exe .cue .bin .chd .psf .m3u .pbp\n\nCopy your PlayStation roms to $romdir/psx\n\nCopy compatble BIOS files to $biosdir"
-rp_module_licence="PROP"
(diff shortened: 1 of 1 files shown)


dankcushions:
> PGXP is set to disabled when using SW rendering, otherwise a warning is shown (and PGXP is disabled anyway).
This option doesn't seem to be exposed through the RetroArch's menu.

just tried on pi4 with latest binary and a clean core options file (ie, only `duckstation_GPU.Renderer = "Software"` set as it would be during install), and no warning displayed or anything in the (verbose) log (and it indeed defaults `duckstation_GPU.PGXPCPU = "false"` after).

rest looks good!

cmitu:
@dankcushions thanks - you're right, starting with the SW renderer as default, `duckstation_GPU.PGXPCPU = "false"` is set automatically. I removed the option, since it shouldn't be needed.

Did a few tests and if the PGPXP option becomes visible only when using a HW GPU renderer. If enabled, it stays enabled when switching back to the SW GPU renderer and it's no longer visible/editable from the options menu. Tested with v1.9.8 and latest core (where the new `Option Categories` are enabled and the core seems to implement them).

joolswills:
Thanks.

PR #3407, opened 2021-09-29
add forced modes to analog_dpad_mode

hlad:
Since v1.9.6 RetroArch changes the way the "analog to digital" works and since then the "forced mode" is need for some cores (eg. lr-vice, lr-puae).

#https://github.com/libretro/RetroArch/pull/12562


joolswills:
Thanks.

PR #3412, opened 2021-10-09
yquake2: update to v8.00 and add add-ons game source

cmitu:
Changes included:
* updated the Quake2 client to 8.00
* switched building to `make`, since the `cmake` method is no longer recommended (see https://github.com/yquake2/yquake2/commit/fb1a2b0ce7b249cd50990225cc484dc2a9f965de)
* build the add-ons sources from the same project.
* (minor) don't run the cleanup in the `quake2` port folder when the shareware version is not downloaded

Notes:
* the version of the package is dependant on the main YQuake2 repo. If the `get_repo_version` from `mupen64plus` (which features the same multi-repository approach) is added as a helper, we might be able to use it here to detected when one of the repos gets a new release. For now, the main repo is the one of interest, so I think it's ok to use it for update detection.
* the add-ons (rogue/xatrix) seem to run without the added `game.so` files, but withouth it there's lots of errors logged. Errors don't appear when the `game.so` is found and loaded.

Supercedes: #3331

### DIFF
--- a/scriptmodules/ports/yquake2.sh
+++ b/scriptmodules/ports/yquake2.sh
@@ -12,7 +12,7 @@
 rp_module_id="yquake2"
 rp_module_desc="yquake2 - The Yamagi Quake II client"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/yquake2/yquake2/master/LICENSE"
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks for this btw - apart from the comments above, it looks good to me.

PR #3416, opened 2021-10-14
mame: update QT5 dependency list

cmitu:
Similar to #3397, replace 'qt5-default' which is no longer present on current Debian/Ubuntu releases.


joolswills:
Thanks.

PR #3418, opened 2021-10-23
CDogs-SDL 1.1.1

IoriBranford:
New features, content and fixes since 0.7.2: https://github.com/cxong/cdogs-sdl/releases

I keep the old one as cdogs-sdl-0.7.2 in case there's a need for it. If not feel free to leave it out.


joolswills:
Thanks. GTK is only needed for the editor - so maybe best excluded unless out, so

joolswills:
gtk 3 is only needed for the editor os dialogs I think so it maybe worth excluding the editor unless building on a platform that uses the desktop. Seems to compile without the protobuf dependencies and the code includes a nanopb library so not sure if that's required unless developing or something - will double check.

PR #3419, opened 2021-10-26
audiosettings: support the ARM-side HDMI driver

cmitu:
Needed for configurations using the 'vc4-kms-v3d' overlay on Pi3/Pi4.
The names are (almost) similar, but the default order of the cards is different:

 * (fkms) the order is HDMI 1[+2], Headphones
 * (kms)  the order is Headphones, HDMI 0[1] (!)

We may have to add a function to set the HDMI0 port as default, let's wait for Bullseye to land and check then.

### DIFF
--- a/scriptmodules/supplementary/audiosettings.sh
+++ b/scriptmodules/supplementary/audiosettings.sh
@@ -40,7 +40,24 @@ function gui_audiosettings() {
     fi
 }
 
+function _reset_alsa_audiosettings() {
+    /etc/init.d/alsa-utils reset
+    alsactl store
+    rm -f "$home/.asoundrc" "/etc/alsa/conf.d/99-retropie.conf"
+    printMsgs "dialog" "Audio settings reset to defaults"
+}
+
+function _move_old_config_audiosettings() {
+    if [[ -f "$home/.asoundrc" && ! -f "/etc/alsa/conf.d/99-retropie.conf" ]]; then
+        if dialog --yesno "The ALSA audio configuration for RetroPie has moved from $home/.asoundrc to /etc/alsa/conf.d/99-retropie.conf\n\nYou have a configuration in $home/.asoundrc - do you want to move it to the new location? If $home/.asoundrc contains your own changes you should choose 'No'." 20 76 2>&1 >/dev/tty; then
+            mv "$home/.asoundrc" "/etc/alsa/conf.d/"
+        fi
+    fi
+}
+
 function _bcm2835_alsa_compat_audiosettings() {
+    _move_old_config_audiosettings
+
     local cmd=(dialog --backtitle "$__backtitle" --menu "Set audio output (ALSA - compat)" 22 86 16)
     local hdmi="HDMI"
 
@@ -89,10 +106,7 @@ function _bcm2835_alsa_compat_audiosettings() {
                 alsactl store
                 ;;
             R)
-                /etc/init.d/alsa-utils reset
-                alsactl store
-                rm -f "$home/.asoundrc"
-                printMsgs "dialog" "Audio settings reset to defaults"
+                _reset_alsa_audiosettings
                 ;;
             P)
                 _toggle_pulseaudio_audiosettings "on"
@@ -103,6 +117,8 @@ function _bcm2835_alsa_compat_audiosettings() {
 }
 
 function _bcm2835_alsa_internal_audiosettings() {
+    _move_old_config_audiosettings
+
     local cmd=(dialog --backtitle "$__backtitle" --menu "Set audio output (ALSA)" 22 86 16)
     local options=()
     local card_index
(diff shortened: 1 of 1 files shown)


cmitu:
Needs an adjustment for Pi3 et. al now that 'vc4-kms-v3d' is default all Pi models.

cmitu:
Fixed the display and detection for models < 4 when they use the new `vc4-kms-v3d` overlay.

cmitu:
Added a slightly different `.asoundrc` variant when `vc4hdmi` driver is used for audio (enabled by the `vc4-kms-v3d` overlay), since the PCM configuration is different - which allows aplications to use the HDMI output without any conversion.

From my testing, the HDMI cards with the new driver will be configured only when audio is detected from HDMI connection, that's why my initial testing on the Pi4 failed (I only have video plugged in).

I'll do some more testing with the new `vc4hdmi` driver on the Pi4 and a HDMI (with audio) connection, just to make sure that the correct output is detected and enabled.

cmitu:
Did some testing on the PI4 and the audio output is fine - switching to Headphones or HDMI[1|2] is working correctly.

What it doesn't work - and I don't think it's something that can be done here - is the mixer control. Strangely it works on a Pi3 (which has only one HDMI card), but not on the Pi4.

joolswills:
@cmitu does this need more changes?

cmitu:
I've been meaning to pick this up again. I'll see if the mixer can be enabled on the Pi4, which is the only thing missing (can't control the volume of the sound card).

joolswills:
@cmitu thanks. No pressure, I was just going through the tickets. Cheers.

joolswills:
Works well for me. For hdmi volume can be controlled on the hdmi device I guess. I'm happy to include this now and can always adjust it later if needed.

joolswills:
Es will complain on stdout if there's no mixer but maybe we can adjust that and just log it instead.

cmitu:
> I'm happy to include this now and can always adjust it later if needed.

I guess it's better than nothing (which is what we have at the moment) :). I have a question opened in the forums, let me see if we can get some info out of it. If nothing conclusive comes up, we can merge it.

joolswills:
Sounds good. Thanks.

cmitu:
@joolswills since the default config doesn't implement the volume control properly, I've added it to the generated `$HOME/.asoundrc`. Tested on Pi3/Pi4 (on _bullseye_, using the VC4 HDMI ARM driver).

This should get rid of EmulationStation errors, the new volume control has the same name (HDMI), it may help the transition of existing configs.

Still needed to be considered:
 * `dmix` is not present, this will brake multiple applications accessing the card
 * defaults for new systems/images. HDMI was the default card before the the VC4 driver was used for HDMI, now it's the analog output (Headphones). Additional configs need to be added on new images if we want to default to HDMI.

joolswills:
Thanks. I can test on a pi 400.

joolswills:
@cmitu Not tested on a RPI400 yet sorry. I was just testing it on a rpi4, and the HDMI volume control works well.

I think we should handle this system-wide via a `/etc/alsa/conf.d/99-retropie.conf`. I think it will be easier to manage, and will allow users to have an additional custom `.asoundrc` if they want. 

I tested doing it like this, and it seemed to work fine.

cmitu:
Using a sistem wide `.conf` sounds ok, you can include any changes to add the functionality.

scriptmodules/supplementary/audiosettings.sh
```diff
@@ -40,6 +40,13 @@ function gui_audiosettings() {
     fi
 }
 
+function _reset_alsa_audiosettings() {
+    /etc/init.d/alsa-utils reset
+    alsactl store
+    rm -f "$home/.soundrc" "/etc/alsa/conf.d/99-retropie.conf"
```
cmitu, line 46: `rm -f "$home/.soundrc"` - should be `rm -f "$home/.asoundrc"`

joolswills:
I added a commit with a requester to ask the user if they want to move their ALSA config. It's not ideal but there's no markers in the file around the RetroPie config etc.

Before if a user had a custom ~/.asoundrc it would be overwritten if switching output before, so this is still overall better I think.

scriptmodules/supplementary/audiosettings.sh
```diff
@@ -89,9 +106,7 @@ function _bcm2835_alsa_compat_audiosettings() {
                 alsactl store
                 ;;
             R)
-                /etc/init.d/alsa-utils reset
-                alsactl store
-                rm -f "$home/.asoundrc"
+                _reset_alsa_audiosettings
                 printMsgs "dialog" "Audio settings reset to defaults"
```
cmitu, line 110: `printMsgs` is redundant here, it's included in the `_reset_alsa_audiosettings`.

cmitu:
> Before if a user had a custom ~/.asoundrc it would be overwritten if switching output before, so this is still an improvement I think.

Yes, seems right. I have just one more observation with the changes, but I'll give it another test before merging.

joolswills:
Thanks. I'll fix it up.and review. I'll squash the last two commits I think.

joolswills:
I squashed the commits keeping your main change and the other changes as a separate commit.

cmitu:
Looks fine to merge.
One minor punctuation correction maybe - the text in the _move file_ has a `.` after `?`, maybe a leftover from when the phrase was not a question.

> You have a configuration in $home/.asoundrc - do you want to move it to the new location?. If $home/.asoundrc contains...

joolswills:
I'll correct that, then merge. Thank you.

PR #3421, opened 2021-10-29
Add cloud sync utility for RetroArch and Ports saves

ridgekuhn:
Hello,

I've written a cloud sync utility using [rclone](https://rclone.org/), systemd, `inotifywait`, and bash. It registers save directories for RetroArch (based on the user's `retroarch.cfg`) and allows the user to sync their saves with an rclone remote either manually or automatically in the background. It also includes support for a few ports, the standalone ppsspp emulator, and multiple builds of RetroArch (ie, if the user is on a 64-bit distro and has both 64-bit and 32-bit builds of RetroArch). I've tried to make it easy to extend, so more ports and standalones should be supported in the future. There is also a version for [ArkOS](https://github.com/ridgekuhn/arklone-arkos).

Video here: https://youtu.be/-jmoS1xwVcI

I couldn't find any documentation on writing a RetroPie module, so I've taken my best guess and would appreciate feedback on this if it seems like a good candidate to be merged in. Any issues with the app itself should go to [the project repo](https://github.com/ridgekuhn/arklone-retropie/).

The only issue I've encountered so far is that the RetroArch build included with the 4.7.1 Raspberry Pi 2/3 image does not respect the `sort_savefiles_by_content_enable` and `sort_savestates_by_content_enable` settings in `retroarch.cfg`. (Building from source or downloading the updated binary works as expected.) Is there any way to get the packages script to update RetroArch while installing this module?

Thanks!

### DIFF
--- a/scriptmodules/supplementary/arklone.sh
+++ b/scriptmodules/supplementary/arklone.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="arklone"
+rp_module_desc="arklone cloud sync utility by ridgek"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/ridgekuhn/arklone-retropie/master/LICENSE.md"
+rp_module_repo="git https://github.com/ridgekuhn/arklone-retropie master"
+rp_module_section="opt"
+
+function sources_arklone() {
+    gitPullOrClone
+}
+
+function install_arklone() {
+    md_ret_files=(
+        'src'
+        'LICENSE.md'
+        'README.md'
+        'install.sh'
+        'uninstall.sh'
+    )
+}
+
+function configure_arklone() {
+    if [[ "${md_mode}" = "install" ]]; then
+        # Run install script
+        "${md_inst}/install.sh"
+
+        # Link to EmulationStation RetroPie menu
+        touch "${home}/RetroPie/retropiemenu/arklone.rp"
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


joolswills:
Sorry but I don't want to include this into core RetroPie but you can include it as an external module via the ext/ folder and provide users with instructions to install etc.

ridgekuhn:
Understood. I would appreciate your negative feedback if you have any.

joolswills:
Well - if we included something like this, I would prefer it to be part of RetroPie-Setup than we maintain rather than an external project. But you can provide instructions for users to install it and use it as needed as an external scriptmodule.

ridgekuhn:
That makes sense. Given some time and a little guidance on how to integrate it properly into the RetroPie module ecosystem, I wouldn’t mind eventually refactoring it for that purpose since that was basically my intent anyway; if that’s something you’re interested in?

One question I have either way, can you point me to some functions for determining which ports are installed? Is there a way to get a list of all of them and their status? Arklone can generate watchers programmatically (and does so for any combination of savefile/savestate settings in the user’s retroarch.cfg), but I wasn’t sure how to implement an interface for automatically doing it for ports and ended up just hardcoding a bunch of common ones. I’m not super happy with that workaround since it means watchers end up getting registered with systemd whether or not the watched path actually exists!

Finally, sorry for this question, but I’m away from my Pi and dev machine for a few days. Will this module script work as-is if placed in the ext/ dir, or does it need different settings defined in the config vars?

joolswills:
I think it's something I would consider, although right now I'm looking at bugfixing and existing PRs before our next image. But it's definitely useful.

All installed ports should be in `/opt/retropie/ports` and the folder name will be the script module ID. However I intend to add to the admin modules so you can query the script module meta data from the command line.

Regarding the `ext` folder. It expects repositories to be checked out there that has the same structure as `RetroPie-Setup`. Eg `ext/YOUR_REPO` that contains `scriptmodules/supplementary/YOUR_MODULE.sh`

There's no facility to update from RetroPie-Setup, or manage external repositories (yet) , but it should handle modules there and list them separately after our modules in the /Manage Packages/ sections. And it will handle installing/uninstalling of external modules.

ridgekuhn:
Got it, that’s all super useful info, thanks so much!

PR #3423, opened 2021-11-02
joy2key: fix a couple of bugs in the SDL version

cmitu:
Added a couple of fixes for issues discovered tracking the errors reported in a forum post [1].

* When a device is disconnected, its events were not cleared. 
   This was problematic for joystick/d-pad movement tracking, which keeps the last move event in the event queue. Disconnecting a device while the movement is active would keep this last move in the event queue forever, so clear all events associated with a device when it disconnects

*  For certain joysticks (axis), SDL will generate spurious axis movement events during initialization.
It seems that for axis where `absmin` and `absmax` are asymmetrical (ex. 0-255 instead of -255 - 255), SDL generates extra events which are not user generated. This can causes issues - i.e. automatically trigger the `runcommand` menu without user interaction, unintended scrolling, etc.

     As a workaround, discard any axis movement events _immediately_ after the device is added.
**NOTE**: this will discard any potential valid axis events from another (already connected) device if they are present, forcing the user to probably re-center/re-position the (other) joystick they were using.

[1] Forum post - https://retropie.org.uk/forum/topic/31569/

### DIFF
--- a/scriptmodules/admin/joy2key/joy2key_sdl.py
+++ b/scriptmodules/admin/joy2key/joy2key_sdl.py
@@ -255,7 +255,7 @@ def filter_active_events(event_queue: dict) -> list:
     for e in event_queue:
         if event_queue[e][0] is None:
             continue
-        
+
         last_fire_time = event_queue[e][2]
         repeat_count = event_queue[e][1]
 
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks

PR #3424, opened 2021-11-02
mupen64plus: enable GLideN64 plugin option EnableInaccurateTextureCoordinates for rpi.

dankcushions:
Massively improves framerate at cost of accuracy.

Tested on Pi 4, GoldenEye menu:
True = 47 FPS
False = 33 FPS

The accuracy cost is essentially the last few months of texture handling improvements. These do fix some long standing display/accuracy bugs, but the framerate cost is so prohibitive on Pi that I doubt there would be cause to use them. Certainly not by default.


joolswills:
Thanks.

PR #3425, opened 2021-11-07
bluetooth: update python scripts and deps for python3

cmitu:
Updated the dependencies and scripts for python3:
    
  * `python-dbus` no longer exists in Debian 11 'bullseye' or Ubuntu > 20.04, installl the `python3` version
  * `python-gobject` has been superseeded by `python(3)-gi`, the current package is just a transitional package that pulls
    `python-gi` and the old `python-gobject-2` (deprecated). Update the dependencies and the scripts to use the new package.
  * set python3 as interpreter for the helper scripts. Fixed the dict syntax for getting items.

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -38,7 +38,7 @@ function _get_connect_mode() {
 }
 
 function depends_bluetooth() {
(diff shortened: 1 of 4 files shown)


cmitu:
This needs a bit of testing on `bullseye`, I'll set it back to draft.

joolswills:
thanks.

PR #3429, opened 2021-11-19
Add advanced wifi module with on screen keyboard

OfficialPhilcomm:
A lot of people (including me) have the problem that you always need a keyboard to connect to a new wifi. Or another device to ssh into the Pi.

This aims to fix this problem.

Built with `ruby` and `curses`, this terminal application allows for connecting to a wifi, without the need of a keyboard.
Since controller buttons are mapped to the arrow keys, enter and space, this app reads them like keyboard inputs for navigation.

This pr just includes the RetroPie-Setup `.sh` file, feel free to `wget -P /home/pi/RetroPie-Setup/scriptmodules/supplementary/ "https://raw.githubusercontent.com/OfficialPhilcomm/retropie-wifi-manager/master/advanced-wifi.sh"
` and try it yourself. I will continue to make improvements on the project's repo, so this can be merged already.

The workflow is as follows:

## Select a WiFi
Screenshot 2021-11-19 at 17 33 44
The app has a simple wifi selection screen. All networks found with `iwlist scan` are listed here.

White networks are unknown to the device, cyan ones are saved in `wpa_supplicant.conf`, green ones you are connected to right now.

For white networks you only have the option to connect.
Green and Cyan additionally have the functionality to be prioritized (`priority=1` in `wpa_supplicant.conf`) and to be deleted. These options come with a confirmation prompt.

## Enter password
Screenshot 2021-11-19 at 15 52 49

On this page you can enter the password. The password must be at least 8 characters long, otherwise it shows an error prompt.

## Saving network config
After clicking on connect, the app writes to the `wpa_supplicant.conf`. It uses `wpa_passphrase` to generate a hashed password, so the one written to the file is not in clear text.

### DIFF
--- a/scriptmodules/supplementary/advanced-wifi.sh
+++ b/scriptmodules/supplementary/advanced-wifi.sh
@@ -18,7 +18,7 @@ rp_module_repo="git https://github.com/OfficialPhilcomm/retropie-wifi-manager.gi
 rp_module_licence="MIT https://raw.githubusercontent.com/OfficialPhilcomm/retropie-wifi-manager/master/LICENSE.md"
 
 function depends_advanced-wifi() {
-  getDepends ruby ruby-dev
+  getDepends ruby ruby-dev libncurses-dev
 }
 
 function sources_advanced-wifi() {


cmitu:
Thanks for submitting this. The scriptmodule needs some modifications if it were to be accepted, but right now I was interested in seeing how it works.

I tried to give it a test, but I get an error when running the `wifi2` entry from the RetroPie system:
```
Traceback (most recent call last):
	2: from wifi2.rb:2:in `<main>'
	1: from /usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb:85:in `require'
/usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb:85:in `require': cannot load such file -- require_all (LoadError)
```

What `ruby` version does it need to run ? Default in RPI OS is `ruby2.7` (_bullseye_) or 2.5 (_buster).

OfficialPhilcomm:
> Thanks for submitting this. The scriptmodule needs some modifications if it were to be accepted, but right now I was interested in seeing how it works.
> 
> I tried to give it a test, but I get an error when running the `wifi2` entry from the RetroPie system:
> 
> ```
> Traceback (most recent call last):
> 	2: from wifi2.rb:2:in `<main>'
> 	1: from /usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb:85:in `require'
> /usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb:85:in `require': cannot load such file -- require_all (LoadError)
> ```
> 
> What `ruby` version does it need to run ? Default in RPI OS is `2.7` (_bullseye_) or `2.5` (_buster_).

Yes, sorry, thanks for pointing that out. The scriptmodule doesn't install the `require_all` gem, completely forgot that.

For now, as a hotfix, just run `sudo gem install require_all`, I will add this to the scriptmodule asap.

What would need to be changed though?

cmitu:
> Yes, sorry, thanks for pointing that out. The scriptmodule doesn't install the `require_all` gem, completely forgot that.

Ok, that did the trick, the interface starts now. I've done some testing, here's what I've noticed:

**Bugs**
* there's a set of warnings thrown by `ruby` when running, the last line appears on all screens at the bottom of the screen:
  >    wifi2/lib/screens/wifi_selection_screen.rb:110: warning: rb_safe_level will be removed in Ruby 3.0
 
  This is especially visible if you do any changes and the screen needs to refresh, there's a scrolling list of warnings that appear but they're overwritten by the on-screen window. This makes the operation tedious, since each screen refresh shows the errors again that and adds about 2 sec of delay for each move - took me some time to add a simple pass. I guess this can be handled in code or just silenced with `2 > /dev/null` (you can run it with `2 > logfile.txt` to see the warnings).

* the list of APs SSIDs is not showing all the APs present. If I use the current _Wifi_ module and look at the current visible APs, I see about 5 entries, the script only shows me the current connected AP. Maybe look into how the `wifi` module gets the list of networks and repurpose it for your script ?

* I'm not familiar with how _priority_ works with _wpa_supplicant_, but it didn't work for me. I had a previously added network (no priority, just SSID/PSK added to the `.conf`) then I added a new one, but after saving it and marked it as priority, it didn't connect to the new network (even after a reboot).

* the password prompt when entering the passphrase is mis-spelled as _Passowrd_
* the list of available characters doesn't cover the range of allowed by WPA (not sure about WEP, but we're leaving that to the past). I can't find a canonical source, just https://en.wikipedia.org/wiki/Wi-Fi_Protected_Access#cite_note-21 as a reference. The list of characters is missing for instance <kbd>Space</kbd> or `[`, `@`, `<`, `>`, etc. Not sure if `§` is a valid character in a WPA passphrase 

**Missing**

* I can't add a network not present in the scanned list. Not a big thing, but some users might have a hidden ESSID or may want to add the network without being there.
* It should be possible to add the network without a passphrase (public networks). I think this is possible in the current `wifi` and if this module aims to replace it, it's maybe useful to add the ability to add un-secured/public networks. 

**UI/UX**

Now, most of these are mostly subjective - so they should be taken with a grain of salt, but I think there's should be some consistency between the existing menus in RetroPie and this script. I've listed some of them below.

* During startup, it should print something like 'Scanning for available networks' instead of an empty screen. Scanning can take some time.

* The currently selected character/string should be made more evident. Right now they are emphasized with Red color, but I think maybe the BG also would need to change to make it more obvious where the character is.
  The same for all the UI elements, I don't know if it's possible to make the screens follow the same color scheme as the one used by `dialog` in other RetroPie menus, but it would be nice if it did.

* There's no 'cancel'/'exit' button on the startup screen - the 'B' button will close the dialog, but I think it needs something next to 'Connect' to make it obvious how to exit.
* The layout should be centered and the screen areas should be separated a bit. They have some separation now, but for instance the _Shift_ and _Connect/Exit_ should not be on the same line.
* If I enter less than 8 chars, the next screen tells me about the error, but I cannot go back to fill the rest or change the password. If 8 chars is minimum, then perhaps warn the user and don't progress to the next screen or maybe go back to fill in the rest of the password ?

**Scriptmodule** related

> Out of curiosity, what would need to be changed?

* What's the purpose of moving the files from the cloned repo to another folder (`/opt/dev_philcomm`) ? The scriptmodule should be under `/opt/retropie/supplementary`, there's no need to duplicate it into another folder.
* the `noinstclean` flag is not needed, `nobin` I think is ok.
  You can just use the `sources_..` function to clone it to `$md_inst` and have `install` just install the needed gems.
* use the `rp_module_repo` variable to set the repository and branch. You can still call `gitPullOrClone` for installation from `sources_..`, and adding it there is useful for the RetroPie-Setup script to check for updates and don't re-install when needed.
* `ruby-dev` is not strictly neede and the `ncurses` gem is packaged as `ruby-curses`. Maybe look to see if `require_all` may be packaged, so there wouldn't be a need to run a `gem install`.
  You can also just run `getDepends <list-of-packages>`, there's no need for an additional variable to hold the list (it's useful when you have to conditionally add a package).
* (optional) add a license and a link to it like other modules, even if minimal. Makes it look nice on https://retropie.org.uk/stats/licences/.
* the module should be available only for RPI so add `rpi` to the flags. Ubuntu uses `network-manager`, while Debian I think it has a different `ctrl_interface` configured for its `wpa_supplicant.conf` service.
* use the `configure` function to add the RetroPie menu entry or remove it. In `configure`, you have the `$md_mode` variable to know whether the package is installed or not. Don't use the `install` function for it, it's not the place there.
* there's no need for a `remove` function, the setup script will take care of removing the install folder (`$md_inst`) and `configure` will be called also (where you can remove the menu entry, as indicated above).

Now, there's some considerations about the naming and the menu entry - whether we can include it as an alternative to the current wifi module or just integrate into it, but I think for now it's enough.

OfficialPhilcomm:
> the list of APs SSIDs is not showing all the APs present. If I use the current Wifi module and look at the current visible APs, I see about 5 entries, the script only shows me the current connected AP. Maybe look into how the wifi module gets the list of networks and repurpose it for your script ?

`iwlist`, which is used to retrieve all the cells, needs sudo. This is a system thing, I cannot change that. Without sudo, it just lists the current connected one.

> I'm not familiar with how priority works with wpa_supplicant, but it didn't work for me. I had a previously added network (no priority, just SSID/PSK added to the .conf) then I added a new one, but after saving it and marked it as priority, it didn't connect to the new network (even after a reboot).

A network inside `wpa_supplicant.conf` can have `priority=1,2,3,4...`. The wifi chip tries to connect in order from the highest to lowest priority. Sometimes it can be that the prioritized network is not available for some reason, but that is a wifi chip thing rather than my code. On a Pi Zero 2 it works like 80% of the time.

> the password prompt when entering the passphrase is mis-spelled as Passowrd

:facepalm: Thank you :heart:

> the list of available characters doesn't cover the range of allowed by WPA (not sure about WEP, but we're leaving that to the past). I can't find a canonical source, just https://en.wikipedia.org/wiki/Wi-Fi_Protected_Access#cite_note-21 as a reference. The list of characters is missing for instance Space or [, @, <, >, etc. Not sure if § is a valid character in a WPA passphrase

I wasn't sure which characters are allowed too. The way I built the keyboard module allows for easy adding of characters, so that can easily be fixed.

> I can't add a network not present in the scanned list. Not a big thing, but some users might have a hidden ESSID or may want to add the network without being there.

Good point, I will add that.

> It should be possible to add the network without a passphrase (public networks). I think this is possible in the current wifi and if this module aims to replace it, it's maybe useful to add the ability to add un-secured/public networks.

That might be tricky, but should be possible. Might need to rewrite some parts.

> During startup, it should print something like 'Scanning for available networks' instead of an empty screen. Scanning can take some time.

Should be easy to add

> The currently selected character/string should be made more evident. Right now they are emphasized with Red color, but I think maybe the BG also would need to change to make it more obvious where the character is.
The same for all the UI elements, I don't know if it's possible to make the screens follow the same color scheme as the one used by dialog in other RetroPie menus, but it would be nice if it did.

I can change the background color of a selected string too. Not sure about a "global color", I would need to read about this in the `curses` documentation. I will come back on this.

> There's no 'cancel'/'exit' button on the startup screen - the 'B' button will close the dialog, but I think it needs something next to 'Connect' to make it obvious how to exit.

I am currently rethinking how the wifi selection works with the options, so I will take this into consideration.

> The layout should be centered and the screen areas should be separated a bit. They have some separation now, but for instance the Shift and Connect/Exit should not be on the same line.

I experimented with layout centering, it doesn't really work well.

> If I enter less than 8 chars, the next screen tells me about the error, but I cannot go back to fill the rest or change the password. If 8 chars is minimum, then perhaps warn the user and don't progress to the next screen or maybe go back to fill in the rest of the password ?

Good point, I will look into it.

OfficialPhilcomm:
> What's the purpose of moving the files from the cloned repo to another folder (/opt/dev_philcomm) ? The scriptmodule should be under /opt/retropie/supplementary, there's no need to duplicate it into another folder.

The reason for me is that I already created another PR for a different addon, which is also located in `/opt/dev_philcomm`. But then I didn't really work with the script modules before. So if this is the way to go, then I will adjust to that.

> the noinstclean flag is not needed, nobin I think is ok.
You can just use the sources_.. function to clone it to $md_inst and have install just install the needed gems.

Is there documentation on how to write scriptmodules?

> use the rp_module_repo variable to set the repository and branch. You can still call gitPullOrClone for installation from sources_.., and adding it there is useful for the RetroPie-Setup script to check for updates and don't re-install when needed.

I was wondering how to make the check for updates part work

> ruby-dev is not strictly neede and the ncurses gem is packaged as ruby-curses. Maybe look to see if require_all may be packaged, so there wouldn't be a need to run a gem install.
You can also just run getDepends <list-of-packages>, there's no need for an additional variable to hold the list (it's useful when you have to conditionally add a package).

Will have a look

> (optional) add a license and a link to it like other modules, even if minimal. Makes it look nice on https://retropie.org.uk/stats/licences/.

Already on it, I plan to put it under the MIT license

> the module should be available only for RPI so add rpi to the flags. Ubuntu uses network-manager, while Debian I think it has a different ctrl_interface configured for its wpa_supplicant.conf service.

Makes sense.

> use the configure function to add the RetroPie menu entry or remove it. In configure, you have the $md_mode variable to know whether the package is installed or not. Don't use the install function for it, it's not the place there.

Got it

> there's no need for a remove function, the setup script will take care of removing the install folder ($md_inst) and configure will be called also (where you can remove the menu entry, as indicated above).

So the way I imagine it working at some point, it will rename to original wifi module to make it invisible and then this module will take it's place.
How would this work when uninstalling, since it would need to put the old module back into place?

> there's no need for sudo in the script, the script is already running as root.

Alright

OfficialPhilcomm:
- [ ] Fix wifi2/lib/screens/wifi_selection_screen.rb:110: warning: rb_safe_level will be removed in Ruby 3.0
- [ ] Fix missing APs
- [ ] Fix stupid password misspelling
- [ ] Add all possible password characters
- [ ] Add ability to add SSIDs not present
- [ ] Add scanning startup screen
- [ ] Change color scheme to make elements more visible
- [ ] Add exit button to wifi selection screen
- [ ] Look into view centering
- [ ] Add better password error handling
- [ ] Improve scriptmodule
  - [ ] Remove `noinstclean` flag
  - [ ] Remove copying of wifi2 folder and make .sh file use `opt/retropie/supplementary/advanced-wifi/`
  - [ ] Make check for updates working
  - [ ] Use `ruby-curses` and possibly `ruby-require_all` package instead of gems
  - [ ] Add license
  - [ ] Add `rpi` flag
  - [ ] Remove `remove` function
  - [ ] Remove sudo

OfficialPhilcomm:
`ruby-curses` is far behind, its a version from 2017 (`1.2.4`), `gem install curses` installs `1.4.2`

OfficialPhilcomm:
https://github.com/ruby/curses/blob/d3b8aaef627dbf0ec1e8b6bf4139198a0ee4e23c/History.md#132--2019-12-20 the warning should not appear anymore says the curses changelog. Can you do `gem list`, search for the 'curses` line and tell me your version? Maybe for some reason you have a different version? Some system tools use curses `1.2.4`, so that is installed by default.

OfficialPhilcomm:
I have serious trouble with the scriptmodule file. It seems to remove the `/opt/retropie/supplementary/advanced-wifi` folder after installation

OfficialPhilcomm:
A couple of comments after a day of programming:

The 'ruby-curses` is too far behind, its a version from 2017 (1.2.4), gem install curses installs 1.4.2, which is the newest.

The warning `wifi2/lib/screens/wifi_selection_screen.rb:110: warning: rb_safe_level will be removed in Ruby 3.0` should not be shown. https://github.com/ruby/curses/blob/d3b8aaef627dbf0ec1e8b6bf4139198a0ee4e23c/History.md#132--2019-12-20 (curses changelog) says this was resolved with version `1.3.2`. What I suspect is that you might not have installed the gem. By default, version `1.2.4` of curses is installed on the pi, it is used by some system commands (htop for example, if I'm not mistaken). So if you don't install the gem manually, the app might try to use this version. You can find out which version ruby is using with `gem list`, which lists all gems installed plus their version.

I did some updates to the scriptmodule. My repo now stays in `supplementary`, but it seems the `noinstclean` flag is needed so it isn't removed after installation. Without the flag, the directory is cleared.
What doesn't work is the check for update functionality. I worked on that for two hours, looking at other modules, but I couldn't make it work. Please send me a link to documentation, if there is none, then please write some example code up.

Otherwise I made some huge improvements, as you can see in https://github.com/RetroPie/RetroPie-Setup/pull/3429#issuecomment-974627462. It doesn't seem like curses works well with background colors, so I will experiment with other ways to make the text more readable (although I think it's perfectly fine).
Tomorrow I will look at view centering.

OfficialPhilcomm:
Styling is now finalized. Due to performance reasons I had to stick with a simple layout, so I cannot make it look like the other retropie modules and make it performant (I tried to do that, but especially the Pi 3 and Zero had major refresh issues).

This is my final draft:

Screenshot 2021-11-20 at 22 19 57

Screenshot 2021-11-20 at 22 20 37

I might change something here and there, but this will be pretty much how far it'll go in beauty without performance problems.

cmitu:
> The warning wifi2/lib/screens/wifi_selection_screen.rb:110: warning: rb_safe_level will be removed in Ruby 3.0 should not be shown. https://github.com/ruby/curses/blob/d3b8aaef627dbf0ec1e8b6bf4139198a0ee4e23c/History.md#132--2019-12-20 (curses changelog) says this was resolved with version 1.3.2

Ok, so the source install will remove the warning, that's good. In this case, you'd also need `libncurses-dev` as dependency, otherwise the source install for the `curses` gem will fail.

> I did some updates to the scriptmodule. My repo now stays in supplementary, but it seems the noinstclean flag is needed so it isn't removed after installation. Without the flag, the directory is cleared.

Yes, it seems so. The directory being cleared is normal, that's the `install` action behavior. I'll take a look later to see why this doesn't work.

> What doesn't work is the check for update functionality. I worked on that for two hours, looking at other modules, but I couldn't make it work. It always says update available. Please send me a link to documentation, if there is none, then please write some example code up.

There's no official documentation, I think it was described in the commit that added this functionality, but basically when installing the package there's a `retropie.pkg` file created which records the date and commit from which the install was done, then it compares it with the last commit from the source repo and the date of the scriptmodule. It does work though, but the scriptmodule has to be commited to the repo so that the check works (comparison with the scriptmodule file date will always fail otherwise):

image

Regarding the `install` function - you don't need to change the permissions on the install folder to the install user (`pi` or otherwise). What's the reason for this ?

I also don't see the removal/configure actions. You've added a GUI configuration for it which will remove the new entry, but I don't think this is needed just for adding/removing the menu entry - just use the `configure` function to create/remove the entry.


> Styling is now finalized. Due to performance reasons I had to stick with a simple layout..

Can't figure out why a simple display would have performance issues - are you constantly refreshing/re-drawing the screen ? That shouldn't be needed unless something is changed (i.e. input is pressed).

The interface IMHO still needs some improvements.

Regarding the `sudo` access - I think that should be sorted out, but I think I found the reason why the AP list is incomplete: your SSID regex is too restrictive:
```
SSID_REGEX = /\s*ESSID:"(?<ssid>[A-Za-z0-9-]+)"/
```
This excludes any names containing e.g. spaces or `.` , which are valid names (e.g. my 2.4ghz AP for Pi3 or later has such characters in the name).

I haven't re-tried the prioritization part again or entering the new ESSID, but I'll gave them a try later on.

OfficialPhilcomm:
> Regarding the install function - you don't need to change the permissions on the install folder to the install user (pi or otherwise). What's the reason for this ?

Since there is no documentation, I looked at other examples, especially `retropie-manager`. How can I know what to copy and what not without documentation? If it is not needed, I will remove it.

> Can't figure out why a simple display would have performance issues - are you constantly refreshing/re-drawing the screen ? That shouldn't be needed unless something is changed (i.e. input is pressed).

The screen refreshes on input. Curses works in windows. Each window can be drawn into, and refreshed seperately. Currently I use two windows, one for the window and the other so there is one padding bar around the box. A third window (which would be needed for a background) had heavy performance decrease on the pi zero.

I could make it work probably by reinventing the wheel, but honestly I don't want to in my free time. I think it is fine the way it is. It is usable, now readable too.

> Regarding the sudo access - I think that should be sorted out, but I think I found the reason why the AP list is incomplete: your SSID regex is too restrictive:

Yep I will rework the regex today. Also ran into some issues there.

joolswills:
Thanks for this - appreciate the work you have put in, but I don't want to include this into the main RetroPie as it is I'm afraid.

I'm not against the inclusion of an OSK in general but I would prefer it implemented differently.

The use of Ruby is the first issue - it's not something I am very familiar with which would be an issue in regards to maintenance, but also I don't want to include Ruby dependencies for core RetroPie-Setup (on top of the ones we already have). I would prefer it to be implemented in something we already include and also that I am happy to maintain. 

I would prefer some reusable code that for example could be called from dialog and then would return a string to be used - it could be a reusable component then if needed elsewhere.

You can set this up though as an external repo (either your own) or an existing one, and have it as something users can install if they want. Modules checked out into the ext/REPONAME/scriptmodules/etc will be available.

OfficialPhilcomm:
Kinda understandable :)

Ruby is one of my main languages, I am really fluent in it, that's why I used it. About the maintenance part, I will maintain this as I have personal interest in it.

If you wanna look into other possibilities together let me know.

cmitu:
> Since there is no documentation, I looked at other examples, especially `retropie-manager`. How can I know what to copy and what not without documentation? If it is not needed, I will remove it.

For the scriptmodules' structure there's no full documentation - I've added some hints some time ago as a response to a forum post - https://retropie.org.uk/forum/topic/27127/is-there-a-guide-somewhere-to-create-install-scripts-if-you-can-get-a-game-to-compile/2

Functions/helpers used in the scriptmodules are mostly documented in https://retropie.org.uk/api/.

ParadoxGBB:
Hope between the architectural and maintenance issues there can be a solution to introduce functionality like this --- my primary fluent script language is PowerShell so I won't be of much help either.

The argument around reusability could help changing hostname, pi's default PW, etc, that would have been very useful in setting up my Pi02W.

Would love to help test this once it's ready --- thanks for keeping me in the loop so far.

OfficialPhilcomm:
@joolswills so I experimented with putting the scriptmodule into `ext/` folder. Everything works except it always shows `Update is available`. I use the exact same code as in `retropie-manager`, and there it works. @cmitu commented before that this only works when the scriptmodule is commited to master, so does that mean that this is also the case for scriptmodules in `ext/`?

PR #3430, opened 2021-11-22
Fix bugs in ScummVM launch script

ctrueden:
As discussed at:
https://retropie.org.uk/forum/topic/29349


cmitu:
For users that already have the `.svm` files in place (`game.svm`), maybe it would be better to skip the creation of a new entry.
They may have scraped already the games and have the gamelist entry present, this would double their gamelist entries.

Gemba:
FWIW, the thread was split and is now https://retropie.org.uk/forum/topic/32058/scummvm-launcher-improvements

ctrueden:
Thanks @Gemba. I didn't realize this was being discussed so actively lately. I've been meaning to circle back around to this PR; I'll try to catch up on the forum first before I push anything new.

PR #3431, opened 2021-11-28
joy2key,wifi: add an on-screen keyboard dialog

cmitu:
(Inspired by https://github.com/RetroPie/RetroPie-Setup/pull/3429)

Currently, the WiFi configuration from RetroPie menu needs a keyboard for configuring a connection.

Added a `python` script to display on-screen keyboard which can be operated with `joy2key` and the gamepad, which allows entering the WiFi network/key/password without a keyboard. The script accepts keyboard input, so it's suited for both keyboard and joystick input. It's using the `urwid` library (https://urwid.org).

The WiFi scriptmodule was modified to use the new OSK script instead of the `dialog` based inputboxes.

Additional modifications to the WiFi module:
 - added a confirmation step before removing the WiFi config and stopping the WiFi
 - added an info message during the WiFi scanning (instead of having a black screen)
 - removed the check for the pass length, it's done by the new OSK script
 - removed the `dhcpcd` hack, not needed since `wpa_supplicant` starts automatically
 - reformatted a bit the 'country' warning
 - after changing the config file, tell `wpa_supplicant` to reload its settings, seems like `ip link up/down` doesn't do that anymore.

NOTE: I've tested the new OSK script with smaller displays, but anything less than 40 cols will not be displayed fully. The OSK keyboard will scroll horizontally (when there's no more space on screen), but this is non-intuitive and might confuse users. _If_ we need to accomodate an even smaller screen size, then the OSK display area should be probably changed to accomodate that (i.e. less letters on a 'keyboard' row).

### DIFF
--- a/scriptmodules/admin/joy2key.sh
+++ b/scriptmodules/admin/joy2key.sh
@@ -15,11 +15,11 @@ rp_module_section="core"
 
 function _update_hook_joy2key() {
     # make sure joy2key is always updated when updating retropie-setup
-    rp_isInstalled "$md_id" && install_bin_joy2key
+    rp_isInstalled "$md_id" && rp_callModule "$md_id"
 }
 
 function depends_joy2key() {
-    local depends=()
+    local depends=(python3-urwid)
     # 'python3-sdl2' might not be available
     # it's packaged in Debian starting with version 11 (Bullseye)
     local p_ver
@@ -33,7 +33,7 @@ function depends_joy2key() {
 
 function install_bin_joy2key() {
     local file
-    for file in "joy2key.py" "joy2key_sdl.py"; do
+    for file in "joy2key.py" "joy2key_sdl.py" "osk.py"; do
         cp "$md_data/$file" "$md_inst/"
         chmod +x "$md_inst/$file"
         python3 -m compileall "$md_inst/$file"
--- a/scriptmodules/supplementary/wifi.sh
+++ b/scriptmodules/supplementary/wifi.sh
@@ -59,7 +59,9 @@ function connect_wifi() {
     local options=()
     i=0
     _set_interface_wifi up 2>/dev/null
+    dialog --infobox "\nScanning for WiFi networks..." 5 40 > /dev/tty
     sleep 1
+
     while read essid; read type; do
         essids+=("$essid")
         types+=("$type")
@@ -72,9 +74,12 @@ function connect_wifi() {
     choice=$("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty)
     [[ -z "$choice" ]] && return
 
+    local osk
+    osk="$(rp_getInstallPath joy2key)/osk.py"
+
     local hidden=0
     if [[ "$choice" == "H" ]]; then
-        cmd=(dialog --backtitle "$__backtitle" --inputbox "Please enter the ESSID" 10 60)
+        cmd=(python3 "$osk" --backtitle "$__backtitle" --inputbox "ESSID" --minchars 4)
         essid=$("${cmd[@]}" 2>&1 >/dev/tty)
         [[ -z "$essid" ]] && return
         cmd=(dialog --backtitle "$__backtitle" --nocancel --menu "Please choose the WiFi type" 12 40 6)
@@ -92,19 +97,18 @@ function connect_wifi() {
 
     if [[ "$type" == "wpa" || "$type" == "wep" ]]; then
         local key=""
-        cmd=(dialog --backtitle "$__backtitle" --insecure --passwordbox "Please enter the WiFi key/password for $essid" 10 63)
+        local key_min
+        if [[ "$type" == "wpa" ]]; then
+            key_min=8
+        else
+            key_min=5
+        fi
+
+        cmd=(python3 $osk --backtitle "$__backtitle" --inputbox "WiFi key/password" --minchars $key_min)
         local key_ok=0
         while [[ $key_ok -eq 0 ]]; do
             key=$("${cmd[@]}" 2>&1 >/dev/tty) || return
             key_ok=1
-            if [[ ${#key} -lt 8 || ${#key} -gt 63 ]] && [[ "$type" == "wpa" ]]; then
-                printMsgs "dialog" "Password must be between 8 and 63 characters"
-                key_ok=0
-            fi
-            if [[ -z "$key" && $type == "wep" ]]; then
-                printMsgs "dialog" "Password cannot be empty"
-                key_ok=0
-            fi
         done
     fi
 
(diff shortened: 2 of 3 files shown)


cmitu:
Some screenshots 


| *Default* view | *Smaller* screen | *Error* message |
| -- | -- | -- |
 |  |

OfficialPhilcomm:
First, nice that you made it work.

Second, there is a reason why I used ruby. You need less code to achieve more and keep readability. For example, you used 600+ lines just for the keyboard, while I used 1200 for the whole wifi manager. But that is more of a personal issue.

Third, I think there should be an option for users to either use the OSK or a keyboard. I made my thing a standalone app so people can choose. So maybe make this into a scriptmodule and let users install it manually.

cmitu:
> Second, there is a reason why I used ruby. You need less code to achieve more while maintaining readability. For example, you used 600+ lines just for the keyboard, while I used 1200 for the whole wifi manager. But that is more of a personal issue.

You're right - that's party because of my overly verbose style with lots of comments, new line separators and extra new lines for code separation and partly because of the `urwid` library style. Code is only 368 lines, which can be further shortened.

> Third, I think there should be an option for users to either use the OSK or a keyboard. I made my thing a standalone app so people can choose. So maybe make this into a scriptmodule and let users install it manually.

You can still use the keyboard - is supports both keyboard and navigating/presssing letters to enter the string.
I thought about having different dialogs for keyboard/gamepad, but it seemed an extra step (or steps) for users and it was easier to just add keyboard input support.

OfficialPhilcomm:
> You can still use the keyboard - is supports both keyboard and navigating/presssing letters to enter the string.
I thought about having different dialogs for keyboard/gamepad, but it seemed an extra step (or steps) for users and it was easier to just add keyboard input support.

It didn't work for me for some reason. I just tried it again, and it worked. Not sure what the issue was before.

joolswills:
Thanks for this. Looks good. Will test and review. Appreciate the efforts and thanks @OfficialPhilcomm for your contributions also.

scriptmodules/admin/joy2key.sh
```diff
@@ -15,11 +15,11 @@ rp_module_section="core"
 
 function _update_hook_joy2key() {
     # make sure joy2key is always updated when updating retropie-setup
-    rp_isInstalled "$md_id" && install_bin_joy2key
+    rp_isInstalled "$md_id" && depends_joy2key && install_bin_joy2key
```
joolswills, line 18: This could probably just be `rp_isInstalled "$md_id" && rp_callModule "$md_id"`

joolswills: `rp_isInstalled "joy2key" && rp_callModule "joy2key"` is fine too of course.

scriptmodules/supplementary/wifi.sh
```diff
@@ -150,9 +156,6 @@ _EOF_
 function gui_connect_wifi() {
     _set_interface_wifi down 2>/dev/null
     _set_interface_wifi up 2>/dev/null
-    # BEGIN workaround for dhcpcd trigger failure on Raspbian stretch
-    systemctl restart dhcpcd &>/dev/null
-    # END workaround
```
joolswills, line 155: I would probably leave this change out or do it in another commit at least.

scriptmodules/supplementary/wifi.sh
```diff
@@ -26,6 +26,8 @@ function _set_interface_wifi() {
             ip link set wlan0 down
         fi
     fi
+    # signal wpa_supplicant to reconfigure the interface
+    wpa_cli -i wlan0 reconfigure 2>/dev/null
```
joolswills, line 30: This causes `OK` to be output and breaks wifi scanning for me. With this line I get

```wlan0     Interface doesn't support scanning : Device or resource busy```

It's run just before scanning so maybe it's in the wrong place ?

```
    _set_interface_wifi up 2>/dev/null (which includes the call)
    dialog --infobox "\nScanning for WiFi networks..." 5 40
```

**Comment by joolswills** [comment]
A few comments about the RetroPie-Setup side. The OSK works great - thanks. I think I would prefer renaming Erase to Maybe Delete or Del ?

File: scriptmodules/admin/joy2key.sh (hunk 1 above)
Comment (cmitu) [reply scriptmodules/admin/joy2key.sh]: yeah, it was an copied from the initial version where I've added the 'depends' on the `retropiemenu` update hook, but using the 1st version is shorter. Applied.

File: scriptmodules/supplementary/wifi.sh (hunk 2 above)
Comment (cmitu) [reply scriptmodules/supplementary/wifi.sh]: Ok, sure.

File: scriptmodules/supplementary/wifi.sh (hunk 3 above)
Comment (cmitu) [reply scriptmodules/supplementary/wifi.sh:30]: It's supposed to tell `wpa_supplicant` to re-read the conf file. 
It's probably better to move the call somewhere else, since this gets called before scanning and it's not supposed to. I'll take a look and change it.

**Comment by cmitu** [comment]
> A few comments about the RetroPie-Setup side. The OSK works great - thanks. I think I would prefer renaming Erase to Maybe Delete or Del ?

Sure, I have no preferences. I replaced it with <kbd>Delete</kbd>.

File: scriptmodules/supplementary/wifi.sh (hunk 3 above)
Comment (cmitu) [reply scriptmodules/supplementary/wifi.sh]: I've removed it and it looks like the `dhcpcd` restart takes care of the reloading. Probably that's reconfiguration  failed in my tests, since I removed the `dhcpcd` reloading.

Looks like there's still a hook run for `wpa_supplicant` in `/lib/dhcpcd/dhcpcd-run-hooks` that performs the reloading, so adding back the `dhchp` reload takes care of re-configuring `wpa_supplicant` after any config files are changed.

File: scriptmodules/supplementary/wifi.sh [hunk 4]
```diff
@@ -59,7 +59,9 @@ function connect_wifi() {
     local options=()
     i=0
     _set_interface_wifi up 2>/dev/null
+    dialog --infobox "\nScanning for WiFi networks..." 5 40
```
joolswills, line 62: This needs to be `dialog --infobox "\nScanning for WiFi networks..." 5 40 >/dev/tty` or it wont work via retropie_setup.sh due to the logging.

cmitu: Should be fine now.

joolswills:
Thanks!

scriptmodules/supplementary/wifi.sh
```diff
@@ -214,6 +218,8 @@ The file should contain two lines as follows\n\nssid = \"YOUR WIFI SSID\"\npsk =
                     connect_wifi
                     ;;
                 2)
+                    dialog --defaultno --yesno "This will remove the WiFi configuration and stop the WiFi.\n\nAre you sure you want to continue ?" 12 35
```
joolswills, line 221: Sorry this needs the redirection also.

joolswills:
Sorry I noticed another dialog needing changing. Ideally the dialog improvements etc would be better in another commit. So it's separate from the osk stuff. But it's not a big issue.

scriptmodules/supplementary/wifi.sh
cmitu: Should be ok now. Ideally I should have split the extra changes into a separate commit. If anymore issues come up, I can do it so it's easier to inspect.

joolswills:
Thanks for your work on this.

cmitu:
Thank you, @meleu, appreciated.

PR #3432, opened 2021-11-29
runcommand: display launch menu dialog if no image is found

s1eve-mcdichae1:
This change allows the launch menu dialog to still be shown when launch art is enabled if no image is found. Current behavior is that when art is enabled, the dialog is never shown; therefore if no image is found, nothing is displayed at all.

This line to show the dialog is only called if no image is found in the first place, so checking if art is enabled at this point seems superfluous. We didn't get one, for one reason or another; either because it wasn't there, or because we didn't even look. By eliminating this test, we can fall back to show the dialog whenever an image is not obtained, for WHATEVER reason.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -1245,7 +1245,7 @@ function show_launch() {
         else
             fbi -1 -t "$IMAGE_DELAY" -noverbose -a "$image" </dev/tty &>/dev/null
         fi
-    elif [[ "$DISABLE_MENU" -ne 1 && "$USE_ART" -ne 1 ]]; then
+    elif [[ "$DISABLE_MENU" -ne 1 ]]; then
         local launch_name
         if [[ -n "$ROM_BN" ]]; then
             launch_name="$ROM_BN ($EMULATOR)"


s1eve-mcdichae1:
The "ports" system never shows art except `all/launching` or `ports/launching`. Even if there are `{rom}-image` files in the `downloaded_images/ports` folder, it doesn't use them (I tried to fix that, too, but it didn't work like I thought it would.)

This change allows the dialog box to be shown for all "Ports" roms, since the images never are.

joolswills:
> This change allows the launch menu dialog to still be shown when launch art is enabled if no image is found.

This was by design

> This line to show the dialog is only called if no image is found in the first place, so checking if art is enabled at this point seems superfluous

You can have artwork without enabling artwork - launching images work outside of this setting.

s1eve-mcdichae1:
> You can have artwork without enabling artwork - launching images work outside of this setting.

This isn't about showing artwork without enabling artwork. Rather the reverse: showing the DIALOG box WITH artwork is enabled, instead of a blank screen (only when an image isn't able to be found though).

good for any games that don't have an individual artwork, and for ports, which won't use it anyway even if they do have one.

if there is a "launching" image it will be shown in these cases. My idea just puts the dialog back in cases where even the "launching" art is not found.

joolswills:
> Rather the reverse: showing the DIALOG box WITH artwork is enabled, instead of a blank screen (only when an image isn't able to be found though).

yes I know - the black screen is by design though. if you enable to use art, the launch text dialog is disabled. I think it's cleaner than showing the dialog for when a game is missing an image.

joolswills:
As for the ports, I guess that's a bug that can be fixed. I will check.

s1eve-mcdichae1:
> yes I know - the black screen is by design though. if you enable to use art, the launch text dialog is disabled. I think it's cleaner than showing the dialog for when a game is missing an image.

cleaner, sure, but not more useful. A blank notebook is cleaner than one with notes in. How is the user to know when to press their button? The rule is "while the dialog is shown or immediately after the image disappears." How does that parse when there is no dialog or image shown at all?

joolswills:
Well you can start tapping a button almost right away.  I've not had an issue. Sorry but I'm not going to change the current behaviour.

s1eve-mcdichae1:
> As for the ports, I guess that's a bug that can be fixed. I will check.

This looked like it would be a very simple fix, but I am very noob and it didn't work like I thought it would. I've documented what I tried at:
https://retropie.org.uk/forum/post/265877

I'm not sure why that didn't work like I thought it would, but it didn't seem to do anything at all.

s1eve-mcdichae1:
> I think it's cleaner than showing the dialog for when a game is missing an image.

You know, when I was just *testing ports*, I was of the impression that something is better than nothing, but after actually *playing games* with this change, I do have to agree with you. After playing a few console titles I became used to the images, and then had a start when I launched a port and saw the infobox, I thought it was an error at first...

PR #3433, opened 2021-11-30
joy2key: fix crash on older pythons/py-sdl2 and init update

cmitu:
* Added a local fix for a crash when using older Python and Py-SDL2 versions.
 Upstream bug: https://github.com/marcusva/py-sdl2/issues/75.
 Upstream fixed the issue, but older distros (Ubuntu 18.04 based) don't have an updated `python3-sdl2` package.

* It seems that initializing the video is not really required for the joystick subsystem to work.
 On newer SDL versions (tested with 2.0.16), initialization will fail with video added if the video cannot be initialized.
 This prevents starting `joy2key_sdl.py` if ES is already started or trying to run it headless (i.e. no KMS/X available).
 While this is not a use case we're covering, it's better to remove the video initialization if it's not needed.


joolswills:
Thanks

PR #3434, opened 2021-11-30
set Dolphin's build type to Release

cobalt2727:
noticeable performance increases with these flags on

do not expect any playable performance in more than a few games on a pi 4 with a 64-bit OS, but hey, it's in "experimental" for a reason


scriptmodules/emulators/dolphin.sh
```diff
@@ -15,7 +15,7 @@ rp_module_help="ROM Extensions: .gcm .iso .wbfs .ciso .gcz .rvz .wad .wbfs\n\nCo
 rp_module_licence="GPL2 https://raw.githubusercontent.com/dolphin-emu/dolphin/master/COPYING"
 rp_module_repo="git https://github.com/dolphin-emu/dolphin.git :_get_branch_dolphin"
 rp_module_section="exp"
-rp_module_flags="!all 64bit"
+rp_module_flags="!all 64bit aarch64"
```
joolswills, line 18: The aarch64 flag shouldn't be needed as 64bit should be set on aarch64

scriptmodules/emulators/dolphin.sh
```diff
@@ -25,7 +25,7 @@ function _get_branch_dolphin() {
 }
 
 function depends_dolphin() {
-    local depends=(cmake pkg-config libao-dev libasound2-dev libavcodec-dev libavformat-dev libbluetooth-dev libenet-dev liblzo2-dev libminiupnpc-dev libopenal-dev libpulse-dev libreadline-dev libsfml-dev libsoil-dev libsoundtouch-dev libswscale-dev libusb-1.0-0-dev libxext-dev libxi-dev libxrandr-dev portaudio19-dev zlib1g-dev libudev-dev libevdev-dev libmbedtls-dev libcurl4-openssl-dev libegl1-mesa-dev qtbase5-private-dev)
+    local depends=(ca-certificates qtbase5-dev qtbase5-private-dev git cmake make gcc g++ pkg-config udev libudev1 libavcodec-dev libavformat-dev libavutil-dev libswscale-dev libxi-dev libxrandr-dev libudev-dev libevdev-dev libsfml-dev libminiupnpc-dev libmbedtls-dev libcurl4-openssl-dev libhidapi-dev libsystemd-dev libbluetooth-dev libasound2-dev libpulse-dev libpugixml-dev libbz2-dev libzstd-dev liblzo2-dev libpng-dev libusb-1.0-0-dev gettext libao-dev libenet-dev libopenal-dev libreadline-dev libsoil-dev libsoundtouch-dev libxext-dev portaudio19-dev zlib1g-dev libegl1-mesa-dev)
```
joolswills, line 28: dependencies for make gcc g++ ca-certificates are not needed (and maybe others).

scriptmodules/emulators/dolphin.sh
```diff
@@ -38,9 +38,9 @@ function sources_dolphin() {
 function build_dolphin() {
     mkdir build
     cd build
-    cmake .. -DCMAKE_INSTALL_PREFIX="$md_inst"
+    cmake .. -DCMAKE_INSTALL_PREFIX="$md_inst" -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS=-mcpu=native -DCMAKE_C_FLAGS=-mcpu=native -DCMAKE_C_FLAGS_INIT="-static"
     make clean
-    make
+    make -j$(nproc)
```
joolswills, line 43: This should be already handled by RetroPie-Setup (via MAKEFLAGS)

scriptmodules/emulators/dolphin.sh
```diff
@@ -38,9 +38,9 @@ function sources_dolphin() {
 function build_dolphin() {
     mkdir build
     cd build
-    cmake .. -DCMAKE_INSTALL_PREFIX="$md_inst"
+    cmake .. -DCMAKE_INSTALL_PREFIX="$md_inst" -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS=-mcpu=native -DCMAKE_C_FLAGS=-mcpu=native -DCMAKE_C_FLAGS_INIT="-static"
```
joolswills, line 41: we don't want to hardcode cpu flags here.

joolswills (changes requested):
Sorry but this can't be accepted as is as there are multiple issues.

We can change the build type to release but the other changes need looking into.

joolswills:
What was wrong with the previous dependencies? please provide some details for all the dependency changes.

Also why the static linking required ?

scriptmodules/emulators/dolphin.sh
```diff
@@ -25,7 +25,7 @@ function _get_branch_dolphin() {
 }
 
 function depends_dolphin() {
-    local depends=(cmake pkg-config libao-dev libasound2-dev libavcodec-dev libavformat-dev libbluetooth-dev libenet-dev liblzo2-dev libminiupnpc-dev libopenal-dev libpulse-dev libreadline-dev libsfml-dev libsoil-dev libsoundtouch-dev libswscale-dev libusb-1.0-0-dev libxext-dev libxi-dev libxrandr-dev portaudio19-dev zlib1g-dev libudev-dev libevdev-dev libmbedtls-dev libcurl4-openssl-dev libegl1-mesa-dev qtbase5-private-dev)
+    local depends=(qtbase5-dev qtbase5-private-dev git cmake pkg-config udev libudev1 libavcodec-dev libavformat-dev libavutil-dev libswscale-dev libxi-dev libxrandr-dev libudev-dev libevdev-dev libsfml-dev libminiupnpc-dev libmbedtls-dev libcurl4-openssl-dev libhidapi-dev libsystemd-dev libbluetooth-dev libasound2-dev libpulse-dev libpugixml-dev libbz2-dev libzstd-dev liblzo2-dev libpng-dev libusb-1.0-0-dev gettext libao-dev libenet-dev libopenal-dev libreadline-dev libsoil-dev libsoundtouch-dev libxext-dev portaudio19-dev zlib1g-dev libegl1-mesa-dev)
```
joolswills, line 28: What was wrong with the previous dependencies? please provide some details for all the dependency changes

scriptmodules/emulators/dolphin.sh
```diff
@@ -38,7 +38,7 @@ function sources_dolphin() {
 function build_dolphin() {
     mkdir build
     cd build
-    cmake .. -DCMAKE_INSTALL_PREFIX="$md_inst"
+    cmake .. -DCMAKE_INSTALL_PREFIX="$md_inst" -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS_INIT="-static"
```
joolswills, line 41: Why is the static linking required ?

scriptmodules/emulators/dolphin.sh
cobalt2727: Staying in line with https://github.com/dolphin-emu/dolphin/wiki/Building-for-Linux

scriptmodules/emulators/dolphin.sh
cobalt2727: It's not _required_, but I figured this might help reduce the chance of a build failing.

cmitu:
The CMake build file sets the CMAKE_BUILD_TYPE to Release by default, so I don't think the option is really needed:

https://github.com/dolphin-emu/dolphin/blob/dc0814ae4622313d513468bdc377ee9c031de199/CMakeLists.txt#L422-L425

PR #3439, opened 2021-12-06
added missing gettext dependency

hatl:
Tried to install RetroPie on a clean minimal Kubuntu installation.
Had to add gettext to the list of dependencies.


joolswills:
Thanks.

PR #3440, opened 2021-12-06
fixed compilation with newer gcc (required for ubuntu 21.10)

hatl:


hhromic:
I believe this should be better suited to be sent upstream instead of a local patch here?

cmitu:
@hhromic it's actually included in upstream's `master` (https://github.com/amadvance/advancemame/commit/f8688cae05799a30cd5337e626a07fa9f004c0a3), but it's not part of the v3.9 release we install. 
Perhaps we should switch to `master`, since the 3.9 release was too far back and lots of small fixes have been incorporated since.
@hatl is the current `master` working for you on Ubuntu 21.10 ? It may be better to use that instead of adding a new patch.

hatl:
tried to compile master -> yes, works :+1:

cmitu:
In this case, can you modify the PR to pull from master and drop the patch ? I think it's better this way. 
If you do so, can you squash the modifications into a single commit ?

joolswills:
Thanks.

hatl:
> In this case, can you modify the PR to pull from master and drop the patch ? I think it's better this way. If you do so, can you squash the modifications into a single commit ?
> 
> EDIT: this patch would however be useful for the `advmame-1.4` (and possibly `advmame-0.94`), which are no longer maintained upstream. `advmame-1.4` compiles ok with the patch, haven't tested `advmame-0.94` yet.

Thanks for the hint.
I'll look into advmame 1.4 and 0.94 when I have time - and create a new pull request.

cmitu:
> Thanks for the hint.

That was more of a note for me :), so I'll remember to add a PR for the older versions.

PR #3443, opened 2021-12-09
wolf4sdl: (bug) additional checksum for registered version game data

s1eve-mcdichae1:
Additional mission 3 checksum for installation media: "Spear of Destiny - Super CD Pack"


s1eve-mcdichae1:
The `vswap` file for Mission 3 has a different checksum on the "Spear of Destiny - Super CD Pack" than what is in the docs, but it still runs with the same command. Simply inserting this additional checksum into the script allows it to launch through ES.

joolswills:
Thanks.

PR #3445, opened 2021-12-11
Update function game_data_wolf4sdl()

s1eve-mcdichae1:


s1eve-mcdichae1:
Alternatively, make it work with either name.

Renaming to `.sd1` broke compatibility with RetroPie-Extra's **ecwolf** package.

**wolf4sdl** *can* run it with either name but the command is different (`vswap.sd1` needs to have the `--mission 1` option, and `vswap.sod` needs to *NOT* have that option.)

This change:
- uses the correct command for each filename,
- checks for all three versions (sod, sd1, sdm) before downloading shareware data, and
- creates the proper launch script for each version that exists.

PR #3450, opened 2021-12-20
lr-flycast: update to switch to upstream project

cmitu:
The libretro fork is no longer maintained and the libretro project uses the upstream project now (_flyinghead/flycast_).
This commit switches the scriptmodule to build from the upstream project's repository.

Changes:
 * updated description to match upstream
 * removed mention to `dc_flash.bin` BIOS file, it's no longer needed.
 * switched the build to use `cmake`, matching upstream and updated the dependencies.
 * removed the module renames hook, it's been 2.5 years since the change, `lr-reicast` hasn't been used in a long time and `lr-beetle-dc` was just a temporary (1 month) name.

Flycast changes:
  * OpenGL - support for high refresh rate monitors (120+ Hz) (1.2)
  * Support for WinCE games on 32-bit ARM platforms
  * Support for Core Options categories
  * Input mapping updates for Naomi/Atomiswave games, separate mapping profiles for Arcade/Dreamcast


cesarpuig:
Hello, in my case it does not work for me.

I get the following error:

`/opt/retropie/supplementary/runcommand/runcommand.sh: línea 1304:  5712 Violación de segmento  /opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-flycast/flycast_libretro.so --config /opt/retropie/configs/dreamcast/retroarch.cfg "/home/pi/RetroPie/roms/dreamcast/Bangai-O v1.001 (2000)(Virgin)(PAL)(M3)[!].chd" --verbose --appendconfig /dev/shm/retroarch.cfg`

I can solve it by substituting the content of the emulators.cfg file for:

`lr-flycast = "/opt/retropie/emulators/retroarch/bin/retroarch -L /opt/retropie/libretrocores/lr-flycast/flycast_libretro.so </dev/null --config /opt/retropie/configs/dreamcast/retroarch.cfg %ROM%"`

My system is:
RPI4
Retropie 4.7.19
Raspbian 10 (Buster)

cmitu:
Hm, looks like the STDIN redirection is still needed after all, as mentioned.

scriptmodules/libretrocores/lr-flycast.sh
```diff
@@ -10,61 +10,51 @@
 #
 
 rp_module_id="lr-flycast"
-rp_module_desc="Dreamcast emulator - Reicast port for libretro"
-rp_module_help="Previously named lr-reicast then lr-beetle-dc\n\nDreamcast ROM Extensions: .cdi .gdi .chd .m3u, Naomi/Atomiswave ROM Extension: .zip\n\nCopy your Dreamcast/Naomi roms to $romdir/dreamcast\n\nCopy the required Dreamcast BIOS files dc_boot.bin and dc_flash.bin to $biosdir/dc\n\nCopy the required Naomi/Atomiswave BIOS files naomi.zip and awbios.zip to $biosdir/dc"
-rp_module_licence="GPL2 https://raw.githubusercontent.com/libretro/flycast/master/LICENSE"
-rp_module_repo="git https://github.com/libretro/flycast.git master"
+rp_module_desc="Multi-platform Sega Dreamcast, Naomi and Atomiswave emulator derived from Reicast"
```
joolswills, line 13: Minor, but do we need to mention multi platform ?

scriptmodules/libretrocores/lr-flycast.sh
```diff
@@ -82,7 +72,6 @@ function configure_lr-flycast() {
 
     local def=0
     isPlatform "kms" && def=1
-    # segfaults on the rpi without redirecting stdin from </dev/null
```
joolswills, line 85: Can we keep a comment about the redirect - as it's not obvious from the code why it's needed.

scriptmodules/libretrocores/lr-flycast.sh
```diff
@@ -10,61 +10,51 @@
...
 rp_module_section="opt"
 rp_module_flags="!armv6"
 
 function depends_lr-flycast() {
-    local depends=(zlib1g-dev)
+    local depends=(zlib1g-dev libzip-dev)
     isPlatform "videocore" && depends+=(libraspberrypi-dev)
     isPlatform "mesa" && depends+=(libgles2-mesa-dev)
+    isPlatform "x11" && depends+=(libgl-dev)
     getDepends "${depends[@]}"
```
joolswills, line 25: Missing a cmake dependency.

joolswills:
Thanks for the contribution. A couple of minor remarks and a dependency and I think it's good to go. I've not build on anything else other than RPI 1/2/4 yet.

joolswills:
Also misses dependency libraspberrypi-dev for videocore targets - fails on my build machine with `/usr/bin/ld: cannot find -lbrcmGLESv2`

joolswills:
Sorry - I see it does have that - so I will need to debug further why I get that error. Apologies - will get back about that,

scriptmodules/libretrocores/lr-flycast.sh
cmitu: I've updated the desc from upstream. Maybe we can change it to 'Open Source ...' instead ?

scriptmodules/libretrocores/lr-flycast.sh
cmitu: Add, thanks.

scriptmodules/libretrocores/lr-flycast.sh
cmitu: Yes, initially I removed the redirection (there was a commit about input being done on a separate thread), but it looks like it's still needed.

cmitu:
> Also misses dependency libraspberrypi-dev for videocore targets - fails on my build machine with `/usr/bin/ld: cannot find -lbrcmGLESv2`

That may happen if the `/opt/vc/lib` is not added to the linker search path, but `-DUSE_VIDEOCORE=On` should take care of that. Add `VERBOSE=1` to the `make` command to see the complete link command line.

joolswills:
Please can you rebase this over the current master as the image.sh is out of date, and makes it easier for me to test the build in my chroot build environment.

joolswills:
(That's why it threw that error btw - it downloaded a bullseye image as image.sh missed my November 7th changes).

cmitu:
@joolswills try now, I've rebased over the current `master`.

joolswills:
Thanks. Binaries have been updated.

PR #3456, opened 2022-01-01
lower-case required for recent ubuntu package

hatl:
seems like the binary was changed to lower-case in the recent package:
pcsx2-unstable 3:1.7.0~git202112312251+202201010334~ubuntu21.10.1


joolswills (approved):

joolswills:
Thanks.

PR #3457, opened 2022-01-04
retroarch - update to v1.10.0

joolswills:
Switch to RetroPie fork to manage our patches there. Remove included patch files.


joolswills:
Binaries and assets have been updated.

Issue #67, opened 2012-10-09
Have a closer look at prboom package


petrockblog:
http://www.raspberrypi.org/phpBB3/viewtopic.php?p=190291#p190291

PR #3459, opened 2022-01-08
Patched pcsx-rearmed installation script to fix lack of sound do to error that should be a warning for PulseAudio

akoerner:
## Problem
The emulator pcsx_rearmed has no audio on RetroPi 4.7.20 with a Raspberry Pi Zero W resulting in the following error in /dev/shm/runcommand.log:

```bash
alsa: refusing to run under PulseAudio's emulation 
selected sound output driver: none
```

## Impact after merge
The installer  script scriptmodules/emulators/pcsx-rearmed.sh patches the source code of pcsx_rearmed by commenting out an error that should be a warning. The sound works after this pull request.  With line # 67(return -1) commented out in [plugins/dfsound/alsa.c](https://github.com/notaz/pcsx_rearmed/blob/master/plugins/dfsound/alsa.c) the sound card initialized correctly and the emulator produces sound.

## Context
Raspbian GNU/Linux 10 (buster)
RetroPi 4.7.20
Hardware        : BCM2835
Model           : Raspberry Pi Zero W Rev 1.1

## Root cause
The emulator throws an error, which should only be a warning. If  the function alsa/soundlib::snd_ctl_card_info_get_name() returns a sound card identification string containing the words "PulseAudio" . 

## The fix
Patching the file [plugins/dfsound/alsa.c](https://github.com/notaz/pcsx_rearmed/blob/master/plugins/dfsound/alsa.c)  before compile time for all raspberry pies as I have done in this pull request.
The file [plugins/dfsound/alsa.c](https://github.com/notaz/pcsx_rearmed/blob/master/plugins/dfsound/alsa.c) in the project pcsx_rearmed at https://github.com/notaz/pcsx_rearmed throws an error if the function "snd_ctl_card_info_get_name(info)" returns a string with the value "PulseAudio" in it.  This results in the emulator outputting no audio.  This should be a warning only and not an exception at least for this context. I don't know the impact of removing commenting out the line #67  on other systems so sending a pull request to the pcsx_rearmed team may not make sense.

**Update**
Issue can also be resolved by adding a preprocessor directive to select the libretro sound driver
```
    if isPlatform "neon"; then
        ./configure --sound-drivers=libretro --enable-neon
    else
        ./configure --sound-drivers=libretro --disable-neon
    fi
```

## Comments
precompiled binaries need to be patched also.

### DIFF
--- a/scriptmodules/emulators/pcsx-rearmed.sh
+++ b/scriptmodules/emulators/pcsx-rearmed.sh
@@ -26,13 +26,12 @@ function sources_pcsx-rearmed() {
 }
 
 function build_pcsx-rearmed() {
-    sed -i '67 s|.|//&|' plugins/dfsound/alsa.c
     if isPlatform "neon"; then
-        ./configure --sound-drivers=alsa --enable-neon
+        ./configure --sound-drivers=libretro --enable-neon
     else
-        ./configure --sound-drivers=alsa --disable-neon
+        ./configure --sound-drivers=libretro --disable-neon
     fi
-    make clean 
+    make clean
     make
     md_ret_require="$md_build/pcsx"
 }
--- a/scriptmodules/emulators/pcsx-rearmed.sh
+++ b/scriptmodules/emulators/pcsx-rearmed.sh
@@ -26,13 +26,15 @@ function sources_pcsx-rearmed() {
 }
 
 function build_pcsx-rearmed() {
+    sed -i '67 s|.|//&|' plugins/dfsound/alsa.c
     if isPlatform "neon"; then
         ./configure --sound-drivers=alsa --enable-neon
     else
         ./configure --sound-drivers=alsa --disable-neon
     fi
     make clean
-    make
+   
+   make
     md_ret_require="$md_build/pcsx"
 }
 
--- a/scriptmodules/emulators/pcsx-rearmed.sh
+++ b/scriptmodules/emulators/pcsx-rearmed.sh
@@ -32,9 +32,8 @@ function build_pcsx-rearmed() {
     else
         ./configure --sound-drivers=alsa --disable-neon
     fi
-    make clean
-   
-   make
+    make clean 
+    make
     md_ret_require="$md_build/pcsx"
 }
 


cmitu:
> The emulator throws an error, which should only be a warning. If the function alsa/soundlib::snd_ctl_card_info_get_name() returns a sound card identification string containing the words "PulseAudio" .

I think this was a deliberate decision by the upstream project and it's not a bug - there is a separate PulseAudio output plugin that can be used (when enabled).

Generally, we prefer source patches to be sent upstream and not kept in RetroPie-Setup, and in this case it looks like just enabling PulseAudio (`--sound-drivers=alsa,pulseaudio`) would solve your problem. 

Note that the RetroPie image doesn't rely on PulseAudio for sound and the it's enabled in the Pi images we offer on retropie.org.uk/download.

akoerner:
First off thank you for the response.  I have dug into the issue further and have identified a misconfigured preprocessor declaration for the sound driver in the package script from **alsa** to **libretro**.

This change is in line with the makefile in lr-pcsx_rearmed (see [Makefile.libretro line #496](https://github.com/libretro/pcsx_rearmed/blob/master/Makefile.libretro#L496))

Upstream pcsx_rearmed supports the libretro sound driver so the new proposed change resolves the issue.

One final comment is that pcsx_rearmed is the only PSX emulator that runs full speed on the Raspberry Pi Zero W.  lr-pcsx_rearmed does run but not at a playable speed.  For this I think it is important that it works out-of-the-box.

cmitu:
> First off thank you for the response. I have dug into the issue further and have identified a misconfigured preprocessor declaration for the sound driver in the package script from **alsa** to **libretro**.

Where exactly is that ?

> First off thank you for the response. I have dug into the issue further and have identified a misconfigured preprocessor declaration for the sound driver in the package script from **alsa** to **libretro**.
> 
> This change is in line with the makefile in lr-pcsx_rearmed (see [Makefile.libretro line #496](https://github.com/libretro/pcsx_rearmed/blob/master/Makefile.libretro#L496))

Don't confuse the 2 repositories - the upstream (standalone) emulator may have had at some point a _libretro_ stub added, but there's no implementation. Trying to compile the standalone emulator with `--sound-drivers=libretro` results in an error during the final linking stage:
```
/usr/bin/ld: plugins/dfsound/out.o: in function `SetupSound':
/home/pi/src/pcsx_rearmed/plugins/dfsound/out.c:34: undefined reference to `out_register_libretro'
```

Did you test the changes you've done - e.g. installed the emulator from source with your version of the scriptmodule ?

> 
> Upstream pcsx_rearmed supports the libretro sound driver so the new proposed change resolves the issue.

It doesn't seems like it does to me - the compilation fails. Furthermore, that's a _libretro_ implementation, which would need a _libretro_ frontend (i.e. RetroArch) for sound output, which is obviously not included in the standalone.

> One final comment is that pcsx_rearmed is the only PSX emulator that runs full speed on the Raspberry Pi Zero W. 

Yes, that's why it's still included and the default PS1 emulator on Pi0/1 platforms.

> lr-pcsx_rearmed does run but not at a playable speed. For this I think it is important that it works out-of-the-box.

Yes, it's important to work out-of-the-box, that's why I mentioned in my first comment that we're not including (or requiring) PulseAudio in the RetroPie base image, the emulator should work with the default ALSA audio output. Did you find that the default RetroPie image for Pi 0/1 has PulseAudio included/working ?

akoerner:
>Where exactly is that ?

In scriptmodules/emulators/pcsx-rearmed.sh
https://github.com/RetroPie/RetroPie-Setup/pull/3459/commits/0f58e3966934416226082f68665e76623d3b41a9

>Don't confuse the 2 repositories - the upstream (standalone) emulator may have had at some point a libretro stub added, but there's no implementation. Trying to compile the standalone emulator with --sound-drivers=libretro results in an error during the final linking stage:

I understand they are different projects and that lr-pscx-rearmed is a derivative  
In the case of your error the refrence "out_register_libretro" is defined on [line #236 of pcsx_rearmed/frontend/libretro.c](https://github.com/notaz/pcsx_rearmed/blob/master/frontend/libretro.c#236)

> Did you test the changes you've done - e.g. installed the emulator from source with your version of the scriptmodule?

Yes, I compiled from source with a cleaned environment with a modified scriptmodules/emulators/pcsx-rearmed.sh


>Yes, it's important to work out-of-the-box, that's why I mentioned in my first comment that we're not including (or requiring) PulseAudio in the RetroPie base image, the emulator should work with the default ALSA audio output. Did you find that the default RetroPie image for Pi 0/1 has PulseAudio included/working ?

From my perspective it does not work out-of-the-box. I have installed RetroPie using the Raspberry Pie Foundations Raspberry Pi Imager and have modified no underlying system configuration. This circles around to my original point of writing the MR

cmitu:
Closing since there's no response. 
I don't think the patch is necessary (or correct) in the current RetroPie configuration, if PulseAudio gets added to the base image, then we'll have to adjust the build options accordingly.

PR #3460, opened 2022-01-09
xroar: update to 1.0.7

cmitu:
Changes:
 * Added support for Tandy Color Computer 3 (NTSC/PAL).
 * Added support for Tandy MC-10
 * Emulator tries multiple SDL video renderers in order
 * Tape play/pause function
 * New snapshot format to support CoCo 3, MC-10
 * Add Control+Shift+D to flush disk images
 * RACE Computer Expansion Cage support (-cart mpi-race)
 * Support leading "~/" in filenames, not just path element
 * IDE, NX32, MOOH now all require user to specify an image

Scriptmodule changes:
 * Added explicit dependencies for ALSA and ZLIB. PulseAudio dependency added for `x11` platforms
 * Added emulator entries for CoCo3 models (PAL/NTSC)
 * Removed OSS support explicitely during build

### DIFF
--- a/scriptmodules/emulators/xroar.sh
+++ b/scriptmodules/emulators/xroar.sh
@@ -11,14 +11,16 @@
 
 rp_module_id="xroar"
 rp_module_desc="Dragon / CoCo emulator XRoar"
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

joolswills:
The libasound2-dev dependency was misspelt - I have fixed it in 362d8620 - binaries will be updated shortly

joolswills:
Binaries are updated.

PR #3466, opened 2022-01-24
Added upstream / current stella libretro core - lr-stella

joolswills:
I put this together rather quickly (and it needs testing), based on the lr-stella2014 script module and referencing the post on https://retropie.org.uk/forum/topic/32067/installing-lr-stella-upstream-on-raspberry-pi-4b/3?_=1643005126251 - which was similar.

If it's useful, it's an easy addition.

### DIFF
--- a/scriptmodules/libretrocores/lr-stella.sh
+++ b/scriptmodules/libretrocores/lr-stella.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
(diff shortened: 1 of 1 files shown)


cmitu:
Looks like the makefile enables LTO by default, which yields some warnings during linking. Might be worth disabling LTO on ARM targets with `make LTO=""`.

joolswills:
Thanks.

joolswills:
I just disabled it on everything. It uses more memory too I have noticed. I doubt it will change the performance drastically, but we can always re-enable if needed.

PR #3467, opened 2022-01-24
scummvm: update to fix GL runtime detection

cmitu:
Version 2.5.1 incorporated the OpenGL based scalers by default and this causes a runtime error on GLES platforms.
Update to a post-2.5.1 commit which changes the OpenGL loader from GLEW to GLAD and fixes this problem.

Minor - upstream license changed to GPL3+.


joolswills:
Thanks.

PR #3470, opened 2022-02-07
Update esthemes.sh with a new theme

RetroHursty69:
Theme is called ShadowClean and is suited for 16:9 monitors


joolswills:
thanks.

PR #3471, opened 2022-02-08
Fix gles2rice plugin entry.

s1eve-mcdichae1:
This var is not set locally, it is inheriting wrong value "main packages" from ../admin/setup.sh#L380, resulting in:

    mupen64plus-gles2ricemain packages = "...


joolswills:
Thanks.

PR #3472, opened 2022-02-09
eduke32: patch a gameplay bug in E1L4

cmitu:
Cherry-picked d8be6964fa81e9 to fix the shrinker ray getting stuck and breaking gameplay.
Reported in https://retropie.org.uk/forum/topic/32180/

Upstream info: https://voidpoint.io/terminx/eduke32/-/commit/d8be6964fa81e98ef370c82a2127d1381373950f


joolswills:
Thanks.

PR #3473, opened 2022-02-10
eduke32: (QOL) fix log location

s1eve-mcdichae1:
eduke32.log is saved to `.` so it ends up in `~`. If we alter the command to pushd into configdir before launching, it ends up where it should.


joolswills:
Thanks. I would like to implement this slightly differently but I need to test first.

s1eve-mcdichae1:
What did you have in mind? I initially had it set to just move the logfile after runtime but this felt cleaner, to me, to just create it in the right place to begin with.

joolswills:
Using `$md_conf_root/$portname` - the script is also used by ironfury. plus it's simpler to use the location we symlink to directly.

s1eve-mcdichae1:
I see, yeah that makes sense.

PR #3474, opened 2022-02-10
mupen64plus: (QOL) Link config dir for network access to saves, etc.

s1eve-mcdichae1:
For backup over network.


joolswills:
Thanks.

PR #3476, opened 2022-02-10
Remove lr-duckstation

joolswills:
Upstream has ceased development and the libretro binaries are no longer available


anzipex:
Damn..

EzeRinaldi:
yeah but... why? :(

ViperAcidZX:
This removal sucks.

PR #3480, opened 2022-02-13
Add lr-pcsx2 package configuration

mbudnek:
This adds a new experimental package configuration for the in-development [libretro port of the PCSX2 PlayStation2 emulator](https://docs.libretro.com/library/pcsx2/).  This allows much tighter integration with the rest of the RetroPie than is available using the standalone version of PCSX2.

So far I've tested this on Ubuntu 20.04 and 21.10.  Both are able to launch and play the games I tested.

### DIFF
--- a/scriptmodules/libretrocores/lr-pcsx2.sh
+++ b/scriptmodules/libretrocores/lr-pcsx2.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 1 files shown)


litaniast:
I'm using Ubuntu 22.04 but can't build it, main branch is gone, but instead, there is master branch so I tried that but failed.

snewell:
> I'm using Ubuntu 22.04 but can't build it, main branch is gone, but instead, there is master branch so I tried that but failed.

There's a libretro branch that builds for me, but nothing will load.  If I switch the repo to https://github.com/libretro/LRPS2, everything builds and runs as expected (tested for only about an hour with one game).

mbudnek:
Sorry, this PR didn't really get any attention, and I'm no longer running RetroPie on my machine, so I don't plan on keeping it updated.  I also haven't kept up with the lr-pcsx2/LRPS2 split.

PR #3481, opened 2022-02-13
Revert "sdl2 - updated to retropie-2.0.20 branch"

joolswills:
Reverts RetroPie/RetroPie-Setup#3390

Seems to have some issues with (some) OpenGL software on the RPI4. eg Darkplaces quake, alephone.

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -16,11 +16,11 @@ rp_module_section="depends"
 rp_module_flags=""
 
 function get_ver_sdl2() {
-    echo "2.0.20"
+    echo "2.0.10"
 }
 
 function get_pkg_ver_sdl2() {
-    local ver="$(get_ver_sdl2)+1"
+    local ver="$(get_ver_sdl2)+5"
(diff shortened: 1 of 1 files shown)


nickdesaulniers:
I think this is breaking builds on my pi zero w 2, due to https://github.com/libsdl-org/SDL/issues/5132 which was fixed in the 2.0.20 release IIUC.
```
/home/pi/src/SDL-release-2.0.18/src/video/raspberry/SDL_rpivideo.h:47:5: error: unknown type name ‘EGL_DISPMANX_WINDOW_T’
   47 |     EGL_DISPMANX_WINDOW_T dispman_window;
```

cmitu:
@nickdesaulniers we don't ship SDL version `2.0.18`, so I don't see how this is related to RetroPie's SDL version.
If you're using RaspiOS `bullseye` release or other distro, then I'm afraid is not something we support right now.

nickdesaulniers:
> If you're using RaspiOS bullseye release or other distro, then I'm afraid is not something we support right now.

I am; is there a milestone tracking bullseye support? I've been able to get everything working minus the autostart script.

Issue #3483, opened 2022-02-17
speech and alternate music don't work in uqm port


atfrase:
[The Ur-Quan Masters](http://sc2.sourceforge.net/) can be configured to use fully voiced dialogue and alternate music from the original 3DO version of the game, but these do not work in the RetroPie port. The installer downloads the relevant content files (`uqm-*-voice.uqm` and `uqm-*-3domusic.uqm`), and the game's setup menu shows options to enable them, but they do not take effect.

It looks to me like this is because RetroPie's `uqm.sh` install script (and the RetroPie-hosted pre-built binary archive produced from that script) install those two archives to the wrong directory, where the game does not find them. I'll submit a PR momentarily to correct this for installations from source, but the hosted pre-built archive will have to be updated as well.

cmitu:
Fixed with the mentioned PR

PR #3484, opened 2022-02-17
fix for uqm optional 3DO music and speech

atfrase:
As reported in #3483, the `uqm.sh` install script currently puts all content archives in the installed `.../content/packages/` directory, which seems eminently reasonable. However, despite the upstream project website and INSTALL guide having no obvious documentation to this effect, it is not correct: in fact only the `uqm-*-content.uqm` file goes here; `uqm-*-voice.uqm` and `uqm-*-3domusic.uqm` have to go in the `.../content/addons/` directory or the game won't find them.

This patch corrects the path when installing uqm from source, but the pre-compiled binary archive still contains all three files in the `packages/` directory and will have to be similarly updated.

I might also suggest adding the "remix" music packs to the installation, which are found at http://sc2.sourceforge.net/downloads.php under "The official Ur-Quan Masters remix packs"; they also have to go under `.../content/addons/`. I haven't added that to this PR since those files would also have to be mirrored in the RetroPie archives first.

### DIFF
--- a/scriptmodules/ports/uqm.sh
+++ b/scriptmodules/ports/uqm.sh
@@ -36,10 +36,12 @@ function sources_uqm() {
     downloadAndExtract "$(rp_resolveRepoParam "$md_repo_url")" "$md_build" --strip-components 1
     local packages="$md_build/content/packages"
     mkdir -p "$packages"
+    local addons="$md_build/content/addons"
+    mkdir -p "$addons"
     local ver="$(_get_ver_uqm)"
     download "$__archive_url/uqm-${ver}-content.uqm" "$packages"
-    download "$__archive_url/uqm-${ver}-voice.uqm" "$packages"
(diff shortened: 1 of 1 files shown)


cmitu:
Indeed, according to the [Wiki](https://wiki.uqm.stack.nl/The_Ur-Quan_Masters_Technical_FAQ#How_do_I_use_add-on_packages.3F), the proper location for addons is `content/addons`. 
The same page mentions the addons are added with the `--addon` parameter - is the parameter needed for the 2 addons to be enabled/usable ? If so, we might add them to the launch command.

atfrase:
No, I don't think the --addons switch is required unless the add-ons are in a different directory. If they're in the expected place (content/addons/) they should be found automatically.

This may mean that it might work to leave the files under packages/ and then specify --addons .../content/packages/, if that's easier for some reason, but I haven't tested that.

cmitu:
Re-tested this and it seems to work, doesn't need the `--addons` parameter added.

PR #3486, opened 2022-02-20
DuckStation standalone PSX emulator

dankcushions:
Standalone version using the (archived) repo. Draft as there's some issue:

1. GUI (escape key) is not navigable on the keyboard for me - i'm unable to get to the different horizontal 'tabs' within the settings menu. this makes binding a controller impossible.
2. Sometimes it seems the GUI does not load the graphics correctly, and all the menu items are just black boxes, making it impossible to navigate. however i'm unable to recreate this consistently

i would still like to generate a default controller setup via an emulationstation script - i started doing this via the evdev backend here: https://github.com/dankcushions/RetroPie-Setup/tree/duckstation-nogui - didn't get very far, but i believe it's possible.

### DIFF
--- a/scriptmodules/emulators/duckstation.sh
+++ b/scriptmodules/emulators/duckstation.sh
@@ -65,7 +65,6 @@ function configure_duckstation() {
     fi
     # SettingsVersion = 3 stops overwrite of any settings when version number doesn't match
     iniSet "SettingsVersion" "3"
-    iniSet "ControllerBackend" "evdev"
     if ! grep -q "\[BIOS\]" "$config"; then
         echo "[BIOS]" >> "$config"
     fi
--- a/scriptmodules/emulators/duckstation.sh
+++ b/scriptmodules/emulators/duckstation.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="duckstation"
+rp_module_desc="PlayStation emulator DuckStation"
+rp_module_help="ROM Extensions: .pbp .cue .bin .chd .img .ecm .mds .mdf\n\nCopy your PlayStation roms to $romdir/psx\n\nCopy the required BIOS file to $biosdir"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/stenzek/duckstation/master/LICENSE"
+rp_module_section="exp"
+rp_module_flags=""
+
+function depends_duckstation() {
+    local depends=(cmake libsdl2-dev libsnappy-dev pkg-config libevdev-dev libgbm-dev libdrm-dev)
+    getDepends "${depends[@]}"
+}
+
+function sources_duckstation() {
+    gitPullOrClone "$md_build" https://github.com/stenzek/duckstation.git
+}
+
+function build_duckstation() {
+    cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_DISCORD_PRESENCE=OFF -DUSE_X11=OFF -DUSE_DRMKMS=ON -DBUILD_NOGUI_FRONTEND=ON -DBUILD_QT_FRONTEND=OFF .
+    make clean
+    make
+
+    md_ret_require="$md_build/bin/duckstation-nogui"
+}
+
+function install_duckstation() {
+    md_ret_files=(
+        'LICENSE'
+        'README.md'
+        'bin'
+    )
+}
+
+function configure_duckstation() {
+    mkRomDir "psx"
+
+    # needed?
+    chown -R $user:$user "$md_inst/bin"
+
+    local config="$md_conf_root/psx/duckstation.ini"
+
+    addEmulator 0 "$md_id" "psx" "$md_inst/bin/duckstation-nogui -portable -settings $config -- %ROM%"
+    addSystem "psx"
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    # create config file
+    touch "$config"
+    chown -R $user:$user "$config"
+
+    # set config defaults
+    iniConfig " = " "" "$config"
+    if ! grep -q "\[Main\]" "$config"; then
+        echo "[Main]" >> "$config"
+    fi
+    # SettingsVersion = 3 stops overwrite of any settings when version number doesn't match
+    iniSet "SettingsVersion" "3"
+    iniSet "ControllerBackend" "evdev"
+    if ! grep -q "\[BIOS\]" "$config"; then
+        echo "[BIOS]" >> "$config"
+    fi
+    iniSet "SearchDirectory" "$biosdir"
+    if ! grep -q "\[MemoryCards\]" "$config"; then
+        echo "[MemoryCards]" >> "$config"
+    fi
+    iniSet "Directory" "$romdir/psx"
+    if ! grep -q "\[Display\]" "$config"; then
+        echo "[Display]" >> "$config"
+    fi
+    iniSet "LinearFiltering" "false"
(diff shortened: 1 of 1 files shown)


scriptmodules/emulators/duckstation.sh
```diff
@@ -0,0 +1,85 @@
...
+        'README.md'
+        'bin'
+    )
+}
+
+function configure_duckstation() {
+    mkRomDir "psx"
+
+    # needed?
+    chown -R $user:$user "$md_inst/bin"
```
cmitu, line 48: Is this needed because `-portable` is used ?

scriptmodules/emulators/duckstation.sh
```diff
@@ -0,0 +1,85 @@
...
+    addEmulator 0 "$md_id" "psx" "$md_inst/bin/duckstation-nogui -portable -settings $config -- %ROM%"
+    addSystem "psx"
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    # create config file
+    touch "$config"
+    chown -R $user:$user "$config"
+
+    # set config defaults
```
cmitu, line 61: IMHO the entire config handling should be similar to how we do it in `retroarch`: create a default file and copy it when it doesn't exist (i.e. on first installs, otherwise it's added with `.rp-dist`).

scriptmodules/emulators/duckstation.sh
```diff
@@ -0,0 +1,85 @@
...
+    local depends=(cmake libsdl2-dev libsnappy-dev pkg-config libevdev-dev libgbm-dev libdrm-dev)
+    getDepends "${depends[@]}"
+}
+
+function sources_duckstation() {
+    gitPullOrClone "$md_build" https://github.com/stenzek/duckstation.git
+}
+
+function build_duckstation() {
+    cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_DISCORD_PRESENCE=OFF -DUSE_X11=OFF -DUSE_DRMKMS=ON -DBUILD_NOGUI_FRONTEND=ON -DBUILD_QT_FRONTEND=OFF .
```
cmitu, line 29: Shouldn't the `X11` support be enabled for `x11` platforms ? (this might need additional dependencies also).

cmitu:
IMHO trying to create an `evdev` backend mapping is not so important.
Do note that `evdev` input codes are not 100% the same as the ones reported by SDL's joystick input IDs, so the mapping will not always be accurate (see my comment [here](https://github.com/RetroPie/RetroPie-Setup/pull/2992#issuecomment-587286442)).

scriptmodules/emulators/duckstation.sh
mrcmunir: Yes it's needed X11 enabled option platform also will be great option enable option for Desktop QT/SDL2 BUILD frontend will be better for some custom option in the GUI.
And DRMKMS only for only KMS platform .

scriptmodules/emulators/duckstation.sh
```diff
@@ -0,0 +1,85 @@
...
+
+function configure_duckstation() {
+    mkRomDir "psx"
+
+    # needed?
+    chown -R $user:$user "$md_inst/bin"
+
+    local config="$md_conf_root/psx/duckstation.ini"
+
+    addEmulator 0 "$md_id" "psx" "$md_inst/bin/duckstation-nogui -portable -settings $config -- %ROM%"
```
mrcmunir, line 52: Here I recommended will add, for the duckstation-qt build on the X11/wayland platforms.

dankcushions: I don't have an x11 platform to test but for a RetroPie context i would have thought the controller-driven NOGUI front end would be more appropriate more than a mouse-driven one which I presume QT is?

scriptmodules/emulators/duckstation.sh
dankcushions: i don't have an X11 platform to test but have enabled x11, and drmkms where appropriate

scriptmodules/emulators/duckstation.sh
dankcushions: agree! I think I've done this now

scriptmodules/emulators/duckstation.sh
dankcushions: i removed it and everything seems to still work, so hopefully all ok!?

scriptmodules/emulators/duckstation.sh
mrcmunir: I mean similar this 
addEmulator 1 "$md_id" "psx" "$md_inst/bin/duckstation-nogui -portable -settings $config -- %ROM%"
addEmulator 0 "$md_id" "psx" "$md_inst/bin/duckstation-qt -portable -settings $config -- %ROM%"

https://github.com/stenzek/duckstation/wiki/Command-Line-Arguments

Will needed -fullscreen if you have switch fullscreen Window .

In any case, this would be an alternative option to the no interface and it is the user who chooses which one to use sometimes qt build it's more friendly custom automatic remaps gamepads options than nogui in Desktop Options .

As soon as I have some time I try it on my x86-64 PC or aarch64 platform, which are the devices I have at hand.

dankcushions: ok, yes that makes sense giving the choice, although as I don't have a device to test this is not something I would want to add in this PR, although please feel free to add it yourself later! :)

gdniw:
Many thanks, I test it and it is working very well!

cmitu:
I've re-tested this with the latest changes, seems to work fine, with a few notices:

* Conf file generation - should be simplified, no need to use the `ini` routines, just write the config with `cat << EOF >> $config`. It's easier to read and we only copy it once, on the 1st install, so we don't have to account for existing settings. 
   Some defaults I see changed:
     * _Start Fullscreen_ should be on
     * Don't load cheats automatically (_Automatically Load Cheats_).
     * Lower the logging level to ERROR, _WARN_ produces a lot of messages and it spams `/dev/shm/runcommand.log`.
     * _Confirm Power Off_ should be off
     * _Save State on Exit_ should be off
   
* Controller setup. I was expecting some auto-config for in-game controls to work (via SDL's _GameControllerApi_), but it didn't work for me. However, the settings menu is navigable OOB - L1/R1 are working to switch between the settings menu tabs. The Back/Action buttons are reversed compared to RetroArch's RGUI - _East_ is back/cancel and _South_ is Activate/Enter. It may be possible to change this in the source (see [here](https://github.com/stenzek/duckstation/blob/51041e47f70123eda41d999701f5651830a0a95e/src/frontend-common/fullscreen_ui.cpp#L4699)).
   It's possible to map the settings menu to a button, but an extra button would be needed (i.e. Xbox button or the PS button.
   Tested with a pair of DS3/DS4 controllers, don't have an Xbox one to see.

jeifi:
i am trying to test it on a Raspberry PI3B+ with RetroPie 4.8 but the  compilation always fails around 60%. Is it only for RPI4/400 ?

abarjj:
I don't know if the team is still working on this emulator or not. I use it so far on all my systems and I prefer it to the libretro variant. After the massive updates the emulator has gotten a few days ago, it is not possible to build it anymore. I would be obliged if someone can guide me in finding the solution to the problem. Here is the log : https://pastebin.com/hPJ9tSf7.  Thanks

amparg:
> I don't know if the team is still working on this emulator or not. I use it so far on all my systems and I prefer it to the libretro variant. After the massive updates the emulator has gotten a few days ago, it is not possible to build it anymore. I would be obliged if someone can guide me in finding the solution to the problem. Here is the log :https://pastebin.com/q9sCZVB0. Thanks

Change the source to` Latest `and it will work. 

 `gitPullOrClone "$md_build" https://github.com/stenzek/duckstation.git Latest`

mrcmunir:
SDL  nogui seems deprecated and not recommended by author fot latest improvements
https://github.com/stenzek/duckstation/commit/68866f3174228194cd86c9caef1038e1721c96dc

For Qt build  Needed qt6.1 and recommended sdl2 2.0.22 with HIDAPI ON .

IoanZzZ:
Changing this line: `gitPullOrClone "$md_build" https://github.com/stenzek/duckstation.git Latest` to `gitPullOrClone "$md_build" https://github.com/stenzek/duckstation.git legacy` it allows the installation of this emulator.

PR #3491, opened 2022-03-05
lr-ppsspp: preset a save directory

cmitu:
`lr-ppsspp` uses the `savefile_directory` path in order to create its own save structure.
Game saves are not regular `.srm` files, but they're saved in a `PSP/SAVEDATA/<GAME_ID>` folder structure.

Without setting a `savefile_directory`, the core will try to save to `/PSP/SAVEDATA/<GAME_ID>`, which obviously fails.
Setting this to `$HOME/.config/ppsspp` makes it possible to share the savedata with the standalone emulator.

### DIFF
--- a/scriptmodules/libretrocores/lr-ppsspp.sh
+++ b/scriptmodules/libretrocores/lr-ppsspp.sh
@@ -44,6 +44,11 @@ function configure_lr-ppsspp() {
         mkUserDir "$biosdir/PPSSPP"
         cp -Rv "$md_inst/assets/"* "$biosdir/PPSSPP/"
         chown -R $user:$user "$biosdir/PPSSPP"
+
+        # the core needs a save file directory, use the same folder as standalone 'ppsspp'
+        iniConfig " = " "" "$configdir/psp/retroarch.cfg"
+        iniSet "savefile_directory" "$home/.config/ppsspp"
+        mkUserDir "$home/.config/ppsspp"
     fi
 
     addEmulator 1 "$md_id" "psp" "$md_inst/ppsspp_libretro.so"


scriptmodules/libretrocores/lr-ppsspp.sh
```diff
@@ -44,6 +44,11 @@ function configure_lr-ppsspp() {
         mkUserDir "$biosdir/PPSSPP"
         cp -Rv "$md_inst/assets/"* "$biosdir/PPSSPP/"
         chown -R $user:$user "$biosdir/PPSSPP"
+
+        # the core needs a save file directory, use the same folder as standalone 'ppsspp'
+        iniConfig " = " "" "$configdir/psp/retroarch.cfg"
+        iniSet "savefile_directory" "$home/.config/ppsspp"
+        mkUserDir "$home/.config/ppsspp"
```
joolswills, line 51: We symlink "$home/.config/ppsspp" so maybe it's better to directly use the retropie location at `/opt/retropie/configs/psp` instead ?

cmitu: In my test it didn't work, I think `ppsspp` has some path handling and if the path ends in `psp` resets it (?). I'll give it another test, but that was the reason `$home/.config/ppsspp` was used.

cmitu: Ok, I re-tested and what happens is that `ppsspp` strips the final `psp` from the save path (some compatibility checks probably). So instead of saving to `/opt/retropie/configs/psp/PSP/SAVEDATA`, it creates the saves folder as `/opt/retropie/configs/psp/SAVEDATA` (and also adds the `SYSTEM` and `PPSSPP_STATE` folders in the same config folder).

Using `$home/.config/ppsspp` 'tricks' `ppsspp` into creating the `PSP` main folder with all the necessary sub-folders (as documented the [core doc](https://docs.libretro.com/library/ppsspp/#directories)).

joolswills: Thanks

joolswills (approved):

PR #3493, opened 2022-03-10
add cdi emulator build with lr-mame2015

XenuIsWatching:
This adds support for the Phillips CD-i. This is just a build of the lr-mame2015 that subtargets the CD-i.


XenuIsWatching:
In order to run, you'll need to "find" a cdimono1.zip to be placed in the BIOS directory.

cmitu:
`cdimono` support is already present in `lr-mess` and `mame`. What's the advantage of using an older (and un-maintained) build or MAME for it ?

XenuIsWatching:
I see what you mean. lr-mame2015 only just added a convenient subtarget... but lr-mess should be used instead

PR #3500, opened 2022-03-16
Update esthemes.sh with new theme

RetroHursty69:
Theme is called Vinyl-Hits. Works well on Pi4. VRAM at 130mb. Theme is a tribute to Vinyl LP's.


joolswills:
Thanks.

Issue #3496, opened 2022-03-14
Detecting initial Desktop/Monitor resolution might fail in X11 when using connected (but disabled) screen


SjoerdV:
When using Retropie with multiple connected monitors of which at least one is turned off in X11 settings, the wrong monitor (the disabled one) might be detected by runcommand.sh. because it will only test if a monitor is connected and not if it is active.

I propose rewriting this section:
https://github.com/RetroPie/RetroPie-Setup/blob/6e83a7d57517b7fbabc1e30309167fc52f85b4f1/scriptmodules/supplementary/runcommand/runcommand.sh#L399-L403
to:

```bash
# determine current output
mode_id[0]="$($XRANDR --verbose | awk '/ connected.*\(0x[a-f0-9]{1,}\)/ { print $1;exit }')"
# determine current mode id & strip brackets
mode_id[1]="$($XRANDR --verbose | awk '/ connected.*\(0x[a-f0-9]{1,}\)/ {print;exit}' | grep -o "(0x[a-f0-9]\{1,\})")"
mode_id[1]="$(echo ${mode_id[1]:1:-1})"
```

cmitu:
Thanks for the heads up, I'll give it a test and add the modifications into `runcommand`.

PR #3501, opened 2022-03-19
runcommand: check only active displays on X11

cmitu:
When mutiple displays are connected, some of them may be disabled - skip them when considering the available video modes.

Fixes: #3496 .


joolswills:
Thanks.

PR #3502, opened 2022-03-19
retropiemenu: fix launching RetroArch from the menu

cmitu:
Due to the recent addition of the GameMode support, RetroArch needs access to the user's DBUS session socket.
When starting RetroArch from the RetroPie menu, `su` is used to launch it as the install user, but the DBUS socket path is not preserved (`DBUS_SESSION_BUS_ADDRESS`), leading to a en error and then a crash.

Prevent the crash by setting the XDG_RUNTIME_DIR when RetroArch is started with `su`.

NOTE: this primarily affects X11/PC users, where GameMode is part of the desktop install.


joolswills:
Thanks.

PR #3504, opened 2022-03-20
Update Redream to Universal arm (32/64bits)

mrcmunir:
Hello As you know, a zip  build has been released  about 5 days that  in same link   for arm32-bit and  arm64 64-bit.
It doesn't change much else since the Redream script takes care of loading aarch32.elf or aarch64.elf if you are under 32bits or 64bits. 

Only supported on boards the GPU driver can run context  opengl 3.1 up to 4.5  (some extensions optional) or Opengl es 3.1 context.

### DIFF
--- a/scriptmodules/emulators/redream.sh
+++ b/scriptmodules/emulators/redream.sh
@@ -14,10 +14,10 @@ rp_module_desc="Redream Dreamcast emulator"
 rp_module_help="ROM Extensions: .cdi .cue .chd .gdi .iso\n\nCopy your Dreamcast roms to $romdir/dreamcast"
 rp_module_licence="PROP"
 rp_module_section="exp"
-rp_module_flags="noinstclean !all rpi4 !aarch64"
+rp_module_flags="noinstclean !all gles31 aarch64"
 
 function __binary_url_redream() {
-    echo "https://redream.io/download/redream.aarch32-raspberry-linux-latest.tar.gz"
+    echo "https://redream.io/download/redream.universal-raspberry-linux-latest.tar.gz"
 }
 
 function install_bin_redream() {


joolswills:
Thanks.

cmitu:
The new version hangs when starting a game - similar to https://gitlab.com/inolen/redream/-/issues/1274.
Maybe should revert this until this is fixed. @mrcmunir is the new version working for you ?

mrcmunir:
> The new version hangs when starting a game - similar to https://gitlab.com/inolen/redream/-/issues/1274. Maybe should revert this until this is fixed. @mrcmunir is the new version working for you ?

It's working fine for me tested under ubuntu armv7/aarch64 18.04/20.04 and  ubuntu up system 
Screenshot from 2022-03-23 11-07-32

In that link it shows the problem for x86-64
For now our script does not have support for the x86-64 binary download.

cmitu:
I get a similar hang to the one in the linked issue on the Rpi4 (armv7l - 32 bit).  We have a similar report in the forums that - after the update - `redream` is not working.

mrcmunir:
> I get a similar hang to the one in the linked issue on the Rpi4 (armv7l - 32 bit). We have a similar report in the forums that - after the update - `redream` is not working.

 Then we can do one thing to revert the changes until it is fixed if it gives problems on rpi4 32bits.

Or The other alternative would be to separate the discharge having control for
armv7 (32bits) and aarch64 (64bits) and also support x86-64 in each of the downloads.

cmitu:
> > I get a similar hang to the one in the linked issue on the Rpi4 (armv7l - 32 bit). We have a similar report in the forums that - after the update - `redream` is not working.
> 
> Then we can do one thing to revert the changes until it is fixed if it gives problems on rpi4 32bits.

 That would be the fastest option right now.

> Or The other alternative would be to separate the discharge having control for armv7 (32bits) and aarch64 (64bits) and also support x86-64 in each of the downloads.

PC (x86-64) could be added, but I think we'd rather have **inolen**'s accord to include it - like we have for the ARM/PI binaries.

mrcmunir:
Hi @cmitu Try testing this to see if it downloads correctly for you. 
```
if isPlatform "rpi4" && isPlatform "32bit"; then
         echo "https://redream.io/download/redream.aarch32-raspberry-linux-latest.tar.gz"
elif isPlatform "aarch64"; then
         echo "https://redream.io/download/redream.universal-raspberry-linux-latest.tar.gz"
fi
```
If the binary is downloaded correctly, I make a pull request

cmitu:
@mrcmunir I don't have a Pi system right now to test, but looks fine - though I'd rather revert the PR to make it simple to re-apply again. I'll try to open an issue to see if we can solve this fast - before reverting.

cmitu:
Reported upstream in https://gitlab.com/inolen/redream/-/issues/1276.

cmitu:
Looks like there is an updated version that fixes the problem, we'll just wait for more confirmations in the forum before closing the upstream issue.

PR #3506, opened 2022-04-01
alephone (update): Use latest version game data.

s1eve-mcdichae1:
(Unless using old-version game engine, then use old-version data too.)

### DIFF
--- a/scriptmodules/ports/alephone.sh
+++ b/scriptmodules/ports/alephone.sh
@@ -54,18 +54,22 @@ function install_alephone() {
 }
 
 function game_data_alephone() {
-    local release_url="https://github.com/Aleph-One-Marathon/alephone/releases/download/release-20150620"
+    local version="20150620"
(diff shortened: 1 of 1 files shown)


cmitu:
Looks ok, thank you.

scriptmodules/ports/alephone.sh
```diff
@@ -54,18 +54,22 @@ function install_alephone() {
 }
 
 function game_data_alephone() {
-    local release_url="https://github.com/Aleph-One-Marathon/alephone/releases/download/release-20150620"
+    local version="20150620"
+    if compareVersions "$__os_debian_ver" ge 9 || [[ -n "$__os_ubuntu_ver" ]]; then
+        local version="20220115"
```
joolswills, line 59: This doesn't need the additional "local".

s1eve-mcdichae1: You're right, of course. Thanks.

joolswills:
I squashed the commits. Thanks.

PR #3507, opened 2022-04-01
fixes jzintv background color to black on pillar boxed 4:3 emulator output and 16:9 displays

Gemba:
-  sets 4:3 display format for jzintv automagically (launcher file)
-  fixes a UI freeze bug when a mouse is attached and moved on a RPI3 when jzintv is running

ref: https://retropie.org.uk/forum/topic/32433/jzintv-has-black-border-or-full-screen-color


scriptmodules/emulators/jzintv.sh
```diff
@@ -44,8 +52,9 @@ function build_jzintv() {
     mkdir -p jzintv/bin
     cd jzintv/src
 
+    isPlatform "rpi" && extra='EXTRA=-DPLAT_LINUX_RPI'
```
cmitu, line 55: `extra` should be declared as `local`.

scriptmodules/emulators/jzintv.sh
```diff
@@ -61,17 +70,60 @@ function install_jzintv() {
 }
 
 function configure_jzintv() {
-    mkRomDir "intellivision"
+    local system="intellivision"
```
cmitu, line 73: No need to make `system` a variable, since it's not changed.

scriptmodules/emulators/jzintv.sh
```diff
@@ -61,17 +70,60 @@ function install_jzintv() {
...
+    mkRomDir $system
 
-    local options=(
-        --displaysize="%XRES%x%YRES%"
-        --quiet
-        --rom-path="$biosdir"
-        --voice=1
-    )
+    local start_script="$romdir/$system/+Start jzintv.sh"
```
cmitu, line 76: This shouldn't be in in the ROMs folder unless the emulator/program should be started standalone.
If a wrapper launcher is needed (to parse and transform the arguments), then it should reside with the main executable, in `$md_inst`.

scriptmodules/emulators/jzintv.sh
```diff
@@ -61,17 +70,60 @@ function install_jzintv() {
...
+    options+=(\$1); shift
+done
+
+rom=\$1
+echo "Launching: \$jzintv_bin \${options[*]} \"\$rom\"" >> /dev/shm/runcommand.log
+
+pushd "$romdir/$system" > /dev/null
+\$jzintv_bin \${options[*]} "\$rom"
+popd
+_EOF_
```
cmitu, line 122: Can you explain better what is the launcher wrapper supposed to do ? I'm guessing one function would be to set the display size to a 4:3 video mode, but what does the rotation has to do with it ?

scriptmodules/emulators/jzintv/01_rpi_pillar_boxing_black_background_sdl2.patch
```diff
@@ -0,0 +1,15 @@
+--- a/gfx/gfx_sdl2.c
```
cmitu, line 1: Is this a SDL bug ? Then perhaps should be fixed in SDL repo (retropie/sdl).

scriptmodules/emulators/jzintv.sh
Gemba: It is more a DRY approach. I don't like to maintain the same information on n locations. But I can remove it or set at least readonly. What is expected the "RetroPie-way"?

scriptmodules/emulators/jzintv.sh
Gemba: Ah, ok. I had another scriptmodule as reference. But good to know. Will be moved to that location.

scriptmodules/emulators/jzintv.sh
Gemba: Right. Will be reflected in update to this PR.

scriptmodules/emulators/jzintv/01_rpi_pillar_boxing_black_background_sdl2.patch
Gemba: No it is how the emulator calls/uses SDL. SDL works right. The emulator assumes two things, and if both are false the issue (colored pillars) do show up. One: Physical display is always 4:3, two: The ROM does use black as background color.

scriptmodules/emulators/jzintv.sh
Gemba: Right, one is to limit the emulator display to 4:3 on a physical display with different aspect ratio.

The rotation is an educated guess. On 90/270 degree I assume the XRES and YRES are flipped (ie. 1080 and 1920). If not, what options do exsist to detect 90/270 mounted displays?

scriptmodules/emulators/jzintv.sh
```diff
@@ -61,17 +70,60 @@ function install_jzintv() {
...
+#    --gfx-verbose
+    --displaysize="\${intv_w}x\${intv_h}"
+    --rom-path="$biosdir"
+    --voice=1
+)
+
+# ingest additional options
+while [[ \$# -gt 1 ]] ; do
+    options+=(\$1); shift
+done
```
joolswills, line 114: if you have shifted the input params why just do something like `options+=("$@")`

scriptmodules/emulators/jzintv.sh
```diff
@@ -27,6 +27,14 @@ function sources_jzintv() {
     # jzintv-YYYYMMDD/ --> jzintv/
     mv jzintv-[0-9]* jzintv
     cd jzintv/src
+    # archive files have CRLF linendings, patch has CR only
+    find . -type f \( -iname "*.c" -o -iname "*.h" \) -exec flip -u {} \+
```
joolswills, line 31: I'm not sure I like the converting of all the files to LF to apply the patch. Seems overkill. Also so you mean the patches has LF only not CR ?

scriptmodules/emulators/jzintv.sh
Gemba: The jzintv options are "space-free", while the ROM (last parameter) could contain spaces. The last parameter should be quoted to avoid shell interpretation.

joolswills: I don't see how that wouldn't work with my line above - please can you give an example ?

scriptmodules/emulators/jzintv.sh
```diff
@@ -44,8 +52,9 @@ function build_jzintv() {
     mkdir -p jzintv/bin
     cd jzintv/src
 
+    isPlatform "rpi" && local extra='EXTRA=-DPLAT_LINUX_RPI'
```
joolswills, line 55: You need to declare the variable as local for all platforms nor just for the RPI.

scriptmodules/emulators/jzintv.sh
Gemba: Well, yes. LF (0x0a) was ment.
Some other options I considered here: Change the source zip archive to LF (hosted at your site, must be remembered when a new source archive is deployed), convert patches to CRLF (awkward as RaspiOS uses UNIX style endings), use `--ignore-whitespace` in applyPatch (may have knock-on effects).

scriptmodules/emulators/jzintv.sh
```diff
@@ -61,17 +70,60 @@ function install_jzintv() {
...
+_EOF_
+    chown $user:$user "$start_script"
+    chmod u+x "$start_script"
 
-    addEmulator 1 "$md_id" "intellivision" "$md_inst/bin/jzintv ${options[*]} %ROM%"
-    options+=(--ecs=1)
-    addEmulator 0 "${md_id}-ecs" "intellivision" "$md_inst/bin/jzintv ${options[*]} %ROM%"
-    addSystem "intellivision"
+    addEmulator 1 "$md_id"     "$system" "bash '$start_script' %XRES% %YRES% %ROM%"
+    addEmulator 0 "$md_id-ecs" "$system" "bash '$start_script' %XRES% %YRES% --ecs=1 %ROM%"
```
joolswills, line 127: The file is marked as executable and lives on a native linux fs so you don't need to use "bash" to run it now.

scriptmodules/emulators/jzintv.sh
```diff
@@ -61,17 +70,60 @@ function install_jzintv() {
...
+# ingest additional options
+while [[ \$# -gt 1 ]] ; do
+    options+=(\$1); shift
+done
+
+rom=\$1
+echo "Launching: \$jzintv_bin \${options[*]} \"\$rom\"" >> /dev/shm/runcommand.log
+
+pushd "$romdir/$system" > /dev/null
+\$jzintv_bin \${options[*]} "\$rom"
```
joolswills, line 120: If using my example above this would need to be `\${options[@]}`  btw - Unless I'm missing something this should work ok.

joolswills:
I have made some comments but I'm not sold on including this as it is currently due to the size, and what seems to be unnecessary complexity with the launch script etc.

scriptmodules/emulators/jzintv.sh
Gemba: Right. Leftover from a sample scriptmodule I peeked on unreflected (dosbox.sh).

scriptmodules/emulators/jzintv.sh
```diff
@@ -18,7 +18,7 @@ rp_module_section="opt"
 rp_module_flags="sdl2"
 
 function depends_jzintv() {
-    getDepends libsdl2-dev libreadline-dev
+    getDepends libsdl2-dev libreadline-dev flip
```
joolswills, line 21: BTW we don't use flip anywhere else. dos2unix is used though so that would be a preference. But you can probably just use sed though to change the files that are needed (or include the patch with CRLF line endings and avoid having to change the source).

scriptmodules/emulators/jzintv.sh
Gemba: I'd like to challenge this. Works also fine when the initial clause is not true.

joolswills: It may work, but it will mean the variable is used undeclared for non RPI platforms and if the variable had something in it in a parent function that data would be included. So this is not correct.

scriptmodules/emulators/jzintv.sh
joolswills: That's because the script in dosbox is included in `$romdir` which may be on a fat32 fs so this is needed in that case. As the script is moved now it's not needed here.

scriptmodules/emulators/jzintv.sh
cmitu: Leave it as-is then. A variable is used if the emulator supports multiple systems, but it's not the case here and I'd like that `intellivision` is explicit for both `addSystem` and `addEmulator.`

scriptmodules/emulators/jzintv.sh
cmitu: If the display is flipped, then wouldn't the emulator display also be flipped ? I don't see how this would be useful unless the emulator has a rotation feature (like ES or RetroArch).
For detecting if the display is rotated, I don't think that's possible outside X11 (using `xrandr`). For a RPI-03 with `dispmanx`, the display is rotated by the firmware and it's transparent to the system, for KMSDRM there's no global rotation configuration that works, each DRM client should handle the orientation hint (DSM prop) accordingly.
Leaving aside the rotation, which would bring something new, wouldn't it be better if the user decided (via `runcommand`'s video mode selection) the aspect ration ? Your patch would handle the letterboxing when the user selects a 4:3 aspect ration and we can pass %XRES% and %YRES% directly to `jzintv`, without the need for a wrapper script ?

scriptmodules/emulators/jzintv.sh
Gemba: Ok. I have put it as it was before the PR.

scriptmodules/emulators/jzintv.sh
Gemba: Ah. Now I understand. Corrected it.

scriptmodules/emulators/jzintv.sh
Gemba: I had no success when changing the patch line endings. `patch` hinted to use either `--ignore-whitespace` or `--binary`.
I changed it to `dos2unix` and limited the conversion to the source file which will be patched.

scriptmodules/emulators/jzintv.sh
cmitu: OK, that looks cleaner, but I think it would be cleaner to replace `bc` and `awk` calls with `python` (which is a system package and always available).
* The `bc` calculation can be replaced with:
   ``` shell
   local do_pillarboxing="$(python -c 'print(800/600>=4/3)')"
   if "$do_pillarboxing" == "True"; then 
  ...
   ```
* The `awk` calls can be done with `$(python -c 'printf(round(\$var*1.3333))')"`. This will make it easier to understand the intentions, since there's no need for workarounds to get a rounded value (i.e. add `+0.5` and count on `printf` to truncate down).

Gemba: Did not consider python in the first place but right, it removes a dependency and is less awk-ward, changed with pleasure. Used python3 explicitly as with python2 there would have been additional float/int casts needed.

cmitu: Yeah, I think it's simpler now. Thanks for the changes.

Gemba:
Flipped the content of the patch files as the filename was not reflecting the patch content.

joolswills:
I have added a commit to remove the source line conversion, and have adjusted the patches to CRLF so they apply cleanly. If you are happy, I will manually squash, update the commit message accordingly, and merge.

scriptmodules/emulators/jzintv.sh
```diff
@@ -44,6 +49,11 @@ function build_jzintv() {
     mkdir -p jzintv/bin
     cd jzintv/src
 
+    if isPlatform "rpi" ; then
+        local -r extra='EXTRA=-DPLAT_LINUX_RPI'
+    else
+        local -r extra=''
+    fi
```
joolswills: What is `extra` used for ? It doesn't seem to be used in the function?

joolswills:
Ah, the `extra` was used, but got removed when @Gemba  merged from master in the second commit. I will rebase it to restore it. This isn't ready for merge just yet.

Also I think I would prefer we use `awk` over python for those calculations. Although I do still feel like a lot of this would be better handled in the main code really.

cmitu (approved):
I think it's ok as is right now.

We could improve over by removing the `python` code; I think the reason `awk` was replaced it was the lack of a `round` function, which made the calculation a bit more complicated.

PR #3511, opened 2022-04-09
attractmode: dependency update

cmitu:
`libavresample-dev` has been obsoleted by `libswresample-dev` on recent (21.10 and later) Ubuntu versions and Debian Testing.
Also explicitely add the dependency on curl's libs.


joolswills:
Thanks.

PR #3513, opened 2022-04-10
love2d/love switched to main branch

timred:
love2d/love now using main (default) branch not master.
See: https://github.com/love2d/love/branches/all


joolswills:
Thanks.

PR #3514, opened 2022-04-10
PUAE: use 2021 version for ARM platforms

cmitu:
Since upstream switched to WinUAE 4.5.1, the performace requirements are higher for ARM platforms (Pi4).
Use the 'legacy'/2.6.1 branch [1] for ARM devices and track `master` for everything else.

[1] https://github.com/libretro/libretro-super/commit/0105d6bc2749b27e


joolswills:
I would prefer an additional module. I will sort this.

https://retropie.org.uk/forum/topic/32502/can-retropie-please-track-legacy-lr-puae-now/3?_=1649614461878

PR #3517, opened 2022-04-18
lr-flycast: add to arcade system

s1eve-mcdichae1:
**lr-flycast** can be used to play certain NAOMI and Atomiswave arcade roms from the MAME sets. Currently, it is set up to load these only from the `dreamcast` rom dir and not from `arcade`. I believe this should remedy that (tested RPi4).

### DIFF
--- a/scriptmodules/libretrocores/lr-flycast.sh
+++ b/scriptmodules/libretrocores/lr-flycast.sh
@@ -74,21 +74,13 @@ function configure_lr-flycast() {
         mkRomDir "$sys"
         ensureSystemretroconfig "$sys"
         addSystem "$sys"
+        # platform-specific
+        if isPlatform "gl"; then
+            iniConfig " = " "" "$configdir/$sys/retroarch.cfg"
+            iniSet "video_shared_context" "true"
+        fi
     done
     mkUserDir "$biosdir/dc"
-
-    # system-specific
-    if isPlatform "gl"; then
-        local configs=(
-            "$configdir/dreamcast/retroarch.cfg"
-            "$configdir/all/retroarch/config/Flycast/arcade.cfg"
-        )
-        local configfile
-        for configfile in $configs; do
-            iniConfig " = " "" "$configfile"
-            iniSet "video_shared_context" "true"
-        done
-    fi
 
     local def=0
     isPlatform "kms" && def=1
--- a/scriptmodules/libretrocores/lr-flycast.sh
+++ b/scriptmodules/libretrocores/lr-flycast.sh
@@ -69,20 +69,30 @@ function install_lr-flycast() {
 }
 
 function configure_lr-flycast() {
-    mkRomDir "dreamcast"
-    ensureSystemretroconfig "dreamcast"
-
+    local sys
+    for sys in dreamcast arcade; do
+        mkRomDir "$sys"
+        ensureSystemretroconfig "$sys"
+        addSystem "$sys"
+    done
     mkUserDir "$biosdir/dc"
 
     # system-specific
     if isPlatform "gl"; then
-        iniConfig " = " "" "$configdir/dreamcast/retroarch.cfg"
-        iniSet "video_shared_context" "true"
+        local configs=(
+            "$configdir/dreamcast/retroarch.cfg"
+            "$configdir/all/retroarch/config/Flycast/arcade.cfg"
+        )
+        local configfile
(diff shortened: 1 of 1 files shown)


cmitu:
This looks ok, though I'd rather have the `video_shared_context` option set at the system level, instead of setting it as folder override. I don't think other arcade core use HW rendering, like `flycast`, so it's safe to use it in the system's `retroarch.cfg`.

cmitu:
This looks ok, though I'd rather have the `video_shared_context` option set at the system level, instead of setting it as folder override. I don't think other arcade core use HW rendering, like `flycast`, so it's safe to use it in the system's `retroarch.cfg`.

s1eve-mcdichae1:
Like this?

cmitu:
Yes, that's what I had in mind.

scriptmodules/libretrocores/lr-flycast.sh
```diff
@@ -69,20 +69,22 @@ function install_lr-flycast() {
...
-        iniConfig " = " "" "$configdir/dreamcast/retroarch.cfg"
-        iniSet "video_shared_context" "true"
-    fi
-
     local def=0
     isPlatform "kms" && def=1
     # segfaults on the rpi without redirecting stdin from </dev/null
     addEmulator $def "$md_id" "dreamcast" "$md_inst/flycast_libretro.so </dev/null"
-    addSystem "dreamcast"
+    addEmulator 0 "$md_id" "arcade" "$md_inst/flycast_libretro.so </dev/null"
```
joolswills, line 89: addSystem should be called after addEmulator.

Note I am changing the setting of `video_shared_context` in this file as part of another PR (#3404) which I intend to look at again and update, so it may just be easier for me to implement this later or we can revisit when the other PR is done.

s1eve-mcdichae1: I see, because in "remove" mode, addSystem will remove it from the FE (ex: `es_systems.cfg`), but only if addEmulator has already removed the system's `emulators.cfg` first, yeah?

s1eve-mcdichae1: I see, because in "remove" mode, addSystem will remove it from the FE (ex: `es_systems.cfg`), but only if addEmulator has already removed the system's `emulators.cfg` first, yeah?

joolswills: Yep. That's correct.

s1eve-mcdichae1: Based on your WIP. What do you think of:

```
function configure_lr-flycast() {
    local sys
    local systems=(dreamcast arcade)
    for sys in $systems; do
        local def=0
        if isPlatform "kms" && [[ "$sys" == "dreamcast" ]]; then
            def=1
        fi
        # segfaults on the rpi without redirecting stdin from </dev/null
        addEmulator $def "$md_id" "$sys" "$md_inst/flycast_libretro.so </dev/null"
        addSystem "$sys"
    done

    for sys in $systems; do
        mkRomDir "$sys"
        ensureSystemretroconfig "$sys"

    # platform-specific
    if isPlatform "gl"; then
        iniConfig " = " "" "$configdir/$sys/retroarch.cfg"
        iniSet "video_shared_context" "true"
    fi
    done

    mkUserDir "$biosdir/dc"
}
```

Or, with `defaultRASystemConfig`:

```
function configure_lr-flycast() {
    local sys
    local systems=(dreamcast arcade)
    local def
    for sys in $systems; do
        def=0
        if isPlatform "kms" && [[ "$sys" == "dreamcast" ]]; then
            def=1
        fi
        # segfaults on the rpi without redirecting stdin from </dev/null
        addEmulator $def "$md_id" "$sys" "$md_inst/flycast_libretro.so </dev/null"
        addSystem "$sys"
    done

    local params=()
    # system-specific
    if isPlatform "gl"; then
        params+=("video_shared_context" "true")
    fi
    for sys in $systems; do
        mkRomDir "$sys"
        defaultRASystemConfig "$sys" "${params[@]}"
    done

    mkUserDir "$biosdir/dc"
}
```

Is it okay to have "local" inside the "for" loop like in the first one, or should I put `local def` outside the loop and only have `def=0` inside it like in the second?

Do I need quotes on `systems=("dreamcast" "arcade")` or, (probably not on `sys in "$systems"`, that would pass it as a single arg, right?)

PR #3522, opened 2022-05-05
On 64-bit OSs compiling Mupen64bit requires more swap

GeorgeMcMullen:
Hello,

I've recently been testing out some 64-bit OSes and different boards. In my testing, I found that mupen64bit would throw errors while compiling. After further investigation, I found that just increasing swap when on a 64-bit system would alleviate the issue. This is similar to how more swap is needed when compiling MAME on a 64-bit platform.

Side note: seems this pull request somehow also has merges from the base repo. It doesn't look like it causes any issues, but please let me know how I can clean it up if needed. Thanks in advance!


joolswills:
You should rebase it over master to get rid of the unwanted merge commits.

PR #3523, opened 2022-05-06
ppsspp/lr-ppsspp: (QOL) link save file dir for libretro core, and (BUG) resolve conflicts on removal

s1eve-mcdichae1:
Stand-alone already does this with this (shared) dir, but if you only use the lr-core, it doesn't link it.

### DIFF
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -172,10 +172,19 @@ function configure_ppsspp() {
     fi
 
     mkRomDir "psp"
-    moveConfigDir "$home/.config/ppsspp" "$md_conf_root/psp"
-    mkUserDir "$md_conf_root/psp/PSP"
-    ln -snf "$romdir/psp" "$md_conf_root/psp/PSP/GAME"
+    if [[ "$md_mode" == "install" ]]; then
+        moveConfigDir "$home/.config/ppsspp" "$md_conf_root/psp"
+        mkUserDir "$md_conf_root/psp/PSP"
+        ln -snf "$romdir/psp" "$md_conf_root/psp/PSP/GAME"
+    fi
 
     addEmulator 0 "$md_id" "psp" "pushd $md_inst; $md_inst/PPSSPPSDL ${extra_params[*]} %ROM%; popd"
     addSystem "psp"
+
+    # if we are removing the last remaining psp emu - remove the symlink
+    if [[ "$md_mode" == "remove" ]]; then
+        if [[ -h "$home/.config/ppsspp" && ! -f "$md_conf_root/psp/emulators.cfg" ]]; then
+            rm -f "$home/.config/ppsspp"
+        fi
+    fi
 }
--- a/scriptmodules/libretrocores/lr-ppsspp.sh
+++ b/scriptmodules/libretrocores/lr-ppsspp.sh
@@ -48,9 +48,16 @@ function configure_lr-ppsspp() {
         # the core needs a save file directory, use the same folder as standalone 'ppsspp'
         iniConfig " = " "" "$configdir/psp/retroarch.cfg"
         iniSet "savefile_directory" "$home/.config/ppsspp"
-        mkUserDir "$home/.config/ppsspp"
+        moveConfigDir "$home/.config/ppsspp" "$md_conf_root/psp"
     fi
 
     addEmulator 1 "$md_id" "psp" "$md_inst/ppsspp_libretro.so"
     addSystem "psp"
+
+    # if we are removing the last remaining psp emu - remove the symlink
+    if [[ "$md_mode" == "remove" ]]; then
+        if [[ -h "$home/.config/ppsspp" && ! -f "$md_conf_root/psp/emulators.cfg" ]]; then
+            rm -f "$home/.config/ppsspp"
+        fi
+    fi
 }


cmitu:
Can you expand on what QoL improvements the patch for the `libretro` core is adding ? I can only think of one thing - making the saves accessible over the `config` folder, which may be useful for backup/restore.

Removing user configurations is a no-no IMHO. Even if the user removes the emulator(s), we shouldn't remove the config folder - they may install the emulator (or core) later on and we don't want to zap their saves or configurations.
I'm also not sure why you're trying to replicate the `moveConfigDir ` function - is there that's not covered by it ?

s1eve-mcdichae1:
> Can you expand on what QoL improvements the patch for the `libretro` core is adding ? I can only think of one thing - making the saves accessible over the `config` folder, which may be useful for backup/restore.

Yes, exactly that; so that saved games will persist across backups.

You've said this before, to the effect of "the only reason I could think to do this would be if you want to backup your saves." As if that's not reason enough by itself, but while it is, in fact, the exact and only reason I've done it. Is that not something people would normally want to have done? Don't you backup your saves? I do want to.

> Removing user configurations is a no-no IMHO. Even if the user removes the emulator(s), we shouldn't remove the config folder - they may install the emulator (or core) later on and we don't want to zap their saves or configurations.

I'm not removing any configs. I'm ensuring the link *to* the configs does *not* get removed prematurely. The actual saves and configs are still intact in `$md_conf_root`, only the link *to* them in `$home/.configs` is removed (just as `moveConfigDir` would do) and only if nothing else is using it. At least, that was my intent. Is that not what I've done?

> I'm also not sure why you're trying to replicate the `moveConfigDir ` function - is there some thing that is not covered by the existing helper ?

1: Stand-alone and libretro-core both use the same `$home/.configs/ppsspp` dir for saves/configs. Stand-alone already links this with `moveConfigDir`; libretro doesn't, but if both are installed (regardless of order), the end result is that the configs exist in `$md_conf_root` and a symlink is left behind in `$home/.configs`.

2: If, then, stand-alone (or, after my first change here, either one) is removed, then the symlink is removed by `moveConfigDir` at https://github.com/RetroPie/RetroPie-Setup/blob/5cf9b95ab58f0bf6e808b58f3fcb753e17d3906f/scriptmodules/helpers.sh#L526-L530

3: This leaves the remaining (libretro or, after my first change, either) emulator looking for configs in `$home/.configs` where it can't find them because there are now neither actual configs nor the symlink to them at that location.

THE PROBLEM:

The problem, as I can see it, is in using `moveConfigDir`, in "remove" mode, on a location that is still used by another emulator.

The solution, is to not do that; we can't use `moveConfigDir` in "remove" mode on a location that is still used by another emulator.

I suppose the "install" part doesn't need to be replicated, we could just use the actual helper as long as it's nested inside an "install" section so it's not called on removal.

On "remove," we can't use the helper. We could either check that nothing else is using the symlink before removing it as I've done (I just checked for an `emulators.cfg`; if it's not there, it means there are no remaining PSP emulators and so it should be safe to remove the link), or just not remove it ever I suppose, but it looks like attempts are usually made to clean up after ourselves and so I was attempting to preserve that.

cmitu:
> Yes, exactly that; so that saved games will persist across backups.
> 
> You've said this before, to the effect of "the only reason I could think to do this would be if you want to backup your saves." As if that's not reason enough by itself, but while it is, in fact, the exact and only reason I've done it. Is that not something people would normally want to have done? Don't you backup your saves? I do want to.

You've read this wrong - I'm not debating whether this is an improvement or not. I was asking that the improvement(s) should be mentioned in the commit/PR, so it's easier to find the reason the new code. So if there were other advantages of symlink-ing the config in the `$md_conf_root`, they should be expanded on. 

> > Removing user configurations is a no-no IMHO. Even if the user removes the emulator(s), we shouldn't remove the config folder - they may install the emulator (or core) later on and we don't want to zap their saves or configurations.
> 
> I'm not removing any configs. I'm ensuring the link _to_ the configs does _not_ get removed prematurely. The actual saves and configs are still intact in `$md_conf_root`, only the link _to_ them in `$home/.config` is removed (just as `moveConfigDir` would do) and only if nothing else is using it. At least, that was my intent. Is that not what I've done?
> 
> > I'm also not sure why you're trying to replicate the `moveConfigDir ` function - is there some thing that is not covered by the existing helper ?
> ...
> The problem, as I can see it, is in using moveConfigDir, in "remove" mode, on a location that is still used by another emulator.

Ok, now I see the issue. You latest code iteration does it better than the previous version.
Yes, removing `$home/.config/ppsspp` will only remove the symlink, though I'm not sure it's really needed.

s1eve-mcdichae1:
> You've read this wrong - I'm not debating whether this is an improvement or not. I was asking that the improvement(s) should be mentioned in the commit/PR, so it's easier to find the reason for adding the new code in the future. So if there were other advantages of symlink-ing the config in the `$md_conf_root`, they should be expanded on.

Aye, understood. Yeah that (having network/backup access to saved games) was really all I had in mind for this, before I noted the conflict that occurs on removal.

s1eve-mcdichae1:
> Yes, removing `$home/.config/ppsspp` will only remove the symlink, though I'm not sure it's really needed.

Removing the link is what `moveConfigDir` *would* do if it was used here, and it is only *not* used here because the link shouldn't be removed *too early*. But it *can* be removed once nothing is using it, just like all the other emulators that *do* use the helper function.

It does seem to be working as intended. I remove just one PSP emulator, and the link remains. When I remove the remaining emulator, the link is removed with it. My `.configs` dir is cleaner. My savegames remain in `/opt/retropie/...` and when I reinstall the emulator, it links back to them.

Do you think it's too much?

cmitu:
> Do you think it's too much?

I'm ok either way.

joolswills:
Thanks.

PR #3524, opened 2022-05-06
Update srb2.sh

dpark12300:
Simplified and added latest build 2.2.10

### DIFF
--- a/scriptmodules/ports/srb2.sh
+++ b/scriptmodules/ports/srb2.sh
@@ -7,53 +7,13 @@
 #
 # See the LICENSE.md file at the top-level directory of this distribution and
 # at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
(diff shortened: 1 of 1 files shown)


dpark12300:
I added the official install code.

[http://wiki.srb2.org/wiki/Source_code_compiling/Makefiles](url)

dpark12300:
This is adding the 2.2.10 patch/fix from the developers, STJR.

cmitu:
Sorry, but this is not how a scriptmodule works in RetroPie-Setup - you've deleted all the required functions that are used by RetroPie-Setup.

If you wish to bump the version to `2.2.10`, then you should change
```
rp_module_repo="git https://github.com/STJr/SRB2.git SRB2_release_2.2.9"
```
to
```
rp_module_repo="git https://github.com/STJr/SRB2.git SRB2_release_2.2.10"
```

PR #3526, opened 2022-05-14
emulationstation: build omxplayer only on dispmanx platforms

cmitu:
Starting with RaspiOS 'bullseye', the `omxplayer` RPI video player is no longer supported.
Use the new EmulationStation build option (`OMX`) to enable the `omxplayer` bits only for `dispmanx` platforms. The new options is added in https://github.com/RetroPie/EmulationStation/pull/792. The `RPI` build option is used only for pre-setting some program settings on the Raspberry Pi platforms (audio/video memory).

Complements https://github.com/RetroPie/EmulationStation/pull/792.


joolswills:
Thanks.

PR #3527, opened 2022-05-14
Create supertuxkart.sh

derryktutt:
Tell me if I need anything.

### DIFF
--- a/scriptmodules/ports/supertuxkart.sh
+++ b/scriptmodules/ports/supertuxkart.sh
@@ -1,0 +1,34 @@
+
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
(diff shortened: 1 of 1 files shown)


s1eve-mcdichae1:
RetroPie-Extra has an STK module that's using an `xorg` window; does this run full-screen without it?

https://github.com/Exarkuniv/RetroPie-Extra/blob/master/scriptmodules/ports/supertuxkart.sh

s1eve-mcdichae1:
RetroPie-Extra has an STK module that's using an `xorg` window; does this run full-screen without it?

https://github.com/Exarkuniv/RetroPie-Extra/blob/master/scriptmodules/ports/supertuxkart.sh

s1eve-mcdichae1:
Actually, this isn't loading at all for me on RPi4 without using `xorg`, it just kicks me back to ES. Are you testing on x86, or...?

Also, I think you need a fourth arg for `addPort`, so it can configure the `emulators.cfg` command, like:

    addPort "$md_id" "supertuxkart" "SuperTuxKart" "/usr/games/supertuxkart"

PR #3528, opened 2022-05-15
system - use architecture for __platform instead of x86

joolswills:
system - use architecture for __platform instead of x86
    
For architectures not handled by custom logic, we previously checked `uname -m` for i686/x86_64/amd64 and then set __platform=x86.
    
This change switches to using the actual output of `uname -m` as the `__platform` variable. If a`platform_$__platform` function exists, we will call it, otherwise we call platform_native (which used to be platform_x86).
    
In platform_native we check if the platform is i386/i686/x86_64 and then set an x86 flag.
    
This seems more logical and allows us to more easily check for x86_86 in scriptmodule flags. It also resolves redream incorrectly having x86_64 added to the flags in d877a81 (which work after this change).
    
The error: fatalError "Unknown platform - please manually set the __platform variable to one of the following is now removed. As we default to platform_native this could potentially allow RetroPie-Setup to work on something we don't specifically support.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -430,18 +430,10 @@ function get_platform() {
                             ;;
                     esac
                 else
-                    case $architecture in
-                        i686|x86_64|amd64)
-                            __platform="x86"
-                            ;;
-                    esac
+                    __platform="$architecture"
                 fi
                 ;;
         esac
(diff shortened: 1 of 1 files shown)


joolswills:
I have changed the way this is implemented.

PR #3530, opened 2022-05-16
ps2: add support for CHD disc images

cmitu:
Supported by `pcsx2` since https://github.com/PCSX2/pcsx2/pull/4314.


joolswills:
Thanks.

PR #3531, opened 2022-05-17
splashscreen.sh: (QOL) updated menu interface

s1eve-mcdichae1:
Updated wording to remove ambiguity.

Slightly reworked randomizer function.

Added mode label (custom, etc.) to randomizer menu title.

Fixed path dialog on randomizer choices.

Disable randomizer upon selecting default splashscreen.

### DIFF
--- a/scriptmodules/supplementary/splashscreen.sh
+++ b/scriptmodules/supplementary/splashscreen.sh
@@ -185,6 +185,7 @@ function choose_splashscreen() {
 
 function randomize_splashscreen() {
     options=(
+        0 "Disable splashscreen randomizer"
         1 "Randomize RetroPie splashscreens"
         2 "Randomize own splashscreens (from $datadir/splashscreens)"
         3 "Randomize all splashscreens"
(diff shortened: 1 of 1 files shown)


joolswills:
If we are going to use `currently:` for `Show splashscreen on boot` then I think it would look better to use this for option 3 - `Splashscreen randomizer` also perhaps ? But maybe it would just be easier to drop the "currently".

joolswills:
Apart from that it looks good to me - thanks.

s1eve-mcdichae1:
I'm probably putting way too much thought into a simple menu option, but here's what I was thinking.

As a toggle, it felt awkward to me. Without "currently:", I am expecting it to open up another menu and make a selection from available options. Somehow, that extra word signals to me that it just flips on and off like a switch, instead.

The randomizer function, *does* open up another menu where you choose an option so it didn't feel necessary but I'm not opposed to add it, either. How about:

    3 "Randomizer options (currently: ${ini_value^})

I think this also allows to eliminate the use of `$random` variable, too.

(...also, should option 4 "Use default splashscreen" additionally set the splashscreen to "enabled"?)

joolswills:
> The randomizer function, _does_ open up another menu where you choose an option so it didn't feel necessary but I'm not opposed to add it, either. How about:
> 
> ```
> 3 "Randomizer options (currently: ${ini_value^})
> ```
> 
> I think this also allows to eliminate the use of `$random` variable, too.
> 
> (...also, should option 4 "Use default splashscreen" additionally set the splashscreen to "enabled"?)

ok - I'm happy with that also.

joolswills:
Thanks :)

PR #3532, opened 2022-05-18
Macintosh: (BUG) Add missing exts for Basilisk

s1eve-mcdichae1:
Ref:

https://github.com/RetroPie/RetroPie-Setup/blob/1eedc5ad2c1f344182d425ed8ec04c4664b1db12/scriptmodules/emulators/basilisk.sh#L14

https://www.reddit.com/r/RetroPie/comments/urxlb0/emulating_macintosh_no_roms_shows_up

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -100,7 +100,7 @@ intellivision_fullname="Intellivision"
 love_exts=".love"
 love_fullname="Love"
 
-macintosh_exts=".txt .dsk"
+macintosh_exts=".txt .dsk .img .rom"
 macintosh_fullname="Apple Macintosh"
 
 mame_exts=".zip .7z"


cmitu:
BasiliskII is always started with `disk.img` and `mac.rom`; adding more entries to the list would work just for listing purposes, since launching them would not produce anything different then 'Start Basilisk'.

`.rom` files are firmware files, not game ROMs, so I don't see why they should be exposed. I guess we can accomodate listing `.img` files, but without an adjusted emulator command it may cause confusion for users that would expect launching the emulator for the `.img` to actually _run_ the image.

s1eve-mcdichae1:
@cmitu ohh, yeah I see how that works, now.

So when this person said "this works," they hadn't gotten any farther than seeing the list of roms, to actually launch one.

Well I guess I can forgive them, because **I** didn't get any farther than "somebody else said it worked."

Issue #963, opened 2015-08-15
[RP3.0] Emulationstation doesn't recognize zipped Wonderswan games


zerojay:
When launching Emulationstation, if my Wonderswan games are zipped, a Wonderswan section does not show up whatsoever. When unzipped, they are recognized and added to Emulationstation.

Likely a small change to a list of recognized ROM extensions.

HerbFargus:
Thank you for the input. I have submitted a fix: https://github.com/RetroPie/RetroPie-Setup/pull/964

Issue #962, opened 2015-08-15
[RP3.0] lr-handy segmentation faults on launch every time


zerojay:
Running the lr-handy core with any verified good Lynx dump (No-Intro) results in the lr-handy core outputting the following:

Invalid cart.
Invalid ROM.

And then segfaults. Does not appear to make a difference if the ROMs are zipped or unzipped.

HerbFargus:
I just added a fresh image and I've had no problems. you may have to add lynxboot.img to the bios folder.

joolswills:
Works for me too.

zerojay:
I already have lynxboot.img in the BIOS folder (which is not mentioned in the Lynx setup page on the Wiki, by the way), also taken from the No-Intros rom set and it's still segfaulting on launch.

zerojay:
This appears to be an upstream bug with the original Handy core. If your Lynx ROMs do not contain a header - as No-Intros do not - Handy crashes and burns.

As there is no work being done on Handy anymore, it may be a good idea to mention on the Lynx wiki page that the core will only work with ROMs that have a header.

HerbFargus:
Here's the current repo.

https://github.com/libretro/libretro-handy

I'll make a note of it in the wiki.

PR #3533, opened 2022-05-22
amiberry: update to version 5.2

cmitu:
* New:
  - Added new CD32 config with 8MB Fast, use that with --autoload option
  - Enable RTC automatically if enabled expansions have RTC
  - Support for Cycle-Exact mode from XML options (#959)
  - Ability to set default soundcard (#963)

* Bugs fixed
  - Hotkey mappings didn't work with controller buttons (#949)
  - Only check for hotkey buttons if they have been mapped (#949)
  - When using Auto-crop and virtual mouse driver, the pointer would not reach the edges of the screen (#962)
  - Launch Amiberry binary directly in macOS app bundle (#970)

* Improvements
  - Updated bundled AROS ROMs to latest version
  - Updated WHDBooter XML and WHDLoad to latest versions
  - Picasso96 new features are now optional. DACSWITCH fix.


nemo93:
@cmitu thank you so much for the lightspeed PR creation! And amazing support as always. I'm following closely the controller issue(s) some users are reporting. 5.2 should address those issues indeed. Also emulation accuracy with few games will be improved. Nothing much to add except I got 5.2 running great for some time now still without issues. Thanks again!

joolswills:
Thanks.

joolswills:
RPi binaries are updated.

PR #3535, opened 2022-05-23
lr-flycast: (QOL) add to Arcade system (redux)

s1eve-mcdichae1:
Ref: https://github.com/RetroPie/RetroPie-Setup/pull/3517

(Revisited for `defaultRAConfig`)

### DIFF
--- a/scriptmodules/libretrocores/lr-flycast.sh
+++ b/scriptmodules/libretrocores/lr-flycast.sh
@@ -69,15 +69,20 @@ function install_lr-flycast() {
 }
 
 function configure_lr-flycast() {
-    local def=0
-    isPlatform "kms" && def=1
-    # segfaults on the rpi without redirecting stdin from </dev/null
-    addEmulator $def "$md_id" "dreamcast" "$md_inst/flycast_libretro.so </dev/null"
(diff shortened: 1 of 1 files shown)


joolswills:
Why are you not using the original PR ?

s1eve-mcdichae1:
> Why are you not using the original PR ?

Still getting my GIT legs, I guess. I figured the old one was outdated since the code had changed "underneath" it, so to speak. So I made this new one based on the most recent version.

is it preferable, then, to continue building off the original, even when the current version diverges from that?

joolswills:
It's possible to make the changes in the old branch, then squash them together and force push back. But if new to git, that's ok. This change looks good thanks - I will give it a test then merge.

joolswills:
I will squash this also.

cmitu:
@s1eve-mcdichae1 don't use the web interface. 
From the CLI, run `git rebase -i <commit-ish>` to squash all commits (including the one chosen) into a new one. Then push again (using `--force`) to your branch the new commit.

s1eve-mcdichae1:
@cmitu huzzah!

joolswills:
Thanks.

PR #3536, opened 2022-05-28
xpadneo: upgrade to v0.9.2

cmitu:
New stable version:

* improved compatibility with the new BLE enabled controller firmware
* better compatibility with kernels 5.12 or later and latest DKMS versions. On kernels 5.12 or later should no longer Bluetooth ERTM is not unconditionally disabled anymore. For better stability, it's generally recommended to upgrade controllers to the new BLE firmware variant (version 5.x and above) for improved connection stability
* fixed rumble throttling for modern (BLE enabled) firmware
* fixed an issue conforming to the Linux input spec better which improves compatibility with jinput
* improved documentation


joolswills:
Thanks.

PR #3538, opened 2022-06-02
xpadneo: upgrade to v0.9.3

cmitu:
This is a maintenance release, fixing the Xbox button for SDL apps using the Gamecontroller API.


joolswills:
Thanks

PR #3541, opened 2022-06-09
simcoupe: update the branch selection to fix building

cmitu:
Updated the default branch, since upstream changed the project repository's default to 'main'.
Since the `cmake` requirements have been bumped to 3.14 some time ago, select the closest tag before the change in order to fix building on Ubuntu 18.04.
Note also that current code requires C++17 'filesystem' support, which is not present in Ubuntu 18.04.

As an aside, the reason for the `cmake` upgrade was the support for _FetchContent_. Dependencies are now downloaded at build time, so we'll be able to remove the `libspectrum-dev` dependency at some point, when removing support for older `cmake` versions.


joolswills:
Thanks.

PR #3542, opened 2022-06-10
Run as root by default

twojstaryzdomu:
Requiring users to type sudo is unnecessary. This patch makes running the script a bit easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,9 @@ The script is executed with:
 
 ```shell
 cd RetroPie-Setup
-sudo ./retropie_setup.sh
+./retropie_setup.sh
 ```
+Note: the script is running as root by default.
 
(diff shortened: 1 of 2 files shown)


joolswills:
Sorry but I don't want to change this.

twojstaryzdomu:
> Sorry but I don't want to change this.

It isn't a reasonable argument. Why do you require users to type an extract command that adds no value here really?

priiduneemre:
> It isn't a reasonable argument. Why do you require users to type an extra command that adds no value here really?

As not to obfuscate the context and make it verbatim what/how is really executed? Typing `sudo` should be a concious decision, not hidden in a script somewhere. IMHO, the [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) applies well here.

PR #3543, opened 2022-06-10
Suppress some confirmation prompts with IAMSURE shell env

twojstaryzdomu:
Following menu item selection, some prompts (e.g. to update or install) are     redundant and it ought to be possible to suppress them. This patch enables     suppressing such prompts by setting the IAMSURE environment variable. Typically users who have seen such a prompt more than once will want this. Users who run the script explicitly with 'sudo' need to use '-E' parameter to export the current environment properly.

The other PR for running as root by default ensures the script picks up a non-root user's environment properly without the need to specify 'sudo -E' at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,9 @@ The script is executed with:
 
 ```shell
 cd RetroPie-Setup
-sudo ./retropie_setup.sh
+./retropie_setup.sh
 ```
+Note: the script is running as root by default.
 
 When you first run the script it may install some additional packages that are needed.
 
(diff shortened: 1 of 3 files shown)


joolswills:
You can use `__nodialog=1` - and retropie_packages.sh directly bypassing the retropie_setup.sh for automating updates. Sorry but I won't want to include this.

twojstaryzdomu:
@joolswills care to explain why? Your proposal is an internal variable that suppresses all dialogs while mine suppresses a select few most users would care about. It is also safer as it leaves the removal dialogs intact.

PR #3544, opened 2022-06-10
Non-interactive convenience scripts for updates

twojstaryzdomu:
Scripts allowing updating either the packages or the script itself without
invoking the GUI. The scripts execute the main functions of the setup module.

Users may want to schedule them as
`cd ~/RetroPie-Setup && ./retropie_update_script.sh && ./retropie_update.sh`
to enable non-interactive updates, possibly capturing standard output for
later inspection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,9 @@ The script is executed with:
 
 ```shell
 cd RetroPie-Setup
-sudo ./retropie_setup.sh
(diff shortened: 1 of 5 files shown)


joolswills:
This includes unrelated commits - but it's not functionality I want to include. Sorry.

PR #3546, opened 2022-06-12
emulationstation: remove non existent branch v2.7.6

gizmo98:
Remove deprecated debian version check. Our emulationstation repo has no v2.7.6 branch anymore. If branch is not existent default branch "master" is used. Fix this and always use branch "stable".


cmitu:
I don't think the repo had a `v2.7.6` branch, `gitPullOrClone` pulls the `v2.7.6` [tag](https://github.com/RetroPie/EmulationStation/tree/v2.7.6) for older Debian versions.

gizmo98:
You are right. I have only looked after branches.

PR #3547, opened 2022-06-12
emulationstation: fix OpenGL 2.1 for x11

gizmo98:
Add new build option for OpenGL 2.1 renderer on x11 platforms.
https://github.com/RetroPie/EmulationStation/pull/795/files


joolswills:
Thanks.

Issue #3161, opened 2020-06-07
Problems with netplay guest


alessandropaggi:
I've posted many times on the forum about this issue, but I've still got no answer so here I am. I can't connect to netplay rooms with Retropie despite running the same core and the same rom (same CRC) of the host. Every time I try to connect to a netplay room I get the the messages:
"Compatible content found"
"Couldn't find a suitable core or content file, load manually"
I've also ran this test: I am running Retropie on a Pi4 and Retroarch on my laptop (same Retroarch version, same core, same rom). I did not perform any rom scan on neither system (see below), I was just launching the same rom on the same emulator. On the host side I went in Retroarch menu Netplay->Host->Start Netplay Host. On the guest side I went in Retroarch menu Netplay->Refresh Room List and selected the host room.

- Retropie is able to serve as a netplay host, that is, I was able to join the Retropie netplay host room as a guest from Retroarch running on the laptop.

- Retropie cannot join netplay as a guest, that is, when I try to join the host room of Retroarch running on the laptop, I get the messages shown before. This also happen on any netplay host room I see in the room list.

I also tried installing Retropie on an Ubuntu virtual machine, obtaining the same results.
I don't know if this related, but I also tried to create a playlist by scanning the rom directory, by going in Retroarch menu Main Menu->Load Content->Playlist->Scan Directory, selecting the rom directory and then Scan This Directory. On Retroarch running on the laptop this correctly created a playlist with the roms found in the directory, while Retropie is not able to find any content.

cmitu:
I managed to reproduce this, but it's not fully clear why it happens. The content is matched based on CRC (so I think a playlist won't help), but if fails with 
```
[...]
[INFO] [CONTENT LOAD]: CRC32: 0xc3508f7e .
[INFO] [Lobby]: Testing CRC matching for: C3508F7E|crc
[INFO] [Lobby]: Current content CRC: C3508F7E|crc
[INFO] [Lobby]: CRC match C3508F7E|crc with currently loaded content
[INFO] [Lobby]: Couldn't find a suitable core
```
which is puzzling, since the core is already loaded (!). The error message is from https://github.com/libretro/RetroArch/blob/01bd76405fc1441c200de54f5cfe920c6da44fcd/tasks/task_netplay_find_content.c#L145, but tracing back to see where `state->core_path` is set didn't show anything in particular.

The error shows up even with a vanilla RetroArch configuration (i.e. no RetroPie specific configs).
[...]

PR #3548, opened 2022-06-13
runcommand: add more parameters for RetroArch

cmitu:
Added 2 new parameters for the runtime-generated `retroarch.cfg`:

* `libretro_log_level` will add more debug messages from the libretro core, in addition to the frontend debugging messages.
   Cand be useful to diagnose core issues, since `--verbose` activates just the frontend logging.

* `libretro_directory` is set to the core file folder (i.e. `/opt/retropie/libretrocores/lr-<core>`).
   It is used by during netplay, as a client, when RetroArch attempts to re-load the core used during netplay.
   Basically, when connecting to a remote netplay host, RetroArch will try to locate a `core.info` file and then try to re-load the core file. If any of them is missing, the connection fails with "Couldn't find a suitable core or content file".

   The netplay issue was reported and diagnosed in #3161. Rather than symlink each installed libretro core to `$HOME/.config/retroarch/cores` (the default value for `libretro_directory`), we'll just set it at runtime with the above parameter. We still need to add the `.info` files for the installed cores, but that's a separate patch.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -701,6 +701,7 @@ function main_menu() {
                 ;;
             L)
                 COMMAND+=" --verbose"
+                VERBOSE=1
                 return 0
                 ;;
             U)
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -1101,6 +1101,14 @@ function retroarch_append_config() {
         iniSet "video_fullscreen_y" "${dim[1]}"
     fi
 
+    # set `libretro_directory` to the core parent folder
+    local core_dir=$(echo "$COMMAND" | grep -Eo " $ROOTDIR/libretrocores/.*libretro\.so " | head -n 1)
+    core_dir=$(dirname "$core_dir")
+    [[ -n "$core_dir" ]] && iniSet "libretro_directory" "$core_dir"
+
+    # if verbose logging is on, set core logging to INFO
+    [[ $COMMAND =~ "--verbose" ]] && iniSet "libretro_log_level" "1"
```
joolswills, line 1110: Might be safer/easier to set a global variable when we add the --verbose to avoid the comparison.

cmitu: Makes sense - I added a global var, similar to how `$NETPLAY` is set.

joolswills:
Thanks.

PR #3549, opened 2022-06-13
retroarch: add core info files to configuration

cmitu:
Core info files are needed starting from 1.10.1 in order to enable save/load state, rewind, runahead, (disk control ?).
In addition to that, they're needed for netplay clients to work (see issue #3161).

NOTE: this PR and #3548 should fix issue #3161.


cmitu:
Adding the info files will enable the start-up info to pop-up:

image

If we don't want that, I can add the `retroarch.cfg` parameter to disable it, should be easy to find (tm).

joolswills:
I'm not sure. It may be best to disable it by default to keep current behaviour.

cmitu:
@joolswills ok, added the config option to quiesce the info blurb.

I also checked the libretro cores that we ship and the upstream repository is missing a few `.info` files:
 - `mess` and `mess2016`. We can clone them based on `mame` and `mame2016`.
 - `superflappybirds`. No idea what are the capabilities, I'll have to give it a spin.
 - `mupen64plus`. We can clone it from `mupen64plus-next`.

My idea for the missing core info files is to generate them from the scriptmodule, via `configure`, instead of adding them in `retroarch`. Apart from the `mupen64plus` core, they're not installed by default.

joolswills:
Thanks.

PR #3552, opened 2022-06-15
helpers.sh & bashwelcometweak.sh: avoid proprietary DNS hosts to determine own IP

twojstaryzdomu:
The `getIPAddress` function shouldn't depend on third party proprietary DNS servers that are not part of any Internet standard and may be blocked in some jurisdictions for political reasons.

If the servers behind 8.8.8.8 are inaccessible for any reason, technical or otherwise, the parts of the script that require the IP address to be known won't be functional. It is far more prudent to rely on the 13 root DNS servers that are independent of a company's whim and less likely to be blocked across the world.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -1618,16 +1618,28 @@ function dkmsManager() {
 ## @details This function first tries to obtain an external IPv4 route and
 ## otherwise tries an IPv6 route if the IPv4 route can not be determined.
 ## If no external route can be determined, nothing will be returned.
-## This function uses Google's DNS servers as the external lookup address.
+## This function uses no proprietary DNS, resorting to root name servers to determine a route.
 function getIPAddress() {
     local dev="$1"
     local ip_route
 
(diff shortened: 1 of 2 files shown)


joolswills:
> If the servers behind 8.8.8.8 are inaccessible for any reason, technical or otherwise, the parts of the script that require the IP address to be known won't be functional.

This isn't true. It doesn't matter if there is anything online at 8.8.8.8. The IP used isn't connected to.

twojstaryzdomu:
> > If the servers behind 8.8.8.8 are inaccessible for any reason, technical or otherwise, the parts of the script that require the IP address to be known won't be functional.
> 
> This isn't true. It doesn't matter if there is anything online at 8.8.8.8. The IP used isn't connected to.

It does matter, 8.8.8.8 is a third party, proprietary DNS and as such prone to all sorts to nefarious misdeeds by a rogue man in the middle.

And what if a router owner chooses to set up a static route to 127.0.0.1 to force DNS off of 8.8.8.8?

Please do not link against 8.8.8.8 in code as its bad practice. Using the standard set of root servers is more proactive.

PR #3555, opened 2022-06-15
runcommand.sh: refactor output directory into a global variable, use /tmp as fallback

twojstaryzdomu:
This is to allow another directory for `runcommand.sh` outputs for cases when outputs are worth preserving (i.e. crash). Also, /dev/shm may not be accessible on all environments, for user discretionary reasons. When `/dev/shm` doesn't exist or isn't writable, Fall back to `/tmp`.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -79,7 +79,11 @@
 
 ROOTDIR="/opt/retropie"
 CONFIGDIR="$ROOTDIR/configs"
-LOG="/dev/shm/runcommand.log"
+OUTPUTDIR="/dev/shm"
+[ -d "${OUTPUTDIR}" ] \
+  && [ -w "${OUTPUTDIR}" ] \
+    || OUTPUTDIR=/tmp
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -79,7 +79,11 @@
 
 ROOTDIR="/opt/retropie"
 CONFIGDIR="$ROOTDIR/configs"
-LOG="/dev/shm/runcommand.log"
+OUTPUTDIR="/dev/shm"
+[ -d "${OUTPUTDIR}" ] \
+  && [ -w "${OUTPUTDIR}" ] \
+    || OUTPUTDIR=/tmp
```
joolswills, line 85: Note we use `[[` for RetroPie bash scripts.

joolswills:
I would prefer to make the log directory configurable rather than just falling back to another location. I will do a change for this.

scriptmodules/supplementary/runcommand/runcommand.sh
twojstaryzdomu: It's a bit of a moot comment since you're not merging it unless you want to be picky.

Issue #3554, opened 2022-06-15
runcommand.sh: does not capture logs when `/dev/shm` is unavailable or restricted


twojstaryzdomu:
`/dev/shm` isn't traditionally the location to dump data to. On some environments it may not be available (it is after all an in-memory shared location). Please enable logs to be written elsewhere (`/tmp`) if `/dev/shm` isn't available.

cmitu:
This is by design - a memory backed location was chosen in order to prevent writing to the sdcard on the Pi (the primary usage for RetroPie). 
From the systems that RetroPie supports, is there a distro where `/dev/shm` is not available and writable ?

twojstaryzdomu:
> 
I do not question the intent. It's all very well and in fact what most users want, not to write to flash too often.

I question the code - do you need to hard-code /dev/shm every time in your script? Why not define it via a variable and set it at the top of the script, enabing an easy change to a different location.

Let me give you another scenario: my game crashes the box and I would like it to capture output to disk. Nothing you write to memory would be preserved across reboots. Do you understand this case?

twojstaryzdomu:
> I think it would just be simpler to allow this to be set in the runcommand configuration file - your change is just hardcoding more locations. I don't think it's a big issue though. I can make a suitable change.

Very well but you haven't got rid of the hard-coded `/dev/shm` that are all over the place in `runcommand.sh`.

joolswills:
> 

I missed two references which I will fix.

twojstaryzdomu:
> > 
> 
> I missed two references which I will fix.

Thanks! Please see #3557, I think it's only fair that it's a value that a user may modify easily via the menu like the others.

PR #3556, opened 2022-06-15
runcommand - allow the runcommand log directory to be changed.

joolswills:
If an entry log_dir is made in /opt/retropie/configs/all/runcommand.cfg it will be used by runcommand for saving runcommand.log/runcommand.info files instead of the default /dev/shm

#3554


twojstaryzdomu:
Very good but if you want to do it right, you need to:
1. Get rid of the hard coded `/dev/shm`. Have a single log/scratch area.
2. Add a menu item that configures this, similar to the other menu items.

scriptmodules/supplementary/runcommand/runcommand.sh
```diff
@@ -119,6 +118,10 @@ function get_config() {
         iniGet "image_delay"
         IMAGE_DELAY="$ini_value"
         [[ -z "$IMAGE_DELAY" ]] && IMAGE_DELAY=2
+        iniGet "log_dir"
+        LOGDIR="$ini_value"
+        [[ -z "$LOGDIR" ]] && LOGDIR="/dev/shm"
```
twojstaryzdomu, line 123: Please modify `[[ -z "$LOGDIR" ]] && LOGDIR="/dev/shm"` to `[[ -d "$LOGDIR" ]] || LOGDIR="/dev/shm"`.

If the `$LOGDIR` is unset or not a directory, default to `/dev/shm`. Though mind `/dev/shm` isn't guaranteed to be mounted. `/tmp` would be the safest bet.

twojstaryzdomu:
@joolswills your commit is inconsistent with the other options being controlled via the GUI but not this one.
The commit #3557 added the GUI configuration part for the parameter but regrettably you have chosen to close it.

PR #3557, opened 2022-06-16
runcommand: replace hard coded /dev/shm with log_dir, add a runcommand configuration menu item

twojstaryzdomu:
This PR extends #3556 with a retropie menu configuration entry for the log/scratch area `log_dir` variable and replaces the remaining hard-coded `/dev/shm` values with the `log_dir` variable.

### DIFF
--- a/scriptmodules/supplementary/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand.sh
@@ -43,6 +43,7 @@ function install_bin_runcommand() {
         iniSet "governor" ""
         iniSet "disable_menu" "0"
         iniSet "image_delay" "2"
+        iniSet "log_dir" ""
         chown $user:$user "$configdir/all/runcommand.cfg"
     fi
     if [[ ! -f "$configdir/all/runcommand-launch-dialog.cfg" ]]; then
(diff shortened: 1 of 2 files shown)


joolswills:
I'm not sure it needs to be in the GUI. But if I will add that I would prefer to do it in my previous PR rather than yours based on mine.

twojstaryzdomu:
I would consider the PR on its sole merit, not the fact of not being its author.

What is the case for the other runcommand.sh options in the GUI that have their configuration file equivalent? I considered the case of users who need persistency for their logs across crashes and the ability to update it quickly.
Your PR requires to dabble in manual editing configuration files, so it is hacky and inconsistent with the fact other options can be changed in the GUI.

twojstaryzdomu:
> Well I'd already started on the changes. I don't think it needs to be multiple commits and I'd like to implement it in the way I want. If that's ok.

Absolutely, you're entitled to prefer your own commits, even if they could be done better.

Had your PR been up to scratch, I wouldn't have submitted mine. The inconsistencies with the other configuration items being in the GUI is a fact.

It's up to you to decide but with my multiple PRs closed without merging. 

I just wonder if it's a pet project thing where maintainers settle for worse regardless of the actual merit. Is there a systemic problem with external contributions?

PR #3558, opened 2022-06-20
runcommand: remove spurious space from `libretro_directory`

cmitu:
Fix the extra spaces added in the value of `libretro_directory`,caused by the `grep` pattern in 553c4fef.
The extra space at the beginning crashes RetroArch when trying to initiate a netplay connection (as a client), see https://github.com/RetroPie/RetroPie-Setup/issues/3161#issuecomment-1160146932.


joolswills:
Thanks

PR #3566, opened 2022-07-23
amiberry: add wget dependency

cmitu:
Amiberry uses it to download the WHDLoad and GamesControllerDB updates.
Since we switched to `curl` for downloads, `wget` may not be installed and updates would fail.


joolswills:
Thanks.

PR #3569, opened 2022-08-06
scraper: fix building the package

cmitu:
One of the depedencies updates has broke the `scraper` build [1].
Until the compatibility is restored (could happen) or scraper is updated (unlikely), just use a previous version of the dependent Go library.

[1] https://github.com/J-Swift/thegamesdb-swagger-client-go/commit/06967c20ca1d18e7fbba5929d1b4bf930052956d#commitcomment-75241057


joolswills:
Thanks.

PR #3570, opened 2022-08-09
lr-pcsx-rearmed: fix build for NEON platforms

cmitu:
Upstream introduced recently an ASM optimized path for NEON platforms, enabled with a new `define`.
Added the new define for NEON platforms, otherwise the build fails because it picks the wrong ASM path.

Introduced in https://github.com/libretro/pcsx_rearmed/commit/047899a4f30d4a11e467e6b5aa0ba4cbf94c3f6b

NB: standalone doesn't seem affected, since the `makefile` is created properly by `configure`.

### DIFF
--- a/scriptmodules/libretrocores/lr-pcsx-rearmed.sh
+++ b/scriptmodules/libretrocores/lr-pcsx-rearmed.sh
@@ -35,7 +35,7 @@ function build_lr-pcsx-rearmed() {
         params+=(ARCH=aarch64 DYNAREC=ari64)
     fi
     if isPlatform "neon"; then
-        params+=(HAVE_NEON=1 BUILTIN_GPU=neon)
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

theofficialgman:
this is for arm64 neon support
if I understand retropies `scriptmodules/system.sh` correctly, there aren't any current arm64 (os and kernel) platforms in retropie with neon enabled. all armv8 platforms fall into the else here:
```
    if isPlatform "32bit"; then
        __default_cpu_flags+="  -mfpu=neon-fp-armv8"
        __platform_flags+=(arm armv8 neon)
    else
        __platform_flags+=(aarch64)
    fi
```

gizmo98:
Neon for armv8 is not the same thing. You cannot use old neon assembler with aarch64. GCC should use the neon unit by default because the aarch64 neon unit is IEEE754 compliant now.

PR #3573, opened 2022-08-15
retroarch: add extra core info files for missing upstream files

cmitu:
Upstream core info repository doesn't contain core info files for:

 * Mupen64Plus (obsolete upstream, the Plus/Next variant is the only one that has core info). Based on Mupen64Plus-Next core info file
 * Mess (upstream doesn't build/distribute it). Based on the Mame core info file
 * Mess2016 (upstream doesn't build/distribute it). Based on the Mame2016 core info file

This should complement 55aab04a6bb220e001bb0f5cbbafab0a62ecee08, where core info files have been added to RetroPie's RetroArch installation.
In the end, I think the `retroarch` package is the place to add the extra info, since they're only consumed by the front-end, plus it's less files needed to be modified.


joolswills:
Thanks.

PR #3574, opened 2022-08-20
scraper: fix scraping from ScreenScraper.fr

cmitu:
Added the patch from @Torben2000 [1] which fixes scraping from the ScreenScraper source.
Can be removed when/if upstream (@sselph) accepts the changes in the main repository.

[1] https://github.com/sselph/scraper/pull/265


joolswills:
Thanks

PR #3575, opened 2022-09-07
DOSBox Staging 0.79 update 

nemo93:
Full changelogs: 
* major release: https://dosbox-staging.github.io/v0-79-0/
* minor release: https://dosbox-staging.github.io/v0-79-1/

TL;DR: many improvements done on the audio front where the SB16 will now sound exactly as it's supposed to, compressor to prevent clipping, etc. Inclusion of a nice mouse driver for Windows 3.1. Performance improvement for low-spec devices (eg. RPi) and other graphical improvements (CGA, shaders, etc). Localization improvements. xbox-compatible controller mapperfiles for 180+ games.

**All credits to DOSBox' community and of course to the many excellent contributors of Staging.**

Script updated to include new required files and better default values for RPi3/4 devices. 

* `libspeexdsp-dev` has been added as a new dependency. 
* New required files (shaders, keyboards, translations, mapperfiles, etc) are copied over the install dir to `resources`.
* bump from "exp" to "opt".

### DIFF
--- a/scriptmodules/emulators/dosbox-staging.sh
+++ b/scriptmodules/emulators/dosbox-staging.sh
@@ -14,7 +14,7 @@ rp_module_desc="modern DOS/x86 emulator focusing on ease of use"
 rp_module_help="ROM Extensions: .bat .com .exe .sh .conf\n\nCopy your DOS games to $romdir/pc"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/dosbox-staging/dosbox-staging/master/COPYING"
 rp_module_repo="git https://github.com/dosbox-staging/dosbox-staging.git :_get_branch_dosbox-staging"
-rp_module_section="exp"
+rp_module_section="opt"
 rp_module_flags="sdl2"
 
 function _get_branch_dosbox-staging() {
@@ -22,7 +22,7 @@ function _get_branch_dosbox-staging() {
 }
 
 function depends_dosbox-staging() {
-    getDepends cmake libasound2-dev libglib2.0-dev libopusfile-dev libpng-dev libsdl2-dev libsdl2-net-dev meson ninja-build
+    getDepends cmake libasound2-dev libglib2.0-dev libopusfile-dev libpng-dev libsdl2-dev libsdl2-net-dev libspeexdsp-dev meson ninja-build
 }
 
 function sources_dosbox-staging() {
@@ -30,16 +30,10 @@ function sources_dosbox-staging() {
 }
 
 function build_dosbox-staging() {
-    local params=(-Dbuildtype=release -Ddefault_library=static --prefix="$md_inst")
+    local params=(-Dprefix="$md_inst" -Ddatadir="resources")
 
-    # Fluidsynth (static)
-    cd "$md_build/contrib/static-fluidsynth"
-    make
-    export PKG_CONFIG_PATH="${md_build}/contrib/static-fluidsynth/fluidsynth/build"
-
-    cd "$md_build"
     meson setup "${params[@]}" build
-    ninja -C build
+    meson compile -j${__jobs} -C build
 
     md_ret_require=(
         "$md_build/build/dosbox"
@@ -56,15 +50,25 @@ function configure_dosbox-staging() {
 
     [[ "$md_id" == "remove" ]] && return
 
+    local config_dir="$md_conf_root/pc"
+    chown -R $user: "$config_dir"
+
+    local staging_output="texturenb"
+    if isPlatform "kms"; then
+        staging_output="openglnb"
+    fi
+
     local config_path=$(su "$user" -c "\"$md_inst/bin/dosbox\" -printconf")
     if [[ -f "$config_path" ]]; then
         iniConfig " = " "" "$config_path"
         if isPlatform "rpi"; then
             iniSet "fullscreen" "true"
-            iniSet "fullresolution" "desktop"
-            iniSet "output" "texturenb"
+            iniSet "fullresolution" "original"
+            iniSet "vsync" "true"
+            iniSet "output" "$staging_output"
             iniSet "core" "dynamic"
-            iniSet "cycles" "25000"
+            iniSet "blocksize" "2048"
+            iniSet "prebuffer" "50"
         fi
     fi
 }


scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -22,24 +22,19 @@ function _get_branch_dosbox-staging() {
...
 function build_dosbox-staging() {
-    local params=(-Dbuildtype=release -Ddefault_library=static --prefix="$md_inst")
-
-    # Fluidsynth (static)
-    cd "$md_build/contrib/static-fluidsynth"
-    make
-    export PKG_CONFIG_PATH="${md_build}/contrib/static-fluidsynth/fluidsynth/build"
+    local params=(--prefix="$md_inst")
 
     cd "$md_build"
```
cmitu, line 35: the `cd` is not needed anymore, was needed because of the separate build step for `fluidsynth`.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -56,15 +51,23 @@ function configure_dosbox-staging() {
 
     [[ "$md_id" == "remove" ]] && return
 
+    # copy required resources
+    local config_dir="$md_conf_root/pc"
+    cp -R "$md_build/contrib/resources/"{freedos-cpi,freedos-keyboard,glshaders,translations} "$config_dir/"
```
cmitu, line 56: Shouldn't the resources be copied to `$md_inst/resources` ?
I've noticed `meson install` doesn't copy them, which is unfortunate.

scriptmodules/emulators/dosbox-staging.sh
```diff
@@ -56,15 +51,23 @@ function configure_dosbox-staging() {
...
+    chown -R $user:$user "$config_dir"
```
cmitu, line 57: Change to
```
chown -R $user: "$config_dir"
```
otherwise will have to rebase after #3577 gets merged.

cmitu:
@nemo93 thanks for the updated script. I've left a few comments on it, looks fine otherwise.

One question about the default config - is `openglnb` recommended now instead of `texturenb` ?
Haven't got a chance to test this setting on a RPI3, I know it works fine on a RPI4.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Nice spot. This one slipped through.

scriptmodules/emulators/dosbox-staging.sh
nemo93: No. For that release everything to be copied over the "config" directory (which is `/home/$user/.dosbox` by default).  If that's an issue we would have to think about creating a sub-dir into .dosbox eg. `.dosbox/staging` or similar.

scriptmodules/emulators/dosbox-staging.sh
nemo93: Thanks for catching this. Script updated.

nemo93:
@cmitu thanks a lot for your reviews and comments. I've pushed an update to reflect the changes. Also used the opportunity to promote Staging to "opt". If that's an issue please tell me.

Regarding `openglnb` vs `texturenb` the truth is that maintainers will drop `surface/texture*` in the future. Or at least rework them. Therefore all optimizations and efforts have been put to ensure `opengl*` is running fine on as many devices/configs as possible. This has been tested also on both RPI3/4 with official Retropie images and OS. If you or any user face a regression on RPi3 this will be thoroughly investigated.

Thanks again!

scriptmodules/emulators/dosbox-staging.sh
cmitu: According to https://github.com/dosbox-staging/dosbox-staging/blob/main/docs/README.resources, it should be possible to ship it as `$md_inst/resources/` (option 2. from the list). I'd prefer that resources shipped by the application to be copied in the install rather than to the user's folder. 

On my installation (for the previous version) I've used `$HOME/.config/dosbox/`'s folders (`glshader` and `mt32-roms`) for my local files, but I'll give it a test with the `$md_inst/resources` for the new resources.

cmitu:
I've finally managed to build and run on a Pi3 and it works fine (with the default config).

One issue I've encountered during building from source is that `ninja` likes to max out the # of jobs to 6 (despite the `man ninja` promising that the default is [#ofCPUs](https://github.com/ninja-build/ninja/blob/d4017a2b1ea642f12dabe05ec99b2a16c93e99aa/src/ninja.cc#L250)). This has a very high chance of maxing out the memory and crashing the compiler due to insufficient RAM. I wasn't able to complete a build until I've either added more memory or changed the build command to:
```
meson compile -j${__jobs} -C build
```
@nemo93 did you have any similar issues during building on a Pi3 ? Or did you use the Pi4 for 'cross-compiling' ?

scriptmodules/emulators/dosbox-staging.sh
nemo93: I stand corrected and I'm glad to catch this one up. I feel bad to be honest :( You're absolutely right and it's much better to have all those files under the "install dir" instead of .config. It's much cleaner plus it simplifies even further the script. Thanks again for that and apologize for not only missing that but for being wrong... 

Script edited accordingly.

nemo93:
> @nemo93 did you have any similar issues during building on a Pi3 ? Or did you use the Pi4 for 'cross-compiling' ?

Another good call. Well I'm the one testing on RPi4 hence no issue while building on that hardware. After checking with the core team member who tests on RPi3, it's definitely known that building on RPi3 takes much longer. Quote: "The big killer on the pi3 is the software scalar which uses a ton of macro-expansions; I remember that file alone takes about 6 minutes to compile w/ -O3 on the pi3, and will soak up a lot of memory." I've edited the script accordingly thanks to your input.

There's a [PR](https://github.com/dosbox-staging/dosbox-staging/pull/1950) which will have `meson install` to copy all of the `resources` files over the install dir. I'll have to edit the script once it's in.

scriptmodules/emulators/dosbox-staging.sh
cmitu: No worries, should be fine now.

cmitu:
@nemo93 almost there :).

I've tested more the `openglnb` option on a Pi3 and it's not working correctly (on RetroPie). My guess is that testing on the Pi3 was done with a current RaspiOS (based on Debian _bullseye_), while RetroPie still uses the 'legacy' _buster_ version. The difference is:

 * _bullseye_ uses the Mesa VC4 driver which supports both OpenGL 2.1 and GLES 2.0 on Pi3
 * _buster_ uses the proprietary Broadcom drivers which support only GLES 2.0 on Pi3.

Since `dosbox-staging` seems to be relying on OpenGL support for shaders and the `opengl[nb|pp]` outputs, they won't work properly in RetroPie on the Pi3. On my 1080p display, `openglnb` shows a tiny unscaled screen in the upper-left corner and the log shows there's no GL support detected:
``` log
2022-09-21 16:17:10.156 | SDL: version 2.0.10 initialized (RPI video and alsa audio)
2022-09-21 16:17:10.174 | CONFIG: Loaded primary conf file /home/pi/.config/dosbox/dosbox-staging.conf
2022-09-21 16:17:10.196 | RENDER: Using GLSL shader 'sharp'
2022-09-21 16:17:10.198 | DISPLAY: Initialized 1024x768 window-mode using Nearest-neighbour scaling on 1080p display-0
2022-09-21 16:17:10.226 | OPENGL: Vendor: unknown
2022-09-21 16:17:10.226 | OPENGL: Version: 0.0.0
2022-09-21 16:17:10.226 | OPENGL: GLSL version: unknown
2022-09-21 16:17:10.227 | OPENGL: Pixel buffer object: missing
2022-09-21 16:17:10.227 | OPENGL: NPOT textures: not supported
...
```

My suggestion is to replace the `openglnb` default for `output` with
```sh
if isPlatform "kms"; then
   iniSet "output" "openglnb"
else
   iniSet "output" "texturenb"
fi
```
and this will take care of future OS upgrades in RetroPie, while keeping current (_buster_) users unaffected.

Thanks so much for the support, sorry for nitpicking.

joolswills:
Thanks for your work on this @nemo93 @cmitu

nemo93:
@cmitu please keep on nitpicking! This is quite useful and helps to improve Staging. Thanks a lot for that. I've added your condition based on Pi model. I was unsure of the best style. Please shout if it's not fine. Also that this PR is not ready yet to be merged in as the PR improving the "resources dir" handling is still in review. I'll give you a nudge once it's there.

@joolswills anytime. It's a breeze thanks to @cmitu excellent testing and feedback.

cmitu:
@nemo93 thank you for the changes, they're fine.
Looks fine now, I suppose - as you said - the build changes for the `resources` installation have to land in a stable release, in order for us to take advantage of that. I've built the current `main` and the resources are copied correctly. Is there any intention to have the linked PR (now merged) also included in a future `0.79.1` ?

nemo93:
Thanks again for all testing and feedback. The upcoming `0.79.1` is expected to land end of this week/quite soon. I'll update that thread as well as OP with updated changelog once it's there.

nemo93:
0.79.1 has been released. We should be good now. I've updated the OP to reflect a bit more the changes. As a bonus there are plenty of "mapperfiles" added for those using a xbox-compatible controller. This means users will be able to play DOS games with just a controller which is definitely Retropie-friendly 😉 

Thanks again for the time, testing and feedback.

cmitu:
@nemo93 that was fast :). Re-tested again the installations and it's ok now, no more hiccups on the Pi3. Should be ready to merge (nudge @joolswills).

>  This means users will be able to play DOS games with just a controller which is definitely Retropie-friendly 😉

Hah, from a cursory look it seems the mapping have the _Start + Select_ set to exit the emulator ? If yes, that's neat.

nemo93:
> Hah, from a cursory look it seems the mappings have the _Start + Select_ set to exit the emulator ? If yes, that's neat.

That's right. Objective was to mimick the Retropie/Retroarch combo.

nemo93:
Anything I should do/you're expecting from me? Thanks!

cmitu:
@nemo93 no, it's fine.
@joolswills I think this is ready to be merged.

joolswills:
Thanks everyone.

joolswills:
This new version failed to build on the Raspberry Pi 1 due to the speexdsp meson wrap defaulting to enabling simd (neon). See https://github.com/RetroPie/RetroPie-Setup/pull/3590

priiduneemre:
Just a heads up, this doesn't work OOTB on Ubuntu 18.04. I tested this on an ODROID XU4, but the Raspberry Pi 3/4 probably has the same issue.

The newest version of `meson` available in the [offical repos](https://pkgs.org/search/?q=meson) (for 18.04) is `0.45.1`, from March of 2018. I solved it by installing `0.61.5` manually via `sudo -H pip3 install meson`. Don't know what the best solution would be but just wanted to point that out.

The error, specifically:
```
meson.build:150:16: ERROR: lexer
os_family_name = {
                ^

A full log can be found at /opt/retropie-setup/tmp/build/dosbox-staging/build/meson-logs/meson-log.txt
usage: meson [-h] [--prefix PREFIX] [--libdir LIBDIR]
...
meson: error: unrecognized arguments: -j3 -C build
```

Not sure if the latest `meson` for 20.04 supports these params; currently didn't have a chance to check.

priiduneemre:
Just to add to my comment above, I also had to disable OpenGL support via: 
```
-    local params=(-Dprefix="$md_inst" -Ddatadir="resources")
+    local params=(-Dprefix="$md_inst" -Ddatadir="resources" -Duse_opengl="false")
```

After that, the scriptmodule built fine on ODROID XU4 w/ Ubuntu 18.04.

I suspect the OpenGL issue is an ODROID XU4-specific one (the vendor files are wonky, maybe? no idea). Could not find an obvious way to get it to work with OpenGL enabled. At least it builds without it, though.
* Full logs for Meson issue: see [here](https://github.com/RetroPie/RetroPie-Setup/files/9898685/rps_2022-10-23_144230.log).
* Full logs for OpenGL issue: see [here](https://github.com/RetroPie/RetroPie-Setup/files/9898654/rps_2022-10-30_160253.log).

Also, sorry for the thread necromancy. I will stop now :).

PR #3577, opened 2022-09-09
Fix chown to use primary group of user for group name

kmcfate:
Sometimes there is a 'users' group, not a group per user. This makes it use primary group of user, not the explicit user name for group name.

### DIFF
--- a/scriptmodules/admin/apidocs.sh
+++ b/scriptmodules/admin/apidocs.sh
@@ -47,7 +47,7 @@ function build_apidocs() {
 
 function install_apidocs() {
     rsync -a --delete "$md_build/html/" "$__tmpdir/apidocs/"
-    chown -R $user:$user "$__tmpdir/apidocs"
+    chown -R $user: "$__tmpdir/apidocs"
 }
 
 function upload_apidocs() {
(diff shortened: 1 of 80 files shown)


joolswills:
Thanks.

Yeah, it was always designed around there being a group with the same name (as RetroPie originated on an OS where this was the default). The changes look good, but I will need to give it a test etc.

joolswills:
After a discussion with some other developers, we would prefer this to be implemented with a variable `$group`. eg doing `group=$(id -ng)` in `retropie_packages.sh` and then doing `chown $user:$group`. Primarily because it's a clearer more well known syntax than leaving the group name out of the chown.

I'm happy if you want to change this PR to implement it like that, or if you prefer I can make the changes. Thanks again.

laiti:
@joolswills I made a [pull request ](https://github.com/RetroPie/RetroPie-Setup/pull/3612) following that guideline.

joolswills:
Going to do this with some variable name changes in addition. Sorry for delays but will look at this soon.

cmitu:
Implemented in https://github.com/RetroPie/RetroPie-Setup/pull/3951

PR #3579, opened 2022-09-18
lr-pcsx-rearmed: trust the project's makefile to detect defaults

notaz:
The libretro/pcsx_rearmed NEON gpu plugin now can also run on arm64 and x86, and it's makefile should be able to detect when to enable those things. Remove the 'else' case and let it do the detection.

libretro/pcsx_rearmed#690


joolswills:
Thanks.

PR #3581, opened 2022-09-19
Add Bunsen Linux to supported flavours

eouthwaite:
This allowed the emulator compilations to work for me using Bunsen Linux on a fairly old PC which is unable to run Ubuntu.

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -170,7 +170,7 @@ function get_os_version() {
 
     local error=""
     case "$__os_id" in
-        Raspbian|Debian)
+        Raspbian|Debian|Bunsenlabs)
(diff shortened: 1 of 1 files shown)


cmitu:
If Bunsen Linux is a direct Debian spin-off, then there's no need for `__os_ubuntu_ver`.

eouthwaite:
Thanks! I admit I tried it first with the intention of digging into the reasoning behind the options if it failed...
I'll re-run without that line and update the file

eouthwaite:
@cmitu the PR has been simplified quite considerably - Bunsenlabs sits on top of Debian, so shares the __os_release

joolswills:
Looks good. thanks.

PR #3587, opened 2022-10-05
Add support for Pine64 RockPro64 board.

christopherkobayashi:
Pretty much what it says on the tin -- add a minimal definition to enable support for the RockPro64.

Tested under vanilla Debian 11 (bullseye) with stella and mame2003 cores.


joolswills:
Thanks.

PR #3588, opened 2022-10-05
enable crypto extensions for TX1 and TX2

theofficialgman:
crc and simd and enable by default in all versions of gcc so would be redundant to add here.

Cortex-A57 armv8.0 has optional crypto extensions so they are not enabled by default. TX1 and TX2 have these extensions available in hardware.


mrcmunir (approved):

joolswills:
Thanks.

PR #3590, opened 2022-10-21
dosbox-staging - fix building for armv6

joolswills:
The speexdsp meson wrap enables armv7/neon instructions when building for armv6 (Raspberry Pi 1). This causes compilation to fail.

Patch meson.build to disable the simd option for older arm platforms (that don't have the neon platform flag).

### DIFF
--- a/scriptmodules/emulators/dosbox-staging/speexdsp_simd_disable.diff
+++ b/scriptmodules/emulators/dosbox-staging/speexdsp_simd_disable.diff
@@ -1,0 +1,13 @@
+diff --git a/meson.build b/meson.build
+index 8db4dd4..e5e7549 100644
+--- a/meson.build
++++ b/meson.build
+@@ -396,7 +396,7 @@ endif
+ if not speexdsp_dep.found() or not is_system_speexdsp_reliable
+     speexdsp_dep = subproject(
+         'speexdsp',
+-        default_options: default_wrap_options,
++        default_options: default_wrap_options + ['simd=false']
+     ).get_variable('speexdsp_dep')
+     speexdsp_summary_msg = 'built-in'
+ endif
--- a/scriptmodules/admin/image.sh
+++ b/scriptmodules/admin/image.sh
@@ -174,7 +174,7 @@ function _init_chroot_image() {
     trap "_trap_chroot_image '$chroot'" INT
 
     # mount special filesystems to chroot
-    mkdir -p "$chroot"/dev/pts
+    mkdir -p "$chroot"{/dev/pts,/proc}
     mount none -t devpts "$chroot/dev/pts"
     mount -t proc /proc "$chroot/proc"
 
--- a/scriptmodules/emulators/dosbox-staging.sh
+++ b/scriptmodules/emulators/dosbox-staging.sh
@@ -27,6 +27,10 @@ function depends_dosbox-staging() {
 
 function sources_dosbox-staging() {
     gitPullOrClone
+    # patch the dosbox-staging meson.build script to disable neon instructions for older arm
+    if ! isPlatform "neon"; then
+        applyPatch "$md_data/speexdsp_simd_disable.diff"
+    fi
 }
 
 function build_dosbox-staging() {


scriptmodules/emulators/dosbox-staging/speexdsp_simd_disable.diff
```diff
@@ -0,0 +1,13 @@
+diff --git a/meson.build b/meson.build
+index 8db4dd4..e5e7549 100644
+--- a/meson.build
++++ b/meson.build
+@@ -396,7 +396,7 @@ endif
+ if not speexdsp_dep.found() or not is_system_speexdsp_reliable
+     speexdsp_dep = subproject(
+         'speexdsp',
+-        default_options: default_wrap_options,
++        default_options: default_wrap_options + ['simd=false']
```
kcgen, line 10: Thanks for catching this @joolswills and for flagging it @nemo93.
Nice to see the RPI1 still supported in RetroPie!

For the next Staging release, we'll find a way to check for NEON inside meson.

joolswills: Thanks.

nemo93:
For the record: https://github.com/mesonbuild/wrapdb/pull/723 > thanks @kcgen. Upon next release, the aforementioned patch can be dropped from Retropie's side for the Raspberry Pi 1. Thanks again.

nemo93:
@joolswills https://github.com/mesonbuild/wrapdb/pull/723 has been merged in. This should check for instructions from meson directly. Can you please confirm? If so the above should not be required any longer. Thanks.

Thanks a lot @kcgen.

kcgen:
Thanks for the heads up @nemo93. 

Yes; the default release build of DOSBox Staging's `main` branch on non-NEON-capable ARM systems should build the embedded SpeexDSP without NEON instructions.

cmitu:
@nemo93 @kcgen FWIW, the added check in the _wrapdb_ forumla fails because of a linker error with the new release:
```
....
Running compile:
Working directory:  /home/pi/RetroPie-Setup/tmp/build/dosbox-staging/build/meson-private/tmpzhy9qlp_
Command line:  cc /home/pi/RetroPie-Setup/tmp/build/dosbox-staging/build/meson-private/tmpzhy9qlp_/testfile.c -o /home/pi/RetroPie-Setup/tmp/build/dosbox-staging/build/meson-private/tmpzhy9qlp_/output.exe -pipe -mcpu=cortex-a72 -mfpu=neon-fp-armv8 -O2 -D_FILE_OFFSET_BITS=64 -O0

Code:

#include <stdint.h>
#include <arm_neon.h>
#include <math.h>

static void square4_serial(float*);
static void square4_vectorized(float*);
static int compare4(const float*, const float*);

static void square4_serial(float* array) {
    for (uint8_t i = 0; i < 4; ++i) {
        const float f = array[i];
        array[i] = f * f;
    }
}

static void square4_vectorized(float* array) {
    float32x4_t f = vld1q_f32(array);
    f = vmulq_f32(f, f);
    vst1q_f32(array, f);
}

static int compare4(const float* a, const float* b) {
    float sum_of_diffs = 0.0f;
    for (uint8_t i = 0; i < 4; ++i) {
        sum_of_diffs += fabsf(a[i] - b[i]);
    }
    const int rounded_sum = (int)(roundf(sum_of_diffs));
    return rounded_sum;
}

int main() {
    float values1[4] = {1.0f, 2.0f, 3.0f, 4.0f};
    square4_serial(values1);

    float values2[4] = {1.0f, 2.0f, 3.0f, 4.0f};
    square4_vectorized(values2);

    return compare4(values1, values2);
}

Compiler stdout:

Compiler stderr:
 /usr/bin/ld: /tmp/ccye1m2e.o: in function `compare4':
testfile.c:(.text+0x1a0): undefined reference to `roundf'
collect2: error: ld returned 1 exit status

Could not compile test file /home/pi/RetroPie-Setup/tmp/build/dosbox-staging/build/meson-private/tmpzhy9qlp_/testfile.c: 1
...
``` 

This is an a Pi4 with the latest RPI OS (`bullseye`, _gcc version 10.2.1 20210110 (Raspbian 10.2.1-6+rpi1)_). Though compiling on RetroPie would fail right-away because of the new Meson (0.57) requirement :(.

kcgen:
@cmitu ; darn - thanks for this heads up.

I've been running and testing 0.80 prior to its release on a retropie system:

`lsb_release -a`

```
Distributor ID:	Raspbian
Description:	Raspbian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
```

`gcc -v`

```
Target: arm-linux-gnueabihf
Thread model: posix
gcc version 8.3.0 (Raspbian 8.3.0-6+rpi1) 
```

`uname -a`

```
Linux retropie 5.15.56-v7l+ #1575 SMP Fri Jul 22 20:29:46 BST 2022 armv7l GNU/Linux
```

The SpeexDSP wrap's SIMD check is non-fatal (fix for this is PR'd  [here](https://github.com/mesonbuild/wrapdb/pull/848)), but the wrap should still build.  Atleast, it builds on my side.

As for Meson, newer versions are backward compatible, and have plenty fixes. 

It's just a Python PIP package, so ideally, `pip3` could be used instead of the `.deb` as an easy way to stay up-to-date. 

Alternatively, it looks like `bullseye`'s back-ports are right up to date! https://packages.debian.org/bullseye-backports/meson (maybe another option there).

I'm actually using PIP with Meson 0.64.x on my buster retropie box, and it's building DOSBox-Staging.

kcgen:
We try to hold back package versions based on constraints from the oldest still-in-support OSes: Ubuntu 18.04, Debian buster (with PIP :-)), Windows 8.1, and macOS 10.15 Catalina (arm + x86_64).

On every PR, we have CI jobs that build DOSBox Staging under all of these environments, and we only merged when it's passing on all (plus other constraints). 

Right now though - retropie on raspberry pi is not on that list - but we'd like it to be! The reason is only because GitHub doesn't have a CI runner that offer such an environment.

Do you know how we can automate and test our builds under an official retropie environment? (Right now, I periodically run my own build - but my Pi has local packges, my own settings, etc.. )

cmitu:
> The SpeexDSP wrap's SIMD check is non-fatal (fix for this is PR'd  [here](https://github.com/mesonbuild/wrapdb/pull/848)), but the wrap should still build.  At least, it builds on my side (maybe Meson 0.56 is less forgiving?).

Sorry, I should have added more details - it does build/run when lowering the Meson version requirement. The check in `speexdsp` is, of course, non fatal for the build, but it was mis-detecting the NEON capabilities of the build host.

Thanks for sending a patch for the _speexdsp_ WrapDB formula, should help the NEON detection on ARM platforms (including Raspi OS).


> Do you know how we can automate and test our builds under an official retropie environment? (Right now, I periodically run my own build - but my Pi has local packges, my own settings, etc..

That's a bit more involved to setup on a Github (x86_64) base install. Technically it should be possible to cross-compile using RetroPie-Setup scripts, but cross compiling still needs a Raspi OS installation and installing it just to run a CI job is resource/time consuming and it has its quirks.

Thanks for the comments and testing. I'll take a look at the changes in RetroPie for the new version (`libsdl2-image` is required now anyway so it needs a new dependency) and maybe fix our Ubuntu 18.04 build.

PR #3592, opened 2022-10-26
Adding the new Qt version of PCSX2

raelgc:
This script will install and update the PCSX2 emulator with the new [Qt interface](https://www.reddit.com/r/emulation/comments/uvf7cd/pcsx2_first_public_qt_release_is_now_available/) (AVX2 version).

The new Qt interface has several improvements over Wx, including:
- a fresh and modern user interface with dark mode available
- per-game settings
- new and simplified gamepad configuration panel with an auto-mapping feature
- and much more.

This will add an entry named `pcsx2-qt` under **Experimental**, to not conflict with the Wx version (which still exists, but has no more active development).

This script works querying the api.pcsx2.net site for the latest build.

The latest build is always offered as an [AppImage](https://en.wikipedia.org/wiki/AppImage), which is, in a summary, a portable app (no root required) built in a single compressed file including all libraries the application depends, similar to an OSX `.dmg` file. It works in most distros without additional requirements (usually just add the executable permission). The AppImage has a nice [FAQ for users](https://docs.appimage.org/user-guide/faq.html).

I know we prefer compile emulators from the source, but the binary this script uses is the official from the PCSX2 site, not an external source.

Anyway, feel free to request changes or simply close this PR if you think it doesn't fit here :)

### DIFF
--- a/scriptmodules/emulators/pcsx2-qt.sh
+++ b/scriptmodules/emulators/pcsx2-qt.sh
@@ -29,7 +29,7 @@ function install_bin_pcsx2-qt() {
 
 function configure_pcsx2-qt() {
     mkRomDir "ps2"
-    addEmulator 0 "$md_id" "ps2" "$md_inst/bin/rpcs2.AppImage %ROM% -fullscreen -batch"
-    addEmulator 1 "$md_id-nogui" "ps2" "$md_inst/bin/rpcs2.AppImage %ROM% -fullscreen -nogui"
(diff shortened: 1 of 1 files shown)


cmitu:
This should be superseded by the https://github.com/RetroPie/RetroPie-Setup/pull/3615, unless we want to embrace the AppImage path completely.

joolswills:
This isn't needed now due to #3615 as mentioned, so I'm going to close.

PR #3598, opened 2022-11-26
kodi: fix installation on Ubuntu 22.04 and later

cmitu:
The Kodi PPA for Jammy (22.04) and later doesn't include `kodi-inputstream-rtmp` anymore. Trying to install the version from the Ubuntu repositories will produce a conflict due to mismatched dependencies with the versions from the PPA, so  don't try to install the package.

NOTE: the addon can be later installed from inside Kodi, which is also recommended in its installation page: https://kodi.tv/addons/matrix/inputstream.rtmp


joolswills:
Thanks.

PR #3599, opened 2022-11-27
lzdoom: update to 3.88b

cmitu:
Updated to 3.88b (released on Feb 2022) to include recent fixes, should fix building with recent `gcc` (Ubuntu 22.04, Debian testing).

Changed the license URL to the location on the main branch and removed the extra quotes around the `snd_mididevice` parameter.

### DIFF
--- a/scriptmodules/ports/lzdoom.sh
+++ b/scriptmodules/ports/lzdoom.sh
@@ -11,8 +11,8 @@
 
 rp_module_id="lzdoom"
 rp_module_desc="lzdoom - DOOM source port (legacy version of GZDoom)"
-rp_module_licence="GPL3 https://raw.githubusercontent.com/drfrag666/gzdoom/g3.3mgw/docs/licenses/README.TXT"
-rp_module_repo="git https://github.com/drfrag666/gzdoom 3.87c"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/drfrag666/gzdoom/master/LICENSE"
+rp_module_repo="git https://github.com/drfrag666/gzdoom 3.88b"
 rp_module_section="opt"
 rp_module_flags=""
 
@@ -67,7 +67,7 @@ function add_games_lzdoom() {
 
     # FluidSynth is too memory/CPU intensive
     if isPlatform "arm"; then
-        params+=("+'snd_mididevice -3'")
+        params+=("+snd_mididevice -3")
     fi
 
     if isPlatform "kms"; then


joolswills:
Thanks.

cmitu:
@joolswills I'll take a look, maybe we can instruct `cmake` to use system libraries.

joolswills:
@cmitu Cheers.

cmitu:
I've looked over and it's not as straightforward as just using the system libraries.

The build fails because the included `lzma` library was updated to [21.06](https://github.com/drfrag666/gzdoom/commit/b20e72544ad6d898b0b43b61c7665b3c3a0f8e7b). A simple change to the build file (`CMakeLists.txt`) to include the system's `lzma` libraries is not enough, since Debian (& friends) include only version [9](https://packages.debian.org/search?keywords=lzma-dev) and they're not compatible, or rather the usage of the library in `lzdoom`'s source relies on the newer SDK version.

The new `lzma` SDK version seems to be miss-compiled with the _buster_ `gcc` (it uses the ARM7 CRC intrinsic implementation, but it's correctly compiled with the _bullseye_ toolchain (for _rpi1_/_armv6_). So the root issue seems to be that the new SDK requires a more recent toolchain.

During the tests with the `rpi1` platform, I found that `gzdoom`'s `cmake` build file are using [Pi2 compatible](https://github.com/drfrag666/gzdoom/blob/4ce19a12164ebd6802b7fb2f94d700a6e736d0fb/src/CMakeLists.txt#L337-L343) compiler flags for non-ARMv8 ARM platforms, so I'm wondering if the binary package we built before this version were really able to run on a Pi1/Zero.

I'll keep digging to see if we can force the `lzma` SDK to compile without the CRC optimizations and how we can patch the main `CMakeLists.txt` to make it `armv6` compatible.

cmitu:
Unfortunately without some patches I don't see a way around. Something like [this](https://github.com/RetroPie/RetroPie-Setup/compare/master...cmitu:RetroPie-Setup:lzdoom-armed?expand=1), which disable the hardware ARM CRC32 intrinsic functions and remove the Pi3/Pi2-only flags for ARM platforms.

joolswills:
Yeah I think that's the only way - seems odd they check for `#if !defined(__ARM_FEATURE_CRC32)` and then define it / force it - since this should automatically be handled by GCC (this preprocessor define is set correctly depending on GCC -mcpu`)

PR #3601, opened 2022-12-01
systems: add support Linux Mint LMDE

cmitu:
Add the Debian version for the Debian Edition of Linux Mint.
  - version 4 is based on Debian 10 'buster
  - version 5 is based on Debian 11 'bullseye'

Older versions (1 to 3) are no longer supported upstream.


joolswills:
Thanks.

PR #3607, opened 2022-12-06
sdl2 - updated to retropie-2.26.3 branch for bullseye

joolswills:
Due to https://github.com/libsdl-org/SDL/issues/5399 which breaks darkplaces-quake in RetroPie we are keeping Buster at 2.0.10 until there is a resolution.

Added libsamplerate0-dev dependency to module.

Branch includes reworking the patch for kms modesetting via env variables, some minor changes to our Mali and RPI patches, and some removal of no longer needed changes.

It also includes removal of some debian/control build dependencies which were added to upstream to align with Debian's packaging. The additions caused dependencies like libwayland-dev and libpulse-dev to be required for all platforms. As we handle adding dependencies in our sdl module code, these have been removed in our retropie-2.0.16+ sdl2 branches.

Include fix for detection of videocore due to our renamed include folders (thanks @cmitu).

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -16,11 +16,20 @@ rp_module_section="depends"
 rp_module_flags=""
 
 function get_ver_sdl2() {
-    echo "2.0.10"
+    if [[ "$__os_debian_ver" -ge 11 ]]; then
+        echo "2.26.3"
+    else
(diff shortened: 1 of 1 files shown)


joolswills:
@cmitu you happy for this to be merged now?

cmitu:
Yep, I think we can merge it.

PR #3609, opened 2022-12-09
Update Yamagi Quake 2 to version 8.20

protocultor:
I know you've just updated to the 8.10 version, but today is Quake II's 25th anniversary, and Yamagi released 8.20, which adds even more improvements to controllers. Now, you can even change sticks layout and deadzone sizes from the menu. Playing on a RPi 3 B+, no performance lost from previous versions.

To make this PR have a little more value, I've also let the expansions' source clone from a certain tag, instead of doing it from "master". This let the expansion take into account a certain version, just like the main repository.


joolswills:
Thanks.

PR #3610, opened 2022-12-13
lzdoom: fix the build for RPI1/ARMv6 platforms

cmitu:
The changes added enable `lzdoom` to build (and run) on RPI1/RPI0:

  - remove the `ARMV8` build option, since it used upstream as a toggle between RPI2 or RPI3 CPU parameters. Since we're passing the platform CPU flags to `gcc` ourselves, they're not really needed and they'll add the wrong options for other ARM platforms
  - patch the included LZMA SDK CRC calculation to not assume every ARM platform supports hardware-assistet CRC (i.e. ARMv8). This will fix the build for RPI1/ARMv6 platforms

### DIFF
--- a/scriptmodules/ports/lzdoom.sh
+++ b/scriptmodules/ports/lzdoom.sh
@@ -27,16 +27,19 @@ function depends_lzdoom() {
 
 function sources_lzdoom() {
     gitPullOrClone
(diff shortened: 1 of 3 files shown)


joolswills:
Thanks for this.

PR #3612, opened 2022-12-24
Fix chown to use primary group of user for group name by using group as variable

laiti:
Fix the assumption that user group matches the user name in installation scripts by setting group variable in `retropie_packgages.sh` script. 

Based on the [PR made by kmcfate](https://github.com/RetroPie/RetroPie-Setup/pull/3577) - thanks for the prework!


joolswills:
Thanks for this, but I would prefer to make these changes myself when I'm ready.

PR #3613, opened 2023-01-03
xboxdrv: upgrade the build process to python3

cmitu:
Python3 is the default in current Debian & friends distros (and `python2` was also recently [removed](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1027108) from Debian _sid_), but the current fork from zerojay/xboxdrv.git doesn't support it.

Upstream has moved back to Github [1] and added the necessary fixes for building with `python3`, so merge the upstream code with the various fixes added in zerojay/xboxdrv under retropie/xboxdrv.git. This fixes building under RaspiOS _bullseye_.

Since the new Sconsfile doesn't build with `python2` anymore, make the `python3` usage explicit for the `deps` and `build` phases.

[1] https://github.com/xboxdrv/xboxdrv


joolswills:
Thanks.

PR #3615, opened 2023-01-10
pcsx2: fix installation and configuration for recent 1.7 packages

cmitu:
* The Ubuntu PPA has 64bit packages for some time, now defaulting to 64bit only for recent 22.04 and newer. Modified the module to take this into account and install the 64bit package when available.

* The new PCSX2 versions have a QT UI by default and the PPA installs it automatically. The parameters and commands are different from the previous WXWidgets based versions, so configuration detects the version installed and sets the correct commands for both versions.


joolswills:
Thanks.

PR #3618, opened 2023-01-15
cgenius: upgrade to 3.2.0 & misc module improvements

cmitu:
Updated to the latest stable version (3.2.0), with mostly bugfixes and stability changes.

Scriptmodule changes:

 * changed the build to be in a separate folder
 * removed the install prefix since it's unused
 * removed the `gcc` workaround since it's not needed anymore
 * added a minimal config to configure the games location
 * removed the gamesdir symlink, not needed with the above modifications. Should prevent copying errors when backing up the SMB folders since the `games` config sub-folder was a symlink outside the `config` share.


joolswills:
Thanks for this. I didn't like the symlink so good to lose it. Cheers!

joolswills:
Thanks for this. I didn't like the symlink so good to lose it. Cheers!

PR #3621, opened 2023-01-18
emulationstation: fix OpenGL version 2.1 check

gizmo98:
Emulationstation does not run with -DUSE_GL21=On if gpu only supports GL <= 2.0. Fix OpenGL version check.


cmitu:
May be better to use `compareVersions` for comparison, has more clarity and `bc` is not a base package on Debian(s).

gizmo98:
Thanks @cmitu at pointing this out. I will rewrite and test the MR with compareVersions.

gizmo98:
@cmitu the MR is updated with compareVersions.

PR #3623, opened 2023-01-23
emulationstation: use SDL2 wayland videodriver if applicable

gizmo98:
If a Gnome/Wayland or Weston/Wayland session is started under Ubuntu 22.04, SDL2 falls back to x11 videodriver. There is ARM hardware which runs much faster, and without graphical glitches, if the SDL2 wayland video driver is used. To prevent SDL2 from using x11 video driver, export "SDL_VIDEODRIVER=wayland", if Wayland session is detected.


joolswills:
Thanks

PR #3624, opened 2023-01-24
emulationstation: allow gles for all platforms

gizmo98:
gles can be used under x11, wayland, kms, dispmanx and fb. Respect gles platform flag and build emulationstation with gles renderer if requested.


joolswills:
Thanks.

daveamato:
Thank you!

PR #3626, opened 2023-01-25
Add recognition of NXP imx socs i.MX6 dual lite, i.MX6 quad and i.MX8M mini 

gizmo98:
-not officially supported.
-tested with NXP i.MX8M mini and i.MX6 dual lite with kernel 6.1 mainline and Ubuntu 22.04.   
-use mesa etnaviv driver if kernel uses etnaviv kernel driver.
-not tested with NXP gpu kernel module and driver blob vivante/galcore.


joolswills:
Thanks.

PR #3630, opened 2023-02-10
yabasanshiro: remove missing python-pip package

gizmo98:
Debian >= 10 has no python-pip package. Use python3-pip instead.


cmitu:
I'm pretty sure that `pip` is not needed at all and can be removed. There's no codepath in the `pi4` branch that would need it. The `libboost-all-dev` though might benefit by replacing it with just the `system filesystem date_time locale` libs.

gizmo98:
Ok. I try to build it without pip.

gizmo98:
@cmitu you are right. It compiles and runs without python-pip!

PR #3631, opened 2023-02-11
Fix file permissions of fs-uae and yabasanshiro

gizmo98:
Unify file permissions of all scriptmodules. Most scriptmodules have file permission 100644. Only fs-uae and yabasanshiro have 100755.


cmitu:
Yep, looks right. Thanks !

PR #3635, opened 2023-02-18
vice - switch to v3.7.1 release

schiele:
Tested on RPI 4 with no issues found so far.
Sound issues mentioned in commit 81fc3a964b79d321285e4e3560a7eb80f3a35270 were not observed by myself so far with that version.


joolswills:
Tested also RPI3 and seems fine. Thanks.

PR #3636, opened 2023-02-18
raspbiantools: Add option to enable compressed memory (ZRAM)

gizmo98:
Add an option which enables ZRAM. ZRAM will which can hold up to 50% of the physical ram.

Compressed memory (ZRAM) can be used with low memory embedded devices. ZRAM has compression ratio of up to 4:1. 256MB of physical memory can store up to 1024MB. It is only used if free memory is getting low. It is much faster than disk swapping. It also prevents linux from using disk swap until ZRAM is full.
https://wiki.debian.org/ZRam
https://haydenjames.io/raspberry-pi-performance-add-zram-kernel-parameters/

Tested with a 2GB Raspberry Pi 4, emulationstation, some retroarch cores and some memory intensive source builds.

### DIFF
--- a/scriptmodules/supplementary/raspbiantools.sh
+++ b/scriptmodules/supplementary/raspbiantools.sh
@@ -125,8 +125,38 @@ function enable_modules_raspbiantools() {
     done
 }
 
+function enable_zram_raspbiantools() {
+    if [[ "$__os_id" == "Raspbian" ]] || [[ "$__os_id" == "Debian" ]]; then
+        aptInstall zram-tools
+        # Use 50% of the current memory for ZRAM
+        local percent="50"
+        iniConfig "=" "" "/etc/default/zramswap"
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. I think it might be nice to have an option to disable it (and perhaps show whether it's enabled or not ?)

gizmo98:
> Thanks. I think it might be nice to have an option to disable it (and perhaps show whether it's enabled or not ?)

I take a look into /proc/swaps now if a zram device exists. If no zram device exists i only allow to enable zram. If a zram device exists i only allow to disable zram. Please take a look. If everything is right i will squash all commits.

joolswills:
Looks good thanks. Happy to include it.

gizmo98:
Thanks for review! Everything is squashed now. Commit message is updated.

joolswills:
Thank you.

PR #3638, opened 2023-02-19
netplay: use the OSK to configure netplay options

cmitu:
Replaced the `dialog` based inputboxes for host/nick/port in the netplay configuration prompts with the new `inputBox` helper, which uses the OSK (when installed).
Should make it easier to configure netplay without a keyboard.

Additional changes:
* the `osk.py` script was modified to accept an existing value instead of always starting with an empty input.
* modified the `inputBox` and added some notes on return value. Also changed the `dialog` based input prompt to be similar to how `osk.py` shows it (_Enter the `$value`_).

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -138,17 +138,22 @@ function editFile() {
 ## @param text default text
 ## @param minchars minimum chars to accept
 ## @brief Opens an inputbox dialog and echoes resulting text. Uses the OSK if installed.
+## @details The input dialog has OK/Cancel buttons and can be cancelled by the user.
+## The dialog will enforce the minimum number of characters expected, re-prompting the user.
+## @retval 0 when the user entered the text and chose the OK button
+## @retval != 0 when the user chose the Cancel button
+
 function inputBox() {
     local title="$1"
     local text="$2"
     local minchars="$3"
     [[ -z "$minchars" ]] && minchars=0
-    local params=(--backtitle "$__backtitle" --inputbox "$title")
+    local params=(--backtitle "$__backtitle" --inputbox "Enter the $title")
     local osk="$(rp_getInstallPath joy2key)/osk.py"
 
     if [[ -f "$osk" ]]; then
         params+=(--minchars "$minchars")
-        text=$(python3 "$osk" "${params[@]}" 2>&1 >/dev/tty) || return $?
+        text=$(python3 "$osk" "${params[@]}" "$text" 2>&1 >/dev/tty) || return $?
     else
         while true; do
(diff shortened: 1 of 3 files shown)


joolswills:
Thanks for this.

I noticed an issue the other day with the OSK. As It's part of the joy2key module and run from the "installed" location, if that module is removed, then the wifi configuration won't work. This would also happen with the netplay config.

I think this needs to be addressed before we add the OSK to other areas. Perhaps separating it from the joy2key module or providing some fallback if it's not installed.

joolswills:
(I'm happy to look into this, but I don't have a lot of free time currently.)

cmitu:
> I think this needs to be addressed before we add the OSK to other areas. Perhaps separating it from the joy2key module or providing some fallback if it's not installed.

I'll take a look. Maybe we can try and run it from the source, if not installed, but that won't work unless we `python3`/`urwid` as part of the mandatory dependencies. 

> (I'm happy to look into this, but I don't have a lot of free time currently.)

No rush.

joolswills:
I started on this. Assuming it works but I'm doing a helper function for it to fall back on dialog if the osk isn't installed. That's the plan - will continue tomorrow.

cmitu:
@joolswills yes, a helper is what I've been thinking about also.
Let me know if you need some modifications to `osk` to make it easier to keep the parameter order/number similar.

cmitu:
I've modified the PR to use the new `inputBox` function now, seems to be working fine.

scriptmodules/helpers.sh
```diff
@@ -148,10 +153,10 @@ function inputBox() {
 
     if [[ -f "$osk" ]]; then
         params+=(--minchars "$minchars")
-        text=$(python3 "$osk" "${params[@]}" 2>&1 >/dev/tty) || return $?
+        text=$(python3 "$osk" "${params[@]}" "$text" 2>&1 >/dev/tty) || return $?
     else
         while true; do
-            text=$(dialog "${params[@]}" 10 60 "$text" 2>&1 >/dev/tty) || return $?
+            text=$(dialog "${params[@]}" 10 60 "Enter the $text" 2>&1 >/dev/tty) || return $?
```
joolswills, line 159: This is the default text so it should just be the `$text` variable as before. Eg for netplay.

Screenshot_2023-02-25_17-54-49

cmitu: I knew I forgot something...

cmitu: Should be fixed now, but I'll need to modify the OSK to remove the built-in _Enter the_. This is what I forgot.

cmitu: Ok, fixed the `osk.py` also, should be ok now.

joolswills:
Thanks.

PR #3639, opened 2023-02-19
scummvm: update to version 2.7.0

schiele:
Changes in 2.7.0:

 New games:
   - Added support for Soldier Boyz.
   - Added support for C64 and ZX Spectrum versions of GLK Scott Adams Interactive Fiction games.
   - Added support for GLK Scott Adams adventures 1-12 in the TI99/4A format.
   - Added support for Obsidian.
   - Added support for Pink Panther: Passport to Peril.
   - Added support for Pink Panther: Hokus Pokus Pink.
   - Added support for Adibou 2 "Environment", "Read/Count 4 & 5" and "Read/Count 6 & 7".
   - Added support for Driller/Space Station Oblivion (DOS/EGA/CGA, Amiga, AtariST, ZX Spectrum and Amstrad CPC versions).
   - Added support for Halls of the Dead: Faery Tale Adventure II.

 New platforms:
   - Added support for the RetroMini RS90 under OpenDingux beta.
   - Added support for the 1st generation Miyoo (New BittBoy, Pocket Go and PowKiddy Q90-V90-Q20) under TriForceX MiyooCFW.
   - Added support for the Miyoo Mini.
   - Added support for KolibriOS.

 General:
   - Reduced amount of false positives in Mass Add.
   - Updated the Roland MT-32 emulation code to Munt mt32emu 2.7.0.
   - Added support for shader-based scalers.
   - Added option for mono sound output (via --output-channels=CHANNELS command line option).
   - Improved cursor scaling in OpenGL mode.
   - Fix crash when browsing folders containing files with \1 in the names.
   - Added possibility to specify RNG seed via GUI or command line option.
   - Added possibility to run ScummVM in autodetection mode by renaming the executable starting with 'scummvm-auto' or by providing an empty file named 'scummvm-autorun' next to the ScummVM executable.
   - Added possibility to supply command line parameters which will be picked up automatically. Put them one per line in a file named 'scummvm-autorun'.
   - Added possibility to customize the default settings by specifying an initial configuration file to load if no configuration file exists in the usual location (via --initial-cfg=FILE or -i command line option).
   - Added support for loading game resources which are bigger than 2GB on more platforms.

 AGI:
   - Improved support for French translations.

 AGOS:
   - Added option to disable the fade-out effects on room transition for Simon1 and 2.

 AGS:
   - Added support for the original installer files for Maniac Mansion Deluxe and The New Adventures of Zak McKracken.

 Director:
   - Support for Pippin version of L-Zone.
   - Fix a bug caused by use of slash in filename.

 Dreamweb:
   - Support playing from the original installer floppies.

 Hadesch:
   - Added support for a 1997 release.

 Kyra:
   - Added support for the Korean version of Legend of Kyrandia 1.
   - Support multi-floppy mac kyra1 both as installer floppies and as installed directory.
   - Added support for the Hebrew version of Legend of Kyrandia 3.

 MADS:
   - Added support for original floppy installer file layout for Rex Nebular.

 Neverhood:
   - Added support for Japanese version of Neverhood.
   - Support localizations by -=CHE@TER=- & Rigel.

 Plumbers:
   - Fixed crash with windows version.

 Private:
   - Allow playing mac version directly from installer CD.
   - Added support for Japanese/Mac version

 Queen:
   - Added option for using a better font in Hebrew version.

 SAGA:
   - Added support for Chinese Inherit the Earth.
   - Added support for Chinese I Have no Mouth and I Must Scream.
   - Added support for Korean I Have no Mouth and I Must Scream.
   - Added support for playing directly from floppy installer for ITE.
   - Support for Amiga (AGA/ECS, Retail/Demo, English/German) Inherit the Earth.

 SCI:
   - Improved text rendering for Macintosh titles.
   - Added support for Casio MT-540, CT-460 and CSM-1 MIDI devices for the SCI0 games that originally supported it.

 SCUMM:
   - Added support for CGA, CGA Composite, CGA black & white and Hercules modes for SCUMM 1 versions of Zak McKracken and Maniac Mansion.
   - Improved accuracy of CGA and Hercules modes for SCUMM 2 (enhanced) versions of Zak McKracken and Maniac Mansion.
   - Improved accuracy of CGA and Hercules modes for Monkey Island 1 (EGA version only - the VGA version does not have CGA and Hercules modes).
   - Fixed some minor glitches for the CGA mode of Loom.
   - Added EGA dithering mode for VGA versions of Loom, Monkey Island 1 and 2 and Indiana Jones 4.
   - Fixed a possible dead-end in the Ultimate Talkie Edition of Monkey Island 2, if one doesn't pick up a required item before Captain Dread brings Guybrush back to Scabb Island, at the end of Part II.
   - Fixed various original game bugs, oversights and continuity errors (only when using the "Enable game-specific enhancements" setting).
   - Improved the decoding of some Western European game strings when they're displayed through ScummVM's interface, such as when pausing a game.
   - Fixed the navigator head not pointing to some directions in Monkey Island 1, in the original releases without the enhanced verb interface.
   - Fixed slightly inaccurate text position in v4 games and in Loom v3.
   - For Sam & Max, it is now possible to shoot down the text lines of the final credits with the crosshair mouse cursor, just like in the original.
   - Fixed lipsync issues in the final scene of Freddi Fish 4.
   - Fixed The Dig and COMI loading cursors not being shown when they should.
   - Improved the accuracy of some audio drivers, which was notably impacting the pitch bending effect in the DOTT intro music.
   - In COMI, only let Guybrush read the clock of Puerto Pollo in the English, Italian and (fan-made) Russian versions of the game, matching the behavior of the original interpreters (probably because of the poor results in the other languages).
   - Improved support for Hebrew HE game localizations.
   - Fixed Roland MT-32 support in Sam & Max.
   - Implemented original GUI and save menus for LucasArts games (DOS, Windows, Amiga, Macintosh, FM-Towns, SegaCD, Atari ST, NES and Commodore 64 versions). Also activate the general "Ask for confirmation on exit" option for a more authentic '90s experience!
   - Fixed minor timing issues for the SMUSH video engine, mostly affecting Full Throttle.
   - Added a low latency audio mode to Full Throttle, The Dig and The Curse of Monkey Island; this can improve audio performance expecially in non-desktop devices, but it is also a little less accurate than the original.
   - Implemented reloading CD audio for Monkey Island 1 and Loom (CD versions), when reloading a save state.

 Sherlock:
   - Added support for Chinese Rose Tattoo.

 Sky:
   - Added support for Chinese Beneath a Steel Sky.

 Sword1:
   - Support Novy Disk Russian translation.
   - Fixed launching various demos.
   - Switched detection to md5-based. Submit your unrecognized versions!

 Sword2:
   - Switched detection to md5-based. Submit your unrecognized versions!

 Tinsel:
   - Fixed Discworld II subtitle colors on big-endian ports.

 Toon:
   - Made game menus behave like in the original.

 TwinE:
   - Fixed rendering issue with doors.
   - Fixed invalid music pause in behaviour and inventory menu.
   - Fixed giving kashes instead of hearts as fallback.
   - Fixed item flashing when they drop.
   - Fixed meca penguin angle at spawn.
   - Fixed background redraw when watching cutscenes at the television.
   - Fixed recenter the screen on activating an inventory item.
   - Fixed pressing W to talk to people also makes you jump.

 Ultima8:
   - Added support for saving and resizing of minimap.
   - Adjust cursor to behave closer to the original.
   - Adjust item quantity slider to behave closer to the original.

 Xeen:
   - Several crash fixes for Clouds of Xeen.
   - Wait until farewell finishes before leaving shops.
   - Don't reload map after leaving character creation.
   - Implement correct clouds falling logic for Swords of Xeen.
   - Fix GateMaster monster in Underworld map.

 3DS port:
   - Update relocation parser to support PREL31 that are emitted by new compiler.
   - Do more autoconfiguration in ./configure.

 Android port:
   - Added Storage Access Framework support.
   - Improved support for game controllers.

 iOS port:
   - Added pointer device support.
   - Improved suppport for touchpad mode.
   - Added support for games that use 32 bit pixel formats.

 Nintendo DS port:
   - Added a splash screen to the top screen when the launcher is active.

 OpenDingux port:
   - Added support for dynamic plugins.

 PS Vita port:
   - Added support for dynamic plugins.

 PSP port:
   - Improved support for games that use 32 bit pixel formats.

 RISC OS port:
   - Added support for 26-bit versions of RISC OS.


cmitu:
Thanks for the PR.

Note that while the `v2.7.0` tag has been added upstream, 2.7.0 hasn't yet been officially released. I'd wait for the release announcement until we bump this to 2.7.0. Also note that `scummvm-sdl1.sh` needs also to be bumped (and tested).

schiele:
Ok, I updated and debugged scummvm-sdl1 until reaching a point where I got it working. I didn't consider that initially since I found it out-of-sync with scummvm before and the upstream ScummVM project seems to have given up on the sdl1 port on RaspberryPI according to comments in the code. Finally, I got it working again by disabling OpenGL, similar to the previous kmp patch but was no longer applicable due to restructuring the OpenGL handling in the configure script. In the current patch, I disabled OpenGL for all RaspberryPI systems for sdl1, independent of the kmp flag. If you want me to do that on kmp only I can change the patch accordingly but since I don't have such a system I thought it better to be safe than sorry.

scriptmodules/emulators/scummvm-sdl1/01_rpi_sdl1.diff
```diff
@@ -1,11 +1,13 @@
...
+diff --git a/configure b/configure
+index 6f44008abdb..25b3713e243 100755
+--- a/configure
++++ b/configure
+@@ -3379,7 +3379,7 @@ if test -n "$_host"; then
  			# We prefer SDL2 on the Raspberry Pi: acceleration now depends on it
  			# since SDL2 manages dispmanx/GLES2 very well internally.
  			# SDL1 is bit-rotten on this platform.
 -			_sdlconfig=sdl2-config
 +			_sdlconfig=sdl-config
```
schiele, line 10: BTW: I was wondering whether it would be better to achieve the result of the patch with a simple one-line sed command instead. In my opinion, this would be easier to maintain since such a sed scriptlet is more robust against small changes in the context of the change but for sure is also more likely to get silently ignored if the structure of the code gets changed in a substantial way. What is your opinion on that?

schiele: To be more precise I thought about replacing the patch with something like this:
sed -ie 's/\([[:space:]]_sdlconfig=sdl\)2\(-config\)/\1\2/' configure

cmitu: `sed` would have worked, I think even with a simpler substitution.
However, I think the better option would be to pre-set `SDL_CONFIG` when calling `configure`, since [find_sdl](https://github.com/scummvm/scummvm/blob/4da739e26e43de2e7fd2efbf4c026c6b14493c09/configure#L451-L454) will prioritize it before `$_sdlconfig`.
I.e.
```
SDL_CONFIG=sdl-config ./configure --params
```
will pick `sdl1` before `sdl2`. This will allow use to drop the patches for `configure.

cmitu:
Thanks for the changes on `scummvm-sdl1`, using the `none` for `--opengl-mode` works fine on `rpi` (using the `dispmanx` video driver).

Starting with `2.6.x`, `scummvm` uses [glad](https://github.com/Dav1dde/glad/) to do the GL context initialization and then does a detection of runtime of the supported context (OpenGL vs OpenGLES vs OpenGLES2). I was thinking of a using `--opengl-mode`:

1. ScummVM
   - `--opengl-mode=auto`. I thin all platforms we support have an OpenGL(ES2) renderer in `sdl2`.
  I've tested this on RPI3 and RPI4 and works fine. GLES2 (via the `RPI` video driver) and OpenGL (via the `kms` driver) are working fine w.r.t. context auto-detection.

2. ScummVM-SDL1
   - `--opengl-mode=none` for the platforms where there's no OpenGL renderer in `sdl1`, like the `rpi`. This may as well mean just `x11`.
   - `--opengl-mode=auto` for the rest.
 
I'm also thinking of booting `scummvm-sdl1` from the `kms` platform, since from _bullseye_ on there will be no video driver to make it work (unless we add something like `sd12-compat`, but that's a totally different story). @joolswills what do you think ?

cmitu:
@schiele thank you for the changes.
For now it's ok, I'll have a following PR to disable the `scummvm-sdl1` on `kms` platforms and a minor patch to fix a crash with the RPI SDL2 driver, but those are minor additions.

Can you squash the commits into a single commit ?

schiele:
> Can you squash the commits into a single commit ?

Done.

cmitu:
Since 2.7.0 has been (officially) released, I think we can merge this.

@joolswills I have a few minor additions and a patch for the RPI/Dispmanx driver, can we wait until tomorrow before building the binaries ?

joolswills:
Maybe add the commit to this PR? We can merge it together.

cmitu:
I'd rather leave the patch separate, just in case we need to revert it. But the other modifications I had in mind can be added:
 - exclude `-sdl1` from `kms`
 - add some swap for building on platform with less than 512 Mb (was reported last week in the forums). Not sure which is the upper limit, I'm about to test to see how much memory is used during the build.

joolswills:
@schiele Do you still have some changes to make ? I was going to implement some changes for @cmitu but I see you just force pushed so wanted to make sure you were finished before I did anything.

schiele:
> @schiele Do you still have some changes to make ? I was going to implement some changes for @cmitu but I see you just force pushed so wanted to make sure you were finished before I did anything.

No, I was just rebasing to current master, no changes from my side. Sorry if this was causing some confusion.

joolswills:
No worries. Thanks.

joolswills:
@cmitu I cherry-picked your commit. Please merge if you are happy.

joolswills:
Also - binaries will be built when merged. You mentioned not building binaries, so let me know before merging, as I will have to disable the build script on my rpi4. Or alternatively we can build binaries and it will rebuild when you add additional changes in a new PR if that is your plan.

cmitu:
@joolswills thanks. Unfortunately my patch doesn't seem to work with 2.7.0 (or maybe it didn't work before and some config changes made me think it did), I'll take a look again at the issue. Binaries away !

@schiele thank you again for the PR.

joolswills:
Cheers.

PR #3640, opened 2023-02-21
packages: amend update detection for untracked modules

cmitu:
Added a fallback for getting the `modified_date` of a module if the module's file is not tracked by `git`. If the module is just copied in the `scriptmodules` folder or under `ext`, but not part of a `git` repo, use the modification date of the file. This avoids re-building the module when a mass update is run.

Noticed this in [this forum topic](https://retropie.org.uk/forum/topic/33922/need-help-with-openbor-retropie-extra-exarkuniv/15), seems like [RetroPie-Extra](https://github.com/Exarkuniv/RetroPie-Extra) copies selected module files under `ext`, while the main repository folder is kept separately.

### DIFF
--- a/scriptmodules/packages.sh
+++ b/scriptmodules/packages.sh
@@ -644,6 +644,8 @@ function rp_hasNewerModule() {
                 local repo_dir="$scriptdir"
                 [[ "$vendor" != "RetroPie" ]] && repo_dir+="/ext/$vendor"
                 local module_date="$(sudo -u "$user" git -C "$repo_dir" log -1 --format=%cI -- "${__mod_info[$id/path]}")"
+                # just in case the module is not known to git, get the file last modified date
+                [[ -z "$module_date" ]] && module_date="$(date -Iseconds -r "${__mod_info[$id/path]}")"
                 if rp_dateIsNewer "$pkg_date" "$module_date"; then
                     ret=0
                 fi


scriptmodules/packages.sh
```diff
@@ -644,6 +644,8 @@ function rp_hasNewerModule() {
                 local repo_dir="$scriptdir"
                 [[ "$vendor" != "RetroPie" ]] && repo_dir+="/ext/$vendor"
                 local module_date="$(sudo -u "$user" git -C "$repo_dir" log -1 --format=%cI -- "${__mod_info[$id/path]}")"
+                # just in case the module is not known to git, get the file last modified date
+                [[ -z $module_date ]] && module_date="$(date -Iseconds -r "${__mod_info[$id/path]}")"
```
joolswills, line 648: Although not technically required please can you quote `$module_date`  (as we do elsewhere)

joolswills:
Going back on what I said on the forum, I was thinking to do a test to check if a file is part of a repo and then doing the git log or date, but actually it's not needed, and this seems fine. Just the minor comment about quoting. Thank you.

cmitu:
@joolswills thanks for the review. I'll amend the PR.

scriptmodules/packages.sh
cmitu: Should be ok now.

joolswills:
Thanks.

PR #3641, opened 2023-02-21
lr-scummvm: fix interactive prompt from unzip

cmitu:
Modified the `unzip` command for the core system files to overwrite unconditionally the files, otherwise we get a prompt when upgrading.


joolswills:
Thanks.

PR #3642, opened 2023-02-21
emulationstation: modify wayland detection

gizmo98:
I have done some tests running weston/wayland from CLI with a rpi4/raspbian-lite. After startup of weston $XDG_SESSION_TYPE is still "tty". If I start a weston/wayland session with graphical login (display manager sddm or gdm3) $XDG_SESSION_TYPE is "wayland". $WAYLAND_DISPLAY can be found in both constellations and seems to be more reliable. On a rpi4 $WAYLAND_DISPLAY=wayland-0. Just replace $XDG_SESSION_TYPE with $WAYLAND_DISPLAY.

### DIFF
--- a/scriptmodules/supplementary/emulationstation.sh
+++ b/scriptmodules/supplementary/emulationstation.sh
@@ -257,7 +257,7 @@ if [[ "\$(uname --machine)" != *86* ]]; then
 fi
 
 # use SDL2 wayland video driver if wayland session is detected
-[[ "$XDG_SESSION_TYPE" == "wayland" ]] && export SDL_VIDEODRIVER=wayland
+[[ "\$WAYLAND_DISPLAY" == wayland* ]] && export SDL_VIDEODRIVER=wayland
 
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/emulationstation.sh
```diff
@@ -257,7 +257,7 @@ if [[ "\$(uname --machine)" != *86* ]]; then
 fi
 
 # use SDL2 wayland video driver if wayland session is detected
-[[ "$XDG_SESSION_TYPE" == "wayland" ]] && export SDL_VIDEODRIVER=wayland
+[[ "\$WAYLAND_DISPLAY" == *wayland* ]] && export SDL_VIDEODRIVER=wayland
```
joolswills, line 260: Is the first asterisk needed? (As you say the variable starts with wayland) Maybe it's just enough to see if this variable contains something if it's only set for Wayland?

gizmo98: Thanks @joolswills. The asterik is not needed. I remove it.

Issue #3616, opened 2023-01-12
RetroPie-Setup cant restart Samba


yves84:
"Configure Samba ROM Shares" not working.

```
= = = = = = = = = = = = = = = = = = = = =
Installing dependencies for 'samba' : Configure Samba ROM Shares
= = = = = = = = = = = = = = = = = = = = =


= = = = = = = = = = = = = = = = = = = = =
Running action 'install_shares' for 'samba' : Configure Samba ROM Shares
= = = = = = = = = = = = = = = = = = = = =

Failed to restart samba.service: Unit samba.service not found.

```

The samba unit was renamed to `smbd.service`.

```
pi@retropie:~/RetroPie-Setup $ sudo systemctl list-units | grep -i samba
nmbd.service                                                                                                           loaded active running   Samba NMB Daemon                                                                       
smbd.service                                                                                                           loaded active running   Samba SMB Daemon
```

Ty.

cmitu:
> The samba unit was renamed to `smbd.service`.

The configure script tries both `samba` and `smbd` services, if the first fails, it will restart the other. We'll probably remove the old(er) name at some point from the script, but the restart works.

cmitu:
No need to close it, just leave it open until we modify the scriptmodule.

cmitu:
This should be fixed with #3643

PR #3643, opened 2023-02-22
samba.sh: only restart samba service if service exists

gizmo98:
Fixes misleading restart fail message https://github.com/RetroPie/RetroPie-Setup/issues/3616

### DIFF
--- a/scriptmodules/supplementary/samba.sh
+++ b/scriptmodules/supplementary/samba.sh
@@ -41,7 +41,8 @@ _EOF_
 }
 
 function restart_samba() {
-    service samba restart || service smbd restart
+    [[ ! -z $(systemctl list-units --full -all | grep smbd.service) ]] && service smbd restart
+    [[ ! -z $(systemctl list-units --full -all | grep samba.service) ]] && service samba restart
 }
 
 function install_shares_samba() {


scriptmodules/supplementary/samba.sh
```diff
@@ -41,7 +41,8 @@ _EOF_
 }
 
 function restart_samba() {
-    service samba restart || service smbd restart
+    [[ ! -z $(systemctl list-units --full -all | grep smbd.service) ]] && service smbd restart
+    [[ ! -z $(systemctl list-units --full -all | grep samba.service) ]] && service samba restart
```
joolswills, line 45: Not sure I like the duplication. Do we really need to check - can't we just hide the error ?

``
    local service
    for service in smbd samba; do
        systemctl restart $service 2>/dev/null
    done
``

If the check is needed that could be done in the loop I guess without duplicating the command.

gizmo98: Redirect output to /dev/null should work as well.

cmitu:
My idea when keeping the linked issue was to remove `samba` as service name and do something like:
```
systemctl -q is-active smbd.service && systemctl -q reload-or-restart smbd.service
```

gizmo98:
@cmitu this does only work if there is no samba.service. Debian stretch has smbd.service and samba.service. Do we remove backwards compatiblity for stretch? Debian bullseye and bookworm have only smbd.service. So we could add a debian version check for samba.service.
https://packages.debian.org/search?mode=path&suite=bookworm&section=all&arch=any&searchon=contents&keywords=samba.service

cmitu:
> Debian stretch has smbd.service and samba.service. Do we remove backwards compatiblity for stretch?

On _stretch_, the `samba.service` unit file is a `/dev/null` symlink - see https://sources.debian.org/src/samba/2%3A4.5.16%2Bdfsg-1%2Bdeb9u2/debian/samba.links/ and [here](https://sources.debian.org/src/samba/2%3A4.5.16%2Bdfsg-1%2Bdeb9u2/debian/samba.install/) for the files included in the package).

I don't think we'll loose compatiblity with _stretch_ by removing support for it, restarting `smbd` works on _stretch_ and should produce the same result.

gizmo98:
Ok. Then your solution looks much better. I will close this PR.

PR #3644, opened 2023-02-23
helpers - added inputBox function that uses the OSK

joolswills:
If the OSK is installed it will be used for input, otherwise it will fall back to using dialog --inputbox

function takes 3 parameters:

The title of the dialog, the default text and a minimum number of characters to accept.

Adjusted wifi module to use the new inputBox. There is no need to redirect cmd output with 2>&1 >/dev/tty as the inputBox function does this.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -133,6 +133,33 @@ function editFile() {
     [[ -n "$choice" ]] && echo "$choice" >"$file"
 }
 
+## @fn inputBox()
+## @param title title of dialog
+## @param text default text
+## @param minchars minimum chars to accept
+## @brief Opens an inputbox dialog and echoes resulting text. Uses the OSK if installed.
+function inputBox() {
+    local title="$1"
+    local text="$2"
+    local minchars="$3"
+    [[ -z "$minchars" ]] && minchars=0
+    local params=(--backtitle "$__backtitle" --inputbox "$title")
+    local osk="$(rp_getInstallPath joy2key)/osk.py"
+
+    if [[ -f "$osk" ]]; then
+        params+=(--minchars "$minchars")
+        text=$(python3 "$osk" "${params[@]}" 2>&1 >/dev/tty) || return $?
+    else
+        while true; do
+            text=$(dialog "${params[@]}" 10 60 "$text" 2>&1 >/dev/tty) || return $?
+            [[ "${#text}" -ge "$minchars" ]] && break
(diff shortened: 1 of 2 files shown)


joolswills:
@cmitu Please give this a try. The helper function would need adjusting to include the default text parameter for the python3 OSK.

scriptmodules/helpers.sh
```diff
@@ -133,6 +133,33 @@ function editFile() {
...
+    local title="$1"
+    local text="$2"
+    local minchars="$3"
+    [[ -z "$minchars" ]] && minchars=0
+    local params=(--backtitle "$__backtitle" --inputbox "$title")
+    local osk="$(rp_getInstallPath joy2key)/osk.py"
+
+    if [[ -f "$osk" ]]; then
+        params+=(--minchars "$minchars")
+        text=$(python3 "$osk" "${params[@]}" 2>&1 >/dev/tty)
```
cmitu, line 151: With the changes from https://github.com/RetroPie/RetroPie-Setup/pull/3638/commits/0b41d5f20b5907310548ec22456d7d16cdda0641, `$text` as an optional (last) parameter will work with the OSK. Current version will ignore it, so I think it's safe to change this to:
```
text=$(python3 "$osk" "${params[@]}" "$text" 2>&1 >/dev/tty)
```

joolswills: The current version throws an error with an additional parameter - so my plan was to get this in, and then you could adjust your PR to add the default parameter. With the current version I get

```
python3 /opt/retropie/admin/joy2key/osk.py --backtitle 'test' --inputbox test --minchars 4 "default text"
usage: osk.py [-h] --backtitle BACKTITLE --inputbox INPUTBOX [--minchars [MINCHARS]]
osk.py: error: unrecognized arguments: default text
```

joolswills: Regarding handling "cancel" - thanks. I missed that. I will adjust and we can retest. Appreciate your help.

cmitu: > With the current version I get.

Ugh, you're right - the extra arg is not working with the current version (damn you `argparse` !). OK, I'll redo the changes after this is merged - probably squash the `inputBox` addition and `osk` changes into a single commit.

joolswills: No worries. I'll fix this PR up later today and merge it and then you can rebase/rework your changes. Thank you :-)

joolswills:
@cmitu I added the return changes. Please let me know if you think this is ok and I'll merge (or feel free to do it)

cmitu:
Perfect, merging...

PR #3645, opened 2023-02-24
packages.sh - function rp_hasNewerModule - variable length commit hash

s1eve-mcdichae1:
Allows proper update detection for modules with 7-character commit sha, such as some RetroPie-Extra modules.

### DIFF
--- a/scriptmodules/packages.sh
+++ b/scriptmodules/packages.sh
@@ -601,10 +601,10 @@ function rp_hasNewerModule() {
                     local repo_commit="$(rp_resolveRepoParam "${__mod_info[$id/repo_commit]}")"
                     # if we are locked to a single commit, then we compare against the current module commit only
                     if [[ -n "$repo_commit" && "$repo_commit" != "HEAD" ]]; then
-                        # if we are using git and the module has an 8 character commit hash, then adjust
-                        # the package commit to 8 characters also for the comparison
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. I don't see an issue but is there a reason why we shouldn't use a minimum of 8 characters vs less ?

joolswills:
Apart from the fact RetroPie-Extra modules may not I mean.

https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1

Maybe we should enforce a minimum for modules but I'm happy to include this after I've tested.

joolswills:
(Please can you squash it also)

s1eve-mcdichae1:
> is there a reason why we shouldn't use a minimum of 8 characters vs less ?

The reason I've been using seven, is because that's what GitHub has shown me in places like this:

5E1EA537-B4C7-4114-8927-B965B7BCD2FF


I suppose it probably uses the same values as:

> Git can figure out a short, unique abbreviation for your SHA-1 values. If you pass --abbrev-commit to the git log command, the output will use shorter values but keep them unique; **it defaults to using seven characters but makes them longer if necessary** to keep the SHA-1 unambiguous

Happy to enforce a minimum. It looks like git itself will only provide them seven or longer; on the other hand, `checkout` is working with as few as four in some quick testing just now.

PR #3648, opened 2023-03-01
hypseus: update runtime parameters for ARM

cmitu:
Since SDL has been updated, make sure we don't encounter the performance issue reported on 2.0.16+ with RPI. After 2.0.16, using `SDL_TEXTUREACCESS_TARGET` for a `SDL_Texture` leads to an extra conversion inside SDL, so avoid it by telling `hypseus` to use `SDL_TEXTUREACCESS_STREAMING`.

Previous discussion and analysis of the performance impact:
 * https://github.com/DirtBagXon/hypseus-singe/issues/66
 * https://retropie.org.uk/forum/topic/18505/

### DIFF
--- a/scriptmodules/emulators/hypseus.sh
+++ b/scriptmodules/emulators/hypseus.sh
@@ -72,6 +72,8 @@ function configure_hypseus() {
     ln -sf "$md_conf_root/daphne/hypinput.ini" "$md_inst/hypinput.ini"
 
     local common_args="-framefile \"\$dir/\$name.txt\" -homedir \"$md_inst\" -fullscreen \$params"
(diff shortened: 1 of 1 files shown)


joolswills:
SDL is only updated on bullseye+. Just wanted to check this will be ok on older sdl also or if it needs a check.

cmitu:
I know.
The change was tested at the time of the issue and it didn't break existing install (i.e. with exiting 2.0.10 or newer). See the comments from https://github.com/DirtBagXon/hypseus-singe/issues/66#issuecomment-1042002504. I actually suggested to be made default

joolswills:
Thanks

PR #3649, opened 2023-03-01
lr-beetle-pce: add accurate PC Engine/CD/SuperGrafx core

nemo93:
Following the [discussion](https://retropie.org.uk/forum/topic/32181/add-the-accurate-version-of-beetle-pce/13), let's try adding that nice core into Retropie. 

This one truly shines, at least on RPi4 and allows for accurate emulation of PC Engine/TurboGrafx-16, PC Engine CD and SuperGrafx. If you ever had issue with a (CD-ROM²) game you should give this one a try.

* Single core for 3 systems: PC Engine/TurboGrafx-16, PC Engine CD and SuperGrafx.
* Accurate given any games I've tried ran without issues or glitches (unlike lr-beetle-pce-fast) including CD-ROM².
* Work with `.chd` and `.7z` without issue.
* More info on the [core page](https://github.com/libretro/beetle-pce-libretro).

About the script itself, I've tried my best to follow the Retropie's conventions. Anyhow the script was already robust enough thanks to @Darksavior (who originally created the script).

* Few core options are set otherwise you might end up with messy screen or gfx artifacts (eg. Art of Fighting/Ryuuko no Ken)
* Multitap is also disabled as some games won't work with it (eg. Cho Aniki).
* Tested only on RPi4 where all games run fullspeed (roughly 400+ titles tested). I don't know if it works the same on RPi3 or older models. Hence why the `!armv6`.
* [lr-beetle-pce-fast](https://github.com/libretro/beetle-pce-fast-libretro) as well as [lr-beetle-supergrafx](https://github.com/libretro/beetle-supergrafx-libretro) remain valid alternatives for very low-spec devices.

For those interesting in testing try first the [Arcade CD-ROM² titles](https://necretro.org/CD-ROM%C2%B2_systems#Arcade_CD-ROM.C2.B2). Those should be more demanding.

Thanks again to @Darksavior for the work done.

### DIFF
--- a/scriptmodules/libretrocores/lr-beetle-pce.sh
+++ b/scriptmodules/libretrocores/lr-beetle-pce.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-beetle-pce"
+rp_module_desc="PC Engine/CD/SuperGrafx emulator - Mednafen PCE port for libretro"
+rp_module_help="ROM Extensions: .7z .ccd .chd .cue .pce .sgx\n\nCopy your PC Engine roms to $romdir/pcengine\n\nCopy the required BIOS file syscard3.pce to $biosdir"
+rp_module_licence="GPL2 https://raw.githubusercontent.com/libretro/beetle-pce-libretro/master/COPYING"
+rp_module_repo="git https://github.com/libretro/beetle-pce-libretro.git master"
+rp_module_section="main"
+rp_module_flags="!armv6"
+
+function sources_lr-beetle-pce() {
+    gitPullOrClone
+}
+
+function build_lr-beetle-pce() {
+    make clean
+    make
+    md_ret_require="$md_build/mednafen_pce_libretro.so"
+}
+
+function install_lr-beetle-pce() {
+    md_ret_files=(
+        'mednafen_pce_libretro.so'
+        'COPYING'
+        'README.md'
+    )
+}
+
+function configure_lr-beetle-pce() {
+    mkRomDir "pcengine"
+    defaultRAConfig "pcengine"
+
+    addEmulator 1 "$md_id" "pcengine" "$md_inst/mednafen_pce_libretro.so"
+    addSystem "pcengine"
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    # Set core options
+    setRetroArchCoreOption "pce_aspect_ratio" "4:3"
+    setRetroArchCoreOption "pce_multitap" "disabled"
+    setRetroArchCoreOption "pce_scaling" "hires"
+}
+


scriptmodules/libretrocores/lr-beetle-pce.sh
```diff
@@ -0,0 +1,62 @@
...
+rp_module_desc="PC Engine/CD/SuperGrafx emulator - Mednafen PCE port for libretro"
+rp_module_help="ROM Extensions: .7z .ccd .chd .cue .pce .sgx\n\nCopy your PC Engine roms to $romdir/pcengine\n\nCopy the required BIOS file syscard3.pce to $biosdir"
+rp_module_licence="GPL2 https://raw.githubusercontent.com/libretro/beetle-pce-libretro/master/COPYING"
+rp_module_repo="git https://github.com/libretro/beetle-pce-libretro.git master"
+rp_module_section="main"
+rp_module_flags="!armv6"
+
+function _update_hook_lr-beetle-pce() {
+    # move from old location and update emulators.cfg
+    renameModule "lr-mednafen-pce" "lr-beetle-pce"
```
joolswills, line 22: I assume this module is based on copy and paste from the pce-fast module as there is no need for this. We don't have a lr-mednafen-pce module that is being renamed.

scriptmodules/libretrocores/lr-beetle-pce.sh
```diff
@@ -0,0 +1,62 @@
...
+        "pcengine"
+    )
+
+    local system
+    for system in "${systems[@]}"; do
+        mkRomDir "$system"
+        defaultRAConfig "$system"
+        addEmulator 1 "$md_id" "$system" "$md_inst/mednafen_pce_libretro.so"
+        addSystem "$system"
+    done
```
joolswills, line 53: If there is only one system we don't need a loop for this.

scriptmodules/libretrocores/lr-beetle-pce.sh
```diff
@@ -0,0 +1,51 @@
...
+function build_lr-beetle-pce() {
+    make clean
+    make
+    md_ret_require="$md_build/mednafen_pce_libretro.so"
+}
+
+function install_lr-beetle-pce() {
+    md_ret_files=(
+        'mednafen_pce_libretro.so'
+        'README.md'
```
joolswills, line 33: We should include COPYING also - It's not included in the other module but probably should be there.

scriptmodules/libretrocores/lr-beetle-pce.sh
nemo93: Understood, removed.

scriptmodules/libretrocores/lr-beetle-pce.sh
nemo93: removed.

scriptmodules/libretrocores/lr-beetle-pce.sh
nemo93: Added.

joolswills:
Looks good now. If you are happy also I will merge in a bit. Thanks.

nemo93:
You surprised me by already reviewing that one :'D I'm very keen to credit anyone who contributed and certainly don't want to upset anyone.

Edit: I'm happy with it and glad to see that core added. For all PC Engine lovers that is certainly a fine addition.

Thanks.

joolswills:
Well, I think it's not possible to credit everyone as the other module had more than one contributor I'm sure (including me) and since it's really just a copy from the other module with a few additional core options. It's just easier to keep the thanks in the PR. 

No upset, just didn't think it was needed in the commit.

joolswills:
I've just realised this is in "main" section after I merged. I suppose that's ok, but maybe optional would have been better at least initially.

joolswills:
I will move to optional packages for now.

nemo93:
> I've just realised this is in "main" section after I merged. I suppose that's ok, but maybe optional would have been better at least initially.

I would disagree here. I've used 'main' provided how stable, accurate and glitch-free that core is. In fact I consider it more compatible than [lr-beetle-pce-fast](https://github.com/libretro/beetle-pce-fast-libretro). Users really have to use that core if they want true PC Engine/CD emulation (provided they're on RPi4).

Yet the decision is yours and I'll respect it.

joolswills:
Well - it's more involved than that. Since the addEmulator call makes this the default, it would become the default on all platforms in main (including rpi2). It can be made the default for the rpi4, but then we need to change the logic in the other module.

joolswills:
I'm not actually talking about the suitability of it in general.

joolswills:
To clarify - 

If both modules are in "main", we probably need to make the -fast core the default for slower devices (armv7 maybe), and this module the default for armv8. But having them both as the default would mean the one installed first would become the default, no matter which system. So it needs a little more work for that.

nemo93:
Makes sense. If you ask me I'd certainly make this core available only on RPi4 (+any other arch except RPi0/1, 2 and 3). Then -fast and -supergrafx core should be available only on RPi0/1, 2 and 3.

Not sure if this could be done. I'll follow any guidance (I could stay put as well!).

joolswills:
I can sort this - but not now as I lack free time this week. I will open an issue for myself and look at this later. I will test on rpi2/rpi3 and decide the best route.

joolswills:
Not a problem - contributions are welcome. Thanks for submitting it.

PR #3653, opened 2023-03-03
mupen64plus.sh: update GlideN64 config version

gizmo98:
GlideN64 overwrites settings if config version is to old. Current config version is 29. https://github.com/gonetz/GLideN64/blob/1f4d04f43b53739bc9b128ab5577d20e3d60ed6a/src/Config.h#L8

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -345,7 +345,7 @@ function configure_mupen64plus() {
             echo "[Video-GLideN64]" >> "$config"
         fi
         # Settings version. Don't touch it.
-        iniSet "configVersion" "17"
+        iniSet "configVersion" "29"
         # Bilinear filtering mode (0=N64 3point, 1=standard)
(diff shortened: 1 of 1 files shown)


joolswills:
We also have a default in https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/mupen64plus/mupen64plus.sh#L253 (Which is overridden by this).

I think we should extract this from the source and copy it to "$md_inst" for use at the configure stage.

from:
https://github.com/gonetz/GLideN64/blob/1f4d04f43b53739bc9b128ab5577d20e3d60ed6a/src/Config.h#L8

I can look into this if you like. Will save us from having to keep this up to date.

joolswills:
I'll merge this - but will look into a solution for managing this automatically.

joolswills:
Would of course be easier if upstream didn't overwrite it each time - I understand backward compatibility issues, but it would be a lot simpler. It's quite an unusual way of dealing with it to overwrite if the version doesn't match I think.

PR #3666, opened 2023-03-06
emulationstation.sh: workaround Ubuntu 22.04 focus problem

gizmo98:
Emulationstation has focus problems under Ubuntu 22.04's "GNOME on Wayland" session. Don't use SDL2's wayland driver and run emulationstation with --fullscreen-borderless if desktop session is "GNOME on Wayland".


joolswills:
Thanks.

PR #3667, opened 2023-03-07
Update OS versions

gizmo98:
-Update Debian unstable version. Current Debian trixie/unstable has version 13. Debian bookworm/testing has version 12. 
-Add Ubuntu 22.10. Ubuntu >= 22.10 and Debian >= 12 have the package libsdl2-tests. Set __os_debian_ver="12" for Ubuntu versions >= 22.10.


joolswills:
thanks.

PR #3668, opened 2023-03-07
lr-prboom: workaround a recent regression

cmitu:
Until [1] is resolved - one way or another - use a good known commit to build the PrBoom core. Upstream core is relatively static, so we're not loosing much for pinning the version.

[1] https://github.com/libretro/libretro-prboom/issues/181


joolswills:
Thanks.

cmitu:
We can revert this now, upstream fixed the issue.

PR #3671, opened 2023-03-08
>>Text translated from Portuguese (Brazil) into English by Google tra…

flpowergames:
…nslator<<

Hello! I'd like to share a theme I developed for the RetroPie 4.8 update!


flpowergames:
...

PR #3676, opened 2023-03-10
openmsx: update the RetroPie initialization script

cmitu:
* fixed a typo in the joystick name cleanup instruction. When the joystick name contained the cleaned up characters, the auto-loading didn't work
* simplified the regex for both joystick and rom name cleanup instructions
* removed the 'osd' menu hiding from our code, since openMSX is now doing it itself (after a small delay). This prevents a warning message showed if the 'osd' menu cannot be found


joolswills:
Thanks.

PR #3677, opened 2023-03-10
Hello! I'm sharing my project with the RetroPie community

flpowergames:


cmitu:
This is (still) incorrect. You should add just one line for your theme, containing:
```
'flpowergames flpowergamesretro_v1.0'
```
It should look like:
``` sh
...
         'cjonasw raspixel-320-240'
         'crxone 3twenty2fourty'
         'leochely Guilty-Gear'
+        'flpowergames flpowergamesretro_v1.0'
     )
     while true; do
         local theme
...
```

I do think your theme needs more work. 

The gamelist font is incomplete and it doesn't show all characters from the game title. Letters like `-`, or `(`, `)`, `[`, `]` are missing from the game title. Enable the 'Last Played' collection and see how the game titles are showing in the default (Carbon) theme and then compare it to how it looks in your theme.
image

PR #3678, opened 2023-03-11
dolphin: update build requirements

cmitu:
Since upstream started requiring a C++20 capable compiler (gcc 10 min), update our build support based on the available gcc version:

 * minimum gcc version is 8 (Ubuntu 18.04, Debian 10)
 * current master can compile with gcc 10 (Debian 11, Ubuntu 22.04)
 * allow building with gcc 9 (Ubuntu 20.04)

Upstream has transitioned and recommends using Qt6 for building, so use that when available (Ubuntu 22.04 for now, but Debian 'bookworkm' and 'bullseye-backports' also include it).

Dependencies updated to include `gettext` and `liblzma-dev`.


joolswills:
Looks good. Thanks.

PR #3679, opened 2023-03-11
return first match of dpkg-query status

theofficialgman:
fixes multiarch which would previously fail due to incorrect parsing of installed packages. the output of out would look like `install ok installed 2.26.1+dfsg-1install ok not-installed` and this would get parsed as having no version and status of not installed.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -180,10 +180,10 @@ function hasPackage() {
 
     local ver
     local status
-    local out=$(dpkg-query -W --showformat='${Status} ${Version}' $1 2>/dev/null)
+    local out=$(dpkg-query -W --showformat='${Status};${Version};' $1 2>/dev/null)
     if [[ "$?" -eq 0 ]]; then
-        ver="${out##* }"
-        status="${out% *}"
+        ver="$(echo ${out} | awk -F';' '{ print $2 }')"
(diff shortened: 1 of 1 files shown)


joolswills:
I think just doing `local out=$(dpkg-query -W --showformat='${Status} ${Version}' $1 2>/dev/null | head -n1)` would be simpler - please can you test that. Seem to be ok here, and will add the change if it works for you.

joolswills:
Actually scrap that   - Ideally I'd like to avoid calling awk twice so will come up with a solution.

joolswills:
I was missing a LF - try `local out=$(dpkg-query -W --showformat='${Status} ${Version}\n' $1 2>/dev/null | head -n1) `

theofficialgman:
> ~I think just doing `local out=$(dpkg-query -W --showformat='${Status} ${Version}' $1 2>/dev/null | head -n1)` would be simpler - please can you test that. Seem to be ok here, and will add the change if it works for you.~

That won't work because the output it on one line already. It's not getting truncated to one line because of the variable setting, it just actually is one line of output

theofficialgman:
> I was missing a LF - try `local out=$(dpkg-query -W --showformat='${Status} ${Version}\n' $1 2>/dev/null | head -n1) `

yeah this will work

PR #3680, opened 2023-03-12
Updates for next AGS release (v3.6.0)

Gemba:
As discussed in the forum: https://retropie.org.uk/forum/topic/33969/adventuregamestudio-ags-scriptmodule-updates

Maybe you should wait until AGS 3.6.0 is final before merging.

Again: Thanks for all the advise for the scriptmodule.

### DIFF
--- a/scriptmodules/emulators/ags.sh
+++ b/scriptmodules/emulators/ags.sh
@@ -11,14 +11,14 @@
 
 rp_module_id="ags"
 rp_module_desc="Adventure Game Studio - Adventure game engine"
-rp_module_help="ROM Extension: .exe\n\nCopy your Adventure Game Studio roms to $romdir/ags"
+rp_module_help="ROM Extension: .exe\n\nCopy your Adventure Game Studio roms to $romdir/ags/<game>/"
 rp_module_licence="OTHER https://raw.githubusercontent.com/adventuregamestudio/ags/master/License.txt"
-rp_module_repo="git https://github.com/adventuregamestudio/ags.git release-3.5.1"
+rp_module_repo="git https://github.com/adventuregamestudio/ags.git release-3.6.0"
 rp_module_section="opt"
 rp_module_flags="!mali"
 
 function depends_ags() {
-    getDepends xorg pkg-config libaldmb1-dev libfreetype6-dev libtheora-dev libvorbis-dev libogg-dev liballegro4-dev
+    getDepends pkg-config libaldmb1-dev libfreetype6-dev libtheora-dev libvorbis-dev libogg-dev liballegro4-dev
 }
 
 function sources_ags() {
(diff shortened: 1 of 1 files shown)


scriptmodules/emulators/ags.sh
```diff
@@ -11,45 +11,79 @@
 
 rp_module_id="ags"
 rp_module_desc="Adventure Game Studio - Adventure game engine"
-rp_module_help="ROM Extension: .exe\n\nCopy your Adventure Game Studio roms to $romdir/ags"
+rp_module_help="ROM Extension: .exe\n\nCopy your Adventure Game Studio roms to $romdir/ags/<game>/"
 rp_module_licence="OTHER https://raw.githubusercontent.com/adventuregamestudio/ags/master/License.txt"
-rp_module_repo="git https://github.com/adventuregamestudio/ags.git release-3.5.1"
+rp_module_repo="git https://github.com/adventuregamestudio/ags.git master"
```
joolswills, line 16: Assuming they will be doing a branch / tag for the release we should use that here (When it's done).

scriptmodules/emulators/ags.sh
```diff
@@ -11,45 +11,79 @@
...
     make -C Engine
-    md_ret_require="$md_build/Engine/ags"
+    md_ret_require="$md_build/ags"
 }
 
 function install_ags() {
-    make -C Engine PREFIX="$md_inst" install
+    make -C Engine install
+    # install Eawpatches GUS patch set (see: http://liballeg.org/digmid.html)
+    download "http://www.eglebbk.dds.nl/program/download/digmid.dat" - | bzcat >"$md_inst/bin/patches.dat"
```
joolswills, line 41: Just to note - moving this from the configure stage to the install stage means it will be packaged with our pre-built binaries rather than installed after. Not an issue, just noting.

scriptmodules/emulators/ags.sh
```diff
@@ -11,45 +11,79 @@
...
+    )
+
+    cat > "$launcher" << _EOF_
+#!/usr/bin/env bash
+
+# ROM usually denotes "<something>.exe" whereas the game files are expected in
+# "<something>/" folder. <something>.exe and <something>/ are expected to be
+# sibling in the $romdir/ags folder.
+ROM=\$1
+fn=\$(basename "\$ROM")
```
joolswills, line 73: you could use a bash built-in for this as you do with the folder= line.

joolswills:
Thanks. Made a few minor comments. Looks good but will need to test etc.

scriptmodules/emulators/ags.sh
Gemba: Yes. Right. Currently there is a lot of activity at the development/main branch.

scriptmodules/emulators/ags.sh
Gemba: Changed `basename` and `dirname` to bash only commands.

Gemba:
FYI Changed the git revision to "release-3.6.0" in the scriptmodule, as this version has been released two weeks ago.

joolswills:
Thanks - Will give it a test this weekend and merge.

joolswills:
I did some limited testing only so far.Added https://laurahunt.itch.io/if-on-a-winters-night-four-travelers - but I had to add the .ags extension to ES. I think it would be worth adding this as we only have .exe currently. I had some older titles I tested with before but I don't seem to have them now, so I'd like to test with an older titles for backward compatibility etc.

The mentioned game required a mouse. I don't know if there is a way to allow joypad support in the config, but that also may be something worth looking into (not necessarily now but just noting it)

joolswills:
This testing was on a RPI4 with bullseye BTW. I will also test on Buster (and with a rpi3 with legacy drivers). But being SDL2 it should be fine.

Gemba:
Thanks, @joolswills 

FYI I filed a PR over there to get AGS scraping covered: https://github.com/detain/skyscraper/pull/23

joolswills:
Sorry for the delay. Thank you for your work on this.

joolswills:
I've been a bit busy, but I noticed this has some dependencies missing when going through errors from my build system. It's missing SDL2 and also cmake dependencies. I will commit a fix.

PR #3683, opened 2023-03-14
retroarch: fix core info update

cmitu:
When the upstream info repository adds an `.info` file we also generate, `git` will refuse the pull done during the update of the info files. Just remove the files before the update and add them after `git` synchronizes with upstream.

Removed the Mess core info files, which triggered this bug initially.


scriptmodules/emulators/retroarch.sh
```diff
@@ -110,9 +110,17 @@ function update_core_info_retroarch() {
     local dir="$configdir/all/retroarch/cores"
     # remove if not a git repository and do a fresh checkout
     [[ ! -d "$dir/.git" ]] && rm -fr "$dir"
+    # remove our locally generated `.info` files, just in case upstream adds them
+    local info_file
+    for info_file in "$md_data"/*.info; do
+        rm "$configdir/all/retroarch/cores/$(basename $info_file)" 2>/dev/null
+    done
```
joolswills, line 117: What about just doing a `git clean -f -d` ?

cmitu: Yes, that's better. Forgot about `git clean`, I think I used it just a couple of times.
I modified the removal/cleanup for local files to use it.

scriptmodules/emulators/retroarch.sh
```diff
@@ -110,9 +110,14 @@ function update_core_info_retroarch() {
...
+    # remove our locally generated `.info` files, just in case upstream adds them
+    [[ -d "$dir/.git" ]] && git -C "$dir" clean -q -f "*.info"
     gitPullOrClone "$configdir/all/retroarch/cores" https://github.com/libretro/libretro-core-info.git
-    # Add the info files for cores/configurations not available upstream
-    cp -f "$md_data/"*.info "$configdir/all/retroarch/cores"
+    # add our info files for cores not included in the upstream repo
+    for info_file in "$md_data"/*.info; do
+        [[ ! -f "$configdir/all/retroarch/cores/$(basename $info_file)" ]] && \
+            cp "$info_file" "$configdir/all/retroarch/cores/"
+    done
```
joolswills, line 120: What's the reason for this change over the previous copy command ?

joolswills: To not overwrite any files that may be added to the repo at a later date ? Maybe cp -u would be enough to handle that (Since the date would be newer). But if this change is needed in case I missed something it would be best to use a bash built-in over basename (as we do elsewhere).

joolswills: Also it's missing a `local`

joolswills:
I'm wondering whether it may just be easier to fork the retroarch repo, add the additional .info files and handle it there. Might simplify things ?

scriptmodules/emulators/retroarch.sh
cmitu: > To not overwrite any files that may be added to the repo at a later date ?...

Yes, that was the intention. `cp --update` works better, thanks.

cmitu:
@joolswills I thought about using a repo in the beginning, but it seems hardly worth it for just a couple of files (one of which is probably unused by the vast majority of users). Upstream doesn't distribute those files, so they're not really useful for them; they might accept a PR though, since they have `.info` files for cores not directly distributed by the _libretro_ project.

joolswills:
Thanks.

PR #3684, opened 2023-03-15
attractmode: add default font dependency

cmitu:
Added the Free fonts, since `attract` will not start without its default font (FreeSans).
The font search code has changed recently, triggering now an error without FreeSans being present.


joolswills:
Thanks.

Issue #3654, opened 2023-03-03
Extract configVersion for GlideN64 automatically


joolswills:
https://github.com/RetroPie/RetroPie-Setup/pull/3653

We could extract this from the source and store in `$md_inst` for use at the configure stage.

gizmo98:
@joolswills after looking at the current package script i have seen configVersion is already stored in $md_inst/share/mupen64plus/GLideN64_config_version.ini. PR https://github.com/RetroPie/RetroPie-Setup/pull/3688 is taking this into account.

PR #3688, opened 2023-03-19
mupen64plus.sh: use extracted configVersion in GlideN64 config section

gizmo98:
We already extract GlideN64 config version. Use it to finally set right config version in mupen64plus package script. Mupen64plus startup script should do this as well, but it is not wrong to do it here. -set right GlideN64 config version in "$config.rp-dist" -set right GlideN64 config version in "$config" because GlideN64 overwrites user settings if config version is wrong fixes: https://github.com/RetroPie/RetroPie-Setup/issues/3654

### DIFF
--- a/scriptmodules/emulators/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus.sh
@@ -321,6 +321,10 @@ function configure_mupen64plus() {
     # on the rp-dist file. This preserves any user configs from modification and allows us to have
     # a default config for reference
     if [[ -f "$config" ]]; then
+        # patch existing mupen64plus config because Gliden64 overwrites GlideN64 user configs
+        # if configVersion is wrong 
+        grep -q "configVersion" "$config" && iniConfig " = " "" "$config" && iniSet "configVersion" "$(cat $md_inst/share/mupen64plus/GLideN64_config_version.ini)"
+    
         mv "$config" "$config.user"
         su "$user" -c "$cmd"
         mv "$config" "$config.rp-dist"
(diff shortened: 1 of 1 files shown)


scriptmodules/emulators/mupen64plus.sh
```diff
@@ -321,6 +321,10 @@ function configure_mupen64plus() {
     # on the rp-dist file. This preserves any user configs from modification and allows us to have
     # a default config for reference
     if [[ -f "$config" ]]; then
+        # patch existing mupen64plus config because Gliden64 overwrites GlideN64 user configs
+        # if configVersion is wrong 
+        grep -q "configVersion" "$config" && iniConfig " = " "" "$config" && iniSet "configVersion" "$(cat $md_inst/share/mupen64plus/GLideN64_config_version.ini)"
```
joolswills, line 326: This isn't a good one-liner imho. It would be more readable to split it up and have inside an `if`

Also duplicating the path - keeping the config in a variable would be better.

Also note you can do `$(<FILE)` instead of cat.

joolswills:
Related to this, I don't like the logic and ini file handling in the launch script code.

For GLideN64 I think we should just make sure there is a section present. On non RPI systems this will be generated on first launch using GLideN64, but we could create a minimal config in configure_mupen64plus with the configVersion.

Probably we should handle all configs that the startup script can handle and make sure they are already set by default. However initially, just removing the config version handling would be a good start.

I'm going to submit a PR for this (Also as I had assigned the original issue to myself).

PR #3689, opened 2023-03-19
Update dolphin.sh

hotspoons:
Updated dolphin dependencies to use the correct package name for older debian/ubuntu releases for qtbase5-private-dev


cmitu:
@hotspoons thank you !

PR #3692, opened 2023-03-21
Display Memory Usage in MB or GB

Jackfritt:


cmitu:
@Jackfritt thanks, that would be a nice addition.
Wouldn't it be simpler to use `printf` from `awk` instead of manipulating the records ? As an added bonus, it can be used to control the number of decimals (like `free` does).

Jackfritt:
I only stole it chamelessly from here
https://stackoverflow.com/questions/29811297/how-to-display-proc-meminfo-in-megabytes
There are examples with printf. I was fine with my solution, but if someone has more time.
Im sooo lazy ;)

cmitu:
It's easier with just
```
free -h | awk '/2/ { printf("%s (Free) / %s (Total)", $4, $2) }'
```

Jackfritt:
Thats rly nice.
I have Swap Space and when there is free available, it displays different.

free 
              total        used        free      shared  buff/cache   available
Mem:         945364      316620       62824       24028      565920      541816
Swap:       1284100       43008     1241092


free -h | awk '/2/ { printf("%s (Free) / %s (Total)", $4, $2) }'

61Mi (Free) / 923Mi (Total)1,2Gi (Free) / 1,2Gi (Total)


If you only want Free Memory you can try this one.
LANG should display in english for greep to work correct.
 grep mem, exclude other lines.
--giga displays MB and GB.
I like to display Total first and then Free. Also added a newline..

LANG=C free -h --giga | grep "Mem:" | awk '/2/ { printf("%s (Total) / %s (Free)\n", $2, $4) }'

cmitu:
> I have Swap Space and when there is free available, it displays different.

You can instruct `free` to include a 3rd line with the totals (RAM + Swap):
```
free -ht | awk 'NR==4 { printf("%s (Free) / %s (Total)\n", $4, $2) }'
```

Jackfritt:
This displays it all together. But do Emulators run smoothly with swap space?( if its in SSD or similar ok)
I would prefer to see my real RAM instead of "slower" swap space.
I dont know much about Emulator interna. The standard retropie install only uses RAM or?
Maybe let the script only display RAM?
I also found out that LC_ALL is the preferred way to go...

LC_ALL=C free -ht --giga | grep "Total:" | awk '/2/ { printf("%s (Total) / %s (Free)\n", $2, $4) }'

So with my opinion this one should be good?

LC_ALL=C free -h --giga | grep "Mem:" | awk '/2/ { printf("%s (Free) / %s (Total)\n", $4, $2) }'

cmitu:
> This displays it all together. But do Emulators run smoothly with swap space?( if its in SSD or similar ok)
> I would prefer to see my real RAM instead of "slower" swap space.

 In your previous reply you mentioned the command didn't show the swap, I'm not sure which one you'd prefer. If the current command includes the swap space in the total/free report, then it should be included also in the new commands.

> I dont know much about Emulator interna. The standard retropie install only uses RAM or?

By default on a Raspberry Pi there is no swap added, but users can add it themselves so it may be beneficial to add it.

> Maybe let the script only display RAM?
> I also found out that LC_ALL is the preferred way to go...

What locale/language do you have configured on your system ?

cmitu:
I've updated the changes with the simpler version. I've tried a few locales, but I don't see anything affected by using a different one than the default.

PR #3694, opened 2023-03-22
steamlink: update dependencies and runtime config

cmitu:
* removed the launcher through Xorg since Steam Link can now run with a desktop env on both RaspiOS 10 and 11, choosing between the DRM or MMAL drivers. Tested on a RPI 3 (RaspiOS 10) and RPI 4 (RaspiOS 11).

* adjusted the dependencies to remove xorg/matchbox/zenity since they're not strictly necessary and added a few packages which might not be automatically installed on RetroPie. The list of the packages that `steamlink` tries to install on first run in is `$XDG_DATA_HOME/SteamLink/steamlinkdeps.txt`, I just picked the ones that are not already part of existing dependencies (`vlc` will install most of them).

* disabled the package on 64bit, since is not available there (see https://steamcommunity.com/app/353380/discussions/0/1649917420735113143/)

### DIFF
--- a/scriptmodules/ports/steamlink.sh
+++ b/scriptmodules/ports/steamlink.sh
@@ -13,11 +13,11 @@ rp_module_id="steamlink"
 rp_module_desc="Steam Link for Raspberry Pi 3 or later"
 rp_module_licence="PROP https://steamcommunity.com/app/353380/discussions/0/1743353164093954254/"
 rp_module_section="exp"
-rp_module_flags="!all rpi3 rpi4"
+rp_module_flags="!all rpi3 rpi4 !64bit"
 rp_module_help="Stream games from your computer with Steam"
 
 function depends_steamlink() {
-    getDepends python3-dev libinput10 libxkbcommon-x11-0 matchbox-window-manager xorg zenity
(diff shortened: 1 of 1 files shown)


scriptmodules/ports/steamlink.sh
```diff
@@ -13,11 +13,11 @@ rp_module_id="steamlink"
 rp_module_desc="Steam Link for Raspberry Pi 3 or later"
 rp_module_licence="PROP https://steamcommunity.com/app/353380/discussions/0/1743353164093954254/"
 rp_module_section="exp"
-rp_module_flags="!all rpi3 rpi4"
+rp_module_flags="!all rpi3 rpi4 32bit"
```
joolswills, line 16: This will disable for all platforms, then enable for rpi3 and rpi4 and then any 32bit platforms. I don't think that's what you want as it would mean the module is enabled for 64bit on the rpi3/rpi4.

`!all rpi3 rpi4 !64bit` would enable for rpi3 and rpi4 but disable for 64bit.

cmitu: You know, `!64bit` was my first choice. I should have known better. Modified.

PR #3696, opened 2023-03-23
mupen64plus - GlideN64 config version and default config logic changes

### LINKED ISSUE #3654: Extract configVersion for GlideN64 automatically

joolswills:
Add missing "local source" in install_mupen64plus

Set configVersion from GLideN64_config_version.ini in $md_conf_root/n64/mupen64plus.cfg as well as mupen64plus.cfg.rp-dist

Add defaults and values handled by startup script for Video-GLideN64 and Video-Rice to default mupen64plus.cfg and re-order logic.

Remove logic in mupen64plus.sh startup script to set configVersion and only change values if they are present in the config (Rather than checking for a section and then adding it)

Fixes #3654 and closes #3688


scriptmodules/emulators/mupen64plus.sh
```diff
@@ -330,29 +340,35 @@ function configure_mupen64plus() {
...
+    # Create GlideN64 section in .cfg
+    if ! grep -q "\[Video-GLideN64\]" "$config"; then
+        echo "[Video-GLideN64]" >> "$config"
+    fi
+    # Settings version
+    iniSet "configVersion" "$gliden64_ver"
+    # Use native res
+    iniSet "UseNativeResolutionFactor" "1"
+    # Enable legacy blending
+    iniSet "EnableLegacyBlending" "True"
```
gizmo98, line 364: Please remove this line. EnableLegacyBlending helps to get some extra fps on slow devices but it can cause visual problems. If "mupen64plus_compatibility_check" autoconf is set it will be enabled anyway.

dankcushions: In the previous commit EnableLegacyBlending was only true for rpi: https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/mupen64plus.sh#L334-L355

I think that's right as all (current) pis are fighting for performance in GLideN64, and the blacklist will disable it for known problem games. I am not sure why it (and some other settings) been taken out of that block - perhaps unintentional @joolswills  ? :)

however you do raise an issue with the launch script, which seems to always enable, and the disable it if the game is in the blacklist. IMO that's wrong - it should just disable when necessary rather than ignore the cfg setting. i.e. remove this line: https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/emulators/mupen64plus/mupen64plus.sh#L267

gizmo98: @dankcushions it just resets to best „rpi“ default setting for performance before it checks blacklist. Autoset function could be the better location. If you want to preserve user settings you need to read,  store and restore it after m64p closes.

joolswills: I'll change it. I got the wrong default (probably confused due to the compatibility check code). I wanted to make sure we always have the keys that are to be used by the launch script which is why I added them outside.

dankcushions: Yeah it's a little unwieldy! https://github.com/RetroPie/RetroPie-Setup/pull/2960 simplifies this some ;) it adds the headers if they don't exist.

PR #3698, opened 2023-03-30
lr-caprice32: work around an upstream cloning issue

cmitu:
Upstream uses a shallow submodule, but is tracking a rolling branch that changed since the submodule was added. Older `git` versions cannot fetch the ref registered in the parent repo and the recursive clone fails. It affects Debian _buster_ and Ubuntu 18.04, plus other downstream distros (Linux Mint 19.3).

Clone the repo temporarily and remove the submodule, since it's not used actively by the core. Issue was reporeted upstream in https://github.com/libretro/libretro-cap32/issues/116.


joolswills:
Thanks

cmitu:
We can revert this now, upstream has modified the submodule configuration/commit and it's building again on _buster_/older `git` versions.

PR #3701, opened 2023-04-09
Update esthemes.sh with 1 new theme

RetroHursty69:
The theme is called ToggleBobble and is a 16:9 ratio theme that works on Pi.


joolswills:
Thanks.

PR #3703, opened 2023-04-16
openmsx: bump version to 18.0 and update input config

cmitu:
Updated to the 18.0 release, which requires GCC 9 and it's installable on Debian 11 (bullseye)/Ubuntu 20.04+.

The input auto-loading script has been updated to use the joystick's SDL name, while previously was using the 'udev' device name. This fixes the issues when the joystick nane (SDL) is different than the 'udev' reported name and thus prevented the auto-loading.


joolswills:
Thanks.

PR #3704, opened 2023-04-17
skyscraper - prevent hall-of-mirrors effect

s1eve-mcdichae1:
By default, Skyscraper creates a composite image with screenshot, logo, and box art, and then saves this as a "screenshot" itself:

Armored Core (USA) (Reprint)

When the local `esgamelist` is then scraped after that, this already-processed "screenshot" (with logo and box art) has the logo and box art applied to it again, leading to a receding "hall of mirrors" effect. This effect is amplified with repeated scraping:

Armored Core (USA) (Reprint) 2

Armored Core (USA) (Reprint) 3

To prevent this, do not cache these already-processed screenshots when scraping the local `esgamelist` source.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -210,7 +210,10 @@ function _init_config_skyscraper() {
     done
 
     # If we don't have a previous config.ini file, copy the example one
-    [[ ! -f "$scraper_conf_dir/config.ini" ]] && cp "$md_inst/config.ini.example" "$scraper_conf_dir/config.ini"
+    if [[ ! -f "$scraper_conf_dir/config.ini" ]]; then
+        cp "$md_inst/config.ini.example" "$scraper_conf_dir/config.ini"
+        sed -i 's/\[esgamelist\]/[esgamelist]\ncacheScreenshots="false"/' "$scraper_conf_dir/config.ini"
+    fi
 
     # Try to find the rest of the necessary files from the qmake build file
     # They should be listed in the `unix:examples.file` configuration line


cmitu:
This is a user choice/(mis-)configuration, we don't want to force the exclusion of screenshots unless the user requests it. So, no, I'd rather not include this.

s1eve-mcdichae1:
> This is a user choice/(mis-)configuration,

The above images result from default settings so there's not a lot user choice there (besides selecting of the local `esgamelist` source.)

The other option, what I've done locally to prevent this, is to edit the `config.ini` with

    [esgamelist]
    cacheScreenshots="false"

...but I don't think the current `iniSet` function can handle that being in a `[section]` like that, right?

what if we `applyPatch` to put this in the `.ini` at the source. That way the user could edit their file if they *really* want to scrape local images (I would suggest disable all of them by default) from the gamelist\*, but won't get this HoM effect by default.

\*(though I find it's better to use `import` for images, and use `esgamelist` only to pick up manual edits I've made to the textual data name, description, etc.)

cmitu:
> > 
> 
> The above images result from default settings so there's not a lot of user choice there (besides selecting of the local `esgamelist` source.)

There _is_ the choice of excluding screenshots from the scraping, either through the dialog based script or through the config file. Your previous patch would have overridden the user's choice made in the `dialog` script, which is not what the user expects.

I'm ok with setting a default of 'cacheScreenshots = false' in the default `.ini`. I consider using the `esgamelist` source and "advanced" option and the user should understand how to configure scraping from it.

Can you try and do the config without a patch ? Using `sed`, it's a one liner:
```
sed 's/\[esgamelist\]/[esgamelist]\ncacheScreenshots=false/' $configdir/all/skyscr
aper/config.ini
```
but it needs to set just once, the 1st time the `config.ini` is copied/generated.

s1eve-mcdichae1:
> I consider using the `esgamelist` source [an] "advanced" option and **the user should understand how to configure scraping from it.**

There are at least a handful of users I've walked through repairing this who didn't, but I suppose there will always be those few poking around where they don't understand (I'm one of them, to be sure. That's often how I learn.)

> Can you try and do the config without a patch ? Using `sed`, it's a one liner:

...in fact, that's how I changed the file in the first place, before making the patch\*. I don't know why I didn't think to just do it that way *in* the script.

\*(because, when I edited the file with `nano` it added the "newline at end of file" too which wasn't there before, and I was trying to change as little as possible. So I re-tried it with `sed`, although I just inserted at a particular line number. Using the substitution like this instead would be more future-proof, though.)

> but it needs to set just once, the 1st time the `config.ini` is copied/generated.

I think just change it in the "example" `.ini` at the source/`$md_build` folder in the `sources_` function, yeah? It still won't fix existing setups and only help for new installs. I'm not sure how else to work it, though. Otherwise we'd have to check if there's already a `cacheScreenshots` line *and* if that line is in an `[esgamelist]` section and I guess if that were easy, it'd already be in `iniGet`/`iniSet`, right?

Something like this, then.

cmitu:
> I think just change it in the "example" `.ini` at the source, in the `$md_build` folder in the `sources_` function, yeah? 

No quite, it should be during _configure_, when the `config.ini` is first copied/created.

> It still won't fix existing setups and only help for new installs.

The goal is to provide a better default configuration, we're not modifying existing setups.

s1eve-mcdichae1:
> No quite, it should be during _configure_, when the `config.ini` is first copied/created.

Sure, the difference being that only our "live" copy of the file has the edit, and the "example" version in the instal (and config) folders remain untouched. This way, then?

cmitu:
Yes, that's better. Thank you for the modifications!

PR #3705, opened 2023-04-17
Skyscraper - update source repo and scriptmodule

s1eve-mcdichae1:
- Update source repo to active fork: https://github.com/Gemba/skyscraper

- Scriptmodule enhancements and fixes (co-authored by @Gemba)

(Original first post follows):
> Upstream repository is currently not accepting pull requests, to focus on developing other projects. So I've made a couple of patches to update the source with one bug fix and one QOL improvement.
> 
> (BUG) Apple2 exts: normally, most roms are cached by hashing the file content. However, apple2 disk images are volatile and may change content. Patch `01_apple2_exts.patch` adds apple2 exts to list of exceptions, to be cached by file name instead.
> 
> (QOL) tg16 system: Many themes support an independent 'tg16' system folder. Patch `02_add_tg16_system.patch` adds tg16 alias for pcengine titles.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -10,13 +10,13 @@
 #
 
 rp_module_id="skyscraper"
-rp_module_desc="Scraper for EmulationStation by Lars Muldjord"
-rp_module_licence="GPL3 https://raw.githubusercontent.com/muldjord/skyscraper/master/LICENSE"
-rp_module_repo="git https://github.com/muldjord/skyscraper :_get_branch_skyscraper"
+rp_module_desc="Scraper for EmulationStation"
+rp_module_licence="GPL3 https://raw.githubusercontent.com/Gemba/skyscraper/master/LICENSE"
+rp_module_repo="git https://github.com/Gemba/skyscraper :_get_branch_skyscraper"
 rp_module_section="opt"
 
 function _get_branch_skyscraper() {
-    download https://api.github.com/repos/muldjord/skyscraper/releases/latest - | grep -m 1 tag_name | cut -d\" -f4
+    download https://api.github.com/repos/Gemba/skyscraper/releases/latest - | grep -m 1 tag_name | cut -d\" -f4
 }
 
 function depends_skyscraper() {
@@ -44,11 +44,15 @@ function install_skyscraper() {
         'artwork.xml.example2'
         'artwork.xml.example3'
         'artwork.xml.example4'
+        'platforms.json'
+        'screenscraper.json'
+        'mobygames.json'
         'tgdb_developers.json'
         'tgdb_publishers.json'
         'mameMap.csv'
         'aliasMap.csv'
-        'hints.txt'
+        'hints.xml'
+        'docs'
         'import'
         'resources'
         'cache/priorities.xml.example'
@@ -133,7 +137,7 @@ function _get_ver_skyscraper() {
 
 function _check_ver_skyscraper() {
     ver=$(_get_ver_skyscraper)
-    if compareVersions "$ver" lt "3.5" ]]; then
+    if compareVersions "$ver" lt "3.5"; then
         printMsgs "dialog" "The version of Skyscraper you currently have installed is incompatible with options used by this script. Please update Skyscraper to the latest version to continue."
         return 1
     fi
@@ -201,42 +205,21 @@ function _init_config_skyscraper() {
 
     # Make sure the `artwork.xml` and other conf file(s) are present, but don't overwrite them on upgrades
     local f_conf
-    for f_conf in artwork.xml aliasMap.csv; do
-        if [[ -f "$scraper_conf_dir/$f_conf" ]]; then
-            cp -f "$md_inst/$f_conf" "$scraper_conf_dir/$f_conf.default"
-        else
-            cp "$md_inst/$f_conf" "$scraper_conf_dir"
-        fi
+    for f_conf in artwork.xml aliasMap.csv platforms.json screenscraper.json; do
+        copyDefaultConfig "$md_inst/$f_conf" "$scraper_conf_dir/$f_conf"
     done
 
     # If we don't have a previous config.ini file, copy the example one
-    if [[ ! -f "$scraper_conf_dir/config.ini" ]]; then
-        cp "$md_inst/config.ini.example" "$scraper_conf_dir/config.ini"
-        sed -i 's/\[esgamelist\]/[esgamelist]\ncacheScreenshots="false"/' "$scraper_conf_dir/config.ini"
-    fi
+    [[ ! -f "$scraper_conf_dir/config.ini" ]] && cp "$md_inst/config.ini.example" "$scraper_conf_dir/config.ini"
 
-    # Try to find the rest of the necessary files from the qmake build file
-    # They should be listed in the `unix:examples.file` configuration line
-    if [[ $(grep unix:examples.files "$md_build/skyscraper.pro" 2>/dev/null | cut -d= -f2-) ]]; then
-        local files=$(grep unix:examples.files "$md_build/skyscraper.pro" | cut -d= -f2-)
-        local file
+    # Artwork example files
+    cp -f "$md_inst/artwork.xml.example"* "$scraper_conf_dir"
 
-        for file in $files; do
-            # Copy the files to the configuration folder. Skip config.ini, artwork.xml and aliasMap.csv
-            if [[ $file != "artwork.xml" && $file != "config.ini" && $file != "aliasMap.csv" ]]; then
-                cp -f "$md_build/$file" "$scraper_conf_dir"
-            fi
-        done
-    else
-        # Fallback to the known resource files list
-        cp -f "$md_inst/artwork.xml.example"* "$scraper_conf_dir"
-
-        # Copy resources and readme
-        local resource_file
-        for resource_file in README.md mameMap.csv tgdb_developers.json tgdb_publishers.json hints.txt; do
-            cp -f "$md_inst/$resource_file" "$scraper_conf_dir"
-        done
-    fi
+    # Copy remaining resources
+    local resource_file
+    for resource_file in mameMap.csv tgdb_developers.json tgdb_publishers.json hints.xml mobygames.json; do
+        cp -f "$md_inst/$resource_file" "$scraper_conf_dir"
+    done
 
     # Copy the rest of the folders
     cp -rf "$md_inst/resources" "$scraper_conf_dir"
@@ -249,7 +232,7 @@ function _init_config_skyscraper() {
     cp -rf "$md_inst/import" "$scraper_conf_dir"
 
     # Create the cache folder and add the sample 'priorities.xml' file to it
-    mkdir -p "$scraper_conf_dir/cache"
+    mkUserDir "$scraper_conf_dir/cache"
     cp -f "$md_inst/priorities.xml.example" "$scraper_conf_dir/cache"
 }
 
@@ -342,7 +325,7 @@ function _scrape_chosen_skyscraper() {
     # Confirm with the user that scraping can start
     dialog --clear --colors --yes-label "Proceed" --no-label "Abort" --yesno "This will start the gathering process, which can take a long time if you have a large game collection.\n\nYou can interrupt this process anytime by pressing \ZbCtrl+C\Zn.\nProceed ?" 12 70 2>&1 >/dev/tty
     [[ ! $? -eq 0 ]] && return 1
-    
+
     local choice
 
     for choice in "${choices[@]}"; do
@@ -371,7 +354,7 @@ function _generate_chosen_skyscraper() {
     fi
 
     local choices
-    local cmd=(dialog --backtitle "$__backtitle" --ok-label "Start" --cancel-label "Back" --checklist " Select platforms for gamelist(s) generation\n\n" 22 60 16) 
+    local cmd=(dialog --backtitle "$__backtitle" --ok-label "Start" --cancel-label "Back" --checklist " Select platforms for gamelist(s) generation\n\n" 22 60 16)
 
     choices=($("${cmd[@]}" "${options[@]}" 2>&1 >/dev/tty))
 
@@ -523,7 +506,7 @@ function gui_skyscraper() {
     while true; do
         [[ -z "$ver" ]] && ver="v(Git)"
 
-        local cmd=(dialog --backtitle "$__backtitle"  --colors --cancel-label "Exit" --help-button --no-collapse --cr-wrap --default-item "$default" --menu "   Skyscraper: game scraper by Lars Muldjord ($ver)\\n \\n" 22 60 12)
+        local cmd=(dialog --backtitle "$__backtitle"  --colors --cancel-label "Exit" --help-button --no-collapse --cr-wrap --default-item "$default" --menu "   Skyscraper: game scraper for EmulationStation ($ver)\\n \\n" 22 60 12)
 
         local options=(
             "-" "GATHER and cache resources"


cmitu:
> Upstream repository is currently not accepting pull requests, to focus on developing other projects. So I've made a couple of patches to update the source with one bug fix and one QOL improvement.

To be fair, @muldjord didn't accept PRs even when the project was (more) active. The Readme recommends using a more active fork (https://github.com/detain/skyscraper), are these issues/bugs fixed there ? We might consider switching to it instead of using the upstream project.

s1eve-mcdichae1:
> - just stuffing PRs into local patches is not the best solution.

I can see some logic in that. I'll check out the fork if that's something on the table.

s1eve-mcdichae1:
> The Readme recommends using a more active fork (https://github.com/detain/skyscraper), are these issues/bugs fixed there ?

The fork has a feature that lets users add custom systems with a config file, so that's fine.

> We might consider switching to it instead of using the upstream project - just stuffing PRs into local patches is not the best solution.

I've been using [this version](https://github.com/RetroPie/RetroPie-Setup/compare/master...s1eve-mcdichae1:RetroPie-Setup:skyscraper-fork) -- first as-is and then with the Apple2 fix applied manually before the build step -- with no issues. A [PR](https://github.com/detain/skyscraper/pull/18) for the Apple2 bug there has gone unanswered.

s1eve-mcdichae1:
Rebased using the forked version.

s1eve-mcdichae1:
The "active" fork was stalled out too, but now @Gemba has picked it up once more. Both of the issues these patches were meant to address, have been implemented upstream.

Updated PR to use the *new* new repo, with a few additional edits.

cmitu:
@s1eve-mcdichae1 thank you, I'll take a look and test this. However, just with a cursory look some changes are not correct - referencing `$md_build` during `configure` is a no-no since it will break binary installs (where there's no `$md_build`).

Gemba:
Maybe then (for the binary and source install)  it is better to _move_ the config files from `md_inst` instead of copying, as these files are not expected to be in the `md_inst` folder?

Also keep in mind that this scriptmodule here (in contrast to the one in my fork) does purge the cache when Skyscraper is removed. This can void a users scrape cache when reinstalling skyscraper (i.e. when the user runs `remove` first).

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -44,14 +45,19 @@ function install_skyscraper() {
...
         'mameMap.csv'
         'aliasMap.csv'
-        'hints.txt'
+        'hints.xml'
+        'docs'
         'import'
         'resources'
         'cache/priorities.xml.example'
+        'VERSION'
+        'skyscraper.pro'
```
Gemba, line 60: This file is not needed during runtime.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -199,44 +205,32 @@ function configure_skyscraper() {
...
-        for resource_file in README.md mameMap.csv tgdb_developers.json tgdb_publishers.json hints.txt; do
-            cp -f "$md_inst/$resource_file" "$scraper_conf_dir"
-        done
+    iniConfig "=" "" "$md_inst/skyscraper.pro"
+    iniGet "unix:examples.files"
+    local files="$ini_value"
+    if [[ -z "$files" ]]; then
+        files="README.md hints.xml artwork.xml artwork.xml.example1 artwork.xml.example2"
+        files+=" artwork.xml.example3 artwork.xml.example4 aliasMap.csv mameMap.csv"
+        files+=" docs/ARTWORK.md tgdb_developers.json tgdb_publishers.json platforms.json screenscraper.json"
```
Gemba, line 219: file mobygames.json is missing

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -44,14 +45,19 @@ function install_skyscraper() {
         'artwork.xml.example2'
         'artwork.xml.example3'
         'artwork.xml.example4'
+        'platforms.json'
+        'screenscraper.json'
         'tgdb_developers.json'
         'tgdb_publishers.json'
         'mameMap.csv'
         'aliasMap.csv'
```
Gemba, line 53: Please add mobygames.json

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: It's used during `configure` to get the list of files, which means it needs to exist in `$md_inst`.

Or is that the whole point of the fallback list: we read from the file in `$md_build` (if building from source) or, if that file doesn't exist (because binary install), use the fallback list instead?

s1eve-mcdichae1:
> Maybe then (...) it is better to _move_ the config files from `md_inst` instead of copying, 

Then you'd not be able to run `configure` again without `install`(`_bin`) also first.  (I know this isn't the "suggested usage" but right now, it's still possible.)

> Also keep in mind that this scriptmodule here (in contrast to the one in my fork) does purge the cache when Skyscraper is removed. This can void a users scrape cache when reinstalling skyscraper (i.e. when the user runs `remove` first).

Yeah I saw that too and have disabled for now by commenting (not removing yet) the function. I too think it's a bit aggressive to remove the cache but I suppose most regular users aren't goofing around with it as much as we are, so for the "average Joe" maybe it is better to purge automatically, idk. I wonder if the devs have any strong opinion one way or the other?

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: Done, should that be added to the [unix:examples.files](https://github.com/Gemba/skyscraper/blob/7c4677172198d39293e8d74827fd22eb9a0f724c/skyscraper.pro#L14) also?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -25,6 +25,7 @@ function depends_skyscraper() {
 
 function sources_skyscraper() {
     gitPullOrClone
+    echo VERSION=\"$(_get_branch_skyscraper)\" > VERSION
```
cmitu, line 28: what's the point of the `VERSION` file ? Is it really needed ?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -44,14 +45,19 @@ function install_skyscraper() {
...
         'tgdb_publishers.json'
         'mameMap.csv'
         'aliasMap.csv'
-        'hints.txt'
+        'hints.xml'
+        'docs'
         'import'
         'resources'
         'cache/priorities.xml.example'
+        'VERSION'
```
cmitu, line 60: is the `VERSION` really needed ?

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -145,10 +151,10 @@ function _list_systems_skyscraper() {
     find -L "$romdir/" -mindepth 1 -maxdepth 1 -type d -not -empty | sort -u
 }
 
-function remove_skyscraper() {
-    # On removal of the package, purge the cache
-    _purge_skyscraper
-}
+# function remove_skyscraper() {
```
cmitu, line 154: please leave the purge on remove operation. If we decide to remove it, then we'll remove the entire function and call, we don't need to comment it.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -199,44 +205,33 @@ function configure_skyscraper() {
...
-        local resource_file
-        for resource_file in README.md mameMap.csv tgdb_developers.json tgdb_publishers.json hints.txt; do
-            cp -f "$md_inst/$resource_file" "$scraper_conf_dir"
-        done
+    # They should be listed in the `unix:examples.files` configuration line
+    iniConfig "=" "" "$md_build/skyscraper.pro"
+    iniGet "unix:examples.files"
+    local files="$ini_value"
+    if [[ -z "$files" ]]; then
+        files="README.md hints.xml artwork.xml artwork.xml.example1 artwork.xml.example2"
```
cmitu, line 217: documentation files don't need to be copied to the configuration folder. Remove `README.md` and `docs/ARTWORK.md`.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -199,44 +205,33 @@ function configure_skyscraper() {
...
+        files+=" artwork.xml.example3 artwork.xml.example4 aliasMap.csv mameMap.csv"
+        files+=" docs/ARTWORK.md tgdb_developers.json tgdb_publishers.json platforms.json"
+        files+=" screenscraper.json mobygames.json"
     fi
+    local file
+    for file in $files; do
+        case "$file" in
+            config.ini)
+                ;;
+            artwork.xml|aliasMap.csv|platforms.json|screenscraper.json|mobygames.json)
```
cmitu, line 227: IMHO only `artwork.xml` and `platforms.json` should be copied-if-not-found here, since they're the only ones meant for user edits.

cmitu (commented):
I've added a few comments for the modifications added.

As a meta-comment for @Gemba, we can drop the release based installation if you'd like, so you don't have explicitely mark releases in the Github repo. Internally, the version can be bumped when new features/bugs are added.

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: > since they're the only ones meant for user edits.

`screenscraper.json` does takes user edits, which are needed to scrape that module for the custom systems (e.g. **tg16**) that are added to `platforms.json`.

I don't use [aliasMap.csv](https://github.com/Gemba/skyscraper/blob/dev_reloaded/aliasMap.csv) (it [don't work with screenscraper](https://retropie.org.uk/forum/topic/31281/skyscraper-determining-search-name-for-game-entry) and that's my primary source), but it does seem meant for user edits also.

I only wasn't sure about `mobygames.json` but looking inside, it doesn't really seem like there's anything for us to do in that one so I'll remove that one here.

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: (They'll need to be removed from **unix:examples.files** too if the binary/source installs are to remain in sync.)

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: Not really. I guess that's pretty well covered by `retropie.pkg` innit.

scriptmodules/supplementary/skyscraper.sh
Gemba: > what's the point of the `VERSION` file ? Is it really needed ?

It is part of the repo anyway. I guess this line is a leftover, can't remember why I put it there.

scriptmodules/supplementary/skyscraper.sh
Gemba: Adding them to the line of `config.ini)` should be fine.

Gemba:
The mobygames situation:
The` mobygames.json` is 1:1 yarked from their API. An user should add an alias to `platforms.json` containing the `platform_name` of `mobygames.json` whenever in need for an extra system. In contrast `screenscraper.json` has the legacy of being manually edited and the names are not 1:1 as on the screenscraper.fr site, but more geared towards the platform folders of RetroPie. This also means the `screenscraper.json` file is not API complete/does neither reflect all systems screenscraper provides nor the system names screenscraper.fr uses.

Thus, the files have -albeit looking similar- different use cases. Initially I thought to align the `screenscraper.json` with the screenscraper.fr provided systems (making a 1:1 "API" copy like in `mobygames.json`), but then I noticed to have many aliases to set in `platforms.json`. In the end I left `screenscraper.json` as is. 

When adding a new system for screenscraper, edit `screenscraper.json` and add the screenscraper.fr system id and name. If the name does not reflect an existing platform folder, set an alias in `platforms.json`.
When adding a new system for mobygames. Lookup in `mobygames.json` for the `platform_name` and add an alias to `platforms.json`.

TL;DR: `screenscraper.json` -> meant for user edits, `mobygames.json` -> readonly reference for user.

scriptmodules/supplementary/skyscraper.sh
cmitu: They should be added at all - the README.md can be added in the `install_` files and `ARTWORK.md` is already part of the `docs`, which is also copied to the install folder.

Gemba:
One more thing: As the `copyDefaultConfig` does not overwrite existing files, it should however do it once for `platforms.json` when updating from 3.7.7 to 3.8.x. The existing file should be moved, to not void any user changes. With the `platforms_sorting.py` the user's `platforms.json` (from 3.7.7) can be reformatted. The diffs between old and new can then be transferred to the new `platforms.json` by the user manually.

From that point on any incoming update changes to `platforms.json` will go to `platforms.json.rp-dist`.

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: > They should be added at all

We are talking about "adding" them to [this line](https://github.com/RetroPie/RetroPie-Setup/blob/b05af36837499a345312a5226cd3e9441a6709b6/scriptmodules/supplementary/skyscraper.sh#L222) in the `case` statement so that they will *not* be copied to the config dir, even though present in the **unix:examples.files** value.

joolswills:
It would be useful to get some of the commits here squashed / reduced before this is accepted. There's a lot of changes so multiple commits is good, but would be useful to have the fixes for earlier commits in the PR combined.

s1eve-mcdichae1:
> It would be useful to get some of the commits here squashed / reduced before this is accepted. 

Yeah I will clean it up before it's merged but I don't think we're quite there yet.

---

`mobygames.json` is not currently copied from `md_inst` to `scraper_conf_dir` during install-from-source because it is not named in the [unix:examples.files](https://github.com/Gemba/skyscraper/blob/7c4677172198d39293e8d74827fd22eb9a0f724c/skyscraper.pro#L14) in the upstream repo and so it doesn't get included in the `files` to be copied ca. [L220](https://github.com/RetroPie/RetroPie-Setup/blob/b05af36837499a345312a5226cd3e9441a6709b6/scriptmodules/supplementary/skyscraper.sh#L220). Not sure if:

1: @Gemba should add it to that line to include it that way, or

2: we should just copy it over manually in the module, or

3: we should abandon reading from this value altogether and just use the "fallback" `files` list in all cases?

---

> One more thing: As the `copyDefaultConfig` does not overwrite existing files, it should however do it once for `platforms.json` when updating from 3.7.7 to 3.8.x.

I'm not sure how to address this. I can `_get_ver_skyscraper` to see what version we're installing, but how to check what ver was replaced?

Maybe `touch` a file in configdir after doing the overwrite, and only do the overwrite if it's not already there?

Or, in most other cases like this (that is, cases where `copyDefaultConfig` is used), doesn't it fall to the user to decide whether to update the file or not when it's different, whether that's because of their own edits or because the default config has changed "underneath"?

In other words, what if we just, don't overwrite it ever? Let them copy the `.rp-dist` if they care to, or let them continue using the old one...what bad happens? I wasn't even aware it had changed and so I've been still using the old version, which seems to be working fine? Is it just an issue of formatting and the additional systems? This file don't even exist at muldjord's version so this even affect who, users that manually switched to detain's fork and then come back to the official RP script?

That's just my take but I'm a self-confessed I-D-TEN-TEE. What does anyone else think?

scriptmodules/supplementary/skyscraper.sh
cmitu: Sorry, I meant to **not** add them to be copied.

cmitu:
> Thus, the files have -albeit looking similar- different use cases. Initially I thought to align the `screenscraper.json` with the screenscraper.fr provided systems (making a 1:1 "API" copy like in `mobygames.json`), but then I noticed to have many aliases to set in `platforms.json`. In the end I left `screenscraper.json` as is.

IMHO `screenscraper.json` should be similar to the other scraper entries. There's an [API call](https://screenscraper.fr/webapi2.php?alpha=0&numpage=0#systemesListe) that returns the list of ScreenScraper supported systems (and other info regarding the system) which can be used to generate an exhaustive list.

cmitu:
> I can `_get_ver_skyscraper` to see what version we're installing, but how to check what ver was replaced?

You can't, at least not 100%. For binary installs, the scriptmodule installation folder is wiped and the contents replaced with the new version's package. For source, it's possible to run `Skyscraper -version` and pick the version, but you can't reliably determine this in all cases.

joolswills:
> Yeah I will clean it up before it's merged but I don't think we're quite there yet.

Sounds good - and sorry if I pre-empted anything. Also thanks for the contribution and on-going changes :)

Gemba:
> `mobygames.json` is not currently copied from `md_inst` to `scraper_conf_dir` during install-from-source because it is not named in the [unix:examples.files](https://github.com/Gemba/skyscraper/blob/7c4677172198d39293e8d74827fd22eb9a0f724c/skyscraper.pro#L14) in the upstream repo and so it doesn't get included in the `files` to be copied ca. [L220](https://github.com/RetroPie/RetroPie-Setup/blob/b05af36837499a345312a5226cd3e9441a6709b6/scriptmodules/supplementary/skyscraper.sh#L220). Not sure if:
> 
> 1: @Gemba should add it to that line to include it that way, or

I have added it to the mandatory files unix:examples.files. Thanks for the hint.
As long as the `skyscraper.pro` comes with the binary install we could ditch the fallback which is a branch we will never hit with Retropie.

Gemba:
> I'm not sure how to address this. I can `_get_ver_skyscraper` to see what version we're installing, but how to check what ver was replaced?
> 
> Maybe `touch` a file in configdir after doing the overwrite, and only do the overwrite if it's not already there?

If an existing `platforms.json` has far less lines than the 3.8.x `platforms.json` it should be replaced, but keep a copy of the existing one. Or Compare/diff the first few lines to check if it is a 3.7.7 style or 3.8.x already.

Gemba:
> Or, in most other cases like this (that is, cases where `copyDefaultConfig` is used), doesn't it fall to the user to decide whether to update the file or not when it's different, whether that's because of their own edits or because the default config has changed "underneath"?

It might end up in users flooding the forum, with "3.8.x supposed to support platform xy or extension z but it does not"-questions. Even if the migration steps (i.e. review `platforms.json` files manually) would be highlighted in the documentation there will be guys who will not check the docs.

Gemba:
> > Thus, the files have -albeit looking similar- different use cases. Initially I thought to align the `screenscraper.json` with the screenscraper.fr provided systems (making a 1:1 "API" copy like in `mobygames.json`), but then I noticed to have many aliases to set in `platforms.json`. In the end I left `screenscraper.json` as is.
> 
> IMHO `screenscraper.json` should be similar to the other scraper entries. There's an [API call](https://screenscraper.fr/webapi2.php?alpha=0&numpage=0#systemesListe) that returns the list of ScreenScraper supported systems (and other info regarding the system) which can be used to generate an exhaustive list.

I like the idea, as I tried similar before. Let me reconsider this. However, it should not slow down Skyscraper scraping process. 

After that `screenscraper.json` should be only be changed when screenscraper supports more systems, i.e. this file shouldn't be altered by the user.

s1eve-mcdichae1:
> > I'm not sure how to address this. I can `_get_ver_skyscraper` to see what version we're installing, but how to check what ver was replaced?

> If an existing `platforms.json` has far less lines than the 3.8.x `platforms.json` it should be replaced, but keep a copy of the existing one. Or Compare/diff the first few lines to check if it is a 3.7.7 style or 3.8.x already.

I'm not familiar with `json` but is it possible to do something like:

    { "version" : "2.0" }

...at the top? Then the script could `grep` for that and decide to replace or not, instead of just guessing (even though it may be a very good guess) based on lines count or formatting style.

cmitu:
> Then the script could `grep` for that and decide to replace or not, instead of just guessing (even though it may be a very accurate guess most of the time) based on lines count or formatting style.

If the format is not compatible, why not rename the file (i.e. `platforms_v2.json`) ? 

This way it can be copied with `copyConfigFile` to ensure that a correct file is available after upgrade. During the usage a warning can be shown if there's a `platforms.json` in the config folder (i.e. _... found `platforms.json`, but will not be used, please migrate any local configs to `platforms_v2.json`..._).

s1eve-mcdichae1:
(Maybe something like this?)

s1eve-mcdichae1:
> I'm not familiar with `json` but is it possible to do something like:
> 
> ```
> { "version" : "2.0" }
> ```

My initial tests with this aren't yielding any success. I can `grep` and `cut` to find the version number, but if I actually stick this line in the JSON, it won't scrape any platforms. Don't know if I just have the syntax wrong or if this is fundamentally not possible.

s1eve-mcdichae1:
This seems to work, if I nest it inside the outer `{}` curly braces and use a trailing comma:

```
{
    "version":"3.8.0.2309",
    "platforms": [
        {
            "name": "3do",
            "scrapers": [
(...)
```

...and then the scriptmodule can grep for that version and only replace the file if it's ~~less than~~ *not greater than or equal* ("doesn't exist" surely fits the latter, less sure how the former is interpreted here.)

Re: the actual version number. It doesn't need to track the branch version, necessarily. It could either remain pegged to this version, or it could just be "2.0" as in *the second iteration of this particular file*. In any case, it wouldn't need to change until/unless another overwrite becomes necessary.

What do you think?

cmitu:
> What do you think?

It's not worth the additional complexity.

Note that this scriptmodule uses @muldjord's version, which isn't using `platforms.json` - am I right ? So any updates/new installs of `skyscraper` in this case would be ok, since the file will be copied as new anyway.

s1eve-mcdichae1:
@cmitu i think yes, as I mentioned earlier this only affects people who've switched to the other fork and then came back to this scriptmodule.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -199,44 +203,32 @@ function configure_skyscraper() {
...
-            cp -f "$md_inst/$f_conf" "$scraper_conf_dir/$f_conf.default"
-        else
-            cp "$md_inst/$f_conf" "$scraper_conf_dir"
-        fi
-    done
-
     # If we don't have a previous config.ini file, copy the example one
-    if [[ ! -f "$scraper_conf_dir/config.ini" ]]; then
-        cp "$md_inst/config.ini.example" "$scraper_conf_dir/config.ini"
-        sed -i 's/\[esgamelist\]/[esgamelist]\ncacheScreenshots="false"/' "$scraper_conf_dir/config.ini"
```
cmitu, line 215: This was lost somehow and I think should reverted. This prevents the caching of artwork when importing images from EmulationStation.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -199,44 +203,32 @@ function configure_skyscraper() {
...
+    local files="$ini_value"
+    if [[ -z "$files" ]]; then
+        files="hints.xml artwork.xml artwork.xml.example1 artwork.xml.example2 artwork.xml.example3"
+        files+=" artwork.xml.example4 aliasMap.csv mameMap.csv tgdb_developers.json tgdb_publishers.json"
+        files+=" platforms.json screenscraper.json mobygames.json"
     fi
+    local file
+    for file in $files; do
+        case "$file" in
+            config.ini|README.md|docs/ARTWORK.md)
```
cmitu, line 222: Please remove the `.md` document files, they shouldn't be copied into the user's confi folder.

cmitu (commented):
Ok, this is closer to the real thing.

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: This is now implemented upstream in the [config.ini.example file](https://github.com/Gemba/skyscraper/blob/e8969834a6f00fbd06b64112b7363e0606ad4c29/config.ini.example#L169) that is copied from the repo, so I think we no longer need to insert this ourselves: https://github.com/Gemba/skyscraper/blob/e8969834a6f00fbd06b64112b7363e0606ad4c29/config.ini.example#L169

scriptmodules/supplementary/skyscraper.sh
s1eve-mcdichae1: Naming those files at this point I think is necessary in order to ensure that they are *not* copied to the config folder, even if they might be named in `$files` (as during a source install, when `$files` [is read from](https://github.com/RetroPie/RetroPie-Setup/blob/0ab5aa17835bd14a9f82e4499f878cfa515ab5e5/scriptmodules/supplementary/skyscraper.sh#L211-L213) the [qmake build file skyscraper.pro](https://github.com/Gemba/skyscraper/blob/e8969834a6f00fbd06b64112b7363e0606ad4c29/skyscraper.pro#L14) and not from the [fallback list that's written in the scriptmodule](https://github.com/RetroPie/RetroPie-Setup/blob/0ab5aa17835bd14a9f82e4499f878cfa515ab5e5/scriptmodules/supplementary/skyscraper.sh#L214-L217).) Or else @Gemba can remove them from **unix:examples.files** so they're not included in `$files` but if I remove the filenames from this line in the scriptmodule now, while they're still named on that line in the build file, then they *will be copied* to the config folder at [lines 227-228](https://github.com/RetroPie/RetroPie-Setup/blob/0ab5aa17835bd14a9f82e4499f878cfa515ab5e5/scriptmodules/supplementary/skyscraper.sh#L227-L228). Naming them here at [L222-L223](https://github.com/RetroPie/RetroPie-Setup/blob/0ab5aa17835bd14a9f82e4499f878cfa515ab5e5/scriptmodules/supplementary/skyscraper.sh#L222-L223) says "do nothing" with them and ensures that they are *not* copied, instead.

scriptmodules/supplementary/skyscraper.sh
cmitu: It's ok then.

scriptmodules/supplementary/skyscraper.sh
cmitu: Then I think I'd prefer to explicitely list the files based on their copy mode (overwrite or not) and get rid of the `unix:example.files` altogether. Either way is fine.

s1eve-mcdichae1: > explicitely list the files based on their copy mode (overwrite or not) and get rid of the `unix:example.files` altogether.

The downside here being, that if new resource files are ever required (such as the new JSON files, this time), then a source install won't work until the scriptmodule is updated to include them. (Nor the binary, supposing it is just built from source and then archived.)

For the record, the current RP scriptmodule featuring muldjord already copies both of these files in a source install, and explicitly includes README.md in the binary also.

cmitu: Yes, that was certainly the original intention, but in practice the file list doesn't change that often and the required files need to be added to `install_` anyway to accomodate binary installs.
I think it's better to be explicit, makes it easier to understand what files are neeed and their purpose. Thanks for the changes.

cmitu:
Re-tested again, looks good to merge.
Thank you @s1eve-mcdichae1 and @Gemba !

s1eve-mcdichae1:
(I just amended the commit message a little -- the comment that had the typo fixed, isn't there anymore so I don't mention it here.)

joolswills:
Thank you for your work on this.

PR #3706, opened 2023-04-17
mupen64plus: update runtime hotkey configuration

cmitu:
The Mupen64plus start-up script adds 3 'hotkey' combinations for the connected joysticks by looking at the RetroArch auto-config profile(s) for the joysticks.

To handle cases when the SDL joystick name (used by Mupen64Plus) is different than the 'udev' name and the auto-config profile is not found just by looking at the filename, an additional search for the profile .cfg files was added - using the 'vendor' and 'product' IDs of the joystick. This takes advantage of the changes in e392c7ac for the input configuration, which save 'vendor'/'product' IDs to the RetroArch's auto-config profile.

I re-organized a bit the hotkey bindings loop so it's indexed by the number of joysticks detected. This avoids searching for the auto-config profile 3 times (# of hotkey bindings), which can is more expensive when using `grep` to look up the `vendor`/`product` IDs. Since 'bind' is a reserved built-in in Bash, I renamed it to `hotkeys_bind`, similar to the other arrays used for hotkeys set-up.

### DIFF
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -99,11 +99,13 @@ function remap() {
     # read retroarch auto config file and use config
     # for mupen64plus.cfg
     local file
-    local bind
+    local hotkeys_bind=( "" "" "" )
     local hotkeys_rp=( "input_exit_emulator" "input_load_state" "input_save_state" )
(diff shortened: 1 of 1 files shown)


cmitu:
I've set it to draft since it may have merge issues after #3696 is merged, but the implementation (pending review changes) should be ok.

cmitu:
I've removed the draft status, was waiting for https://github.com/RetroPie/RetroPie-Setup/pull/3696 to be merged just to check for conflicts, but that may take more time.

PR #3709, opened 2023-04-23
emulationstation: use OpenGL2 for RPI+Mesa platform

cmitu:
The last runtime detection changes for the build instructions enable the GL21 renderer only on non-RPI platforms. The V3D Mesa driver supports OpenGL 2.1, so enable the GL21 renderer when it's used on a Raspberry Pi.

Without this change, the build will use the GL1.4 renderer.


joolswills:
Thanks.

PR #3718, opened 2023-05-09
lr-x1: switch to the libretro repository

cmitu:
Changed the source repository to the libretro fork of the core, since it gets a bit of maintenance than the original repo (`r-type/xmil-libretro`) and I noticed a few more core options have been added over the time. The original core repository hasn't been updated since 2017.

Added the licence from the libretro fork.


joolswills:
Thanks.

PR #3722, opened 2023-06-03
Fix all commands in the ES onfinish input action being overwritten during configuration

obrie:
This fixes the `emulationstation` scriptmodule to only rewrite the relevant `onfinish` commands in `es_input.cfg` in case the user has added their own.

## Context

In my own system, I have several custom scripts that I execute when a new controller has completed configuration in EmulationStation.  To do this, I end up with an `es_input.cfg` file that looks like this:

```xml
<?xml version="1.0"?>
<inputList>
  <inputAction type="onfinish">
    <command>/opt/retropie/supplementary/emulationstation/scripts/inputconfiguration.sh</command>
    <command>/path/to/script1.sh</command>
    <command>/path/to/script2.sh</command>
  </inputAction>
</inputList>
```

As far as I can tell, this is the only way to integrate additional `onfinish` hooks that are unrelated to retropie's `inputconfiguration.sh`.  The relevant code handling this in EmulationStation is here: https://github.com/RetroPie/EmulationStation/blob/08d74d334523ad10edc4a5b72225aa5da47db6db/es-core/src/InputManager.cpp#L424-L446

When a user updates EmulationStation via Retropie-Setup, it'll maintain the existing `es_input.cfg` configuration for the most part.  However, the piece that handles keeping the `onfinish` command in sync with expectations ends up overwriting any other commands you might have defined for that action.  This is handled here: https://github.com/RetroPie/RetroPie-Setup/blob/0bcf01d46926de78609980e8661da5c971189ab8/scriptmodules/supplementary/emulationstation.sh#L221-L231

When the `configure` action runs for `emulationstation`, it'll re-run `init_input` and then overwrite every command in the `onfinish` input action -- even ones it didn't add.

## How to reproduce

To reproduce, you can run the following:

```bash
cat > $HOME/.emulationstation/es_input.cfg <<EOF
<?xml version="1.0"?>
<inputList>
  <inputAction type="onfinish">
    <command>/opt/retropie/supplementary/emulationstation/scripts/inputconfiguration.sh</command>
    <command>/path/to/script1.sh</command>
    <command>/path/to/script2.sh</command>
  </inputAction>
</inputList>
EOF

sudo $HOME/RetroPie-Setup/retropie_packages.sh emulationstation configure
```

This will overwrite `es_input.cfg` and you'll end up with the following:

```xml
<?xml version="1.0"?>
<inputList>
  <inputAction type="onfinish">
    <command>/opt/retropie/supplementary/emulationstation/scripts/inputconfiguration.sh</command>
    <command>/opt/retropie/supplementary/emulationstation/scripts/inputconfiguration.sh</command>
    <command>/opt/retropie/supplementary/emulationstation/scripts/inputconfiguration.sh</command>
  </inputAction>
</inputList>
```

## Changes

To address this issue, it seemed like there were a few options:

1. Change the `xmlstarlet` command to just overwrite the first command (whatever that command may be)
2. Change the `xmlstarlet` command to overwrite commands that match the expected script name (or at least part of it)

I opted for solution (2) as I felt it would be more reliable and unlikely to break going forward.

After this change, re-running the test above results in the following file:

```xml
<?xml version="1.0"?>
<inputList>
  <inputAction type="onfinish">
    <command>/opt/retropie/supplementary/emulationstation/scripts/inputconfiguration.sh</command>
    <command>/path/to/script1.sh</command>
    <command>/path/to/script2.sh</command>
  </inputAction>
</inputList>
```

The one side effect of this change is that it doesn't address the scenario where the `inputAction` node exists, but there's no command for `inputconfiguration.sh`.  For example, suppose you have the following file:

```xml
<?xml version="1.0"?>
<inputList>
  <inputAction type="onfinish">
    <command>/path/to/script1.sh</command>
  </inputAction>
</inputList>
```

Running `configure` on the `emulationstation` scriptmodule won't add a new command.  However, the existing code has similar behavior anyway if `inputAction` is empty:

```xml
<?xml version="1.0"?>
<inputList>
  <inputAction type="onfinish">
  </inputAction>
</inputList>
```

If you were to run `configure` with the above `es_input.cfg` using what's in master, you end up with the same file -- no new command is added.

Rather than try to address this, I felt that this side effect was acceptable given the above.

## Testing

To test, I just re-ran the above reproduction case and confirmed that the output was what I would have expected.


joolswills:
Thank you for the changes and appreciate your detailed analysis and testing.

@cmitu looks good to me and I will test, but appreciate your feedback.

cmitu:
I'd rather use the approach 1. - replace the 1st `command` node - rather than match the command value. The _XPath_ selector is easier to read and since this is an internal RetroPie thing, the 1st value is expected to be replaced with own script.

What's the use case for the additional scripts ? Using events to run them would be better, since it's a documented extension that's meant for user custom scripts.

obrie:
Thanks for the feedback and good question @cmitu :+1:  I should have included that context in the PR description.

For my use case, I have multiple services running in the background on my system that listen to the controls that have been configured for RetroArch and respond according to certain button presses.  One example is a service for showing game manuals on a screen based on a combination of joystick buttons being pressed.  Design-wise, it's similar to how `joy2key.py` is implemented: https://github.com/RetroPie/RetroPie-Setup/blob/0bcf01d46926de78609980e8661da5c971189ab8/scriptmodules/admin/joy2key/joy2key.py#L62-L131

To keep these services updated with the latest configured controls, I added a few `onfinish` hooks.  These hooks will tell my services to take a fresh look at what's configured in RetroArch and reconfigure based on the latest controller configurations.  There are 2 primary reasons for the integration here:

* I don't want my services to be monitoring the filesystem and updating based on changes to files in the retoarch joypad directory.  I prefer to respond to user behavior, rather than filesystem behavior.
* The current `controls-changed` EmulationStation event executes before the RetroArch configurations have been updated: https://github.com/RetroPie/EmulationStation/blob/08d74d334523ad10edc4a5b72225aa5da47db6db/es-core/src/InputManager.cpp#L401-L405

Hope that adds a little more color to the scenario.  You're right there are other implementations we could consider to address this scenario:

* Move the `controls-changed` event after `doOnFinish()`
* Add a new `controls-onfinish` event (or a better name :laughing:) that runs after `doOnFinish()`
* Move `inputconfiguration.sh` to be integrated as a `controls-changed` script (I'd have to ensure that my scripts run *after* `inputconfiguration.sh`)

Some of those options obviously expand the scope here quite a bit.  Thoughts?

obrie:
Regarding the existing PR, I'm on board with changing it so just the first command is modified (target than doing a substring search) if we're okay with just making that change for now 👍

cmitu:
@obrie thanks for the detailed explanation, it's an interesting use case.

For the current PR I think we should change the updating of the `es-input.cfg` so that the 1st command is replaced. @joolswills is it ok with you ?

If you want, you can also send a PR to switch the the ES event to run after the inputscripts are run, looking at your use case I think it's better to run post-configuration, so that the actions from `inputconfiguration.sh` have completed. For now I'm not sure the `config-change` is run with enough parameters to replace the current approach, but we may modify it if we decide to replace the current approach of running the input configuration script with an event based script.

joolswills:
@cmitu sounds ok yeah. Thanks.

obrie:
Branch has been updated with the option (1) implementation.

I'll take a look at updating EmulationStation as well.  I agree that would be the ideal long-term solution for my integrations as well :+1:

cmitu:
@obrie thank you for the changes. Tested and looks good to merge 👍.

joolswills:
Thank you.

PR #3723, opened 2023-06-10
Retroarch now requires libmbedtls, not installed by default on all pl…

hissingshark:
…atforms

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -16,7 +16,7 @@ rp_module_repo="git https://github.com/retropie/RetroArch.git retropie-v1.15.0"
 rp_module_section="core"
 
 function depends_retroarch() {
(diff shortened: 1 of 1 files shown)


cmitu:
On RPI OS/Raspbian, `retroarch` builds without that package.
I don't think the library added is strictly required, what's the failure message during build/configure ?

hissingshark:
> RetroArch already has an mbedtls source copy under deps.

Ah, I think may explain it.  Thank you @cmitu.

It was reported as what seemed like a pretty straight forward runtime error:
`/opt/retropie/emulators/retroarch/bin/retroarch: error while loading shared libraries: libmbedtls.so.12: cannot open shared object file: No such file or directory`

It built without error.  However, the build system for the distributed binary had the  `libmbedtls` package installed for `hyperion.ng` purposes.  I suspect that was detected, so  it didn't get built into the `retroarch` binary and became an unresolved runtime dependency for anyone who didn't have it installed.

Building from the latest commits I can' t reproduce the issue at all, so perhaps it now ignores locally installed versions at build time regardless.

Apologies for the noise.

PR #1103, opened 2015-11-27
Add reicast input script

gizmo98:

PR #3724, opened 2023-06-14
amiberry: update to 5.6.1

### LINKED ISSUE #1077: Enabling option to configure Retroarch controller

### LINKED ISSUE #1080: [lr-fba-next] new release - v0.2.97.37
Hi!! seems that good fellas from FBA just released a new version with bugfixes, new drivers, etc. (read here http://www.fbalpha.com/view/227/). And @barbudreadmon was kind enough to update his libretro port (https://github.com/libretro/libretro-fba). 

I'm building it right now and will try to update this thread to let you know if it's worth it! 

Thanks :)

### LINKED ISSUE #1081: [RPi2 Retropie 3.2.1] Error installing lr-mame2010: "mame2010_libretro.so not found"
Hello, hope you can help me.
I'm trying to install experimental lr-mame2010 package trough RetroPie setup, but I can't. It says it couldn't find mame2010_libretro.so in tmp/build/lr_mame2010.
What am I doing wrong?

Thank you.

### LINKED ISSUE #1088: lr-mednafen-pce is requesting my github login and password when compiling from source
Log started at: Tue Nov 17 16:27:27 EST 2015
/home/pi/RetroPie-Setup/tmp/build/lr-mednafen-pce /home/pi/RetroPie-Setup

Getting sources for 'lr-mednafen-pce' : PCEngine emu - Mednafen PCE core port for libretro

git clone --depth 1 "https://github.com/petrockblog/mednafen-pce-libretro.git" "/home/pi/RetroPie-Setup/tmp/build/lr-mednafen-pce"
Cloning into '/home/pi/RetroPie-Setup/tmp/build/lr-mednafen-pce'...
Username for 'https://github.com':

### LINKED ISSUE #1100: Add reicast input script
Reicast is not really usable at the moment. There is no in game setup menu and joypad config looks a little bit strange. I created a input script, so joypad setup shouldn't be complicated anymore:
https://github.com/gizmo98/RetroPie-Setup/blob/8c5f3bbf7cb09f91b85c0de41d115540dd42ae17/scriptmodules/supplementary/emulationstation/configscripts/reicast.sh

Reicast seems to have no hat support. Just map any left axis device to analog stick if possible. Quit emulator with "select".

cmitu:
Among others, fixes some input issues, one of them reported in the forums, and a Vero4k black screen during emulation. Full changelog at https://github.com/BlitterStudio/amiberry/releases/tag/v5.6.1

Bugfixes:

- Hardfile properties were always set to default (fixes #1077)
- cycle exact was incorrect enabled sometimes (fixes #1075)
- Fix foreign disk image support when drive is standard Amiga 3.5" DD
- hsync events should be before misc events
- fixed uaegfx board parameters
- uaegfx board name is UAE, not uaegfx.card
- Fix helptext formatting in CPU panel (fixes #1088)
- use double-quotes to denote inches in GUI
- when using --autoload with a filename in the current path, Amiberry would crash (fixes #1091)
- fix build errors with GCC 13.1 (fixes #1100)
- savestate cmd line option would not use filename (fixes #1081)
- second controller buttons would not be registered in retroarch (fixes #1092)
- Fullscreen toggle would no longer work since v5.3 (fixes #1103)
- reverted copying of prefs on hard reset (fixes #1068)

Improvements:

- VGA mode resolution autoswitch update
- VGA autoswitch update to support new OS 3.2 monitors
- Upgraded FloppyDriveBridge to v1.4
- improve message when Main ROM is not found (fixes #1095)
- improve Floppy GUI panel (fixes #1094)
- use BGR888 for 32-bit modes (fixes #1080)

### DIFF
--- a/scriptmodules/emulators/amiberry.sh
+++ b/scriptmodules/emulators/amiberry.sh
@@ -13,7 +13,7 @@ rp_module_id="amiberry"
 rp_module_desc="Amiga emulator with JIT support (forked from uae4arm)"
 rp_module_help="ROM Extension: .adf .chd .ipf .lha .zip\n\nCopy your Amiga games to $romdir/amiga\n\nCopy the required BIOS files\nkick13.rom\nkick20.rom\nkick31.rom\nto $biosdir/amiga"
 rp_module_licence="GPL3 https://raw.githubusercontent.com/BlitterStudio/amiberry/master/LICENSE"
-rp_module_repo="git https://github.com/BlitterStudio/amiberry v5.6.0"
+rp_module_repo="git https://github.com/BlitterStudio/amiberry v5.6.1"
 rp_module_section="opt"
 rp_module_flags="!all arm rpi3 rpi4"
 


joolswills:
Thank you.

PR #3726, opened 2023-06-26
Create lr-melods.sh

Unknown-programer:
This script should add the libretro core for Melon DS. The emulator no longer requires bios to run.

### DIFF
--- a/scriptmodules/libretrocores/lr-melods.sh
+++ b/scriptmodules/libretrocores/lr-melods.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


cmitu:
Not sure if we'd want the libretro core of MelonDS - which is terribly out-of-date compared to the upstream MelonDS emulator -, but did you test the script ?

Unknown-programer:
Unfortunately, I don't have a raspberry pi avaliable to test this right now, but I will in the near future. However I am very confident that this script will work.The only potential issue would be that the libretro.so file isn't named correctly, but that can easily be fixed. I know for a fact that this will build because melonds does support arm64. I agree this is not the most up to date version, but it does offer accurate DS emulation, and some DSI emulation. Therefore I believe it's worth adding.

cmitu:
Ok, don't submit un-tested scriptmodules. The core name being wrong is the most obvious error, but it's not the only one. Also, the fact that it builds doesn't mean it also runs adequately or at all.

PR #3733, opened 2023-07-23
configedit - exclude 'thumbnails' directory in manual edit

s1eve-mcdichae1:
Can contain many thousands of '.png' images, if cheevos have been enabled.


joolswills:
Thank you.

PR #3734, opened 2023-07-24
golang: upgrade to a newer version

cmitu:
Fixes compilation for 'google.golang.org/protobuf', which needs 'math/bits', available in go 1.9 and later.


joolswills:
Thank you.

PR #3739, opened 2023-07-30
Image creation changes for bullseye

joolswills:
This PR includes some reworking of the image module for compatibility with bullseye and easier maintenance. Changes include (More information in the individual commits):

* add distribution parameter to install_rp
* only set gpu_mem on buster for rpi1-3
* use fat32 for boot partition and remove unneeded boot flag
* use the mke2fs config from the chroot for ext4 filesystem
* set password and disable userconfig in chroot
* disable 64bit kernel
* build additional rpi3 images for bullseye 
* rework logic to use ini configuration for distributions
* use "dmsetup remove" to remove device mapper files
* use bootfs for boot partition name


joolswills:
Thanks.

>     1. The `boot` partition is labeled `bootfs` since [RPi-Distro/pi-gen@5d2c6f3](https://github.com/RPi-Distro/pi-gen/commit/5d2c6f31cefc7710e3bbc44012b9ffb843294e34)  for Bullseye images.

I can change this to the most recent naming - it doesn't actually matter though as it's referred to by partition id in fstab etc.

>     2. The `arm_64bit` option is necessary only for the Pi4 systems, so maybe can be moved earlier in the `pi4` test.

As it doesn't do any harm for other models, I decided it's easiest just to include it for all.

IoanZzZ:
I was testing the latest Buster weekly image for RPi4/Pi400 and in `config.txt` file, it have`gpu_mem` settings. I report it just in case that this is a issue. Thanks

```
[pi4]
# Enable DRM VC4 V3D driver on top of the dispmanx display stack
dtoverlay=vc4-fkms-v3d
max_framebuffers=2

[all]
#dtoverlay=vc4-fkms-v3d
gpu_mem_256=128
gpu_mem_512=256
gpu_mem_1024=256
overscan_scale=1
arm_64bit=0

```

PR #3740, opened 2023-08-06
Create lr-citra2018

Unknown-programer:
This is an older version of citra that according to libretro themselves is "A Nintendo 3DS emulator targeting libretro exclusively. Based on an older snapshotted version of Citra. Lower compilation and build requirements. "

### DIFF
--- a/scriptmodules/libretrocores/lr-citra2018
+++ b/scriptmodules/libretrocores/lr-citra2018
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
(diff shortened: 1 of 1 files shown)


mrcmunir:
I don't think we needed it the performance is worse than latest build  and less compatible emulation in everything also no supported JIT arm64 added in external dynarmic outdated commit.
The only difference is to run in GPUS with Only OpenGL 3.3 more than a decade ago on PC.

cmitu:
Closing, no feedback and I don't see why we'd add this instead of the current Citra version (standalone or libretro core).

PR #3741, opened 2023-08-08
make git <branch> parameter also optional in rp_hasNewerModule() 

Gemba:
The function `rp_hasNewerModule() ` has the git <branch> parameter mandatory and [bails out](https://github.com/Exarkuniv/RetroPie-Extra/pull/161#discussion_r1286379507) when the <branch> is missing. In contrast `gitPullOrClone()` assumes "master" when no branch parameter is present.

This PR fixes this optional/mandatory mismatch between the functions and allows <branch> to be optional in the `rp_module_repo` scriptmodule header for GIT repos.

### DIFF
--- a/scriptmodules/packages.sh
+++ b/scriptmodules/packages.sh
@@ -598,6 +598,9 @@ function rp_hasNewerModule() {
                     ;;
                 git|svn)
                     local repo_branch="$(rp_resolveRepoParam "${__mod_info[$id/repo_branch]}")"
+                    if [[ "$repo_type" == "git" ]] && [[ -z "$repo_branch" ]] ; then
+                        repo_branch="master"
+                    fi
                     local repo_commit="$(rp_resolveRepoParam "${__mod_info[$id/repo_commit]}")"
                     # if we are locked to a single commit, then we compare against the current module commit only
                     if [[ -n "$repo_commit" && "$repo_commit" != "HEAD" ]]; then


joolswills:
gitPullOrClone is an old function and this was included a long time ago so is kept for backwards compatibility (from when it was called directly with parameters and before we added rp_module_repo)

It shouldn't be optional in the rp_module_repo config.  Also "master" is not a good fallback anymore. this also breaks the check_repos functionality in tools.

joolswills:
I may consider making gitPullOrClone throw an error if the branch is empty but if there are some improperly coded modules out there that miss the branch out that will stop them working (even though they are not technically correct anyway)

s1eve-mcdichae1:
As a fallback, if the field is not present in `rp_module_repo`, instead of "bailing out" and disabling install options as it does, perhaps it can do the "update may be available (unable to check for this package)" and still allow an update to be attempted, like I see in some other cases?

joolswills:
It would be easier just to fix up the incorrect module surely ? I'm not really keen on adding workarounds for when a module is missing required meta data.

joolswills:
If anything I'd like to steer towards being stricter and ensuring the modules have the info or throwing a warning/error. Or disabling them.

s1eve-mcdichae1:
> If anything I'd like to steer towards being stricter and ensuring the modules have the info **or throwing a warning/error**. Or disabling them.

I think it would be as simple as: 19faa60

joolswills:
Sorry, but I don't want to implement it like that.

s1eve-mcdichae1:
NP, just a passing suggestion.

Gemba:
> Thank you, but gitPullOrClone is an old function and this was included a long time ago so is kept for backwards compatibility (from when it was called directly with parameters and before we added rp_module_repo)
> 
> It shouldn't be optional in the rp_module_repo config. Also "master" is not a good fallback anymore. This also breaks the check_repos functionality in tools.

Thanks for the heads up. If it is mandatory on both functions it is fine for me, with the legacy 'failsafe' as implemented now in the gitPullOrClone. Thus for gitPullOrClone the parameter SHOULD be provided, whereas for the rp_hasNewerModule it MUST be provided.

PR #3742, opened 2023-08-12
audiosettings: ensure ALSA configuration folder exists

cmitu:
On RaspiOS (10/11), the ALSA configurations folder doesn't exist, so make sure we're creating it before saving/moving the ALSA configuration file we generate with the scriptmodule.


joolswills:
Thanks. I modified the commit to remove the trailing slashes. Trailing slash isn't required when using mkdir and it matches calls elsewhere.

PR #3744, opened 2023-08-29
helpers: skip APT recommended packages by default

cmitu:
Instruct `apt` to install without recommended packages. `getDepends` already does it, but `aptInstall` does not.
Modified `aptInstall` to run `apt` with `--no-install-recommands` and removed the parameter from any other `aptInstall` calls.

### DIFF
--- a/scriptmodules/supplementary/raspbiantools.sh
+++ b/scriptmodules/supplementary/raspbiantools.sh
@@ -23,7 +23,7 @@ function apt_upgrade_raspbiantools() {
 }
 
 function lxde_raspbiantools() {
-    aptInstall --no-install-recommends xorg lxde
+    aptInstall xorg lxde
     aptInstall raspberrypi-ui-mods rpi-chromium-mods gvfs
     # On `buster`, disable PulseAudio since it messes up the audio settings
     # remove the lxpanel plugin for PulseAudio volume, to prevent a crash due to missing PulseAudio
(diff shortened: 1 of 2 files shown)


scriptmodules/supplementary/raspbiantools.sh
```diff
@@ -23,7 +23,7 @@ function apt_upgrade_raspbiantools() {
 }
 
 function lxde_raspbiantools() {
-    aptInstall --no-install-recommends xorg lxde
+    aptInstall xorg lxde
     aptInstall raspberrypi-ui-mods rpi-chromium-mods gvfs
```
joolswills, line 27: The change may well affect the selection of packages installed here.

joolswills (commented):
aptInstall is designed to just call apt-get install without prompting the user. This change will affect other uses of it, and I don't really see the benefit of this as it now makes it impossible to use the call if you wanted to call it with the default behaviour.

joolswills:
eg. aptInstall openttd on Ubuntu will previously install `openttd openttd-opensfx`

with this change openttd-opensfx will be excluded, and that's not desired.

joolswills:
It could be argued that we should specify all the packages in every module, for the sake of package dependency removal - but that's another issue imho, and would require a lot of work. Also this may well affect 3rd party modules. So I'd prefer to keep this core function as it is, sorry.

cmitu:
> It could be argued that we should specify all the packages in every module, for the sake of package dependency removal - but that's another issue imho, and would require a lot of work

Yes, that was the idea. `getDepends` was already removing the _recommended_ packages, only `aptInstall` would have been affected.

> Also this may well affect 3rd party modules. So I'd prefer to keep this core function as it is, sorry.

Sure, no problem.

PR #3745, opened 2023-08-31
mame: force 'char' to be signed on ARM platforms

cmitu:
Forcing 'char' to be signed fixes crashing in some drivers, needed on ARM.

The default behavior for 'char' is arch/ISA specific, see: https://stackoverflow.com/questions/46463064

### DIFF
--- a/scriptmodules/emulators/mame.sh
+++ b/scriptmodules/emulators/mame.sh
@@ -44,8 +44,9 @@ function build_mame() {
         rpSwap on 4096
     fi
 
-    # Compile MAME
     local params=(NOWERROR=1 ARCHOPTS=-U_FORTIFY_SOURCE PYTHON_EXECUTABLE=python3)
+    # when building on ARM enable 'fsigned-char' for compiled code, fixes crashes in a few drivers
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks, but this doesn't look correct sorry. The `__cxxflags` variable is a global used to set up the default `CXXFLAGS` and shouldn't be changed here and I don't think it will have any effect.

cmitu:
You may be right. Let me check and fix this, before the 0.258 binaries are built.

cmitu:
Should be better now. `OVERRIDE_CXX` work in addition to both `OVERRIDE` and `CXXFLAGS` so it's safe to use.

joolswills:
Thanks. Looks good.

I would add that the signed char issue used to affect older mame before systems that defaulted to unsigned char were well known, and I thought this was resolved many years ago for mame.

If mame and some of the drivers still assume this I think it's a valid thing to mention upstream.

It's possible that additions have been made and upstream is not aware. I will take a look.

For reference we also forced this for older mame (eg. Advancemame), and also some older emulators such as lr-prosystem.

PR #3746, opened 2023-09-05
scraper: change build method to fix FTBS

cmitu:
Changed the build method by using Golang's module support and creating a module from the upstream sources. Fixes the current build errors due to a transitive depedency which requires upgrading Go to 1.17+. 

This simplifies applying the patch and rolling back (downgrading) the TGDB module dependency. As a bonus, we can use the normal way of retrieving the sources and generating the scriptmodule `.pkg` info file.


joolswills:
This failed first time because my version of go was older on the buster install I was just testing on. Looks good, but maybe it's worth us forcing the golang module upgrade if the version is too old. 
```

go: unknown subcommand "mod"
Run 'go help' for usage.
```

cmitu:
Yes, the `golang` upgrade fails due to the version comparison. I've modified it to use `compareVersions`, it should work now.

joolswills:
Thank you.

PR #3749, opened 2023-09-09
aarch64: add PS1 and SNES emulators to default installation

cmitu:
Included `lr-snes9x` and `lr-pcsx-rearmed` in the main section when `aarch64` is detected, otherwise they won't be installed during a basic installation.


joolswills:
Thanks.

PR #3752, opened 2023-09-21
dolphin: update dependencies for latest version

cmitu:
Since May 2023, Qt6's SVG component is required, make sure we install the corresponding dependency.


joolswills:
Thank you.

PR #3759, opened 2023-10-04
audiosettings: add helper service to set ALSA defaults on RPI

cmitu:
Added a helper method that installs a `systemd` service meant to automatically configure ALSA with the HDMI 0 port when using the ARM `vc4hdmi` RPI driver. The service will run once then disable itself.

Since the default ALSA configuration on RaspiOS goes through the analog output (Heaphones) when the `vc4-kms-v3d` overlay is present, if we want to configure the HDMI output by default there needs to be an additional configuration.  The additional configuration needs also to be applied when only one audio output is present, since the ALSA config is a bit more involved.

There are a couple of reasons for which is a service:
 - the `vc4hdmi` device index can be different depending on the Pi model. Pi0(w)/Pi400 has no analog output by default and CM models may not have it also)
 - the card name can also be different (Pi4 with 2 devices vs Pi3/Pi2/Pi1 with 1 device)

Note that the service will configure the 1st HDMI port/device, on Pi models with multiple HDMI ports (Pi4/Pi 400/Pi5/CM4) the user will still need to use the Audio settings dialog to change it if they want to. There may be a way to detect which HDMI port is active and has audio, but that's not implemented.


joolswills:
Thanks for this. Where is this triggered from currently ?

cmitu:
@joolswills for now it's not hooked anywhere, my initial thought is to add it to the image generation, like [xpad](https://github.com/RetroPie/RetroPie-Setup/blob/a43e29c1d2e4ca038dbe1bc0be283df351b04623/scriptmodules/admin/image.sh#L153), based on the platform and image (i.e. only add it to RPI and _bullseye_ and later).

joolswills:
Thanks. I wonder if there is a way to do this without the duplication of alsa config. Also a function to remove the service would be needed, but I'm going to have a think about options.

cmitu:
> I wonder if there is a way to do this without the duplication of alsa config. 

Initially the implementation called `audiosettings`'s `_asoundrc_save_audiosettings` method from ther service, with the appropriate parameters, to generate the config. But I thought it would be faster to not load RetroPie's scripts, in order to finish up quickly before the autologin process is started and EmulationStation powers up.

> Also a function to remove the service would be needed, but I'm going to have a think about options.

I though about this also, was thinking to add a parameter to the new function call (`alsa_defaults_service_audiosettings`) to either install or remove the service. Right now it only does the installation. Not sure when to invoke the removal though, maybe at the start `gui_audiosettings`, when the user is configuring the audio themselves.

PR #3760, opened 2023-10-13
wifi: add NetworkManager support and various tweaks

cmitu:
Added support for NetworkManager and predictable network interface device names. NetworkManager is default on RaspiOS 12 and also on Armbian. Armbian doesn't disable the predictable ifnames, so the changes here will help adding support for configuring WiFi on it.

Changes included:

 - interface name is not hardcoded to 'wlan0', but taken from `/sys/net/devices`. Most of the functions had to be modified and an extra parameter for ifname was added.

 - config/remove wifi functions have been added for each management version (wpasupplicant/nm)

 - tweaked the interface in a few places (colors, title, phrasing)

 - the NetworkManager configuration methods will create a connection named 'RetroPie-WiFi'. This helps identifying the connection during removal, since I didn't find a way to add some info/metadata to a connection in order to mark the one we're configuring with our script.

 - the country code is now read with `raspi-config` instead of querying the `wpa_supplicant.conf`, since it might not exist when NetworkManager is active.

Only tested WPA and Open networks, since I lack any WEP capable AP/device.

### DIFF
--- a/scriptmodules/supplementary/wifi.sh
+++ b/scriptmodules/supplementary/wifi.sh
@@ -14,51 +14,88 @@ rp_module_desc="Configure WiFi"
 rp_module_section="config"
 rp_module_flags="!x11"
 
+function _get_interface_wifi() {
+    local iface
+    # look for the first wireless interface present
+    for iface in /sys/class/net/*; do
+        if [[ -d "$iface/wireless" ]]; then
+            echo "$(basename $iface)"
+            return 0
+        fi
+    done
+    return 1
+}
+
+function _get_mgmt_tool_wifi() {
+    # get the WiFi connection manager
+    if systemctl -q is-active NetworkManager.service; then
+        echo "nm"
+    else
+        echo "wpasupplicant"
+    fi
+}
 function _set_interface_wifi() {
-    local state="$1"
+    local iface="$1"
+    local state="$2"
 
     if [[ "$state" == "up" ]]; then
-        if ! ifup wlan0; then
-            ip link set wlan0 up
+        if ! ifup $iface; then
+            ip link set $iface up
         fi
     elif [[ "$state" == "down" ]]; then
-        if ! ifdown wlan0; then
-            ip link set wlan0 down
+        if ! ifdown $iface; then
+            ip link set $iface down
         fi
     fi
 }
 
-function remove_wifi() {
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/wifi.sh
```diff
@@ -14,51 +14,88 @@ rp_module_desc="Configure WiFi"
 rp_module_section="config"
 rp_module_flags="!x11"
 
+function _get_interface_wifi() {
+    local iface
+    # look for the firs wireless interface present
```
joolswills, line 19: Minor typo - should be _first_

scriptmodules/supplementary/wifi.sh
```diff
@@ -108,15 +145,44 @@ function connect_wifi() {
...
+            nmcli connection modify "$con" wifi-sec.psk "$key"
+            ;;
+        wep)
+            nmcli connection modify "$con" wifi-sec.key-mgmt none
+            nmcli connection modify "$con" wifi-sec.wep-key-flags 0
+            nmcli connection modify "$con" wifi-sec.wep-key-type 2
+            nmcli connection modify "$con" wifi-sec.wep-key0 "$key"
+            ;;
+        open)
+            ;;
```
joolswills, line 175: minor but could probably shorten this by putting `nmcli connection modify "$con"` in a variable

joolswills (commented):
Thank you for this. Not tested yet but looks good. Made a couple of minor notes (the duplication one isn't a big deal).

scriptmodules/supplementary/wifi.sh
cmitu: thanks, fixed.

scriptmodules/supplementary/wifi.sh
cmitu: That's a remnant from testing various properties with `nmcli modify` until I found that the `key-flags` are required in order for NM to get the credentials from its config file instead of asking the user.

I've merged the `modify` lines now so it's just called once - it still shows each connection parameter so it's easy to modify/add. Let me know if it looks better.

joolswills: Looks good thanks.

joolswills:
Thank you!

PR #3763, opened 2023-10-14
retoarch - disable cheevos hardcore mode by default

s1eve-mcdichae1:
Since RetroArch 1.10, cheevos hardcore mode is enabled by default. Should this be "corrected" in RP's own defaults?

As I understand, `_set_config_option_retroarch` only sets the option if it's not explicitly set already. So users who use cheevos, but haven't explicitly set hardcore mode, will see it suddenly disabled -- just as it suddenly *enabled* last year when the default was changed.

https://retropie.org.uk/forum/post/273128

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -281,6 +281,9 @@ function configure_retroarch() {
     # enable video shaders
     iniSet "video_shader_enable" "true"
 
+    # disable cheevos hardcore mode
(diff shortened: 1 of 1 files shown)


cmitu:
> Since RetroArch 1.10, cheevos hardcore mode is enabled by default. Should this be "corrected" in RP's own defaults?

We don't have any defaults for cheevos. This was an upstream decision, I don't think we need to 'correct' it. More context [here](https://github.com/libretro/RetroArch/pull/13465).

PR #3764, opened 2023-10-15
systems: add support for Armbian

cmitu:
Added support for detecting if running on a Armbian built system. Armbian is built on Debian/Ubuntu, so we're not changing the OS release/name.

* Added support for Armbian's Linux kernel package naming, in order to be able to install the kernel headers.
* Removed the ZRAM config from our own menu, Since Armbian has its own configuration tool (`armbian-config`) that handles it. ZRAM is configured and enabled by default on Armbian images.

NOTE: I'll probably add some changes later on to add `armbian-config` to the menu, replacing `raspi-config`. I just need to create a proper/usable icon for it.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -261,6 +261,10 @@ function _mapPackage() {
                         isPlatform "rpi5" && pkg="linux-headers-rpi-2712"
                     fi
                 fi
+            elif isPlatform "armbian"; then
+                local branch="$(grep -oP "BRANCH=\K.*"      /etc/armbian-release)"
+                local family="$(grep -oP "LINUXFAMILY=\K.*" /etc/armbian-release)"
+                pkg="linux-headers-${branch}-${family}"
             elif [[ -z "$__os_ubuntu_ver" ]]; then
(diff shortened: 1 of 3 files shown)


cmitu:
PR rebased after the changes in #3765.

scriptmodules/helpers.sh
```diff
@@ -261,6 +261,10 @@ function _mapPackage() {
                         isPlatform "rpi5" && pkg="linux-headers-rpi-2712"
                     fi
                 fi
+            elif isPlatform "armbian"; then
+                local branch="$(sed -n '/^BRANCH/ s/.*=//p' /etc/armbian-release)"
+                local family="$(sed -n '/^LINUXFAMILY/ s/.*=//p' /etc/armbian-release)"
```
joolswills, line 266: Not much simpler but I think I would have gone for `grep -oP "LINUXFAMILY=\K.*"`

joolswills: (We use this elsewhere a fair amount)

joolswills: :-)

joolswills:
Thank you.

PR #3767, opened 2023-10-16
for v3.9.0 release of Skyscraper

Gemba:
Changes for v3.9.0 in the scriptmodule
- new script files to be installed
- new config files (partially replacements)
- move files from install to config folder instead of copying (`/opt/retropie/supplementary/skyscraper` was crowded with files, which Skyscraper even does not read from there)
- some spaces and quotes corrected

Read about the Skyscraper 3.9.0 changes [here](https://github.com/Gemba/skyscraper/blob/dev_reloaded/README_Skyscraper_reloaded.md#tba-version-390).
Once this scriptmodule is merged I will tag the 3.9.0 version of Skyscraper.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -34,29 +34,45 @@ function build_skyscraper() {
 }
 
 function install_skyscraper() {
+    local config_files=($(_config_files_skyscraper))
+
     md_ret_files=(
-        'Skyscraper'
+        'docs'
         'LICENSE'
         'README.md'
-        'config.ini.example'
-        'artwork.xml'
+        'Skyscraper'
+        'supplementary/scraperdata/check_screenscraper_json_to_idmap.py'
+        'supplementary/scraperdata/convert_platforms_json.py'
+        'supplementary/scraperdata/peas_and_idmap_verify.py'
+    )
+    md_ret_files+=("${config_files[@]}")
+}
+
+
+function _config_files_skyscraper() {
+    local config_files=(
+        'aliasMap.csv'
         'artwork.xml.example1'
         'artwork.xml.example2'
         'artwork.xml.example3'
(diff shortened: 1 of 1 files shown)


joolswills:
>move files from install to config folder instead of copying (/opt/retropie/supplementary/skyscraper was crowded with files, which Skyscraper even does not read from there)

I would copy them maybe to a subfolder in md_inst to tidy things up instead perhaps. Although the files are not used there, it needs to be so you can run "configure" again without throwing errors and RetroPie is designed so that all the files needed are kept in md_inst even if they are to be copied elsewhere.

Gemba:
Suggestion in the pushed update. 

I am not happy with having the `config_files` twice to maintain, can it be `_skyscraper_config_files` or outside the functions or a separate function that provides the array?

Gemba:
If there are issues which impede this PR pls let me know.

m33ts4k0z:
@Gemba Please check [here](https://retropie.org.uk/forum/topic/34753/trouble-with-the-new-skyscraper-package/6). Nothing too important, with the workaround mentioned there it works. I am on Ubuntu Mantic 23.10.

m33ts4k0z:
> If there are issues which impede this PR pls let me know.

Hi I didn't know how to contact you otherwise but it seems that something is wrong with the results on 3.9.1. Running this on 3.7.7 returns results as normal but 3.9.1 returns results not found:

`Skyscraper -p switch -s screenscraper --query "romnom=Super%20Mario%20Odyssey%20(World)%20(En,Ja,Fr,De,Es,It,Nl,Zh,Ru)%20(Rev%203).xci" /home/retropie/RetroPie/roms/switch/super-mario-odyssey.xci`

It doesnt matter if you use another rom name with a cleaner name, the results are always empty on 3.9.1 while it works fine on 3.7.7.

joolswills:
Thank you.

PR #3768, opened 2023-10-17
wifi: update search locations for 'wifikeyfile'

cmitu:
Since RaspiOS Bookworm, the `bootfs` (1st) partition is mounted under `/boot/fimware` instead of `/boot`, so look for the WiFi configuration file under that location also.

Minor:
 - modified 'printMsgs' to allow ANSI color escape codes for messages printed


joolswills:
Thanks. I think the modification for printMsgs should be in a separate commit as it's a function used all over the place, but I actually don't think it's needed to add colours here for a single dialog. Please we leave that for now ?

cmitu:
> Please we leave that for now ?

No problem. Modified the commits to leave just the changes for _bookworm_.

joolswills:
Thank you.

PR #3769, opened 2023-10-17
audiosettings: add PipeWire support

cmitu:
Since PipeWire is the default sound server in RaspiOS 12 _bookworm_, make sure we account for its presence and we're able to choose a default card ('output sink') or disable/enable it.

At this time, PipeWire is using emulating a PulseAudio session manager through `pipewire-pulse`, while also running the native `wireplumber` session manager. Most of the clients still think they speak to PulseAudio using the PulseAudio client API, while the actual backend is PipeWire.

Changed:
 - use `pactl` instead of `pacmd`, since the latter is not working through PipeWire's PulseAudio emulation
 - use `pactl` to set the PipeWire defaul output, instead of using the native `pw-cli`
 - added a toggle functions for PipeWire and PulseAudio, chosen based on the detected sound server

NOTE: although techically possible, I haven't added an option to enable PulseAudio if PipeWire is also present, but I don't think it's needed.


joolswills:
Thank you.

PR #3776, opened 2023-10-20
mupen64plus: buildfixes

cmitu:
This includes 2 commits:

 - one to disable the VK API added for `mupen64-plus-core`, since it's built by default. We don't need that since we don't ship (for now) any video plugin or front-end that support Vulkan.
 - the other removes some bitrot from `mupen64plus-video-gles2rice`, allowing it to compile on `armhf` with a newer g++ (like the one from Raspi OS 12 _bookworm_).


cmitu:
This is ready to be merged.

joolswills:
Thank you.

PR #3777, opened 2023-10-21
raspbiantools: bookworm updates

cmitu:
Some changes prompted by the new RaspiOS Bookworm features:

  - added Firefox as a browser, since it's supported now
  - mimic the RaspiOS defaults and on Pi4/5 platforms add an additional entry for starting the Wayland based desktop, using the `wayfire` compositor.


joolswills:
Thank you.

PR #3778, opened 2023-10-21
samba: add a Web Service Discovery service

cmitu:
[From: https://github.com/christgau/wsdd]
With Windows 10 version 1511, support for SMBv1 and thus NetBIOS device discovery was disabled by default. This causes hosts running Samba not to be listed in the Explorer's "Network (Neighborhood)" views. While there is no connectivity problem and Samba will still run fine, users might want to have their Samba hosts to be listed by Windows automatically.

This should fix the discoverability issues with recent Windows versions, especially since the SMBv1 client is not available anymore in Win11 (as some ill-advised instructions are trying to fix this problem).


joolswills:
Thank you.

PR #3779, opened 2023-10-21
Retroarch Config:- Turn on SRM check and autosave feature

retropieuser:
When playing a game on RetroPie if there is a sudden disconnection that isn't clean (power cut, pulling out the power cable etc), the .srm save file isn't updated correctly. The "autosave_interval" feature checks if the .srm has been updated and if it has been updated it overwrites the original file. In the case of the user, this would be expected behaviour (if you saved your game the save file would write then, not when you exit Retroarch).

The feature itself isn't a significant strain on resources, playing N64, PS1 cores on a pi4 I didn't notice any difference in frame rates or performance.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -278,6 +278,9 @@ function configure_retroarch() {
     # disable 'press twice to quit'
     iniSet "quit_press_twice" "false"
 
+    # enable srm check saving at intervals rather than upon exit
+    iniSet "autosave_interval" = "10"
+
     # enable video shaders
     iniSet "video_shader_enable" "true"
 


cmitu:
I don't think enabling this for all configurations is ok. 

You said

>  _The "autosave_interval" feature checks if the .srm has been updated and if it has been updated it overwrites the original file_

but the RetroArch comment for the option says [here](https://github.com/libretro/RetroArch/blob/d46ede42dc5faf55d149c96ac194fbb6ec2e0ad0/config.def.h#L1228C6-L1228C6):

> _Saves non-volatile SRAM at a regular interval .. Flush to file every 10 seconds on modern platforms by default_

which means the SRAM is written every 10 seconds by RetroArch regardless whether has changed or not. I actually tested this with the latest RetroArch version and it's not doing any checking to see if has changed - the `.srm` file is always written anew.
 Note that RetroArch sets this to 10 sec already for Non-SBC (PC systems/iOS/Android) and disables it for systems with limited I/O.

We don't want to cause un-necessary writes (and at such a small time interval) for _all_ users,  when the only benefit would be just for some error cases when RetroArch crashes during gameplay. As such, enabling this is not a good idea when a lot (if not most) of our Pi/SBC users will probably use a microSD card (slow and limited write cycles).

retropieuser:
Yeah I had a ponder of this over the weekend. I forgot I'm working from a NAS for saves and ROMs, so I'm not enduring the constant read and writing of an SD card. I can't envision anyway check that could be made to see if the ROMs folder is a mounted directory. If that were the case this would be a safe option.

Bit of a shame but equally I guess with the pi5 having a shutdown power button, I guess that would run a safe shutdown code and save Retroarch correctly anyway. That would probably solve a lot of the issues of the plug being pulled

retropieuser:
As above comment, unworkable for SD card formats

PR #3780, opened 2023-10-22
lr-gpsp: fix installation due to missing file

cmitu:
Removed obsolete installed file, since it was removed upstream.


joolswills:
Thanks.

PR #3781, opened 2023-10-24
bookworm: fix Mame2015/Mame2016 not building. 

cmitu:
Looks like some deprecation in python3.11 breaks building both cores due to some codegen including python scripts.
Upstream has a PR proposed in https://github.com/libretro/mame2015-libretro/pull/98, but not applied, so create one locally for each scriptmodule.


joolswills:
Thanks.

PR #3785, opened 2023-10-29
systems: add Vulkan flags for Pi4/Pi5

cmitu:
Added Vulkan as supported renderer for Pi4/Pi5 when running on RaspiOS 12 Bookworm. Technically the support was added in Bullseye for Pi4, but it was only available under X11.

Notes:
 - most emulators include the Vulkan headers, so there's no need to add `libvulkan-dev` as a dependency.
 - not all emulators will run with Vulkan from KMS/DRM - `ppsspp`/`dolphin` can use it under x11/wayland, while `retroarch` runs with the`vulkan` video driver on KMS/DRM.
 - switching RetroArch to use the `vulkan` renderer needs to include the installation/configuration of SLANG shaders, which is not handled at the moment.


TheDiscordian:
Do you experience a performance bump using Vulkan? I'm super interested in this and curious about what's holding this PR back.

PR #3787, opened 2023-11-01
crosscomp - add support for building bookwork arm cross compiler

joolswills:
Split out header include fix from bullseye.diff into own patch file as it's also needed on gcc 12.x.

Apply patch for GCC versions greater than 10.


cmitu:
I didn't try it, but the `limits.h` may be fixed with a configure option - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99476.

joolswills:
Thanks - that change was introduced for gcc 13.2 but can look at using it in the future.

PR #3789, opened 2023-11-04
lr-flycast-dev: add upstream libretro core

cmitu:
Added the libretro core from upstream flyinghead/flycast, it supercedes the current libretro/flycast repository used for the current `lr-flycast` core. The new core doesn't work correctly with the legacy VideoCore GLES drivers [1], so it will have to co-exist with the current `lr-flycast` until we drop support for older RaspiOS versions still using the VideoCore drivers.

Due to an internal compiler error, the enw core doesn't build on Buster with the included `gcc8`, but works on `gcc9` and later, which means it can be installed on Debian 11/Ubuntu 20.04 and later.

Tested on RaspiOS _bookworm_ with `vulkan`, `glcore` and `gl` RetroArch video drivers, on a Pi4.

[1] https://github.com/flyinghead/flycast/issues/559

### DIFF
--- a/scriptmodules/libretrocores/lr-flycast-dev.sh
+++ b/scriptmodules/libretrocores/lr-flycast-dev.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
Thank you.

PR #3791, opened 2023-11-06
ppsspp: revert to v1.13.2 for RPI and older Mesa

cmitu:
The 'ubershaders' introduced in v1.14 are not handled efficiently by the Mesa v3d driver, causing a performance regression [1]. Since the slowdown is not present in the Mesa version from Bullseye/Bookworm (which caused in part the version upgrade), install v1.13.2 just on Buster/RPI.

[1] https://github.com/hrydgard/ppsspp/issues/18388


joolswills:
Thanks.

MertDalbudak:
Thank you!

PR #3792, opened 2023-11-07
audiosettings: fix Pi5 detection

cmitu:
Fixed Pi5 detection by checking for `vc4hdmi` cards when determining if audio is enabled. Since the Pi5 doesn't have the analog audio output (Headphones), the detection thinks there's no audio enabled (since `snd_bcm2835` module is not loaded), but it's wrong.

Also fixed a couple of things:
 - the `pactl` listing parsing would enter an infinite loop if no audio was present
 - when checking for Pulse/Pipewire use the `.service` instead of the `.socket`, since the socket can be disabled, but the service is always active.

### DIFF
--- a/scriptmodules/supplementary/audiosettings.sh
+++ b/scriptmodules/supplementary/audiosettings.sh
@@ -22,7 +22,7 @@ function depends_audiosettings() {
 
 function gui_audiosettings() {
     # Check if the internal audio is enabled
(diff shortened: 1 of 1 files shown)


cmitu:
Draft for now, since I don't have a Pi5, though I tested it with `dtparam=audio=off` on the Pi4.

cmitu:
Removed draft status since I got a confirmation it works on a Pi5.

joolswills:
Thanks.

PR #3793, opened 2023-11-09
skyscraper.sh function _get_ver_skyscraper - perform version check as $user, not as root

s1eve-mcdichae1:
Fixes gather and generate actions, which were broken with the new package version.

https://retropie.org.uk/forum/topic/34771/skyscraper-usage-script-doesn-t-work-after-package-update


cmitu:
@s1eve-mcdichae1 works fine, thank you.

PR #3799, opened 2023-11-11
Add wiringPi as RetroPie-Setup Modul

Liontek1985:
Add wiringPi as RetroPie-Setup Modul


cmitu:
How is this module useful for RetroPie ? It's a library that's useful for application development. Furthermore:
 - the library is deprecated since 2019 and the main repo is now archived, without the interim maintainer that took over when the main developer stopped the project back in 2019. It doesn't work with the latest _Bookworm_ Raspi OS.
 - Raspbian (_Buster_) already has a `wiringpi` package, there's no need for a dedicated scriptmodule.

I don't think we need this.

PR #3803, opened 2023-11-13
Added JFDuke3D, Duke Nukem 3D source port for smaller Pis

protocultor:
Homepage: https://jonof.id.au/jfduke3d/

This old source port is the original base of EDuke32, [according to its own wiki](https://wiki.eduke32.com/wiki/JFDuke3D), and provides an OpenGL ES 2 renderer which runs extremely fast on a Raspberry Pi 3 B+; it shouldn't run any slower in older models. On the mentioned Pi 3B+, it can be played on 1280x720 with no visible slowdown, and unlike EDuke32, this resolution (and any resolution) is presented in actual fullscreen.
In greater hardware, it supports GL 2 and Polymost.
Uses SDL2 Game Controller for input (among others), so any gamepad can be supported as long as a _gamecontrollerdb.txt_ file is included in `$romdir/ports/duke3d`.
Downside is: it seems that it doesn't support addons. It does support NAM at least (couldn't test it, I don't have it, or the Duke addons).
Seriously, just give it a go, on any platform.

**_Semi-unrelated note_**: EDuke32 installer doesn't detect `DUKE.GRP` file if it's all uppercase, so it installs the shareware version of Duke, no matter if you have the full version on `$romdir/ports/duke3d`. Having both the full and shareware version of the game at the same time (with filenames using different letter cases) may work with EDuke32, but will crash JFDuke3D. Make sure you have only one installed.
This _jfduke3d_ script properly detects the existence of `DUKE.GRP`, no matter if upper or lowercase. Maybe this section should be used in the _eduke32_ script.

### DIFF
--- a/scriptmodules/ports/jfduke3d.sh
+++ b/scriptmodules/ports/jfduke3d.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="jfduke3d"
+rp_module_desc="lightweight Duke3D source port by JonoF"
+rp_module_licence="GPL2 https://raw.githubusercontent.com/jonof/jfduke3d/master/GPL.TXT"
+rp_module_repo="git https://github.com/jonof/jfduke3d.git master 41cd46bc"
+rp_module_section="exp"
+rp_module_flags=""
+
+function depends_jfduke3d() {
+    local depends=(
+        libsdl2-dev libvorbis-dev libfluidsynth-dev
+    )
+
+    isPlatform "x86" && depends+=(nasm)
+    isPlatform "gl" || isPlatform "mesa" && depends+=(libgl1-mesa-dev libglu1-mesa-dev)
+    isPlatform "x11" && depends+=(libgtk3.0-dev)
+    getDepends "${depends[@]}"
+}
+
+function sources_jfduke3d() {
+    gitPullOrClone
+}
+
+function build_jfduke3d() {
+    local params=(DATADIR=$romdir/ports/duke3d RELEASE=1)
+
+    ! isPlatform "x86" && params+=(USE_ASM=0)
+    ! isPlatform "x11" && params+=(WITHOUT_GTK=1)
+    ! isPlatform "gl3" && params+=(USE_POLYMOST=0)
+
+    if isPlatform "gl" || isPlatform "mesa"; then
+        params+=(USE_OPENGL=USE_GL2)
+    elif isPlatform "gles"; then
+        params+=(USE_OPENGL=USE_GLES2)
+    else
+        params+=(USE_OPENGL=0)
+    fi
+
+    make clean veryclean
+    make "${params[@]}"
+
+    md_ret_require="$md_build/duke3d"
+}
+
+function install_jfduke3d() {
+    md_ret_files=(
+        'duke3d'
+        'build'
+        'README.md'
+        'GPL.TXT'
+    )
+}
+
+function game_data_jfduke3d() {
+    local dest="$romdir/ports/duke3d"
+    if [[ ! -n $(find $dest -maxdepth 1 -iname duke3d.grp) ]]; then
+        mkUserDir "$dest"
+        local temp="$(mktemp -d)"
+        download "$__archive_url/3dduke13.zip" "$temp"
+        unzip -L -o "$temp/3dduke13.zip" -d "$temp" dn3dsw13.shr
+        unzip -L -o "$temp/dn3dsw13.shr" -d "$dest" duke3d.grp duke.rts
+        rm -rf "$temp"
+        chown -R $user:$user "$dest"
+    fi
+}
+
+function configure_jfduke3d() {
+    local config="$md_conf_root/jfduke3d/duke3d.cfg"
+
+    mkRomDir "ports/duke3d"
+    moveConfigDir "$home/.jfduke3d" "$md_conf_root/jfduke3d"
+
+    # params are just for parity with eduke32, last one is not supported
+    addPort "$md_id" "duke3d" "Duke Nukem 3D" "$md_inst/duke3d" "-j$romdir/ports/duke3d -addon 0"
+
+    if [[ "$md_mode" != "install" ]]; then
+        return
+    fi
+    game_data_jfduke3d
+
+    if [[ -f "$config" ]]; then
+        return
+    fi
+
+    # no config file exists, creating one
+    # with alsa as the sound driver
+    cat >"$config" << _EOF_
+[Sound Setup]
+MusicParams = "audio.driver=alsa"
+_EOF_
+    chown -R $user:$user "$config"
+}


protocultor:
From the same author ([Jonathon Fowler](https://www.jonof.id.au/)) there's a Shadow Warrior port, which uses the same Build engine port, obtained in a git submodule. Seeing that its compilation process is not that different from JFDuke3D, I opted to include it here. Its name is [JFShadowWarrior](https://www.jonof.id.au/jfsw/index.html), or JFSW for short.
Credits and thanks to @Exarkuniv for the detection and addition of the game's expansions, which is included in [RetroPie-Extra](https://github.com/Exarkuniv/RetroPie-Extra).
If you're reading this, I hope you don't mind. The difference with it is that with this one detects more platforms than the Pi 4 alone.
In terms of performance, in a RPi 3B+ you'll have to disable _Voxel Sprites_ to get JFDuke3D's speed. Otherwise, just keep it at 640x480 resolution.
Note that here the "new way" of keeping config directories (`/retropie_dir/config/game_name/port_name/`) is followed, something I couldn't do with JFDuke3D since EDuke32 saves to `config/duke3d`.

joolswills:
Thank you. I think we could avoid some duplication as the build logic looks the same. the shadow warrior port could call the build call from the jfduke3d module.

protocultor:
> I think we could avoid some duplication as the build logic looks the same. the shadow warrior port could call the build call from the jfduke3d module.

Done, now JFShadowWarrior is completely dependent on JFDuke3D. I hope you aren't bothered by the parametrization (not a fan of the way it was handled in the eduke32/ionfury pairing).
Note that I removed the specific commit to clone from both projects' repos. They don't update that often, so missing any bug fixing is not worth it.
Also removed forcing the ALSA driver for music on x86.

scriptmodules/ports/jfduke3d.sh
```diff
@@ -0,0 +1,118 @@
...
+    mkRomDir "ports/duke3d"
+    moveConfigDir "$home/.jfduke3d" "$md_conf_root/jfduke3d"
+
+    # params are just for parity with eduke32, last one is not supported
+    addPort "$md_id" "duke3d" "Duke Nukem 3D" "$md_inst/duke3d" "-j$romdir/ports/duke3d -addon 0"
+
+    if [[ "$md_mode" != "install" ]]; then
+        return
+    fi
+    game_data_jfduke3d
```
joolswills, line 116: I think this should just call the eduke32 game data function (if it's the same)

protocultor: It's "mostly" the same, but it can't be used directly. [Look at it](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/ports/eduke32.sh#L79), and note how at the beginning has an `if [[ "$md_id" == "eduke32" ]];` to verify if it has to install it, because `game_data()` is called by `configure()`, and `configure()` is called on both `eduke32` and `ionfury`.
I can do corrections to both if you want.

joolswills: Thanks - yeah, changing it to support your modules as well would make sense.

protocultor: Did the short version: moved the `if` sentence to `configure()` , so `game_data()` only worries about Duke3D data, and not if it should be called or not. That allows it to be called from the jfduke3d module.

scriptmodules/ports/jfduke3d.sh
```diff
@@ -0,0 +1,105 @@
...
+rp_module_flags=""
+
+function depends_jfduke3d() {
+    local depends=(
+        libsdl2-dev libvorbis-dev libfluidsynth-dev
+    )
+
+    isPlatform "x86" && depends+=(nasm)
+    isPlatform "gl" || isPlatform "mesa" && depends+=(libgl1-mesa-dev libglu1-mesa-dev)
+    isPlatform "x11" && depends+=(libgtk3.0-dev)
```
joolswills, line 26: There isn't a package libgtk3.0-dev - should it be libgtk-3-dev ?

protocultor: You're right, thanks for the correction.

protocultor: Just in case: I force pushed the fix to this; the script is referencing the correct package now (libgtk-3-dev).

newhinton:
@protocultor The emulator works great after adding your script. Is this the right place to mention issues for controls specific to this emulator?

protocultor:
@newhinton no, I don't think so, this is for the addition of the scripts to RetroPie. Try the [official documentation](https://jonof.id.au/jfduke3d/readme.html), or better yet, [my post in the official RetroPie forums](https://retropie.org.uk/forum/topic/34804/duke-nukem-3d-and-shadow-warrior-source-ports-for-old-pis). I'll try to help you there.

scriptmodules/ports/jfduke3d.sh
```diff
@@ -0,0 +1,105 @@
...
+    cat >"$config" << _EOF_
+[Sound Setup]
+MusicParams = "audio.driver=alsa"
+_EOF_
+    chown -R $user:$user "$config"
+}
+
+function configure_jfduke3d() {
+    mkRomDir "ports/duke3d"
+    moveConfigDir "$home/.jfduke3d" "$md_conf_root/jfduke3d"
```
s1eve-mcdichae1, line 95: This makes `configs/ports/jfduke3d` right? Should it be `configs/ports/duke3d/jfduke3d` instead? Puts the `jfduke3d` *inside* `configs/duke3d` (which already exists for **eduke32** module), instead of alongside it.

scriptmodules/ports/jfsw.sh
```diff
@@ -0,0 +1,64 @@
...
+        rm -rf "$temp"
+    fi
+    chown -R $user:$user "$dest"
+}
+
+function configure_jfsw() {
+    local gamedir="$romdir/ports/shadowwarrior"
+
+    mkRomDir "ports/shadowwarrior"
+    moveConfigDir "$home/.jfsw" "$md_conf_root/sw/jfsw"
```
s1eve-mcdichae1, line 53: Here, the opposite. Should this just use `configs/ports/sw` like the RP-Extra module, or why the extra layer?

protocultor: I though the same, but `configs/ports/duke3d` points ***directly*** to the config to EDuke32, and your idea would force the config of JFDuke3D to go inside the config of EDuke32 as a subdirectory. Imagine how that would appear inside the config directories of RetroPie (`/opt/retropie/configs/ports`), and most importantly, how the "home" directory would appear:
```
ports
  |
  |- duke3d (1)
        |
        |- jfduke3d (2)

~
|
|- .config
|     |
|     |- eduke32 (1)
|          |
|          |- jfduke3d (2)
|
|- .jfduke3d (2)
```
`(1)` and `(2)` are to see where the links point.

The "_extra layer_" you mention was inspired on how is currently treated for _Quake 1_ and _2_, where the first layer is the name of the game, and inside them is every individual port for that game (compare `tyrquake.sh` and `darkplaces-quake.sh` to understand). It's just better organized and future-proofed, compared with how _Duke 3D_ does it, which just has "the name of the game" as the destination for the config of just one of the ports. It's just something that didn't age well, because of the need for organization when having more than one port for a game.

I don't find that joining all the configs on the different ports under the "first port made" is a good idea, especially if the user wants to back up a config for a certain port; this user might find every single config for every port of that game included. Or if the players want to "reset the config" by cleaning the directory, they will find that they deleted the configs for all of the ports of that game.

Of course, there may be a better solution, but I dealed with a legacy problem the best I could. I just don't find mixing directories is a good idea, but if the owners of the project want that, it'll be done.

s1eve-mcdichae1: So what you're saying is that it *should* actually be `configs/ports/duke3d/eduke32`, but that that ship has sailed...

protocultor: Yeah, exactly, that's what EDuke32 should have used since the beginning. Although I understand no one would have predicted the use of different ports for the same game, it would have been nice if it was considered in some form. In the ideal world, it would have been great if this change could be done without the user having to do anything on their end.

Still, this had me thinking again. What if JFDuke3D, instead of using `configs/ports/jfduke3d`, used `configs/ports/dukenukem3d/jfduke3d`? Even if we have the same game twice, at least there would be a place where the configs of a potential 3rd port of _Duke Nukem 3D_ could be located, without filling more places on the "root" of `ports`. If you have a better idea, I'm all ears.

joolswills: Thanks. The config locations for some things are not ideal, due to later additions. It might be worth re-organising the configs for eduke32. I will come back to you. Sorry for the delay.

PR #3808, opened 2023-11-19
retroarch - use --disable-crtswitchres on videocore

joolswills:
gfx/video_crt_switch.c doesn't currently build on videocore, due to definition of sr_state being in deps/switchres/switchres_wrapper.h which is not included on videocore.

I haven't used this functionality, so this is a quick workaround, rather than a proper fix. The fact that the code has ifdefs for videocore - I guess it did work, but will need to be looked at. But for now, just wanted to get RetroArch building again.

### DIFF
--- a/scriptmodules/emulators/retroarch.sh
+++ b/scriptmodules/emulators/retroarch.sh
@@ -52,6 +52,7 @@ function build_retroarch() {
         isPlatform "gles31" && params+=(--enable-opengles3_1)
         isPlatform "gles32" && params+=(--enable-opengles3_2)
     fi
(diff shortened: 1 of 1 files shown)


joolswills:
I'm merging this now as a temporary measure until we have a proper fix.

cmitu:
The initial version of `crtswitchres` I think was added by @alphanu1 (https://github.com/libretro/RetroArch/pull/8132) and supported a the VC drivers on the Pi and X11. It was later on replaced with a generic method based on [switchres](https://github.com/antonioginer/switchres), which I don't think supports the VC platform. Probably some `#define`s have been left from the initial implementation, but not fully updated/corrected when the new implementation was added.

PR #3810, opened 2023-11-20
ppsspp - fix building on Raspberry Pi with proprietary drivers

joolswills:
GL_COMPRESSED_RGBA_ASTC_4x4_KHR isn't available in the videocore iv gles2 includes, and documentation suggests this is an opengles 3 extension.

Add ifdefs to not include this if it isn't defined.

### DIFF
--- a/scriptmodules/emulators/ppsspp.sh
+++ b/scriptmodules/emulators/ppsspp.sh
@@ -53,6 +53,9 @@ function sources_ppsspp() {
     # ensure Pi vendor libraries are available for linking of shared library
     sed -n -i "p; s/^set(CMAKE_EXE_LINKER_FLAGS/set(CMAKE_SHARED_LINKER_FLAGS/p" cmake/Toolchains/raspberry.armv?.cmake
 
+    # fix missing defines on opengles2
+    applyPatch "$md_data/gles2_fix.diff"
+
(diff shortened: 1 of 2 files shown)


joolswills:
Based on feedback from https://github.com/hrydgard/ppsspp/pull/18433 I will change this so it instead uses `#ifdef GL_COMPRESSED_RGBA_ASTC_4x4_KHR`

microplay-hub:
The fix works only with v1.16 ! i have try it on official RetroPie Image from Raspberry-PI-Imager and i cant Build with buster become a error

The Problem ist the version check, use by buster v1.13.2 and here dont works the diff, i have remove the check and the diff works with 1.16

function _get_release_ppsspp() {
    local tagged_version="v1.16.6"
    #  the V3D Mesa driver before 21.x has issues with v1.14 and later
    if [[ "$__os_debian_ver" -lt 11 ]] && isPlatform "kms" && isPlatform "rpi"; then
        tagged_version="v1.13.2"
    fi
    echo $tagged_version
}

joolswills:
Thanks - will fix it (and ppsspp-1.5.4)

joolswills:
Fixed - https://github.com/RetroPie/RetroPie-Setup/pull/3814

microplay-hub:
perfect np

PR #3817, opened 2023-12-03
runcommand: have XINIT start a wm by default

cmitu:
Before the introduction of backends, emulator commands prefixed with `XINIT` would also start a WM (matchbox-window-manager). Now the WM is started only for emulators configured with backends `x11`/`x11-c`.

This change restores the previous behavior for `XINIT`-ed emulators for which no `x11` based backend is configured (i.e. non-SDL2 emulators like Dolphin or Duckstation).

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -1213,6 +1213,8 @@ function get_sys_command() {
         # remove XINIT:
         COMMAND="${COMMAND:6}"
         XINIT=1
+        # start with a window manager and hide the cursor by default
+        XINIT_WM=1
     fi
 }
 


cmitu:
Actually, we didn't have automatically start a WM before the backends module was introduced - I was mistaken. I'll check to see what's the impact of the change for the modules that call `XINIT` specifically.

cmitu:
I'm going to shelve this change, since it doesn't match my original premise. For modules using the `XINIT` prefix for starting the emulators, the start-up script already includes the start of a WM and this change will yield an error during start-up. While we only have a couple of those modules, the Extras repo has many more in this situation.

@joolswills how about adding a `XINIT_WM` prefix that will `runcommand` can interpret similar to how `XINIT_WM` is set via `backends.cfg` ? For my original intent (for `dolphin`) I think manually calling the _ _set_backend_ function might work, but the modification would get rid of each module including their own WM initialization and the associated `xset` function.

PR #3820, opened 2023-12-09
Update esthemes.sh 2x new themes

RetroHursty69:
Created 2 new themes for ES to suit 16:9. Called AIGEN and AIGEN_PLUS.


joolswills:
Thank you.

PR #3821, opened 2023-12-14
smw: switch to @mmatyas's fork; upgrade to 2.0

cmitu:
The fork from https://github.com/mmatyas/supermariowar is still active and it has the options to use SDL2 instead of SDL1. Switched the scriptmodule to it with updated dependencies and build instructions. New version also changed the configuration location, now in a dedicated folder.

Main changes since 1.8

 * Experimental online mode, including an online lobby server
 * Lots and lots and lots of bugfixes (seriously, there's at least 200)
 * Added Web, ARM and (an experimental) Android ports, and improved support for Windows, Linux and OSX
 * Improved music and sound quality
 * Ported all parts of the game to SDL2, for hardware accelerated drawing and improved support for many platforms
 * Major cleanup of the source code and the build system

NOTES: 
  - I've updated the license to GPL2, since this is what's declared in the CMake build file.
  - ~~On exit, there's a segfault due to the usage of `atexit(SDL_Quit)`, which is due to Mesa's usage of the same exit handler. An older 'bug', which may be corrected upstream.~~ fixed in https://github.com/mmatyas/supermariowar/issues/258

### DIFF
--- a/scriptmodules/ports/smw.sh
+++ b/scriptmodules/ports/smw.sh
@@ -10,14 +10,14 @@
 #
 
 rp_module_id="smw"
-rp_module_desc="Super Mario War"
-rp_module_licence="GPL http://supermariowar.supersanctuary.net/"
-rp_module_repo="git https://github.com/HerbFargus/Super-Mario-War.git master"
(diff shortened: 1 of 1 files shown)


joolswills:
Thank you.

PR #3823, opened 2023-12-21
Detect rpi5 platform based on /proc/device-tree/compatible

ct1n:
/proc/cpuinfo doesn't contain any Hardware line for Raspbery Pi 5, bookworm 64 bit OS image. See https://github.com/raspberrypi/linux/issues/5783.

Use /proc/device-tree/compatible instead, in the same way other platforms are detected.

Contents of /proc/device-tree/compatible on my rpi5:
raspberrypi,5-model-bbrcm,bcm2712

Before:
__platform_flags=([0]="" [1]="64bit" [2]="gl" [3]="vulkan" [4]="x11")

After:
__platform_flags=([0]="rpi5" [1]="64bit" [2]="aarch64" [3]="rpi" [4]="gles" [5]="gles3" [6]="gles31" [7]="mesa" [8]="kms")

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -458,6 +458,9 @@ function get_platform() {
                         *rk3588*)
                             __platform="rk3588"
                             ;;
+                        *bcm2712*)
+                            __platform="rpi5"
+                            ;;
                     esac
                 elif [[ -e "/sys/devices/soc0/family" ]]; then
                     case "$(tr -d '\0' < /sys/devices/soc0/family)" in


ct1n:
Found a random list of compatible strings for Raspbery Pis here:
https://raw.githubusercontent.com/cr-chsn1/raspiced/main/hwbot_prime_benchmark/script/rpi_hwbot.sh

The authoritative list (or closest to it) is probably in the Linux kernel tree, in files like:
https://git.kernel.org/pub/scm/linux/kernel/git/devicetree/devicetree-rebasing.git/tree/src/arm/broadcom/bcm2835-rpi-a-plus.dts
https://git.kernel.org/pub/scm/linux/kernel/git/devicetree/devicetree-rebasing.git/tree/src/arm/broadcom/bcm2711-rpi-4-b.dts
(though rpi5 is missing; maybe not upstreamed yet)

We could either add checks for all these compatible strings, or just check for "brcm", "bcm" or such and refine using the existing code based on Revision from /proc/cpuinfo (though maybe that might change further).

Let me know what you think.

cmitu:
@ct1n we'll probably handle with something similar to https://github.com/cmitu/RetroPie-Setup/tree/pi-detection.

ct1n:
Yup, that's one of the ways I was thinking. Sounds good, I'll close this PR.

cmitu:
@ct1n thanks for the patch and the bug report.

PR #3824, opened 2023-12-22
system: fix RPI detection with recent RPI kernels

cmitu:
The 'Hardware' field has been removed from the output of '/proc/cpuinfo' after [1], so the setup script will not correctly identify it's running on a Raspberry PI. Use '/proc/device-tree/compatible' to check whether we're on a Raspberry Pi, the same method used for other SBCs.

Model calculation based on the Revision field from '/proc/cpuinfo' needs to be run in 2 places, so the code has been put in a separate function.

[1] https://github.com/pelwell/linux/commit/17804c20303c8295b05baf2ab9578432ac5c2735

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -376,37 +376,42 @@ function get_rpi_video() {
     export PKG_CONFIG_PATH="$pkgconfig"
 }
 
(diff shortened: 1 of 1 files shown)


joolswills:
Thank you.

PR #3825, opened 2023-12-26
skyscraper.sh - use 'Skyscraper -h' for version interrogation. Require 'ge' match for version check.

s1eve-mcdichae1:
Re: https://github.com/RetroPie/RetroPie-Setup/pull/3793#issuecomment-1803364524

@cmitu said:

> The version interrogation can also be shortened to just
> 
>     "$md_inst/Skyscraper --version | cut -d' ' -f 2

This no longer returns a coherent value.

In 3.9.2, the `--version` option returns:

```
pi@retropie:~ $ Skyscraper --version 
 _______ __                                                     ___
|   _   |  |--.--.--.-----.----.----.---.-.-----.-----.----.   |"""|
|   1___|    <|  |  |__ --|  __|   _|  _  |  _  |  -__|   _|   |"""|
|____   |__|__|___  |_____|____|__| |___._|   __|_____|__|     |"""|
|:  1   |     |_____|                     |__|       3.9.2     |"""|
|::.. . |                                                      |"""|
`-------' by Lars Muldjord and contributors                   ""''''" 
```

So, with the `cut` command from the function it looks like:

```
pi@retropie:~ $ Skyscraper --version | cut -d' ' -f 2





.
by

pi@retropie:~ $
```

Now, this currently does "work" in that the non-numerical value returned is not "less than" anything, so it doesn't ~~fail the version check~~ *pass the "fails the version-check" check* and so the script will still gather resources and generate gamelists, but it's not actually returning a version number at all. In function `_check_ver_skyscraper`, if I change the test from:

    if compareVersions "$ver" lt "3.5"

...to:

    if ! compareVersions "$ver" ge "3.5"

...(changes "less than" to "not greater-or-equal") then it fails the check in 3.9.2, says not compatible and won't gather/generate (works fine with this change in 3.9.1).

The old way `Skyscraper -h | grep 'Running Skyscraper'` still works in 3.9.2, even with this change to "not greater":

```
pi@retropie:~ $ Skyscraper -h | grep 'Running Skyscraper'
Running Skyscraper v3.9.2 by Lars Muldjord
pi@retropie:~ $ Skyscraper -h | grep 'Running Skyscraper' | cut -d' ' -f 3 | tr -d v
3.9.2
```

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -147,13 +147,13 @@ function _purge_platform_skyscraper() {
 
 function _get_ver_skyscraper() {
     if [[ -f "$md_inst/Skyscraper" ]]; then
-        echo $(sudo -u "$user" "$md_inst/Skyscraper" --version | cut -d' ' -f2 2>/dev/null)
+        echo $(sudo -u "$user" "$md_inst/Skyscraper" -h | grep 'Running Skyscraper' | cut -d' ' -f 3 | tr -d v 2>/dev/null)
     fi
 }
 
 function _check_ver_skyscraper() {
     ver=$(_get_ver_skyscraper)
-    if compareVersions "$ver" lt "3.5"; then
+    if ! compareVersions "$ver" ge "3.5"; then
         printMsgs "dialog" "The version of Skyscraper you currently have installed is incompatible with options used by this script. Please update Skyscraper to the latest version to continue."
         return 1
     fi


cmitu:
@s1eve-mcdichae1 thanks for the modifications.
Why we need more than the modification for the version detection (`lt` vs `ge`) ? If we can always get the right version (regardless of the version installed), isn't it enough ?

s1eve-mcdichae1:
> Why we need more than the modification for the version detection (`lt` vs `ge`) ? If we can always get the right version (regardless of the version installed), isn't it enough ?

Sure. Functionally it don't matter, since, as I said before it technically still works. With the current version of the rule "less-than == fail", then garbage in == "not-less, so not-fail" and the script will continue. If all older versions report a proper version number and only the newer, supported versions return garbage, then it won't ever matter, in a technical sense.

Aesthetically, from a logical/programmatical (/nitpicking) perspective, if we want to test for "the version should be 3.5 or greater, else do not proceed" then we shouldn't accept a non-number like `bleven point yellow` or `[newline][newline].[newline]by[newline]`, probably.

It's the difference between `require: see the big number`, and `require: NOT see the *little* number (but even if we do see some other stuff that isn't a number at all it's still okay as long as we don't also see the small number anywhere at the same time)`.

One of those is precisely what we're looking for; the other, is a roundabout way of probably still guessing it right most of the time.

Which is *probably* still fine, most of the time, if "probably fine most of the time" is what you're after. It's a small change to just make it for-sure, though.

Basically it comes down to: in the scenario where version interrogation returns something other than a coherent version number, should the script attempt to proceed or should it abort?

cmitu:
> Basically it comes down to: in the scenario where version interrogation returns something other than a coherent version number, should the script attempt to proceed or should it abort?

Abort. Fail early and find the issue in order to fix it.

s1eve-mcdichae1:
> Abort. Fail early and find the issue in order to fix it.

Right. That is the purpose of the change here at L156. The current rule `if compareVersions "lt" #then abort` let the error (at 150) slip through (since it wasn't "less", so it didn't abort).

The proposed rule `if ! compareVersions "ge"` would have (does, in testing) caught it, aborting the action and alerting to the error at 150.

cmitu:
Ok, I tested it and looks fine. Thank you !

PR #3829, opened 2024-01-06
runcommand: add a new launch prefix for X11 programs

cmitu:
Extended the `XINIT:` launch prefix with a new one in order to start a window manager before the x11 application is started. This is similar to how the 'x11' backend starts an application, but without needing an additional configuration set in 'backends.cfg'.

Simplified the launching of x11 enabled ports by replacing their explicit launching scripts with a launch command based on the new launch prefix.


joolswills:
Thanks - have reworked the logic (as discussed) and updated the PR. I need to test it though (feel free to check).

cmitu:
It looks fine - wasn't familiar with the 'match and continue/restart' for the `case` options, but it works. The `XINIT` options are working fine from my testing and `zork` also seems unaffected by the changes.

joolswills:
Thank you!

PR #3830, opened 2024-01-06
raspbiantools: pin Bluez to the RPI repos for Buster

cmitu:
Upgrading via 'apt' the Bluez package with a the version from Debian updates will overwriten the RPT. This breaks pairing for some gamepads. Pin the package so it's always installed/updated from archive.raspberrypi.org on RaspiOS 10 'Buster'.

Added a couple of related changes to how 'apt' is run when updating:
 * use 'full-upgrade' instead of 'dist-upgrade', since it will install new packages introduced by the upgrade.
 * add '--allow-downgrades' to make sure that 'bluez' is retrieved from the RPI repos if it was already upgraded by an upstream update.

### DIFF
--- a/scriptmodules/supplementary/raspbiantools.sh
+++ b/scriptmodules/supplementary/raspbiantools.sh
@@ -18,8 +18,11 @@ function apt_upgrade_raspbiantools() {
     # install an older kernel/firmware for stretch to resolve newer kernel issues or unhold if updating to a newer release
     stretch_fix_raspbiantools
 
+    # on Buster, always install the Bluez package from the RPI repos
+    buster_bluez_pin_raspbiantools
+
     aptUpdate
-    apt-get -y dist-upgrade
+    apt-get -y dist-upgrade --allow-downgrades
 }
 
 function lxde_raspbiantools() {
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/raspbiantools.sh
```diff
@@ -73,6 +76,19 @@ function stretch_fix_raspbiantools() {
     fi
 }
 
+function buster_bluez_pin_raspbiantools() {
+    # pin the 'bluez' package to the RPI repos to prevent any Debian updates overwriting it
+    # use Priority 1001 to force the the installation even when the Debian package is installed
+    local pin_file="/etc/apt/preferences.d/01-bluez-pin"
+    if isPlatform "rpi" && [[ "$__os_debian_ver" -eq 10 ]] && ! -f "$pin_file" ; then
```
joolswills, line 83: Should this be `    if isPlatform "rpi" && [[ "$__os_debian_ver" -eq 10 && ! -f "$pin_file" ]] ; then` ?

scriptmodules/admin/setup.sh
```diff
@@ -595,7 +595,7 @@ function update_packages_gui_setup() {
                 rp_callModule raspbiantools apt_upgrade
             else
                 aptUpdate
-                apt-get -y dist-upgrade
+                apt-get -y full-upgrade
```
joolswills, line 598: I think `full-upgrade` is for `apt` not `apt-get` - it works, but isn't in the man file for apt-get (apart from a mention in another parameter), but I think it is equivalent or should be. Doesn't the pinning work with dist-upgrade?

joolswills: https://salsa.debian.org/apt-team/apt/-/blob/1.6.12/cmdline/apt-get.cc#L415

yeah they are the same.

joolswills: (See https://askubuntu.com/questions/770135/apt-full-upgrade-versus-apt-get-dist-upgrade)

scriptmodules/supplementary/raspbiantools.sh
cmitu: Yes, I corrected it.

scriptmodules/admin/setup.sh
cmitu: Yes, it was the `apt` thing that made me change it. But since they're the same, there's no point in changing it - so reverted the changes.

joolswills:
Thank you. Looks good.

PR #3831, opened 2024-01-07
emulationstation: fix fullscreen launching params on Wayland/Gnome

cmitu:
When running under Gnome/Wayland on Ubuntu 22.04 default configuration, the '--fullscreen-borderless' is not actually fullscreen. The top horizontal panel (?) is still visible and the EmulationStation window is actually pushed down.

Running with just '--fullscreen' is enough, but this is the default, thus there's no need to add any additional parameters when Gnome/Wayland is the current session. Tested on an updated Ubuntu 22.04, with the default Gnome session.


joolswills:
Thanks.

PR #3832, opened 2024-01-11
micropolis: fix path and display

cmitu:
Broken by my previuos PR.

Micropolis runs in a maximized window, but doesn't have any fullscreen option and it relies on the WM cursor to operate, so use the WM + Cursor option for it.


cmitu:
Let me do some testing on this - it may need the `XINIT-WMC` option.

PR #3834, opened 2024-01-12
lr-ppsspp - split out lr-ppsspp specific patches

joolswills:
This fixes binary builds not triggering due to the changes being in ppsspp.


cmitu:
Looks good. 👍

PR #3835, opened 2024-01-15
runcommand: disable powersave when XINIT is used

cmitu:
Add the `xset` commands to disable the display powersave and screen blanking, since gamepad activity is ignored by Xorg and the display will go to sleep if no keyboard/mouse is active. 

I removed the script launchers ccbee7c05a, but the `xset` commands haven't been added `XINIT-WM` launchers.


joolswills:
Thank you.

PR #3836, opened 2024-01-16
helpers: map 'xorg' for Pi5 and for 'legacy' exclusion

cmitu:
The 'xorg' (meta)-package will install the necessary packages to run an Xorg server and to launch applications under an Xorg environment. We need to map 'xorg' for:

 - Pi5 needs an additional package that provides configuration for Xorg, without it the server will not start. Unfortunately the package has no reverse dependency (for now), so it must be installed explicitely.
 - One of the recommended packages installed is the 'xserver-xorg-legacy' package, which breaks our x11 apps launched from `runcommand`. We already uninstall it when the setup script starts (regardless of how it was installed), but let's not install it ourselves when 'xorg' is a dependency for one of our own scriptmodules. NOTE: `runcommand` breakage under `-legacy` is due to our redirection of STDIN/STDOUT. The `-legacy` provided wrapper checks whether the current controlling tty is a 'real console' tty by looking at the major/minor device numbers and bails out if it's not 4/1.

### DIFF
--- a/scriptmodules/helpers.sh
+++ b/scriptmodules/helpers.sh
@@ -299,6 +299,13 @@ function _mapPackage() {
         libfreetype6-dev)
             [[ "$__os_debian_ver" -gt 10 ]] || compareVersions "$__os_ubuntu_ver" gt 23.04 && pkg="libfreetype-dev"
             ;;
+        xorg)
+            # outside X11, don't install the 'xserver-xorg-legacy' package, even if recommended by 'apt'
+            # since it breaks launching x11 apps with 'runcommand'
+            ! isPlatform "x11" && pkg+=" xserver-xorg-legacy-"
+            # Pi5 needs an additional package for xserver to start
+            isPlatform "rpi5" && pkg+=" gldriver-test"
+            ;;
     esac
     echo "$pkg"
 }


cmitu:
Set to draft since we have other options for solving the issues with `xorg`:

 - for the Pi5, we can manually generate the `.conf` file needed when the built-in GPU is used (which is probably 99% of the cases).
 - the the `legacy` stuff, it's not a big deal since we already remove the package in `get_retropie_depends` from `system.sh`, but this means the setup script has to be re-run, which might trip users that installed `xorg` for a module and then have the module's application not run because of the `-legacy` being still present.

joolswills:
Looks good to me thanks.

joolswills:
Thank you

cmitu:
This doesn't seem to work, maybe because `_mapPackage` wasn't meant to return more than one package. There's some code to handle `RP <module> <pkg>`, but returning a list of package names to `getDepends` will install only the 1st package. I think it needs reverting.

joolswills:
I will revert it. I'll see if I can do this in another way.

cmitu:
Thanks.

I think we need to leave the `legacy` alone (it's already handled by `setup`, but only at start-up) and add `gldriver-test` as outright dependency on `rpi5`. The `gldriver-test` doesn't have any direct dependency outside `raspi-config`.

PR #3837, opened 2024-01-17
runcommand: ignore KMSTOOL if no DRM card is available

cmitu:
Check if '/dev/dri/card0' exists before running `$KMSTOOL` (now 'modetest') to check for kernel modesettings support. This avoids 'modetest' being executed and removes almost 2 sec from the `runcommand` runtime on a Pi3.

NB: we can probably cache the result of `$KMSTOOL` globally (i.e. store it under `/dev/shm/kms-output.txt`) and avoid `modetest` being executed altogether on subsequent runs. I'm testing right now switching to 'kmsprint' [1] to replace `modetest` and it's fast enough to be un-noticeable on a Pi3 (~ 100-200ms).


joolswills:
Caching the output is a good idea. Looks good, but please use double `[[`

cmitu:
> Looks good, but please use double `[[`

Done.

PR #3838, opened 2024-01-19
Amiberry 5.6.6

rahim:
This bumps Amiberry from 5.6.4 -> 5.6.6.

Release notes:
https://github.com/BlitterStudio/amiberry/releases/tag/v5.6.5
https://github.com/BlitterStudio/amiberry/releases/tag/v5.6.6

Notably this changes the default vsync behaviour, which should fix https://github.com/BlitterStudio/amiberry/issues/1189  – render corruption when launching from EmulationStation on Pi5. (That was my own motivation, but I'm yet to confirm it has the desired outcome).

libportmidi-dev [has been added as a build dependency](https://github.com/BlitterStudio/amiberry/commit/e52bd6f44d776affc7017192b7ef691c496a3420).

### DIFF
--- a/scriptmodules/emulators/amiberry.sh
+++ b/scriptmodules/emulators/amiberry.sh
@@ -13,7 +13,7 @@ rp_module_id="amiberry"
 rp_module_desc="Amiga emulator with JIT support (forked from uae4arm)"
 rp_module_help="ROM Extension: .adf .chd .ipf .lha .zip\n\nCopy your Amiga games to $romdir/amiga\n\nCopy the required BIOS files\nkick13.rom\nkick20.rom\nkick31.rom\nto $biosdir/amiga"
 rp_module_licence="GPL3 https://raw.githubusercontent.com/BlitterStudio/amiberry/master/LICENSE"
-rp_module_repo="git https://github.com/BlitterStudio/amiberry v5.6.4"
+rp_module_repo="git https://github.com/BlitterStudio/amiberry v5.6.6"
 rp_module_section="opt"
 rp_module_flags="!all arm rpi3 rpi4 rpi5"
 
@@ -47,7 +47,7 @@ function _get_platform_amiberry() {
 }
 
 function depends_amiberry() {
-    local depends=(autoconf libpng-dev libmpeg2-4-dev zlib1g-dev libmpg123-dev libflac-dev libsdl2-dev libsdl2-image-dev libsdl2-ttf-dev libserialport-dev wget)
+    local depends=(cmake autoconf libpng-dev libmpeg2-4-dev zlib1g-dev libmpg123-dev libflac-dev libsdl2-dev libsdl2-image-dev libsdl2-ttf-dev libserialport-dev wget libportmidi-dev)
 
     isPlatform "dispmanx" && depends+=(libraspberrypi-dev)
     isPlatform "vero4k" && depends+=(vero3-userland-dev-osmc)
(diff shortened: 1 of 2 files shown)


cmitu:
Thanks for the PR. Which systems has this been tested on ?

You also want to include `cmake` as depenency and squash the commits into a single one. Is there a way to disable MUNT (MT32 emulation) for weaker systems (Pi0/1/2/3) ?

rahim:
Hi @cmitu 

I should have warned I'm not familiar with either project, I've been enjoying tinkering to try and get things behaving for my setup and largely cargo culted this upgrade PR from earlier work. I thought it might save someone a few minutes of wrangling to generate the new patchfile given the upstream Makefile had been updated.

> Which systems has this been tested on?

The only thing I've tested is that compiling amiberry v5.6.6 succeeds on Pi5 (8GB) running on 64 bit Pi OS, 8KB page size, RetroPie installed manually following the guide.

I don't have earlier generation Pis readily available to test with.

> You also want to include cmake as depenency and squash the commits into a single one.

👍 That's easily done, I did wonder about cmake, but presumed it must have been required earlier, so perhaps covered elsewhere.

> Is there a way to disable MUNT (MT32 emulation) for weaker systems (Pi0/1/2/3) ?

My lack of knowledge is exposed by this – that was carried verbatim from the changes in amiberry, I'm not sure of the background to that choice.

rahim:
> Notably this changes the default vsync behaviour, which should fix https://github.com/BlitterStudio/amiberry/issues/1189 – render corruption when launching from EmulationStation on Pi5. (That was my own motivation, but I'm yet to confirm it has the desired outcome).

The changes in 5.6.6 didn't resolve my issue here either as it happens, I need to do some more digging in to that.

rahim:
I've squashed and added `cmake`.

Right now I've not rebuilt with that, but I can't see that creating a problem.

As best I can tell, there's currently not support for compiling without MUNT. Beyond compilation cost, it's not clear to me it would incur runtime cost unless users choose to lean on that feature?

Given the upgrade doesn't resolve the problem I was hoping it would, and given the limited testing picture, merging perhaps represents more trouble than it's worth right now.

cmitu:
> As best I can tell, there's currently not support for compiling without MUNT. Beyond compilation cost, it's not clear to me it would incur runtime cost unless users choose to lean on that feature?

 Yes, it doesn't seem so. As for the cost - it's just removing extra compilation steps and possible support questions from users which don't have powerful enough devices to run it.

> Given the upgrade doesn't resolve the problem I was hoping it would, and given the limited testing picture, merging perhaps represents more trouble than it's worth right now.

The upgrade cannot solve the issue you had, since it's not an Amiberry bug - it's a Linux kernel bug which will be solved in raspberrypi/linux. What's fixed in Amiberry is the _Vsync_ not being read and applied from the configuration file at start-up (option which works around the kernel bug). 

Tested on Pi3 Buster/Pi4 Bookworm and it builds and runs ok. I think we can merge this.

rahim:
> The upgrade cannot solve the issue you had, since it's not an Amiberry bug - it's a Linux kernel bug which will be solved in raspberrypi/linux. What's fixed in Amiberry is the Vsync not being read and applied from the configuration file at start-up (option which works around the kernel bug).

Yes, I eventually got to that conclusion. Great to have a workaround and that the real fix is on the way.

> Tested on Pi3 Buster/Pi4 Bookworm and it builds and runs ok. I think we can merge this.

Excellent, thanks for the steer!

joolswills:
Thank you.

rahim:
Thank you, great project!

PR #3840, opened 2024-01-21
bashwelcometweak: fix GPU temp reading

cmitu:
On RasPI OS 11/12 (bullseye/bookworm) the `vcgencmd` command has moved to `/usr/bin`. Use the normal path first when running `vcgencmd measure_temp` to get the GPU temperature, otherwise nothing is shown.


joolswills:
On Buster /usr/bin/vcgencmd symlinks to /opt/vc/bin/vcgencmd which is pathed, so maybe it's easier to do `which vcgencmd` to test if it's installed, or just check this location ?

cmitu:
OK, I simplified the check and the execution.

joolswills:
Thank you.

PR #3844, opened 2024-01-24
mupen64plus: modify launch script for RPI detection

cmitu:
Since recent kernels don't have the BCM string in `/proc/cpuinfo`, update the Raspberry Pi detection to use the device tree.

Other updates:

 * simplifed the selection of the OMX output plugin and port number. Applies only to Pi models 0-4, since Pi5 cannot use OMX
 * disabled pre-setting the 'pulse' SDL audio driver for non-OMX/ALSA devices. SDL will try 'pulseaudio' ('pulse' is the SDL 1.2 name) then 'alsa' in this order, so there's no need to force it to 'pulseaudio'. Starting with 2.0.22, SDL supports a list of drivers in the AUDIODEVICE hint, but Buster based systems are not there yet.

### DIFF
--- a/scriptmodules/emulators/mupen64plus/mupen64plus.sh
+++ b/scriptmodules/emulators/mupen64plus/mupen64plus.sh
@@ -133,11 +133,11 @@ function remap() {
 }
 
 function setAudio() {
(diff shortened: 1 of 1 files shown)


joolswills:
Thank you!

PR #3845, opened 2024-01-25
backends: add SDL1.2 Compat backend

cmitu:
Added the SDL12-Compat backend, using a shim library which implements the `sdl1.2` API over SDL2.

The new `sdl12-compat` backend should be used for SDL1 applications when:
 * the platform doesn't have `dispmanx` anymore (e.g. `rpi` platforms with RaspiOS Bullseye and later)
 * `sdl2` has a working accelerated video & rendering drivers
 * not running under `x11` (i.e. KMS)

Notes: 

 - for SDL1+Videocore specific applications, which use directly the dispmanx/legacy GL stack (e.g. `pcsx-rearmed`, `advmame-1.4`, etc.), the backend will not help
 - this can help run more emulators/applications on platforms where there was no accelerated framebuffer SDL1 driver (i.e. Mali ?).
 - the PR marks `sdl1` scriptmodule as `dispmanx` only, to avoid installing it on Bullseye/Bookworm + RPI, since there's no `dispmanx` available there.


cmitu:
NOTE: set to draft for now, so I can re-test some of the SDL1 apps (dosbox/fuse) and add them to the PR in a separate commit.

cmitu:
OK, I think it's enough for a start. Modified the following scriptmodules to take advantage of the new backend:

   * Atari800
   * Fuse
   * Osmose
   * Linapple2 (Linapple-Pie). Upstream seems to be still using SDL1.
   * OpenBOR
   * FBZX
   * XRick
   * CapriceRPI

Considered, but not included:
  * Basilisk II. It has a SDL2 version, but it needs a SDL2 version > 2.0.10, which we don't have on Buster. Perhaps we should add a SDL2 version ?
  * XM7 - has issues building the _agar_ library under `aarch64`. Still investigating.
  * UAE4ARM/UAE4ALL - not tested

Not included:
   * Dosbox - we already have Dosbox-SDL2 and Dosbox-Staging. Emulating SDL1 with SDL2 will probably be less performant than straight up using SDL2 with the 2 other emulators.
   * Daphne - we have Hypseus with SDL2 (and x86) support, not tested.

Not tested:
   * any `videocore` emulators, that use the `dispmanx` API directly:
      - `mame4all`
      - `gpsp` (has a libretro core)
      - `pisnes`. For a 'fast' snes, we have the older cores and could also include or supplant them with the 'supafaust' core (basee on MDFN's SNES supafaust).
      - `giana`
      - `quake3`
      - `pcsx-rearmed` - can be coerced into a binary that doesn't depend on the old GPU drivers (as a generic GLES platform), but not sure if it's worth it since we have the _libretro_ core available.

Binned:
     * `np2pi` - we have the libretro core, based on a (very old) ancestor from [NP2Kai](https://github.com/AZO234/NP2kai/). NP2Kai can also be run standalone with SDL2, if we want.
     * `dgen`

Could be upgraded to SDL2:
     * `openttd` - package has SDL2 support since Bullseye, so nothing to do here (?)
     * `dxx-rebirth` - we're using an older branch, newer versions have SDL2 support.
     * `alephone` - untested
     
Wrong classification:
     * `cdogs-sdl` is actually a SDL2 program !

scriptmodules/emulators/atari800.sh
```diff
@@ -94,6 +94,9 @@ function configure_atari800() {
     # if we are on fkms, use the sdl1 dispmanx backend by default for good performance without using X11/opengl
     isPlatform kms && isPlatform "dispmanx" && _backend_set_atari800 "dispmanx"
 
+    # when no dispmanx is available, but still on KMS, use 'sdl12-compat' and go through SDL2
+    isPlatform "kms" && ! isPlatform "dispmanx" _&& backend_set_atari800 "sdl12-compat"
```
joolswills, line 98: Should this be `_backend_set_atari800 "sdl12-compat"` ?

scriptmodules/ports/xrick.sh
```diff
@@ -45,8 +45,8 @@ function configure_xrick() {
 
     # set dispmanx by default on rpi with fkms
     isPlatform "dispmanx" && ! isPlatform "videocore" && setBackend "$md_id" "dispmanx"
-
-    ln -sf "$romdir/ports/xrick/data.zip" "$md_inst/data.zip"
```
joolswills, line 49: Is this not needed? I couldn't see any mention to the removal in the commit msg.

scriptmodules/supplementary/backends.sh
```diff
@@ -18,6 +18,9 @@ function _list_backends() {
     local id="$1"
     backends=()
 
+    # skip the 'depends' packages
+    [[ ${__mod_info[$id/section]} == "depends" ]] && return 1
+
```
joolswills, line 23: Which dependency packages were showing that need to be skipped ?

scriptmodules/supplementary/sdl12-compat.sh
```diff
@@ -0,0 +1,52 @@
...
+function install_sdl12-compat() {
+    md_ret_files=(
+        libSDL-1.2.so."$(_version_sdl12-compat short)"
+        README.md
+        LICENSE.txt
+    )
+}
+
+function configure_sdl12-compat() {
+    ldconfig -nN "$md_inst"
```
joolswills, line 51: If this module is removed, does ldconfig need to be re-run ?

scriptmodules/supplementary/backends.sh
cmitu: I think `sdl1` showed up if `depends` were not skipped.

scriptmodules/supplementary/sdl12-compat.sh
cmitu: No, this is needed just to create a symlink in `$md_id` , `libSDL-1.2.so` -> `libSDL-1.2.so.1.2.68`. We don't rely on the `ld.so` cache for the library to be found.

scriptmodules/ports/xrick.sh
cmitu: I'll check, may be an error.

scriptmodules/emulators/atari800.sh
cmitu: Yes, I'll fix the typo.

scriptmodules/ports/xrick.sh
cmitu: Ok, it was an error. The `data.zip` file from the original game (downloadable from http://www.bigorno.net/xrick/download.html) should be symlinked to `$md_inst`.

I restored the original line that did the symlink.

scriptmodules/supplementary/sdl12-compat.sh
joolswills: Thank you.

scriptmodules/supplementary/backends.sh
joolswills: It shouldn't - The "skip" is fine as we can avoid the extra logic but I would like to reproduce the issue to see what's going on. I just tested and couldn't.

cmitu: Hm, it does show up for me if (from this PR's branch) I remove the line in question. Do you have the `sdl1` module installed ?

joolswills: Ah I didn't use the branch as didn't notice the change to the sdl1 module flags (adding dispmanx to it). Makes sense now.

joolswills:
Thank you.

PR #3846, opened 2024-01-25
lr-dosbox: fix building with GCC 11+

cmitu:
Reported upstream, but not fixed. Workaround is to force C++11 as dialect, suggested in https://github.com/libretro/dosbox-libretro/issues/137


joolswills:
Thank you.

PR #3847, opened 2024-01-26
px68: change upstream project, use SDL2

cmitu:
The original repository has long been abandoned, but several forks are still active. The (most ?) active one has added support for SDL2 (Renderer, Audio, GameController API) and MIDI (Fluidsynth) among other things and it even has some SDL3 support in the works. [1]

This commit switches the source of the emulator to [1] and adjusts the dependencies, build instructions and files included. When building with Fluidsynth, install a minimal soundfont. Generate a minimal configuration when none is found, setting up the ROM/Disk start folder and the menu language.

[1] https://github.com/TurtleBazooka/px68k, based on https://github.com/kenyahiro/px68k, original repository at https://github.com/hissorii/px68k.

Notable changes in this fork.
* Utilizes SDL2 Renderer (GPU), so "sdl-gfx" is unnecessary
* Supports Full-Screen Mode.(F11)
* Soft keyboard (Right click on F12Menu mode)
* SCSI DiskImage support (Can boot from *.HDS)
* Change final screen output to 24bit (RGB565→RGBA8888)
* Support MIDI-Play (Internal/Munt/fluidsynth/USB-MIDI) (NB: didn't find anything in sources about MUNT, I assume is used on Windows as MIDI softsynth)
* Support US-Keyboard layouts. keyconf.dat shoudl be copied into .keropi folder
* Printer output is saved to File
* Can use XBOX like USB-GamePad (hot-pluggable)
* Add support for CyberStick! (DIGITAL/ANALOG mode)
* FileName's UTF8/SJIS automatic detection Japanese display (dedicated table conversion)

NB: not sure of removal of `!mali` is ok - if it's a `sdl2` module now, it should work with SDL2's Mali driver ?

### DIFF
--- a/scriptmodules/emulators/px68k.sh
+++ b/scriptmodules/emulators/px68k.sh
@@ -12,12 +12,15 @@
 rp_module_id="px68k"
 rp_module_desc="SHARP X68000 Emulator"
 rp_module_help="You need to copy a X68000 bios file (iplrom30.dat, iplromco.dat, iplrom.dat, or iplromxv.dat), and the font file (cgrom.dat or cgrom.tmp) to $biosdir/keropi. Use F12 to access the in emulator menu."
-rp_module_repo="git https://github.com/hissorii/px68k.git master"
+rp_module_repo="git https://github.com/TurtleBazooka/px68k.git master"
 rp_module_section="exp"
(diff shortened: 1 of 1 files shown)


joolswills:
Yeah it should work with Mali.

Thank you!

PR #3849, opened 2024-01-27
quasi88: switch upstream repository, upgrade to 0.7.1

cmitu:
Switched the Quasi88 upstream repository to the "Quasi88Kai" fork [1], since it has some new developments and also supports SDL2.

Changes in the fork:

    * General updates to the build system and migration to CMake
    * All files converted to UTF-8
    * Many improvements in Windows, X11, GTK and SDL backends
    * Add a RetroAchievements-compatible version "RAQUASI88" (only for Windows at this time)

Had to apply a build patch since:
  * SDL2 2.0.10 libraries do not provide the SDL:SDL aliases in the `sdl2-config.cmake`
  * Buster's GCC (8) is implementing the filesystem C++17 API in `libstdc++fs`

[1] https://github.com/winterheart/quasi88


joolswills:
Thank you.

PR #3852, opened 2024-01-28
bombermaaan: update to version 2.x and switch to SDL2

cmitu:
Upgraded to the latest version, which uses SDL2 and SDL2-Mixer.

I think we can now safely release the new RP version, since Bombermaaan finally works on Bullseye/Bookworm.


joolswills:
:-) thank you.

PR #3853, opened 2024-01-28
kodi: fix installation on Bookworm/Bullseye

cmitu:
Since Bullseye, `kodi` should be installed from the RPT repos, since it's built with the necessary patches/optimization directly by the RP folks.
However, the version in the repos don't always overrides the version present in the upstream Debian/Raspbian repositories [1] and installation fails.

Added a workaround to always prefer the Kodi packages originating from archive.raspberrypi.com/archive.raspberrypi.org and also fixed the Raspbian detection on Bookworm.

I think Bullseye and previous had 'archive.raspberrypi.org' for RP repositories, while Bookworm has switched to 'archive.raspberrypi.com', so we can't use the URL for pinning. Added a pin based on the 'l'(Location ?) field of the release from the repository.

[1] https://github.com/raspberrypi/bookworm-feedback/issues/144

NB: for Netflix/Amazon/Disney+/etc. (basically DRM-enabled content), `kodi` can use the `libwidevinecdm0` package to enable streaming. Not sure if we want it enabled/added by default or just mention it in the documentation.

### DIFF
--- a/scriptmodules/ports/kodi.sh
+++ b/scriptmodules/ports/kodi.sh
@@ -34,6 +34,14 @@ function depends_kodi() {
                 apt-key del 4096R/BAA567BB >/dev/null
             fi
         fi
+        if [[ "$__os_debian_ver" -gt 10 ]]; then
+            # install Kodi from the RPI repos directly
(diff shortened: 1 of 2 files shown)


joolswills:
Thank you.

PR #3857, opened 2024-01-29
Don't set dispmanx flag on arm64 systems

sworisbreathing:
omxplayer does not have a build available for aarch64 platform. So when building emulationstation from source on arm64 systems we need to skip this dependency, otherwise the dependency installation fails and we can't install emulationstation

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -354,7 +354,7 @@ function get_rpi_video() {
 
     if [[ "$__has_kms" -eq 1 ]]; then
         __platform_flags+=(mesa kms)
-        if [[ -z "$__has_dispmanx" ]]; then
+        if ! isPlatform "aarm64" && [[ -z "$__has_dispmanx" ]]; then
             if [[ "$__chroot" -eq 1 ]]; then
                 # in a chroot default to fkms (supporting dispmanx) when debian is older than 11 (bullseye)
                 [[ "$__os_debian_ver" -lt 11 ]] && __has_dispmanx=1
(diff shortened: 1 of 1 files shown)


joolswills:
Please can you give more details of your set up / os as I wouldn't have thought the dispmanx flag would be set.

sworisbreathing:
I'm on a Raspberry Pi 3B using 64-bit Raspberry Pi OS Lite Buster

cmitu:
There is no 'Raspi OS Lite Buster' with 64bit support, the first Raspi OS with 64bit support was RaspiOS Bullseye in [Feb 2022](https://www.raspberrypi.com/news/raspberry-pi-os-64-bit/). As @joolswills said, `dispmanx` flag should not be set for 64bit systems, since this API (`dismpanx`) is not available at all for this architecture.

How did you get a Buster 64bit system ? Which image did you use for installation ?

sworisbreathing:
@cmitu I seem to have [found one](https://downloads.raspberrypi.org/raspios_lite_arm64/images/raspios_lite_arm64-2021-05-28/)

```bash
uname -a
Linux retropie 5.10.103-v8+ #1529 SMP PREEMPT Tue Mar 8 12:26:46 GMT 2022 aarch64 GNU/Linux
```

cmitu:
That's the [beta](https://forums.raspberrypi.com/viewtopic.php?f=63&t=275372) version which was trialed by the RPI folks. The first supported version was based on Bullseye and doesn't use the `firmware-kms` overlay. 

Technically it's possible to enable `fkms` on 64 bit even with Bookworm, but it's not going to be supported by the RP folks and I don't think we need to. Even so, the correct patch would be to disable `dispmanx` for 64bit in `system.sh` and not patch individually each module.

sworisbreathing:
oops... my bad.

I've updated the MR to skip setting the dispmanx flag on arm64 systems, and removed the offending commit

cmitu:
@sworisbreathing thanks for the changes - they look fine. MR needs also to be re-titled, though.

sworisbreathing:
Thanks for the feedback. I've updated the title.

PR #3858, opened 2024-01-29
bombermaan: fix build for Buster/Bullseye

cmitu:
Added a patch for CMake to fix generating/building Bombermaaan with older SDL2/SDL2_mixer:

 - on Bullseye/Ubuntu 20.4, `sdl2_mixer-config.cmake` doesn't contain the target aliases in the `SDL2_mixer` namespace.
 - on Buster, `sdl2-config.cmake` doesn't contain the target aliases in the `SDL2` namespace and the includes/libraries are not correctly added to the compilation and linkage options.

### DIFF
--- a/scriptmodules/ports/bombermaaan.sh
+++ b/scriptmodules/ports/bombermaaan.sh
@@ -22,6 +22,7 @@ function depends_bombermaaan() {
 
 function sources_bombermaaan() {
     gitPullOrClone
+    applyPatch "$md_data/01-cmake-sdl-mixer.diff"
 }
 
 function build_bombermaaan() {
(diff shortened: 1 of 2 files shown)


joolswills:
Thank you.

bjaraujo:
Hi, I have added this to bombermaaan master so you shouldn't need this anymore:
https://github.com/bjaraujo/Bombermaaan/commit/e24bb325eb30a51f0ccacf520ba9191bff0b8275

joolswills:
> Hi, I have added this to bombermaaan master so you shouldn't need this anymore: [bjaraujo/Bombermaaan@e24bb32](https://github.com/bjaraujo/Bombermaaan/commit/e24bb325eb30a51f0ccacf520ba9191bff0b8275)

Thank you! I have reverted this change.

PR #3860, opened 2024-01-29
New port: OpenLara

dborel:
Includes 3MB demo data. Enhanced water and lighting with OpenGL 3.1 and SDL. Tested on RPI4 with x11 and kmsdrm. For more information, see https://docs.libretro.com/library/openlara .


Exarkuniv:
you might want to open a PR over at https://github.com/Exarkuniv/RetroPie-Extra. so that way it can get in somewhere, where people will be able to use your script. for the time being 

right now with the PI5 and bookworm, they have their hand fill.

PR #3862, opened 2024-02-03
ports: add GZDoom as a Doom source port

cmitu:
Add back GZDoom, the feature centric port for all Doom engine games, based on ZDoom. Since LZDoom is no longer developed, let's try to add back the few still active ZDoom projects.

Since it was removed, besides countless other features addded, it has gained a GLES2/OpenGL2 + Vulkan renderer. Unfortunately, 32bit support has been dropped entirely since 4.9, so 32bit platforms will have to use the 4.8.0 release.

Recent versions auto-detect the available video renderers and can switch to the less intesive GLES2/GL2 renderer automatically, but for old versions (32bit) the preferred renderer has to be set via `vid_preferbackend`.

### DIFF
--- a/scriptmodules/ports/gzdoom.sh
+++ b/scriptmodules/ports/gzdoom.sh
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


joolswills:
Cheers!

Exarkuniv:
i believe it was integrated before i made the script, but i could never get it to work. so i just resorted to useing the older fork and was going to come back at some point.

thanks for poining out the new prefix. i will definitely look into that

PR #3865, opened 2024-02-10
Deployment of Skyscraper 3.10.0 bash completion file.

Gemba:
I just released Skyscraper 3.10.0 which contains a bash completion file. 

This change deploys the completion file during installation.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -66,6 +66,7 @@ function _config_files_skyscraper() {
         'peas.json'
         'platforms_idmap.csv'
         'resources'
+        'supplementary/bash-completion/Skyscraper.bash'
         'screenscraper_platforms.json'
         'tgdb_developers.json'
         'tgdb_genres.json'
@@ -217,13 +218,12 @@ function configure_skyscraper() {
 }
 
 function _init_config_skyscraper() {
-
     local config_files=($(_config_files_skyscraper))
 
     # assume new(er) install
     mkdir -p .pristine_cfgs
     for cf in "${config_files[@]}"; do
-        bn=${cf#*/} # cut off cache/
+        bn=${cf##*/} # cut off all folders
         if [[ -e "$md_inst/$bn" ]]; then
             cp -rf "$md_inst/$bn" ".pristine_cfgs/"
             rm -rf "$md_inst/$bn"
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/skyscraper.sh
```diff
@@ -275,6 +275,12 @@ function _init_config_skyscraper() {
     # Create the cache folder and add the sample 'priorities.xml' file to it
     mkUserDir "$scraper_conf_dir/cache"
     cp -f "$md_inst/.pristine_cfgs/priorities.xml.example" "$scraper_conf_dir/cache"
+
+    # Deploy programmable completion script
+    mkUserDir "$home/.bash_completion.d"
+    cp -f "$md_inst/.pristine_cfgs/Skyscraper.bash" "$home/.bash_completion.d"
+    chown $user:$user "$home/.bash_completion.d/Skyscraper.bash"
+    chmod 644 "$home/.bash_completion.d/Skyscraper.bash"
```
cmitu, line 283: The file is already `0644`, there's no need for the `chmod` here.

scriptmodules/supplementary/skyscraper.sh
```diff
@@ -217,13 +218,12 @@ function configure_skyscraper() {
...
     local config_files=($(_config_files_skyscraper))
 
     # assume new(er) install
     mkdir -p .pristine_cfgs
     for cf in "${config_files[@]}"; do
-        bn=${cf#*/} # cut off cache/
+        bn=${cf##*/} # cut off all folders
         if [[ -e "$md_inst/$bn" ]]; then
             cp -rf "$md_inst/$bn" ".pristine_cfgs/"
             rm -rf "$md_inst/$bn"
```
cmitu, line 229: Why are these files removed from `$md_inst` ? The `configure` step would not be able to re-run incase some files need to be re-installed.

scriptmodules/supplementary/skyscraper.sh
Gemba: Right. Cleared the line.

scriptmodules/supplementary/skyscraper.sh
Gemba: After a `clean` there is a `configure` still successful. The configs are stashed in `.pristine-cfg` as it can be confusing to have the same config files at `/opt/retropie/supplementary/skyscraper` and `/opt/retropie/configs/all/skyscraper`.

cmitu: The `config` folder is user writeable - if the user erases their config, there will be no configuration files there.
I don't think there's any confusion here - the default configs can be customized, so having per-user versions of the same files present in the install folder is normal. Lots of programs deliver a system wide configuration (usually in `/etc`), which the user can then override on their own (i.e. EmulationStation).

Gemba: Yes, that's exactly my point: If (default) config files are kept separate (e.g. /etc) from the binaries I can deduce they might be user configurable files too. But mixing the binary with config files in hte same diretory is a bad idea, as endusers will poke around with sudo to edit that files alongside to the binary and then scratching their head why changes are not applied. It is a safety measure. Because of the large number of config files of Skyscraper I have put them separate.

cmitu: There hasn't been any confusion so far with Skyscraper or any other application included in RetroPie - the `config` folder is the preferrred and 'official' way to edit configuration files and this is reflected in any documentation we have. Even for applications that insist on using the app folder to store their assets/configuration, we symlink them to `config` in order to be accessible (via file shares) and editable (i.e. `amiberry` or `drastic`).

However, this is maybe not the time to change this. We'll accept this as is and we'll re-visit later on the behavior.

Thank you for the changes!

PR #3868, opened 2024-02-16
dolphin: updates to build and configuration

cmitu:
Modified the module to update the dependencies and improve the initial configuration

* Build changes:
   -  build dependencies modified to include SDL2. Latest version (as of Jan 2024) enable SDL2 by default on Linux and use it for input auto-configuration. Gamepads detected using SDL's GameController API are automatically configred based on their mappings.
   - disable analytics and update support durign the build. Updates are handled by RetroPie-Setup and - unfortunately - analytics require a keyboard/mouse to get over the initial agreement pop-up that is presented to the user on the 1st run.

* Configuration updates:
   - on KMS systems, start Dolphin via X11 by using the `XINIT-WM` launc prefix, which instructs `runcommand` to start a minimal desktop env.
   - change the location of the configuration folder, since now Dolphin follows the XDG specs and uses `$XDG_CONFIG_HOME` to store its settings. It will not affect existing installations, since we symlink-it to the configuration folder anyway and Dolphin will look in the previous location first.
   - pre-set a few initial configuration options in the main `Dolphin.ini` configuration file, upating existing options
   - add a new pre-set for the graphics backend configuration file in `GFX.ini` to switch to GLES3 on platforms where it is available. It should help the Pi5 but also other SBCs we support (i.e. OrangePi5 / RK3399) to startw the correct rendering backend.
   - set the '-gui' variant as default. Since the 'nogui' doesn't seem to read the configuration files correctly (i.e. no Fullscreen or Hotkeys are working), use the main program started with '--batch' to run a game

### DIFF
--- a/scriptmodules/emulators/dolphin.sh
+++ b/scriptmodules/emulators/dolphin.sh
@@ -48,6 +48,12 @@ function depends_dolphin() {
     else
         depends+=(qtbase5-private-dev)
     fi
+    # on KMS use x11 to start the emulator
+    isPlatform "kms" && depends+=(xorg matchbox-window-manager)
+
+    # if using the latest version, add SDL2 as dependency, since it's mandatory
+    [[ "$(_get_commit_dolphin)" == "" ]] && depends+=(libsdl2-dev)
+
     getDepends "${depends[@]}"
 }
 
@@ -58,8 +64,8 @@ function sources_dolphin() {
 function build_dolphin() {
     mkdir build
     cd build
-    # use the bundled 'speexdsp' libs, distro versions before 1.2.1 produce a 'cmake' error
-    cmake .. -DBUNDLE_SPEEX=ON -DCMAKE_INSTALL_PREFIX="$md_inst"
+    # use the bundled 'speexdsp' libs, distro versions before 1.2.1 trigger a 'cmake' error
+    cmake .. -DBUNDLE_SPEEX=ON -DENABLE_AUTOUPDATE=OFF -DENABLE_ANALYTICS=OFF  -DUSE_DISCORD_PRESENCE=OFF -DCMAKE_INSTALL_PREFIX="$md_inst"
     make clean
     make
     md_ret_require="$md_build/build/Binaries/dolphin-emu"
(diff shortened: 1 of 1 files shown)


cmitu:
It might break the additions in #3785, but I'll update any of the PRs if a conflict comes up if either of them is merged.
Note that Vulkan is not currently working on Pi4/Pi5, even under X11.

retropieuser:
Maybe I should open this as a separate issue but using dolphin-gui on a Pi 5, this error message box appears, if I click ok it loops, if I ignore the screen becomes garbled as below. It's similar to the garbled image that was affecting SDL2 on ppsspp, cannonball etc, so perhaps it's a similar issue?


PXL_20240216_131844837

PXL_20240216_131813702

This is the text from the bad_ps_OGL_0.txt file

`#version 310 es

#define FORCE_EARLY_Z layout(early_fragment_tests) in



#define ATTRIBUTE_LOCATION(x)
#define FRAGMENT_OUTPUT_LOCATION(x)
#define FRAGMENT_OUTPUT_LOCATION_INDEXED(x, y)
#define UBO_BINDING(packing, x) layout(packing, binding = x)
#define SAMPLER_BINDING(x) layout(binding = x)
#define TEXEL_BUFFER_BINDING(x) layout(binding = x)
#define SSBO_BINDING(x) layout(std430, binding = x)
#define IMAGE_BINDING(format, x) layout(format, binding = x)

#define VARYING_LOCATION(x)




#extension GL_OES_geometry_point_size : enable


`

cmitu:
@retropieuser I'm not familiar with that error and I don't think is related to this PR. The `dolphin-gui` may trigger this problem, but the PR doesn't patch anything in the emulator itself and this may be caused by recent changes in the upstream repository.

cmitu:
I'll set it to draft to check the fullscreen settings again.

retropieuser:
I've fixed that garbled screen error btw, turning off panic handlers in dolphin.ini Interface works for me:-

[Interface]
UsePanicHandlers = False

joolswills:
thank you.

PR #3872, opened 2024-02-24
skyscraper - change location of deployed bash completion file

s1eve-mcdichae1:
Ref: https://serverfault.com/a/1013395

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -48,7 +48,6 @@ function install_skyscraper() {
     md_ret_files+=("${config_files[@]}")
 }
 
-
 function _config_files_skyscraper() {
     local config_files=(
         'aliasMap.csv'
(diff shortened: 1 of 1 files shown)


cmitu:
How about copying it directly into `/etc/bash_completion.d`, it's easier to manage and we get the same result.

joolswills:
Thanks. If someone updates they will be left with the old file so maybe a good idea to delete the old location during install.

cmitu:
> Yes this works as well. I suppose we don't need to `chown` the file here. 

 No.

> Do we ever need to make the directory first or will that always exist?

 In theory, yes, but in practice without `bash-completion` installed (which creates the folder) it won't work anyway. Let's not bother with the creation, but you can add a check to skip the completion if the folder is missing.

s1eve-mcdichae1:
> Thanks. If someone updates they will be left with the old file so maybe a good idea to delete the old location during install.

Do you want to remove the directory too, or just the file?

cmitu:
> Do you want to remove the directory too, or just the file?

I think we want just the file.

joolswills:
Thanks for this - however I decided to merge https://github.com/RetroPie/RetroPie-Setup/pull/3875 which includes this but also has another change.

PR #3874, opened 2024-02-25
mame: modify build instructions and dependencies

cmitu:
Made a few tweaks to optimize bulding MAME:

 - bump the RAM requirements, to ensure there's enough memory to complete the build.
 - don't build the MAME debugger on non-desktop platforms, since it would need a windowing environment to run anyway. This allows us to remove the QT dependencies and debugger compilation steps.
 - use system's flac/zstd libraries (more could be added)
 - tell the linker to remove debugging info, so the binaries don't need a 'strip' call afterwards
 - synchronize the optimization level in MAME with RetroPie's optimization flags (-O2), to avoid extra optimization during compiling (MAME defaults to '-O3').

### DIFF
--- a/scriptmodules/emulators/mame.sh
+++ b/scriptmodules/emulators/mame.sh
@@ -23,13 +23,17 @@ function _get_branch_mame() {
 
 function depends_mame() {
     if [[ "$__gcc_version" -lt 7 ]]; then
-        md_ret_errors+=("Sorry, you need an OS with gcc 7 or newer to compile $md_id")
+        md_ret_errors+=("Sorry, you need an OS with gcc 7.2 or newer to compile $md_id")
         return 1
     fi
 
     # Install required libraries required for compilation and running
     # Note: libxi-dev is required as of v0.210, because of flag changes for XInput
-    getDepends libfontconfig1-dev qtbase5-dev qtchooser qt5-qmake qtbase5-dev-tools libsdl2-ttf-dev libxinerama-dev libxi-dev libpulse-dev
+    local depends=(libfontconfig1-dev libsdl2-ttf-dev libflac-dev libxinerama-dev libxi-dev libpulse-dev libzstd-dev)
+    # build the MAME debugger only on X11 (desktop) platforms
+    isPlatform "x11" && depends+=(qtbase5-dev qtchooser qt5-qmake qtbase5-dev-tools)
+
+    getDepends "${depends[@]}"
 }
 
 function sources_mame() {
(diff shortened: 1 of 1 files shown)


dankcushions:
one thing I never got around to playing with, but it seemed mame's performance via x is (was?) way better than direct to framebuffer (at least on pi4). i am interested to see if adjusting the script to launch it in an xsession improves anything. not a request, just throwing that out there ;)

cmitu:
@dankcushions just modify `emulators.cfg` for the `mame`/`arcade` system(s) and add an entry with:
```
mame-x11 = XINIT-WM:/opt/retropie/emulators/mame/mame %BASENAME%
```
then you can choose to launch MAME via X11 from the _runcommand_ launch menu. Make sure you're running a recent version of RetroPie and install the `xorg matchbox-window-manager` packages before.

roslof:
I updated the script (synced to 46063658) and tried to build lr-mame.
Received this new breaking error near the end of the build:

```
In file included from ../../../../../src/frontend/mame/ui/pluginopt.h:15,
                 from ../../../../../src/frontend/mame/luaengine.cpp:16:
../../../../../src/frontend/mame/ui/menu.h: In member function ‘std::pair<float, float> ui::menu::get_text_dimensions(std::string_view, float, float, float, ui::text_layout::text_justify, ui::text_layout::word_wrapping)’:
../../../../../src/frontend/mame/ui/menu.h:244:9: note: parameter passing for argument of type ‘std::pair<float, float>’ when C++17 is enabled changed to match C++14 in GCC 10.1
  244 |         {
      |         ^
{standard input}: Assembler messages:
{standard input}:276601: Warning: end of file not at end of a line; newline inserted
{standard input}:276602: Warning: zero assumed for missing expression
{standard input}:276188: Error: invalid operands (*UND* and .gcc_except_table sections) for `-'
g++: fatal error: Killed signal terminated program cc1plus
compilation terminated.
make[2]: *** [emu.make:2066: ../../../../libretro/obj/libretro/src/emu/emumem_aspace.o] Error 1
```

Hope this is helpful

cmitu:
> Hope this is helpful

Not really, since this PR is un-related to `lr-mame`/`lr-mess`.

roslof:
> Not really, since this PR is un-related to `lr-mame`/`lr-mess`.

Ugh, yes. I was completely unhelpful. My apologies for not paying attention to the actual affected script. Was in-part confused because lr-mame also requires additional swap to build on 4GB Pis.

PR #3875, opened 2024-02-25
Put programmable completion file to /etc/bash_comletion.d/

Gemba:
cf. https://retropie.org.uk/forum/post/295194

Also the fn remove_skyscraper() was present two times. Merged them together and added a prompt if skyscraper's cache should also be cleared.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -77,7 +77,24 @@ function _config_files_skyscraper() {
 }
 
 function remove_skyscraper() {
+    local ret
+    local msg
+
+    # Prompt for cache clearing
+    msg="\nRemove also Skyscraper's cache data for all platforms:"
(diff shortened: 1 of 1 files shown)


Gemba:
Haha. Just noticed #3872

Anyhow, I also adressed another issue.

joolswills:
@Gemba Sorry - I missed this when I merged. The remove function can't be interactive as it breaks my automatic binary builds (and any other automated scripts that use the commandline interface)

Instead of asking about purging, maybe don't remove the cache folder, and instead set `_INGMSGS+=("Some message")` which will show in a dialog after removing from RetroPie-Setup. This could instruct the user that it's left the cache files and they can remove manually or something ? Or just remove the cache files and inform the user.

Gemba:
Sure thing, no worries. I filed a PR.

I leave it up to the user if the cache should be zapped manually.

PR #3876, opened 2024-02-26
retropie-manager: removal

cmitu:
It was abandoned by the owner and it's already bitrotten by not being compatible with Python3. Judging by the number of (missing) topics in the forums, I don't think it's used that much.

We can still keep the repository at retropie/RetroPie-Manager, though it's probably better suited to archive the repo.


joolswills:
Thank you.

PR #3881, opened 2024-03-07
pegasus: add new module and small updates

cmitu:
Since the static binaries provided upstream are no longer working on newer RaspiOS or on 32bit ARM, create a standard scriptmodule to build Pegasus with `pegasus-fe-dev`. The new scriptmodule is meant to work similarly to the `es`/`es-dev` pair, only one can work at a time and they share a set of common functions. Note that due to the Qt5 version needed, the new scriptmodule is meant to be used on Debian 11/Ubuntu 22.04 and later.

Modified the original scriptmodule and added a theme download function, using @mmatyas's themes. More can be found at [2].

A 2nd commit adds `pegasus-fe` as an auto-start item, similar to Kodi/ES, while fixing the spelling of 'EmulationStation'.

Notes:
 * the `pegasus-fe` launcher doesn't need the previous TTY set-up, it's now handled by `runcommand`.
 * sometimes the `/dev/dri/card0` is not the right card node since the order of the `cardX` inodes is not guaranteed.
   Find the right card node dynamically (see [1]) and tell Qt about it with a minimal config file used in the launcher

[1] https://forums.raspberrypi.com/viewtopic.php?t=351263
[2] https://github.com/mmatyas/pegasus-theme-gallery-db/blob/master/repos.txt

Closes #3586.

### DIFF
--- a/scriptmodules/supplementary/autostart.sh
+++ b/scriptmodules/supplementary/autostart.sh
@@ -10,7 +10,7 @@
 #
 
 rp_module_id="autostart"
-rp_module_desc="Auto-start Emulation Station / Pegasus / Kodi on boot"
+rp_module_desc="Auto-start EmulationStation / Pegasus / Kodi on boot"
 rp_module_section="config"
 
 function _update_hook_autostart() {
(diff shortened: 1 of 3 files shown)


scriptmodules/supplementary/autostart.sh
```diff
@@ -102,20 +105,29 @@ function remove_autostart() {
...
                 x11_autostart=0
             fi
         else
             options=(
-                1 "Start Emulation Station at boot"
-                2 "Start Kodi at boot (exit for Emulation Station)"
+                1 "Start EmulationStation at boot"
+            )
+            [    $has_kodi == 1 ] && options+=(2 "Start Kodi at boot (exit starts EmulationStation)")
+            [ $has_pegasus == 1 ] && options+=(3 "Start Pegasus at boot")
```
joolswills, line 129: Please use double `[[` (I don't think the spacing is necessary - we don't do that elsewhere).

joolswills (commented):
Thanks.

I would have done the "Emulation Station" -> EmulationStation wording in a separate commit as it's not related to the addition of the pegasus changes.

scriptmodules/supplementary/autostart.sh
joolswills: If these variables are 0/1 I would also use `-eq`

cmitu: OK, used `-eq` since it's consistent with the rest of the module.

cmitu:
@joolswills I added the rename for ES in a separate commit.

scriptmodules/supplementary/autostart.sh
```diff
@@ -102,20 +105,29 @@ function remove_autostart() {
...
             fi
         else
             options=(
-                1 "Start Emulation Station at boot"
-                2 "Start Kodi at boot (exit for Emulation Station)"
+                1 "Start EmulationStation at boot"
+            )
+            [[ "$has_kodi" -eq 1 ] && options+=(2 "Start Kodi at boot (exit starts EmulationStation)")
+            [] "$has_pegasus" -eq 1 ] && options+=(3 "Start Pegasus at boot")
+            options+=(
```
joolswills, line 130: You have single `]` and also `[]` now.

cmitu: Sorry, they should be fixed now.

PR #3884, opened 2024-03-11
Replace dialog with info message for headless operation

Gemba:
As mentioned here: https://github.com/RetroPie/RetroPie-Setup/pull/3875#issuecomment-1989100380


cmitu:
I have 2 notes on this change:

- the message can be simplified. Just inform the user that the cache has not been erased and they can erase it by just removing the cache folder or running the `skyscraper`'s purge function.
- don't rename the `_purge` function, it's used internally elsewhere.

Gemba:
Thanks for the hints. Trimmed the PR accordingly.

PR #3885, opened 2024-03-12
lr-mame/lr-mess: modify build parameters

cmitu:
Update the build parameters so:

  - optimization level matches RetroPie's compiler options (-O2)
  - removed `PYTHON` pre-set since it's set now to `python3` by default
  - disable PortAudio, not needed since audio is handled by RetroArch


joolswills:
Thank you.

PR #3886, opened 2024-03-17
Add multi disc functionality by default

retropieuser:
Mentioned here (https://retropie.org.uk/forum/topic/35203/dolphin-gamecube-wiiware-wii-compatibility-on-pi-5/45?_=1710677064715), am I right in thinking the path will be set as /home/ username /RetroPie/roms/gc ?

### DIFF
--- a/scriptmodules/emulators/dolphin.sh
+++ b/scriptmodules/emulators/dolphin.sh
@@ -105,6 +105,10 @@ RenderToMain = True
 KeepWindowOnTop = True
 [Interface]
 ConfirmStop = False
+[General]
+ISOPath0 = "$home/RetroPie/roms/gc"
+[Core]
+AutoDiscChange = True
 _EOF_
     fi
     # use the GLES(3) render path on platforms where it's available
--- a/scriptmodules/emulators/dolphin.sh
+++ b/scriptmodules/emulators/dolphin.sh
@@ -107,6 +107,7 @@ KeepWindowOnTop = True
 ConfirmStop = False
 [General]
 ISOPath0 = "$home/RetroPie/roms/gc"
+ISOPath1 = "$home/RetroPie/roms/wii"
 [Core]
 AutoDiscChange = True
 _EOF_


cmitu:
Looks ok. A couple of notes/questions:
 -  the `wii` path should also be added. Isn't disc changing supported also on the Wii ?
 - isn't `.m3u` (playlist) support a requirement for the disc change to happen automatically ?

retropieuser:
Ah you're right, need to add m3u to platforms.cfg, let me just figure out how to add that to this pull request, rather than making a new one.

With the Wii, it would make sense, except I don't think any multi disc games existed on the Wii (I don't remember it being a thing on the platform back in the day). I had a google and couldn't find any examples of games that needed two Wii discs

aioanx:
If it is possible, can we add and `dolphin-editor = "XINIT-WMC:/opt/retropie/emulators/dolphin/bin/dolphin-emu" `that is adding the ability to open Dolphin GUI ? 
With this way, we can configure all parameters of the emulator and configure the gamepad easily. Thanks.

cmitu:
> Ah you're right, need to add m3u to platforms.cfg, let me just figure out how to add that to this pull request, rather than making a new one.

I'd prefer to have it separate.

> With the Wii, it would make sense, except I don't think any multi disc games existed on the Wii (I don't remember it being a thing on the platform back in the day). I had a google and couldn't find any examples of games that needed two Wii discs

I think they weren't so frequent they could be only a handful that needed a disc change (some of them had additional content which was not directly related to gameplay). Let's just add it for good measure.

cmitu:
> With this way, we can configure all parameters of the emulator and configure the gamepad easily. Thanks.

This is not a forum, please don't use it as such.

retropieuser:
> > Ah you're right, need to add m3u to platforms.cfg, let me just figure out how to add that to this pull request, rather than making a new one.
> 
> I'd prefer to have it separate.
> 
> > With the Wii, it would make sense, except I don't think any multi disc games existed on the Wii (I don't remember it being a thing on the platform back in the day). I had a google and couldn't find any examples of games that needed two Wii discs
> 
> I think they weren't so frequent they could be only a handful that needed a disc change (some of them had additional content which was not directly related to gameplay). Let's just add it for good measure.

Perfect, I created a separate pull request for platforms

Makes sense, have added the Wii path for good measure, there's probably is some obscure game on the console that was 2 discs

cmitu:
@retropieuser thank you for the changes.

Issue #3379, opened 2021-07-26
xpad triggers_to_buttons=1 patch breaks SDL gamepad mappings


cgutman:
Hello, I am the upstream maintainer of [Moonlight Embedded](https://github.com/moonlight-stream/moonlight-embedded). I am working on tracking down a user-reported issue with Xbox 360/One gamepads being mapped incorrectly while running inside RetroPie (and working perfectly fine when launched outside of RetroPie).

I tracked down the cause to the custom xpad which unconditionally maps triggers to button rather than analog axes:
https://github.com/RetroPie/RetroPie-Setup/blob/6d396a3d5b70417869eddcb6d3d74324d5fce689/scriptmodules/supplementary/xpad.sh#L49-L51

https://github.com/RetroPie/RetroPie-Setup/blob/6d396a3d5b70417869eddcb6d3d74324d5fce689/scriptmodules/supplementary/xpad/01_enable_leds_and_trigmapping.diff#L12-L27

The `triggers_to_buttons` option was never meant to be used in this way, exactly because it can break unaware userspace applications. Upstream xpad only allows it to be used for unknown gamepads (and even for that it is not recommended, because many applications and libraries like SDL have heuristic mappings for things resembling Xbox controllers that `triggers_to_buttons=1` will break).

This screws up any SDL gamepad mappings used by applications (like Moonlight) running inside RetroPie, because the SDL gamepad mapping GUID is identical to the version where triggers are mapped as axes. As a result of the change in number of axes and buttons, the mapping becomes completely wrong (not just the triggers).

Additionally, it prevents the use of the triggers as analog inputs for things like racing games where that functionality is important.

Analog triggers are the norm for all modern gamepads (Xbox 360/One/Series X, [DualShock 4](https://github.com/torvalds/linux/blob/master/drivers/hid/hid-sony.c#L424-L425), [DualSense](https://github.com/torvalds/linux/blob/ff1176468d368232b684f75e82563369208bc371/drivers/hid/hid-playstation.c#L926-L927)), and hid-sony/hid-playstation don't allow you to opt out of it.

Can we nuke this `triggers_to_buttons=1` change?

cmitu:
Thank you for the comprehensive write-up and the debugging done to diagnose the problem.

Looking at the history of the RetroPie's included `xpad` driver, looks like the reason for the patch was EmulationStation (the gaming front-end used by RetroPie) didn't properly support the trigger axis on the Xbox gamepads ([1]).

[...]

PR #3891, opened 2024-03-22
xpad: don't force the triggers-as-buttons option

cmitu:
The `xpad` kernel module is patched locally in order for the 'triggers_to_buttons' option to work on any supported controller. During installation, the option is also enabled automatically. Historically, this has been used to work-around EmulationStation not detecting the analog shoulder triggers, but the detection works now and this extra option is not needed for EmulationStation.

Leaving the option enabled by default has the side effect of breaking SDL's Gamecontroller API DB mappings (no of buttons and axis are different then the upstream mapping). Since more and more applications/games/emulators/etc. are using this SDL sub-system, it's better we don't enable it anymore.

This change removes the module configuration handled by the scriptmodule - both installation and removal. Users can manually add/remove the configuration if they wish - I will update the docs after this and maybe the RP version can be bumped to have a milestone for this change.

Closes: https://github.com/RetroPie/RetroPie-Setup/issues/3379

### DIFF
--- a/scriptmodules/supplementary/xpad.sh
+++ b/scriptmodules/supplementary/xpad.sh
@@ -42,14 +42,10 @@ function build_xpad() {
 
 function remove_xpad() {
     dkmsManager remove xpad "$(_version_xpad)"
-    rm -f /etc/modprobe.d/xpad.conf
 }
 
 function configure_xpad() {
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks.

PR #3893, opened 2024-03-27
bluetooth: fix PS3 pairing with recent BlueZ

cmitu:
Due to changes in BlueZ added to fix CVE-2023-45866, the PS3 controllers won't pair/work anymore with BlueZ. Since the path consist in only one change to default option (ClassicBondedOnly default changed, see [1]), it's been quickly added by all distros [2], [3]. This has already been reported in the forums and fixed (for Buster) by downgrading the `bluez` package. The same fix cannot be applied to current distros, so we can switch back the option to the way it works with PS3 controllers.

NOTE:
 * while technically this makes BlueZ vulnerable to CVE-2023-45866, the exploit mentioned works IIF BlueZ is set to 'discoverable' mode. However, this mode is set only during discoveries, which in RetroPie means just the pairing dialog - I think the risk for a real break-in through the vulnerability described is very low. See [4] for an explanation of the conditions needed to exploit it on BlueZ and a PoC (which didn't work for me, despite having the vulnerable config in place).
 * the configuration is set-up only when a PS3 pairing attempt is made.
 * when removing a device, the vulnerable configuration will also be removed if no more PS3 paired devices are left.

[1] https://git.kernel.org/pub/scm/bluetooth/bluez.git/commit/profiles/input?id=25a471a83e02e1effb15d5a488b3f0085eaeb675
[2] https://ubuntu.com/security/CVE-2023-45866
[3] https://security-tracker.debian.org/tracker/CVE-2023-45866
[4] https://github.com/marcnewlin/hi_my_name_is_keyboard?tab=readme-ov-file#linux-keystroke-injection

### DIFF
--- a/scriptmodules/supplementary/bluetooth.sh
+++ b/scriptmodules/supplementary/bluetooth.sh
@@ -103,6 +103,11 @@ function list_available_bluetooth() {
 
     # sixaxis: add USB pairing information
     [[ -n "$(lsmod | grep hid_sony)" ]] && info_text="Searching ...\n\nDualShock registration: while this text is visible, unplug the controller, press the PS/SHARE button, and then replug the controller."
+    # sixaxis: configure workaround for PS3 when we detect one
+    if cat /proc/bus/input/devices | grep Name | grep -qE "(Sony )?PLAY.*3" ; then
+        _bonded_fix_bluetooth "add"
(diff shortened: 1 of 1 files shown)


cmitu:
Forgot to mention that the workaround/configuration has to be in-place even after pairing, otherwise re-connecting the controller doesn't work.

joolswills:
Thank you.

PR #3895, opened 2024-03-30
Add lr-dirksimple setup script

icculus:
This is an setup script to install the [DirkSimple](https://github.com/icculus/DirkSimple) libretro core.

DirkSimple is a laserdisc game player that implements specific titles from scratch, so it uses an arcade game's video content but not its roms. It takes game data in Ogg Theora format, so the "rom" that the core needs is a single .ogv file per game.

Currently, it plays [Dragon's Lair](https://en.wikipedia.org/wiki/Dragon%27s_Lair) and [Cliff Hanger](https://en.wikipedia.org/wiki/Cliff_Hanger_(video_game)), with other titles planned for the future. RetroArch ships DirkSimple as a standard core, including to their Steam builds. It offers a libretro option for playing some laserdisc titles, so you get all the consistent controller support, hotkeys and state serialization magic, in case running Daphne as a standalone emulator is too cumbersome without those features.

This is a draft pull request at the moment, because RetroPie setup scripts are totally new to me, so feedback and criticism is definitely welcome.

This also needs to figure out what to do with EmulationStation, too: should this be a new system, or should this slide in to the Daphne section? etc.

### DIFF
--- a/scriptmodules/libretrocores/lr-dirksimple.sh
+++ b/scriptmodules/libretrocores/lr-dirksimple.sh
@@ -36,6 +36,7 @@ function install_lr-dirksimple() {
     md_ret_files=(
         'build/dirksimple_libretro.so'
         'data'
+        'LICENSE.txt'
     )
 }
 
--- a/scriptmodules/libretrocores/lr-dirksimple.sh
+++ b/scriptmodules/libretrocores/lr-dirksimple.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
+# too numerous to list here. Please refer to the COPYRIGHT.md file distributed with this source.
+#
+# See the LICENSE.md file at the top-level directory of this distribution and
+# at https://raw.githubusercontent.com/RetroPie/RetroPie-Setup/master/LICENSE.md
+#
+
+rp_module_id="lr-dirksimple"
+rp_module_desc="laserdisc emu - DirkSimple"
+rp_module_help="ROM Extensions: .ogv .dirksimple\n\nCopy your laserdisc movies in Ogg Theora format to $romdir/dirksimple"
+rp_module_licence="zlib https://raw.githubusercontent.com/icculus/DirkSimple/main/LICENSE.txt"
+rp_module_repo="git https://github.com/icculus/DirkSimple.git main"
+rp_module_section="exp"
+
+function depends_lr-dirksimple() {
+    local depends=(cmake)
+    getDepends "${depends[@]}"
+}
+
+function sources_lr-dirksimple() {
+    gitPullOrClone
+}
+
+function build_lr-dirksimple() {
+    local params=(
+        '-DDIRKSIMPLE_LIBRETRO=ON'
+        '-DDIRKSIMPLE_SDL=OFF'
+        '-DCMAKE_BUILD_TYPE=Release'
+    )
+
+    isPlatform "neon" && params+=' -DCMAKE_C_FLAGS="-mfpu=neon"'
+
+    rm -fr build && mkdir build
+    cd build
+    echo "cmake" "${params[@]}" ..
+    cmake "${params[@]}" ..
+    make dirksimple_libretro
+    md_ret_require="$md_build/build/dirksimple_libretro.so"
+}
+
+function install_lr-dirksimple() {
+    md_ret_files=(
+        'build/dirksimple_libretro.so'
+    )
+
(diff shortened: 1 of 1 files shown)


scriptmodules/libretrocores/lr-dirksimple.sh
```diff
@@ -0,0 +1,63 @@
...
+
+function install_lr-dirksimple() {
+    md_ret_files=(
+        'build/dirksimple_libretro.so'
+    )
+
+    rm -rf "$biosdir/DirkSimple"
+    mkUserDir "$biosdir/DirkSimple"
+    cp -rf "$md_build/data" "$biosdir/DirkSimple/"
+    chown -R $user:$user "$biosdir/DirkSimple"
```
cmitu, line 53: The copying of the "bios" data should be moved to `configure` and the `data` folder should be declared inside `md_ret_files`.
The reason is that `install` is also run during installation from binaries which we build for the Pi platforms. During a binary installation, the setup doesn't have the sources (there's no `$md_build` at that time) so any other files needed for setting the emulator should be included in the binary package.
Adding 'data' to `$md_ret_files` ensures they'll be included in the binary installation, adding the necessary files from `$md_inst/data` (and not `$md_build/data`) to `$biosdir/DirkSimple` should be done in the `configure_lr-dirksimple` function.

scriptmodules/libretrocores/lr-dirksimple.sh
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function configure_lr-dirksimple() {
+    mkRomDir "dirksimple"
```
cmitu, line 57: Let's use the `daphne` system instead of `dirksimple`.

cmitu (commented):
I've left a few comments in the review, they're mostly minor modifications and should simplify the scriptmodulke.

icculus:
Perfect, I'll fix this up today. Thanks for the feedback!

scriptmodules/libretrocores/lr-dirksimple.sh
icculus: does `defaultRAConfig` become "daphne" too, or does that stay as "dirksimple"?

icculus: (Likewise for `addSystem` and `addEmulator`)

cmitu: All of the them should be changed to `daphne`.

We might change the core to become a _port_ (like [lr-prboom](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-mrboom.sh)), but we'll handle that after the PR is merged.

icculus:
Okay, all the feedback is incorporated as of c3980f305670d66810db3159c2ff141bbb805c4f, so I'm bumping this from a Draft PR to a real pull request!

scriptmodules/libretrocores/lr-dirksimple.sh
```diff
@@ -0,0 +1,54 @@
...
+    cd build
+    cmake -DDIRKSIMPLE_LIBRETRO=ON -DDIRKSIMPLE_SDL=OFF ..
+    make dirksimple_libretro
+    md_ret_require="$md_build/build/dirksimple_libretro.so"
+}
+
+function install_lr-dirksimple() {
+    md_ret_files=(
+        'build/dirksimple_libretro.so'
+        'data'
```
joolswills, line 38: Please can you include the License file also. Or we can add it after if that's easier.

icculus: Added in 9a7d64c1b4f02ef96deec4baa9cc8d99fd0152af

cmitu:
Looks good to merge.
We can add the `.ogv` extension for EmulationStation in `platforms.cfg`. I'll add a few notes in the docs about the new emulator, supported games and how the _Daphne_ `.mpeg` files can be converted.

joolswills:
Sorry for delay. I squashed the commits on merge. Thank you for the contribution :)

icculus:
Not a problem at all; I'm grateful for your attention.  :)

DjDiabolik:
> Looks good to merge.
> We can add the `.ogv` extension for EmulationStation in `platforms.cfg`. I'll add a few notes in the docs about the new emulator, supported games and how the _Daphne_ `.mpeg` files can be converted.

mmmmmm problably at today this it's missing ? I have add a .OGV on my daphne folders. Reboot so many times Emulationstation but the "news" games never appears.
Looking at es_system.cfg in daphne setting apparently the .ogv it's not present.

PR #3897, opened 2024-04-02
Support new zip Singe ROMS

DirtBagXon:


DirtBagXon:
This will allow support for the new zipped Singe ROMs.

This has been used for a few months now without issue, but welcome feedback if you have a preferred alternative.

cmitu:
hello @DirtBagXon, thanks for the changes.

I think you missed escaping the other vars:
``` sh
if [[ -f "$dir/$name.singe" ]]; then
    singerom="$dir/$name.singe"
elif [[ -f "$dir/$name.zip" ]]; then
    singerom="$dir/$name.zip"
fi
```
should probably be:

``` sh
if [[ -f "\$dir/\$name.singe" ]]; then
    singerom="\$dir/\$name.singe"
elif [[ -f "\$dir/$name.zip" ]]; then
    singerom="\$dir/\$name.zip"
fi
```

I also think it's better to use `if [[ -n "$singerom" ]]` instead of `if [[ -v singerom ]]` since the former checks for non-empty content, not only of the existence of the variable.

DirtBagXon:
Thanks @cmitu 

It's been a while since I played with scriptmodules - I forgot how this worked :)

I think that's escaped all the required var $ for the write to file.

cmitu:
Yes, it works fine now.
So, the change is that instead of the `game.singe` folder, the contents can be zipped (as `game.zip`) and placed in the same folder where the `.singe` folder was expected ?

cmitu:
@DirtBagXon ok, thanks for the explanation. My main concern was whether we should add `.zip` as supported extension in the `daphne` system, it's not needed - the `.zip` file goes into `<game>.daphne` , where the `.singe` file resides right now.

PR #3900, opened 2024-04-12
dosbox-staging: pin version for videocore

cmitu:
Latest (0.81) version of Dosbox Staging doesn't go along with the legacy GPU VideoCore drivers, resulting in a SDL window initialization crash. Pin the 0.80.1 version for these devices, maybe a GLES2-only rendering path will be added later on, which would solve this issue.


joolswills:
Cheers!

PR #3901, opened 2024-04-13
platform: updates for amdstrad and daphne

cmitu:
* Amstrad: add more supported extensions
 * Daphne: enable the OGV format supported by DirkSimple


joolswills:
Thank you :-)

PR #3907, opened 2024-04-22
Add a _supported_ID function for modules 

joolswills:
[packages - add a _supported_ID function for modules](https://github.com/RetroPie/RetroPie-Setup/commit/9341be87bdc3642f5d3541f89f0949eb5b70e592)

If a scriptmodule has a _supported_MODULEID() function it will be called when the scriptmodule is added.
If the function returns 1, the module is not supported and will be disabled.

This can be used to do additional OS/GCC version checks.

[Migrate gcc checks in depends_* functions to _supported_*](https://github.com/RetroPie/RetroPie-Setup/commit/7767a0328133f31a7859c4520d64483e82279053)

Migrate the depends_* checks in dolphin, mame, ti99sim, lr-bsnes, lr-flycast-dev and lr-mame to  _supported_* function.

This will disable the modules when the GCC is too old rather than throwing an error when trying to build

[mysticmine - Disable on Debian versions newer than 10 (buster)](https://github.com/RetroPie/RetroPie-Setup/commit/056cc8302d6151d65ba743285e4f38ffab7e47ca)

The scriptmodule requires Python2 and python modules that are no longer packaged.

--

Some other modules could probably benefit from this (eg fs-uae) and maybe some I missed. I was originally going to handle this with some logic/comparisons in the module flags, but I think this is simpler and more flexible.

Needs some testing.


joolswills:
@cmitu I revisited this and decided to redo it using module flags with the format: `not:$_VAR:-lt:7` for excluding based on the comparison. I think I prefer it as a flag but will open another PR for review and comments.

joolswills:
Actually it might be more consistent to use the format `!:$_VAR:-lt:7`.

joolswills:
Closing in favour of #3909

PR #3908, opened 2024-04-23
echo the applied command line options of Skyscraper; add two scrape modules

Gemba:
This goes back to the forum discussion a while ago: Within the scriptmodule there can be options persisted (cf. `Skyscraper.cfg`) which then get applied to the command line, which overrun settings in Skyscraper's `config.ini`. 

To add more visibility about the applied options (and to avoid confusion about different behaviour of Skyscraper from scriptmodule vs. Skyscraper CLI usage) the user gets the actual command echoed before the scraping starts.

Tested it well on my side and did not notice a regression.

Also, for some reasons two already present scraping modules did not made it into the scriptmodule.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -290,17 +290,21 @@ function _init_config_skyscraper() {
     rm -f "$home/.bash_completion.d/Skyscraper.bash"
 }
 
-# Scrape one system, passed as parameter
-function _scrape_skyscraper() {
-    local system="$1"
+# read scriptmodule's skyscraper.cfg and prepare the command line parameters
+function _get_clioptions_skyscraper() {
+    local system=$1
+    local scrape_module=$2
 
-    [[ -z "$system" ]] && return
+    local params=()
+    local flags
 
     iniConfig " = " '"' "$configdir/all/skyscraper.cfg"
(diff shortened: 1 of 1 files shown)


cmitu:
Hi @Gemba, thank you for the PR.

I've taken a look and agree with the changes.
I think though that the implementation of getting the list of parameters can be simplified a bit
 * rename `_prepare_cli_params` to something like `_get_cmdline_skyscraper`, it's more suggestive.
 * return the `$params` from this function directly instead of using a nameref assignment inside the method. We don't use this kind of assigment in RetroPie and it's kind of a 'gotcha' in Bash that not many are aware of -  we prefer things to be simpler. Just return the array of params (as a string) from the cmdline assembly method and print it before usage in `_scrape_skyscraper`.

Gemba:
Thanks for looking into it. I changed it to your suggestions. Works also well on my side.

However, I noticed something during testing that should not happen with an out-of-the-box installation of RP, but can happen when a user creates a platform/system folder with a space in its name. Then the passing of the system name (`-p`) to Skyscraper breaks.

I tried some things, but reverted them, as they were not elegant.
I will leave it aside for now as it should only happen in rare cases, with deliberate user action before.

cmitu:
Yes, seems to work fine, thank you for the changes.

PR #3913, opened 2024-04-30
mame: add GCC requirements for version 0.265

cmitu:
Starting with version 0.265 (April 2024), GCC 10.3 is required for building MAME. Enforce this requirement when building the emulator and libretro cores, Debian Buster will be getting 0.264 as the last version.


joolswills:
Thank you.

PR #3916, opened 2024-05-02
emulationstation: fix OpenGL version check

theofficialgman:
self explanatory. the openGL version check as is only works on MESA drivers. `Extended renderer info` comes from the `GLX_MESA_query_renderer` extension

### DIFF
--- a/scriptmodules/supplementary/emulationstation.sh
+++ b/scriptmodules/supplementary/emulationstation.sh
@@ -173,7 +173,10 @@ function build_emulationstation() {
             compareVersions $gles_ver lt 2.0  && params+=(-DUSE_GLES1=On)
         else
             params+=(-DGL=On)
+            # mesa specific check of OpenGL version
             local gl_ver=$(sudo -u $user glxinfo -B | grep -oP 'Max compat profile version:\s\K.*')
+            # generic fallback check of OpenGL version
+            [[ -z "$gl_ver" ]] && gl_ver=$(sudo -u $user glxinfo -B | grep -oP "OpenGL version string: \K[^ ]+")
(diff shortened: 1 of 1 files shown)


cmitu:
I believe the check can be made simpler with:
``` sh
sudo -u $user glxinfo -B | grep -m1 'OpenGL version string:' | cut -d ' ' -f4
```

joolswills:
Or even `sudo -u $user glxinfo -B | grep -oP "OpenGL version string: \K[^ ]+"`

(Is this string included more than once ? `-m1` is needed if so, but it's not on my system)

scriptmodules/supplementary/emulationstation.sh
```diff
@@ -173,7 +173,10 @@ function build_emulationstation() {
             compareVersions $gles_ver lt 2.0  && params+=(-DUSE_GLES1=On)
         else
             params+=(-DGL=On)
+            # mesa specific check of OpenGL version
             local gl_ver=$(sudo -u $user glxinfo -B | grep -oP 'Max compat profile version:\s\K.*')
+            # generic fallback check of OpenGL version
+            [ -z "$gl_ver" ] && gl_ver=$(sudo -u $user glxinfo -B | grep -m1 'OpenGL version string:' | awk -F: '{print $2}' | awk '{print $1}')
```
joolswills, line 179: In addition to the simplified grep comments, please use double `[[ ]]`

theofficialgman:
> (Is this string included more than once ? `-m1` is needed if so, but it's not on my system)

Its not needed currently for me. However I don't trust glxinfo to not return more output if for example the user has multiple detected GPUs

joolswills:
Thank you.

PR #3917, opened 2024-05-06
Update platforms.cfg pc98 .cmd files

retropieuser:
For certain CD ROM based games (e.g Policenauts), you need to create a .cmd so you can run the system disk, user disk and CD


cmitu:
That's ok, but put the extension at the beginning of the list, so it's sort of alphabetically sorted.

PR #3920, opened 2024-05-13
yabasanshiro: fix building on ARM64

cmitu:
The previous build method on the Pi4 uses a CMake toolchain file which assumes a 32bit ARM system. To make it compile on ARM64, remove the toolchain file usage and add the appropriate defines/options from it to the CMake invocation.

NB: with these changes, we can probably try to install from the main/master branch (which doesn't have the Pi4 CMake toolchain file), but I'll not add it to this PR.


joolswills:
Thank you.

The RPI4 branch also had shader changes (originally from https://github.com/devmiyax/yabause/issues/737)

PR #3924, opened 2024-05-25
mupen64pluus: workaround the GLES GlideN64 regression

cmitu:
Upstream GlideN64 has a regression due to a recent shader related change, which breaks the video output on GLES devices/platforms. Until this issue is fixed in the upstream GlideN64, build the plugin from a commit before the changes that cause the regression.

The problem was reported in the forums [1] and I reported it upstream [2] and though there is a fix for it, it hasn't been added to the main GlidedN64 repository.

[1] https://retropie.org.uk/forum/topic/35444/
[2] https://github.com/gonetz/GLideN64/issues/2837


joolswills:
Thank you!

PR #3926, opened 2024-05-30
Add NAOMI and Atomiswave and new functionality to platforms.cfg

PhillBenoit:
Added new systems.

platforms.cfg can now tag all attributes than can be passed to setESSystem


cmitu:
IMHO we don't need additional systems for Naomi and Atomiwave, they can use the 'arcade' system.

For the other functionality - which should be submitted as a separate change/PR - and you should also explain theuse case. Support for extra/custom user ROM extensions is already implemented (see [here](https://retropie.org.uk/docs/EmulationStation/#platformscfg-edits)). The other modification (extra launch command ?) is not necessary, in RetroPie custom emulators are implemented in `runcommand` and can be chosen via the [runcommand menu](https://retropie.org.uk/docs/Runcommand/#runcommand-launch-menu).

PhillBenoit:
There seems to be a lot of community demand for NAOMI.  Atomiswave is just as easy to add.  This is not like NeoGeo where MAME runs them just fine.  MAME does not run these games at all on the Pi.  Dreamcast and specialty emulators will be in demand for the foreseeable future.  The only reason to even associate these games with arcade is for the purposes of scraping because they conform to the MAME rom set.  This is what a working entry in es_systems.cfg looks like:

```
  <system>
    <name>atomiswave</name>
    <fullname>Sammy Atomiswave Arcade System</fullname>
    <path>/home/pi/RetroPie/roms/atomiswave</path>
    <extension>.zip .ZIP</extension>
    <command>/opt/retropie/supplementary/runcommand/runcommand.sh 0 _SYS_ dreamcast %ROM%</command>
    <platform>arcade</platform>
    <theme>atomiswave</theme>
  </system>
```

I needed to add the additional functionality to make that happen automatically.  My goal is to keep users from having to manually edit XML files in the way you describe.  Especially when it means having to do configurations as obscure as this.  I want people to be able to load up and go as easy as possible.  I don't know what you want.

PhillBenoit:
OK, I was so focused on making my config file look like what works on my machine that I didn't realize what was going on.  I went ahead and took out the changes to helpers.sh.  Once we're done here, I'll bring them back to fix genesis.  For now, I've left the changes to platforms.cfg and the flycast configs.  My primary concern is the user experience.  I want RetroPie to be easy to use.  I want users to be able to drop files into the correct folders and have the system know what to do with them.  It is with that in mind that I would like to see these changes.  The changes are critical so that if people have a problem with their game, they're not trying to load them in a bunch of other emulators that don't work.  This is the same support that NeoGeo gets.

PhillBenoit:
The longer I stare at it, the worse platforms.cfg looks.  I'll make a new request for naomi.  I can add atomiswave when I fix everything else.

PR #3927, opened 2024-05-31
lr-scummvm: build with OpenGL(ES) context support

cmitu:
Upstream has added support for using an OpenGL(ES2) context for 3D games and HW rendering. The core will request a HW GL context from RetroArch when starting. Added the necessary parameters to the build to support the GL context for the platform where RetroArch is running.


joolswills:
Thank you.

PR #3928, opened 2024-06-05
Add NAOMI

PhillBenoit:
Added new system NAOMI.  Keeping Flycast associated with arcade is a bad idea.  Flycast will never open anything but NAOMI (and Atomiswave) subset of arcade games.  In a desperate attempt to just get the system in, I have left it associated with arcade for now in hopes that people who put their NAOMI games in to the appropriate NAOMI folder will never have a problem trying to load an emulator that will never play their game.


joolswills:
There are a couple of unrelated changes in this PR (EOL character from one file and the microsoft VS changes in .gitignore) and there seems to be a number of commits. Please can you squash the commits and create a single commit with just the Naomi changes (with a suitable commit description). If you're unsure how to do that, please create a new branch and re-implement the changes in a new PR.

PhillBenoit:
The changes and new pull request all came at the request of the last person I was dealing with.  There is no way to do what you're requesting on the github web site.  Visual Studio and github desktop both will not allow me to squash the commits.  I am trying to do what you want but there seems to be no way to do it.

retropieuser:
Hi there, I think you can use the squash and merge function on the web browser interface to put these commits into one 👍

retropieuser (approved):

PR #3931, opened 2024-06-07
amiberry: upgrade to 5.7.2; handle dispmanx deprecation

cmitu:
The 5.7.1 version has a couple of regresions (one is a crash), fixed in version 5.7.2. However, DispmanX support was dropped starting with 5.7.2 so just bumping the version would break Buster based Pi installation.

Therefore the module was changed so that:
 - for DispmanX platforms cherry pick the post 5.7.2 fixes manually and lock the version to 5.7.1
 - for every other platform build the 5.7.2 version


joolswills:
Awesome. Thanks.

PR #3938, opened 2024-06-20
Add Amiberry for x86 platforms as well (fixes #3937)

### LINKED ISSUE #3937: Make Amiberry available for x86_64 as well
Is there a reason why Amiberry is not enabled/available for x86_64 platforms a well?
It runs quite well there for a few years now, so perhaps this was missed?

The only difference when compiling from source would be the PLATFORM option (instead of `rpi3`, use `x86-64`) in the Makefile. 
And I have pre-built binaries for releases as well, even though I know you prefer your own binaries. :)

midwan:
Did a quick test and this seems to work, but there might be more changes you want to make.

I've only added `x86-64` as the platform, as there's no 32-bit version for x86 available.
I've also taken away the platform restrictions, but you know best on how to use those (perhaps just add that `x86` is permitted instead).

### DIFF
--- a/scriptmodules/emulators/amiberry.sh
+++ b/scriptmodules/emulators/amiberry.sh
@@ -15,7 +15,7 @@ rp_module_help="ROM Extension: .adf .chd .ipf .lha .zip\n\nCopy your Amiga games
 rp_module_licence="GPL3 https://raw.githubusercontent.com/BlitterStudio/amiberry/master/LICENSE"
 rp_module_repo="git https://github.com/BlitterStudio/amiberry :_get_branch_amiberry"
 rp_module_section="opt"
-rp_module_flags=""
+rp_module_flags="!all arm rpi3 rpi4 rpi5 x86"
 
 function _update_hook_amiberry() {
     local rom
@@ -31,6 +31,8 @@ function _update_hook_amiberry() {
 function _get_branch_amiberry() {
     if isPlatform "dispmanx"; then
         echo "v5.7.1"
+    elif isPlatform "x86"; then
+        echo "v6.3.3"
     else
         echo "v5.7.2"
     fi
--- a/scriptmodules/emulators/amiberry.sh
+++ b/scriptmodules/emulators/amiberry.sh
@@ -15,7 +15,7 @@ rp_module_help="ROM Extension: .adf .chd .ipf .lha .zip\n\nCopy your Amiga games
 rp_module_licence="GPL3 https://raw.githubusercontent.com/BlitterStudio/amiberry/master/LICENSE"
 rp_module_repo="git https://github.com/BlitterStudio/amiberry :_get_branch_amiberry"
 rp_module_section="opt"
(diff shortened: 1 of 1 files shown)


cmitu:
@midwan thanks for the PR.

I have a couple of notes on it, aside from what @joolswills may want to add:

 - don't remove the current `_flags`, just add the new platform (`x86`) to it.
 - you can modify the `_get_branch_amiberry` function and for the `x86` platform add a different branch for building, something like
    ```` shell
    ...
    if isPlatform "dispmanx"; then
         echo "v5.7.1"
    elif isPlatform "x86"; then
         echo "preview"
    else 
         echo "v5.7.2"
    fi
     ...
     ````

cmitu:
> I've modified the PR accordingly.

Thanks, looks fine.

> We should also fix the directories, as it seems there are some outdated things:
> 
>     * There are more directories included in the amiberry repo by default. Most are optional and configurable anyway, but it would be best to exist, even if empty.

 Excluding `plugins`, which is missing, I'm not sure if we're missing much. Looking at the Pi5 release (5.7.3), we're missing _cdrom_, _floppies_,_harddisks_ and _lha_, but these should be located (in RetroPie) in the `amiga` roms folder and we have no use for them in the installation folder.
 
> 
>     * External libraries now live under the new `plugins` directory, so for example `capsimg.so` should be moved there.
> 
>     * The FloppyBridge library is included in the sources and is part of the Makefile/CMake, and also lives under the `plugins` directory, but it's not handled by this script.

 OK, I'll take a look at including the `plugins` folder. 
 Looking at the Pi5 package from https://github.com/BlitterStudio/amiberry/releases/tag/v5.7.3, I see _plugins_ is included, but it only contains the `floppybridge.so` library - is `capsimg.so` missing on purpose there ?

midwan:
Yes, the extra dirs are not needed in RetroPie, and can be configured in the `amiberry.conf` file to point to wherever you want. Their main purpose is the default path used for the file requesters, if you try to load something from within Amiberry's GUI.

The `capsimg.so` is not included in the latest packages, but that was due to an issue with the build pipeline on my end. I'll get that fixed, probably include the sources in the repo while I'm at it (it's currently a submodule). But it _should_ be included inside the `plugins` directory, otherwise you won't be able to load SPS/IPF images.

So essentially, if you can fix the `plugins` directory missing, and optionally update the paths in `amiberry.conf` for the Floppies/CDROMs/HDD/Lha to point to the `roms/amiga` folder, I think we're good.

cmitu:
I think I'll add the game folders configuration and the `plugins` folder inclusion in a separate modification - I haven't had the chance to fully test the changes to implement the former.
If you can squash the commits I think we can merge this PR with the addition of the new (x86)  platform.

midwan:
Sounds good.
Isn't the squashing done at the merging time? Github has an option to Squash and merge in one step.

joolswills:
Thank you.

PR #3940, opened 2024-06-21
Update dolphin.sh to include other settings

retropieuser:
Adds a symlink to .local/share/dolphin-emu into /opt/configs/gc. The local share directory includes items like the memory card data, game specific configs, Wii Nand, Save states etc which users may want easier access to

### DIFF
--- a/scriptmodules/emulators/dolphin.sh
+++ b/scriptmodules/emulators/dolphin.sh
@@ -90,14 +90,14 @@ function configure_dolphin() {
     [[ "$md_mode" == "remove" ]] && return
 
     # Move the other dolphin-emu options, memory card saves etc
-    moveConfigDir "$home/.local/share/dolphin-emu" "$md_conf_root/gc/dolphin-emu"
-    mkUserDir "$md_conf_root/gc/dolphin-emu"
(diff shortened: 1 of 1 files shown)


cmitu:
I think `$HOME/.local/share/dolphin-emu` should be moved under `gc/dolphin-emu/Local`, just like the configuration is under `gc/dolphin-emu/Config`.
Is the 2nd change related or is just a correction ?

retropieuser:
The second change would line the configs all under a dolphin-emu folder inside GC, it's strange, dolphin on linux, it separates out the config folder into it's own location .config/dolphin-emu and leaves the rest of the settings under .local/share. But on Mac I'm looking at it right now and the folders are all under one dolphin-emu:-

Screenshot 2024-06-21 at 12 37 49

cmitu:
@retropieuser that's fine, but it doesn't answer or address my questions.

retropieuser:
It's comestic tbf that's the bit I was driving at above as you end up with a local directory and a config directory when imo it's tidier in one directory inside configs/gc. Anyway I've made the adjustment to only symlink the local directory

cmitu:
Thanks, I'll merge as is.

PR #3943, opened 2024-06-27
audiosettings: improve Pipewire control

cmitu:
Fixes:
 - settings the default sink (audio output) with `pactl` doesn't work like `pacmd`. The index of the sink is ignored and it's sink's `object.id` that's accepted by `pactl set-default-sink <id>`. As such, previous version didn't correctly set the audio output for Pipewire/Pulseaudio so a separate hash map was added to translate from `index` to the `object.id` of the sink.
 - running `alsamixer` doesn't work correctly as `root` when Pipewire/Wireplumber are started as user services. Thus, run it as the install user in the Pulseaudio context

Tweaks:
 - use `card.name` for the name of the card when listed in the dialog. This makes the name of the HDMI outputs nicer, instead of using the `alsa.name` property which outputs `MAI PCM i2s-hifi-0` as the card name (not very user friendly). With the new naming tweaks, the HDMI outputs are shown as `HDMI-0` or `HDMI-1`.

### DIFF
--- a/scriptmodules/supplementary/audiosettings.sh
+++ b/scriptmodules/supplementary/audiosettings.sh
@@ -242,6 +242,7 @@ EOF
 
 function _pulseaudio_audiosettings() {
     local options=()
+    local sinks=()
     local sink_index
     local sink_label
     local sound_server="PulseAudio"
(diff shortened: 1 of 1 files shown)


scriptmodules/supplementary/audiosettings.sh
```diff
@@ -251,16 +251,21 @@ function _pulseaudio_audiosettings() {
         printMsgs "dialog" "PulseAudio is present, but not running.\nAudio settings cannot be set right now."
         return
     fi
-    while read sink_index sink_label; do
+    declare -a sinks
```
joolswills, line 254: We don't this syntax mostly -  `local sinks=()` is better imho (unless I'm missing something)

cmitu: Right, changed it and moved it up.

joolswills:
Thank you.

PR #3944, opened 2024-06-28
Update solarus.sh for Pi5

DTEAM-1:
Solarus works on Pi5 (tested with games)


cmitu:
The 'flags' line should be removed, not commented. 
As I said in the forums, I'll take a look and see why it was marked as such at the time before removing the restrictions.

PR #3945, opened 2024-06-28
amiberry: update build/configuration

cmitu:
Added a couple of changes to accomodate the upstream additions:

 - externally loaded dynamic libraries are now loaded from `$md_inst/plugins`. Include both the CAPSImg / FloppyBridge libraries on installation.
 - extra paths from which various media (floppies/LHA archives/disk or cdrom images) are now set to the `amiga` rom folder. They're useful when loading things from the Amiberry GUI, otherwise launching media from EmulationStation doesn't need these settings.

Also, don't apply the makefile patch on the x86 (preview) branch and fixed the tag name on the `preview` branch.

This is a follow-up on the things discussed during #3938.


joolswills:
Thank you.

PR #3948, opened 2024-07-09
Add Standalone Flycast

retropieuser:
Script to add standalone version of Flycast. I found when building I had an error when Flycast attempts to pull the submodule Spout but I think this was an error on my connection as I seem to sometimes get this error with lr-flycast and lr-flycast-dev.

Can you check this script please and if there's any edits needed let me know

### DIFF
--- a/scriptmodules/emulators/flycast.sh
+++ b/scriptmodules/emulators/flycast.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+
+# This file is part of The RetroPie Project
+#
+# The RetroPie Project is the legal property of its developers, whose names are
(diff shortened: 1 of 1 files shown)


cmitu:
Is this your work or it's copied from my fork ?

retropieuser:
> Is this your work or it's copied from my fork ?

Is there a flycast version on your GitHub fork? Sorry I didn't check there first.

I used the lr-flycast script as a base testing out as flycast's GitHub doesn't list the dependencies needed. But otherwise it's essentially the same except removing the libretro build option, adding in the fullscreen option when running the application and symlinking towards BIOS/dc that's more suitable than ./local/share/flycast imo

PR #3961, opened 2024-07-28
Supermodel3 - Copy "Assets" when using "main" repo

DirtBagXon:


DirtBagXon:
Reference: https://retropie.org.uk/forum/post/298839

PR #3962, opened 2024-07-29
supermodel3: add the Assets folder in x86

cmitu:
The folder contains the crosshairs bitmaps used in the `main` branch, installed only `x86`.

Should replace https://github.com/RetroPie/RetroPie-Setup/pull/3961.


joolswills:
Cheers.

PR #3966, opened 2024-08-09
Update Yamagi Quake 2 to version 8.41

protocultor:
This is a hotfix release, intended to amend [a small bug introduced by yours truly](https://github.com/yquake2/yquake2/pull/1133); it only affects the GL1 renderer.
Also fixes a SDL sound bug in Windows.


joolswills:
Thank you.

PR #3968, opened 2024-08-21
Fix Attract-Mode Launcher

V0rt3x667:
The Attractmode launcher still uses mesa-drm for modesetting for SFML. This PR switchs over to kmsxx.

### DIFF
--- a/scriptmodules/supplementary/attractmode.sh
+++ b/scriptmodules/supplementary/attractmode.sh
@@ -194,11 +194,13 @@ function configure_attractmode() {
     mkUserDir "$md_conf_root/all/attractmode/emulators"
     cat >/usr/bin/attract <<_EOF_
 #!/bin/bash
-MODETEST=/opt/retropie/supplementary/mesa-drm/modetest
-if [[ -z "\$DISPLAY" && -f "\$MODETEST" && ! "\$1" =~ build-romlist ]]; then
-    MODELIST="\$(\$MODETEST -r 2>/dev/null)"
-    default_mode="\$(echo "\$MODELIST" | grep -Em1 "^Mode:.*(driver|userdef).*crtc" | cut -f 2 -d ' ')"
-    default_vrefresh="\$(echo "\$MODELIST" | grep -Em1 "^Mode:.*(driver|userdef).*crtc" | cut -f 4 -d ' ')"
+MODELIST=/opt/retropie/supplementary/kmsxx/kmsprint-rp
+if [[ -z "\$DISPLAY" && -f "\$MODELIST" && ! "\$1" =~ build-romlist ]]; then
+    MODELIST="\$(\$MODELIST 2>/dev/null)"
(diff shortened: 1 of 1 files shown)


cmitu:
Thank you for the PR, it looks fine, though I haven't yet tested it - I'll give it a test soon.
For completion, I think we can rename `MODETEST` (but keep `MODELINE`), but it's not a big deal.

What KMS platform did you use for testing the changes ?

V0rt3x667:
Apologies for the delay in getting back to you. My laptop decided to join the great circuit board in the sky! After setting up a Raspberry Pi 2 running RaspberryPi OS Lite (Bookworm) and using the full KMS kernel driver by default, I found that I need to change the fields from 6 & 8 to 5 & 7 to pass the resolution and refresh rate to the Attract-Mode launcher. When I tested initially it was on my x86 laptop and setting __use_kms=1. Unfortunately I cannot do any further testing on x86 as my laptop is now scrap. If let me know what you want MODETEST changing to I can amend the PR along with changing the fields. Thanks.

cmitu:
`5` and `7` should be the resolution and the refresh rate. On my test system, the resolution line containing the current mode looks like:
```
Mode: 4 crtc 105 3840x2160 @ 30Hz (30.00) 16:9  (297.00 Mhz, hsync+, vsync+, driver)
```
and since `cut` numbers fields from 1, then `5` and `7` are resolution and refresh rate. I wonder if for the Pi2 the output padding is different and the `-d ' '` is not enough (i.e. there are 2 consecutive spaces before the mode index and the `cut` indices are bumped by one). Can you run a `kmsprint-rp` on the Pi2 and post the output ?

Otherwise, the changes are ok and `MODETEST` should be re-written as `MODELIST` to better convey the content of the variable.

cmitu:
@V0rt3x667 thank you for the changes. 

I've given a few more tests and `cut` is not working 100% correct when there are extra spaces involved. Here's an example, from the same system - Pi4 - as before, where I forced a video mode from the list available:
```
Mode: 4 crtc 105  1152x864 @ 75Hz (75.00)  n/a  (108.00 Mhz, hsync+, vsync+, userdef, driver)
```
Because of the extra space between the CRTC index and the video resolution field, the resolution extracted via `grep -Em1 "^Mode: [0-9]+ crtc" | cut -f 5 -d ' '` is returning a blank value.

This can be remedied with `awk`, but it's also pretty simple to extract the values via `grep`:
```
# extract the resolution
/opt/retropie/supplementary/kmsxx/kmsprint-rp | grep -Em1 "^Mode: [0-9]+ crtc" | grep -oE [0-9]+x[0-9]+
# extract the refresh rate
/opt/retropie/supplementary/kmsxx/kmsprint-rp | grep -Em1 "^Mode: [0-9]+ crtc" | grep -oE [0-9]+Hz
```

V0rt3x667:
Thanks for testing further and confirming the issue. I have updated the PR and squashed the commits.

cmitu:
@V0rt3x667 merged, thank you for the changes.

PR #3969, opened 2024-08-28
Dolphin (GC) - Automated controller mapping and hotkey script

retropieuser:
I've created a rough script so far that does the following:-

- Scans for js profile controllers and gets the evdev device name
- Creates a profile ini for that controller
- Updates the Hotkey.ini file with that controller for that play session
- Checks the Rom's gameID then uses it to create a game specific .ini file, which you then use to pull the controller profile from (as Dolphin cannot pull profiles directly from the global dolphin.ini)

Please can you check and test this, as I only own 8bitdo controllers but I've managed to get them all mapping as well as expected making guesses. Roughly as successful as the mapping process is for mupen64plus

Let me know of some changes, I'm happy to tidy this up as it's quite messy right now

### DIFF
--- a/scriptmodules/emulators/dolphin.sh
+++ b/scriptmodules/emulators/dolphin.sh
@@ -80,12 +80,16 @@ function configure_dolphin() {
     isPlatform "kms" && launch_prefix="XINIT-WM:"
 
     addEmulator 0 "$md_id" "gc" "$launch_prefix$md_inst/bin/dolphin-emu-nogui -e %ROM%"
-    addEmulator 1 "$md_id-gui" "gc" "$launch_prefix$md_inst/bin/dolphin-emu -b -e %ROM%"
+    addEmulator 1 "$md_id-gui" "gc" "$launch_prefix$md_inst/bin/dolphin.sh %ROM%"
     addEmulator 0 "$md_id" "wii" "$launch_prefix$md_inst/bin/dolphin-emu-nogui -e %ROM%"
     addEmulator 1 "$md_id-gui" "wii" "$launch_prefix$md_inst/bin/dolphin-emu -b -e %ROM%"
 
     addSystem "gc"
     addSystem "wii"
+
+    # copy hotkey remapping start script
+    cp "$md_data/dolphin.sh" "$md_inst/bin/"
+    chmod +x "$md_inst/bin/dolphin.sh"
 
     [[ "$md_mode" == "remove" ]] && return
 
--- a/scriptmodules/emulators/dolphin/dolphin.sh
+++ b/scriptmodules/emulators/dolphin/dolphin.sh
@@ -1,0 +1,230 @@
+#!/bin/bash
+
+# Directory for controller profiles
+profile_dir="/opt/retropie/configs/gc/Config/Profiles/GCPad"
+gameid_ini_dir="/opt/retropie/configs/gc/local/GameSettings"  # Directory where GAMEID .ini files are stored
+DOLPHIN_TOOL="/opt/retropie/emulators/dolphin/bin/dolphin-tool"
+hotkeys_file="/opt/retropie/configs/gc/Config/Hotkeys.ini"
+
+# Variables to hold the js profile name without the .ini extension
+js_profile_name=""
+
+# This function checks for devices in /proc/bus/input/devices that have joystick or gamepad handlers.
+function find_game_controllers {
+    echo "Scanning for game controllers..."
+
+    while IFS= read -r line; do
+        # Check for a new device block (starts with "I:")
+        if [[ $line == I:* ]]; then
+            device_name=""
+            js_device=""
+        fi
+
+        # Look for a device name
+        if [[ $line == N:* ]]; then
+            device_name=$(echo $line | cut -d'"' -f2)
+        fi
+
+        # Check for handlers that include jsX (joystick)
+        if [[ $line == *Handlers=* ]] && [[ $line == *js[0-9]* ]]; then
+            js_device=$(echo $line | grep -o 'js[0-9]')
+            echo "Detected Game Controller: $device_name ($js_device)"
+            
+            # Create the .ini file if it doesn't exist
+            check_and_create_ini "$device_name"
+
+            # If this is the first detected joystick, store the profile name
+            if [[ -z "$js_profile_name" ]]; then
+                js_profile_name="$device_name"
+            fi
+        fi
+    done < /proc/bus/input/devices
+
+    # If no joystick was found, notify the user
+    if [[ -z "$js_profile_name" ]]; then
+        echo "No joystick found (js0 to js4)."
+    else
+        echo "Using profile for: $js_profile_name"
+        check_and_update_hotkeys "$js_profile_name"
+    fi
+}
+
+# This function checks if an .ini file exists for a controller and creates it if not.
+function check_and_create_ini {
+    local controller_name="$1"
+    local ini_file="$profile_dir/$controller_name.ini"
+    
+    # Check if the .ini file exists
+    if [[ ! -f "$ini_file" ]]; then
+        echo "Creating .ini file for $controller_name..."
+        
+        # Create the .ini file with the modified template content
+        cat <<EOF > "$ini_file"
+[Profile]
+Device = evdev/0/$controller_name
+Buttons/A = SOUTH
+Buttons/B = EAST
+Buttons/X = NORTH
+Buttons/Y = WEST
+Buttons/Z = TR2
+Buttons/Start = \`START\`
+Main Stick/Up = \`Axis 1-\`
+Main Stick/Down = \`Axis 1+\`
+Main Stick/Left = \`Axis 0-\`
+Main Stick/Right = \`Axis 0+\`
+Main Stick/Modifier = \`Shift\`
+Main Stick/Calibration = 100.00 141.42 100.00 141.42 100.00 141.42 100.00 141.42
+C-Stick/Up = \`Axis 3-\`
+C-Stick/Down = \`Axis 3+\`
+C-Stick/Left = \`Axis 2-\`
+C-Stick/Right = \`Axis 2+\`
+C-Stick/Modifier = \`Ctrl\`
+C-Stick/Calibration = 100.00 141.42 100.00 141.42 100.00 141.42 100.00 141.42
+Triggers/L = TL
+Triggers/R = TR
+D-Pad/Up = \`Axis 7-\`
+D-Pad/Down = \`Axis 7+\`
+D-Pad/Left = \`Axis 6-\`
+D-Pad/Right = \`Axis 6+\`
+EOF
+        echo "Created: $ini_file"
+    else
+        echo ".ini file for $controller_name already exists."
+    fi
+}
+
+# This function checks and updates the Hotkeys.ini file
+function check_and_update_hotkeys {
+    local controller_name="$1"
+    local device_line="Device = evdev/0/$controller_name"
+    
+    # Path to Hotkeys.ini file
+    local hotkeys_file="/opt/retropie/configs/gc/Config/Hotkeys.ini"
+    
+    # Check if the Hotkeys file exists
+    if [[ ! -f "$hotkeys_file" ]]; then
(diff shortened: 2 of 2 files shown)


GoldenPalazzo:
I found some issues: firstly, if some gamepads have the same evdev name, hardcoding "evdev/0/{controller_name}" maps the GameCube virtual gamepad to the same real gamepad; also, the idea is nice but probably not replicable for Wiimotes...? I know it's only for GameCube, but would be also nice to handle Wiimote profiles since the emulator is the same.

retropieuser:
> I found some issues: firstly, if some gamepads have the same evdev name, hardcoding "evdev/0/{controller_name}" maps the GameCube virtual gamepad to the same real gamepad; also, the idea is nice but probably not replicable for Wiimotes...?

> EDIT: To solve the evdev issue, an approach I thought about is to create a new GCPad.ini file and paste the profiles directly to dynamically address the gamepad naming issue with the correct index.

Can you show me the example with the correct indexes? I agree if you had 2 of the same evdev name it probably would fail and it'd be nice to have a solution to that scenario 

Also I noticed myself as I did a fresh install, the hotkey.ini file and /Profiles/GCpad.ini isn't necessarily created when installing Dolphin or when booting for the first time, so I'll add some checks for the directories and files existence

> Moreover another issue I got is that the Hat naming convention is SDL exclusive and all the d-pads on my gamepads (offical DS4, official Xbox One controller, Anbernic RG P01, Logitech F710) all report axes instead of hats while using the evdev handler on Dolphin, while using the SDL input handler I can use them with success (with a few tweaks since Dolphin seems to not like "Hat 0 Up" but "Hat 0 N" and so on...). If all your 8bitdo gamepads worked, probably it's because they actually use hats for the dpad on the evdev handler (assuming, haven't still tried).
The new problem is that by using the SDL handler the namings you get at the start of the script is completely wrong (example: my Dualshock 4 exposes to evdev as "Sony Entertainment Wireless Controller" and to SDL as "PS4 Controller"), so now basically the launching script should also be able to get the SDL naming.

I agree the HAT/Analog problem would need a solution as well, not sure how we can know how dolphin will treat the controller, I'm about stuck there as 8bitdo seem to follow their own rules and always use HAT but I'm sure other manufacturers use different things.

And lastly, that naming controller for SSL is the reason I'm using evdev as no idea why Emulationstation and Dolphin appear to name SDL devices different names and it doesn't appear to be universal across applications?

GoldenPalazzo:
> Can you show me the example with the correct indexes? I agree if you had 2 of the same evdev name it probably would fail and it'd be nice to have a solution to that scenario

By creating a fake device with the same name I got

`SDL/0/PS4 Controller`
`SDL/1/PS4 Controller`

Basically puts an index based on who connected first.

> I agree the HAT/Analog problem would need a solution as well, not sure how we can know how dolphin will treat the controller, I'm about stuck there as 8bitdo seem to follow their own rules and always use HAT but I'm sure other manufacturers use different things.

Your script is basically perfect to be used with SDL namings instead execpt for the directions of the hats: simply replace the directions up/down/left/right with the cardinal directions n/s/w/e and it works flawlessly.

> And lastly, that naming controller for SSL is the reason I'm using evdev as no idea why Emulationstation and Dolphin appear to name SDL devices different names and it doesn't appear to be universal across applications?

evdev naming is consistent because it's something external to SDL, it's just a kernel interface; SDL directly pulls names from their `gamecontrollerdb.txt` if I'm not mistaken.
I too would prefer using evdev instead of SDL2, but since input namings are not consistent, it's probably preferable switching to SDL for the sake of consistency.
A way I found to workaround this is by taking account of both SDL gamepad name and evdev gamepad name, the first to dynamically adjust the controller index for Dolphin in case of namesakes, the second to find the autogenerated config by your script.
I also have tried rewriting from scratch the Dolphin launcher to apply what I said earlier and it seems to be working nicely, also by expanding your script to support Wii gamepads.

retropieuser:
forgive my ignorance on GitHub, are you able to add your own commit or suggested changes please?

I'd find it easier taking a look at the code lines than the description if that makes sense

I agree about the wiimote support for most users. I use a dolphin bar and old wiimotes as I find that gives me the easier experience, but that is a niche setup I have

mrobss:
How do you go about installing this? I'd like to give it a try on my build of Dolphin.

GoldenPalazzo:
> forgive my ignorance on GitHub, are you able to add your own commit or suggested changes please?
> 
> I'd find it easier taking a look at the code lines than the description if that makes sense

Yes, as soon as I have a bit more time after my exams I'll try to incorporate my edits with your branch.

> I agree about the wiimote support for most users. I use a dolphin bar and old wiimotes as I find that gives me the easier experience, but that is a niche setup I have

Not as niche as you think, unfortunately my dolphin bar broke so I'm not able anymore to try to support real wiimotes.

GoldenPalazzo:
> How do you go about installing this? I'd like to give it a try on my build of Dolphin.

At the moment you could try putting the `configscripts/dolphin.sh` in `/opt/retropie/supplementary/emulationstation/scripts/configscripts/` and at the moment I'm not really sure about what's the best way to test the dolphin loader

retropieuser:
> > forgive my ignorance on GitHub, are you able to add your own commit or suggested changes please?
> > I'd find it easier taking a look at the code lines than the description if that makes sense
> 
> Yes, as soon as I have a bit more time after my exams I'll try to incorporate my edits with your branch.
> 
> > I agree about the wiimote support for most users. I use a dolphin bar and old wiimotes as I find that gives me the easier experience, but that is a niche setup I have
> 
> Not as niche as you think, unfortunately my dolphin bar broke so I'm not able anymore to try to support real wiimotes.

Good luck with your exams 👍👍

Dolphin bad is 100% the way to go. So many Wii and wiiware games are built with the control scheme in mind. It's just not the same playing Mario & Sonic Olympics without the waggle control

mrobss:
So I've been messing around with the automapping and controller detecting script, I've now been able to get it to work pretty successfully, and I've tested it with a number of different controllers and it is working fine, except for one..... the 8Bitdo Pro 2 controller.

I think this is the issue. When emulationstation detects the controller it sees it as this:
8bitdo
but it saves it as a file called:
8BitDo Ultimate Wireless Pro 2 Wired Controller.ini

and in that file the ID is this:

[Profile]
Device = evdev/0/8BitDo Ultimate Wireless / Pro 2 Wired Controller

And it doesn't load or work. In fact, it gets confused and loads the profile for whatever the last controller I had.

I think it doesn't like the slash "/" in the ID as it may be looking for it in the file name but the file name cannot have that character in it.

I have noticed that the dolphin script needs to have the exact file name match the controller for it to work.

Do you have any idea what database file emulationstation is pulling those Device ID's from and maybe I can try changing the name from "8BitDo Ultimate Wireless / Pro 2 Wired Controller" to "8BitDo Ultimate Wireless Pro 2 Wired Controller" and I have a feeling that will work.

Anyone have any thoughts?

mrobss:
So a little update. I can confirm that the script works just fine with at least ten different controllers I just tried.  The only trouble I'm having is with the above 8Bitdo controller that puts the "/' slash in the device name.

Any ideas on how to remove that or work around that?

Here's all the devices I have tried so far:
Capture

mrobss:
Instead of re-posting here, I have posted all my findings and work arounds for this issue on the RetroPie support forum: https://retropie.org.uk/forum/post/301779

GoldenPalazzo:
> So I've been messing around with the automapping and controller detecting script, I've now been able to get it to work pretty successfully, and I've tested it with a number of different controllers and it is working fine, except for one..... the 8Bitdo Pro 2 controller.
> 
> I think this is the issue. When emulationstation detects the controller it sees it as this: 466011937_760832592924595_1361551897648252842_n
> 
> but it saves it as a file called: 8BitDo Ultimate Wireless Pro 2 Wired Controller.ini
> 
> and in that file the ID is this:
> 
> [Profile] Device = evdev/0/8BitDo Ultimate Wireless / Pro 2 Wired Controller
> 
> And it doesn't load or work. In fact, it gets confused and loads the profile for whatever the last controller I had.
> 
> I think it doesn't like the slash "/" in the ID as it may be looking for it in the file name but the file name cannot have that character in it.
> 
> I have noticed that the dolphin script needs to have the exact file name match the controller device ID for it to work.
> 
> Do you have any idea what database file emulationstation is pulling those Device ID's from and maybe I can try changing the name from "8BitDo Ultimate Wireless / Pro 2 Wired Controller" to "8BitDo Ultimate Wireless Pro 2 Wired Controller" and I have a feeling that will work.
> 
> Anyone have any thoughts?

Both EmulationStation and Dolphin take the evdev name exposed by the device. There is no way to change it.

[Sanitize filename snippet](https://github.com/RetroPie/RetroPie-Setup/blob/a1588740b1e3615e32e51de223873be6e53b2256/scriptmodules/supplementary/emulationstation/configscripts/dolphin.sh#L221) is directly taken from the retroarch autoconfiguration script, so the issue lies more in the selection of the profile when launching the game.
The autoconfiguration basically removes every irregular character when naming a file, but doesn't seem to be the case when [finding the corresponding config when loading the emulator.](https://github.com/RetroPie/RetroPie-Setup/blob/a1588740b1e3615e32e51de223873be6e53b2256/scriptmodules/emulators/dolphin/dolphin.sh#L24C9-L27C11) It just removes trailing and leading whitespaces.

mrobss:
So is there no way to fix it?  Because the file name needs the irregular slash character but the file name cannot have that character?

GoldenPalazzo:
Actually the solution is quite simple, since you just need to run the sanitise filename snippet when selecting the profile.
But how can we get the exact name of the gamepad?
@retropieuser's solution was to manually parse `/proc/bus/input/devices`, but these are evdev names, and as we noticed before EmulationStation actually uses SDL joystick names with SDL_JOYSTICK_HIDAPI=0, which adds another layer of complexity because Dolphin instead uses SDL gamepad name.
The difference lies between the different functions used to get names.

So we need to get both names: we need joystick name to get the config file written by ES, and the gamepad name to write the `Device = SDL/i/gamepadname` line in the config.

At this point, the old approach of putting a PadProfile becomes useless because we need to dynamically adjust the device name, since we are getting it from the sdl functions when loading the game, and also because if there are two or more devices with the same name, all dolphin gamecube gamepads will point to the same gamepad: the one with the index 0.

In my opinion, the entire loading script needs to be rewritten, while the autoconfig one adapted to SDL mapping conventions and add support for Wiimotes.

mrobss:
Well, that seems like a lot of work?  I've already gotten about 30 different controllers to work with this script and having ONE that doesn't work, where I did actually find a work-around for (pairing it via bluetooth with the Pi directly) so maybe just leave it as is?

mrobss:
The script does not seem to work with .m3u files.  Neither Metal Gear Solid or Medal of Honor Rising Sun worked with the controller.

mrobss:
.m3u files do not work.  The game loads up but it does not detect or change the controller profile.

mrobss:
Here is another issue I've found...

When you have two controllers connected only the 2nd controller has the hotkey.  If you try to exit a game with the first controller it won't work.

This isn't a huge issue as you can exit with the second controller but if you try to connect a controller like the PS5 controller, it creates a "2nd profile" for the motion controls and assigns player 2 to that.  So if you connect a PS5, you cannot exit a game with the controller because the hotkey is on a controller that don't actually exist.

mrobss:
I think I just found a pretty significant bug… if you connect four controllers that emulationstation identifies as the same, for example “four XBOX 360 controllers” then all four controllers are controlled by the first controllers. 

If you connect controllers that identify as different, like an XBOX 360, PS3 controller, 8bitdo, and so on…

But two or more controllers of the same ID will all be controlled by the first one. 

Test it on Mario Party 4 and when you try to select a character, the first controller controls all four characters and the other three are dead.

mrobss:
> > Can you show me the example with the correct indexes? I agree if you had 2 of the same evdev name it probably would fail and it'd be nice to have a solution to that scenario
> 
> By creating a fake device with the same name I got
> 
> `SDL/0/PS4 Controller` `SDL/1/PS4 Controller`

I see you were reporting the same bug I was.  What did you mean "by creating a fake device?"  Were you able to find a work around for this to get multiple versions of the same controller type working?

PR #3973, opened 2024-09-26
lr-scummvm: disable GL(ES) context creation with legacy drivers

cmitu:
Due to the old EGL version implemented by the old BRCM GLES drivers, the core cannot obtain a (HW) GL context to be used for accelerated video output. Disable this feature on systems using the old/legacy GLES(2) driver.

See https://github.com/libretro/scummvm/issues/59 for context.


joolswills:
Thank you!

PR #3976, opened 2024-10-02
xpad: force version on kernel module

cmitu:
DKMS in Bookworm has a bug[1] where the autoinstall function on kernel updates fails due to a miscalculated version comparison, thinking always that the kernel version is newer than the custom version and skipping the installation. The message shown during upgrades is:

    xpad.ko.xz:
    Running module version sanity check.
    Module version  for xpad.ko.xz exactly matches what is already found in kernel
    DKMS will not replace this module.
    You may override by specifying --force.

Added a module version (copied from the `dkms.conf`'s package version) to have 'dkms' always install the out-of-tree version.

[1] https://github.com/dell/dkms/issues/296, fixed by https://github.com/dell/dkms/pull/297


joolswills:
Thank you!

PR #3979, opened 2024-10-09
xpad: fix building on Buster kernels

cmitu:
Due to a recent merge of upstream code, the module no longer builds on Linux kernel 5.10, included in Raspberry Pi OS 'buster'. The change needed to compile is minimal, it's caused by a missing event code from 'input-event-codes.h', part of the kernel's UAPI headers.

I've opened an issue at https://github.com/paroj/xpad/issues/291, but in the mean time we can add a small patch for older OSes to fix updating/building the module.


cmitu:
Well, we can back out the patch now, since upstream was more responsive than the usual and added a similar workaround after my issue.

PR #3982, opened 2024-10-12
joy2key: use 'uinput' for emitting keyboard events

cmitu:
Recent Linux kernels (6.2+) have the ability to disable the 'TIOCSTI' ioctl, thus rendering unusable the method of sending keyboard events to the controlling terminal via `fcntl.ioctl` [1]. Not all distributions have left it enabled currently, but Ubuntu 24.04 has it enabled, affecting the `runcommand` ability to generate keyboard events for a joystick.

Use the `python-uinput` module to create a virtual keyboard and send the proper events through it, without relying on the 'curses' or 'fcntl' modules. Now, since the new module doesn't know about Termios codes, which were used previously, I've added a translation table to accomodate scripts using those capability codes. The `python-uinput` uses the Linux event codes [2].

The new method needs the `uinput` Linux kernel module to be loaded beforehand - so add the module to be automatically loaded. The user also needs to be part of the 'input' group, otherwise they won't be able to use the `uinput` interface -  RetroPie doesn't automate that, but assumes the installation user belongs to that group.

We also don't need the `termios`/`fnctl` calls, so the terminal handling part has been removed, simpplifying a bit the code.

OTHER changes:
 - the device path (/dev/jsX) is now ignored. It hasn't been working since the switch to SDL2 for input processing, but was still parsed for compatibility with the old version.
 - when invoking with the debug (--debug|-d) parameter, the script now runs in the foreground, instead of forking and running in the background. It's easier when running it to diagnose issues; running with debugging enabled should not be used for regular usage.

== PR notes

 This change adds a new dependency for `joy2key`, but `python-uinput` is already shipped in Debian/Ubuntu so it shouldn't be an issue. Ubuntu adds `uinput` support into the kernel, so there's no need to load the module.

 It doesn't seem that the new dependency and the `uinput` initialization in Python adds any noticeable overhead (tested on Pi3/Pi4 with sdcard), but it could be impacting very slow systems (slow storage ?).
 
[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=83efeeeb3d04b22aaed1df99bc70a48fe9d22c4d
[2] https://github.com/tuomasjjrasanen/python-uinput/blob/master/src/ev.py

### DIFF
--- a/scriptmodules/admin/joy2key.sh
+++ b/scriptmodules/admin/joy2key.sh
@@ -19,7 +19,7 @@ function _update_hook_joy2key() {
 }
 
 function depends_joy2key() {
-    local depends=(python3-urwid)
+    local depends=(python3-urwid python3-uinput)
     # 'python3-sdl2' might not be available
     # it's packaged in Debian starting with version 11 (Bullseye)
     local p_ver
@@ -78,7 +78,10 @@ esac
 exit 0
 _EOF_
     chmod +x "$wrapper"
-
+    if ! grep -q "uinput" /etc/modules; then
+        addLineToFile "uinput" "/etc/modules"
+    fi
+    modprobe uinput
     joy2keyStart
 }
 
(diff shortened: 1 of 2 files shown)


joolswills:
Thank you for this. I will give it a test

joolswills:
Thank you. Works well for me (Tested on Buster and Bookworm including on a rpi1). It's is slower to start up on the RPI1, but loading RetroPie-Setup is slow on my RPI1 anyway. If people find it an issue for delaying runcommand they can disable the runcommand joystick control.

PR #3984, opened 2024-10-14
gamecondriver: update driver versions and install script

marqs85:


joolswills:
Thank you. This needs a bit more detail in the commit/PR. Looks like you are switching from a source to a packaged install ?

cmitu:
The RetroPie installation also uses `dkms` to register the module(s) sources and the modules are properly compiled when new kernel packages are installed in Debian. The advantage of our module is that it doesn't need a `.deb` build to upgrade, just a source update of (your) upstream repository.
FWIW, a corrected version of our current module to accomodate last week's repositories' updates is https://github.com/RetroPie/RetroPie-Setup/compare/master...cmitu:RetroPie-Setup:gamecondriver-update-3.

marqs85:
Well, either way is fine. It was just my assumption a versioned release package would be preferable, but building from latest source makes sense especially if the script also supports non-debian OSes.

PR #3989, opened 2024-10-19
Update retronetplay.sh with IP4v Address

retropieuser:
Ensure that we specify ipecho.net reports back an IP4v address, instead of IP6v. This way it's easier for the User to provide their Host IP Address to the Client player


joolswills:
Thank you.

Issue #3991, opened 2024-10-19
Unable to enter runcommand menu in Ubuntu 22.04 LTS (since uinput changes in commit d417e069)


carlosefr:
Since the `uinput` changes to `joy2key` in commit d417e069, it is no longer possible to enter the `runcommand` menu in Ubuntu 22.04 (x86_64+X11). I confirmed this by checking out the commit immediately prior to that and reinstalling `joy2key`, which works.

~~Also, in Ubuntu 22.04 `uinput` is builtin and not a module, thus the group change that should be triggered by `/etc/udev/rules.d/80-rpi-uinput.rules` does not work. I ran `chmod 0660 /dev/uinput` and `chgrp input /dev/uinput` manually, but this had no effect. I also tried using my user's own group, which also didn't work.~~

I also confirmed that `python-uinput` is able to generate keyboard events successfully, with the following script:
```
import uinput
import time

d = uinput.Device([uinput.KEY_LEFT, uinput.KEY_RIGHT])
time.sleep(2)
d.emit_click(uinput.KEY_LEFT)
time.sleep(2)
d.emit_click(uinput.KEY_RIGHT)
```

I'm not sure what the problem is. If you have any suggestions, I can help debug this.

**Edit:** After a couple of reboots, `/etc/udev/rules.d/80-rpi-uinput.rules` is being honored and `/dev/input` gets the proper permissions. Perhaps I hadn't rebooted after that rule was installed, so this bit at least is not an issue.

carlosefr:
Testing this a bit more, and there's a difference in behavior between the current `joy2key_sdl.py` and the one prior to d417e06 that I find suspicious...

I'm running this command via SSH, while EmulationStation is visible in the RetroPie (Ubuntu) machine:
```
python3 /opt/retropie/admin/joy2key/joy2key_sdl.py /dev/input/jsX kcub1 kcuf1 kcuu1 kcud1 0x0a 0x20 0x1b 0x09 kpp knp
```

* In the current version, if I press any button on the gamepad, it emits keyboard events to EmulationStation (e.g. pressing the "left" button makes EmulationStation jump two systems to the left: one because of the gamepad's press, another because of the keyboard event from `joy2key` – I have the keyboard configured as another input in EmulationStation) .
* In the version prior to d417e06 it emits keyboard events to my SSH session (e.g. if I press the "up" button, it goes back one item in bash's history due to the keyboard event from `joy2key`).

[...]

PR #3992, opened 2024-10-20
joy2key: allow using legacy TIOCSTI method for emitting keyboard events

carlosefr:
On desktop platforms, `uinput` sends keyboard events to the window in the foreground. If that window isn't the terminal where `runcommand` is running, its menu cannot be triggered. This happens when, for example, a launch image is being displayed (the image viewer will be the the one getting the keyboard events).

On Linux distributions that disable TIOCSTI (e.g. Ubuntu 24.04), using `uinput` is the only option. But if TIOCSTI is available, it can be used instead by setting `JOY2KEY_USE_LEGACY_TIOCSTI=1` or invoking with the `--legacy-tiocsti`/`-t` parameter.

Related to issue #3991.

### DIFF
--- a/scriptmodules/admin/joy2key/joy2key_sdl.py
+++ b/scriptmodules/admin/joy2key/joy2key_sdl.py
@@ -28,6 +28,9 @@ import signal
 import re
 import os
 import uinput
+import curses
+import fcntl
+import termios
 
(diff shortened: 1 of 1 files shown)


cmitu:
Closed in favor of #3994.

PR #3994, opened 2024-10-29
joy2key: allow selecting the non-SDL version

cmitu:
There are 2 (uinput) related changes here:

* The SDL2/Uinput based Joy2Key is too slow for Pi1/0 devices, so allow the user to choose the older version. The older version is simpler since it doesn't need or load SDL2/Uinput, thus faster to load on those systems. By default, on ARMv6 devices the old version will be configured.

* The Uinput based event generation of keyboard events is not able to trigger the `runcommand` menu when launching images splash is done via `feh`, since the image display will get always have focus and thus the keybord events will not reach the `runcommand` terminal. So, in order for the `runcommand` menu to work after the splash image, make sure we stop `feh` before launch, just like `fbi` is stopped on non-desktop systems.

### DIFF
--- a/scriptmodules/admin/joy2key.sh
+++ b/scriptmodules/admin/joy2key.sh
@@ -60,7 +60,7 @@ if [[ "\${#params[@]}" -eq 0 ]]; then
 fi
 
 script="joy2key_sdl.py"
-! python3 -c "import sdl2" 2>/dev/null && script="joy2key.py"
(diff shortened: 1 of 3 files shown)


cmitu:
This should fix #3991 and replace/offer an alternative to #3992.

joolswills:
Thank you.

PR #4001, opened 2024-11-16
Update steamlink.sh

42Swampy:
new version of steam-link is available. Now works with rpi5, bookworm and 64bit.


cmitu:
Thank you @42Swampy.

PR #4006, opened 2024-12-13
Enable Highscores for lr-mame

TheDiscordian:
This keeps the plugins directory and adds configuration to enable the highscores plugin.

This makes every game I tested's highscores work out of the box.

Tested on Debian 11 & 12, 32bit and 64bit ARM.

### DIFF
--- a/scriptmodules/libretrocores/lr-mame.sh
+++ b/scriptmodules/libretrocores/lr-mame.sh
@@ -59,6 +59,7 @@ function install_lr-mame() {
         'COPYING'
         'mamearcade_libretro.so'
         'README.md'
+        'plugins'
     )
 }
 
@@ -70,4 +71,29 @@ function configure_lr-mame() {
         addEmulator 0 "$md_id" "$system" "$md_inst/mamearcade_libretro.so"
         addSystem "$system"
     done
+
+    [[ "$md_mode" == "remove" ]] && return
+
+    mkUserDir "$biosdir/mame"
+    mkUserDir "$biosdir/mame/ini"
+
+    # Create new INI files if they do not already exist
+
+    # Create mame.ini
+    local temp_ini_mame="$(mktemp)"
+    iniConfig " " "" "$temp_ini_mame"
+    iniSet "pluginspath" "$biosdir/mame/plugins"
+    copyDefaultConfig "$temp_ini_mame" "$biosdir/mame/ini/mame.ini"
+
+    # Create plugin.ini
+    local temp_ini_plugin="$(mktemp)"
+    iniConfig " " "" "$temp_ini_plugin"
+    iniSet "hiscore" "1"
+    copyDefaultConfig "$temp_ini_plugin" "$biosdir/mame/ini/plugin.ini"
+
+    # Copy plugins to BIOS directory
+    cp -R "$md_inst/plugins" "$biosdir/mame/"
+
+    # Set permissions
+    chown "$__user":"$__group" -R "$biosdir/mame/"
 }


cmitu:
@TheDiscordian thank you for the PR, this seems like a good addition.
The implementation of it seems a bit over-complicated though:
 - why is `_get_dir_name_lr-mame` needed ? Its result is always the same.
 - does copying the _plugings_ folder is really needed ? Can't the `pluginspath` be adjusted to the `$md_inst/plugins` or is it a core limitation ?
 - don't use `echo blah > config` when modifying configuration files. Use `iniFile` and `iniSet` to deal with int configuration variables, since they don't overwrite existing values - users may want to change the configuration and we don't want to overwrite their choices or the configuration file altogether.

TheDiscordian:
- I did this because I modeled my change after lr-mame2003, I also didn't understand this, but went with the pattern.
- Unfortunately it is, the directory above needs to be writeable. It fits in well with the BIOS directory ultimately.
- Okay thanks, I'll make those changes. The way I did it will prevent overwriting (checks if file exists before echo), but iniFile and iniSet are superior.

Thanks for the review!

cmitu:
> I did this because I modeled my change after lr-mame2003, I also didn't understand this, but went with the pattern. I'll remove it.

It's because the functions in `lr-mame2003` are also used by `lr-mame2003-plus`, so in that case the function is used to determine (dynamically) the name of the various config folders which are based on the core name. It's not the case here though (`lr-mess` has a different set of functions).

TheDiscordian:
Cheers, I've updated this, I will comment again after I test it.

TheDiscordian:
BTW this implementation would effectively copy plugins once ever, and wouldn't update to new plugin changes unless the plugins directory is deleted. Wasn't sure how best to resolve this and I didn't think it would be a huge deal. However if you think this wouldn't be acceptable I wouldn't mind changing. I suppose my biggest issue was I simply don't know if the user changes anything. I figure is someone wants to update plugins, they can just delete the directory and reinstall.

cmitu:
Looks better and simpler. A couple of comments:

 - we don't need a variable for `dir_name`, just use `mame`. 

> If the plugins don't copy, the latest one is in $md_inst anyways too.

 - not with your implementation, which moves the `plugins` folder. Use `cp -rv` to copy them from `$md_inst`, we should be able to run `configure` at any time to restore the plugins folder.

TheDiscordian:
> we don't need a variable for dir_name, just use mame.

On it!

> not with your implementation, which moves the plugins folder. Use cp -rv to copy them from $md_inst, we should be able to run configure at any time to restore the plugins folder.

How do I trigger configure manually? Forgive my newness here...

Also I meant in the instance where a `plugins` directory already exists, the latest one will be left in `$md_inst` as it checks for the presence of an existing `plugins` directory before doing the move. The move won't happen if the directory exists.

cmitu:
> How do I trigger configure manually? Forgive my newness here...

You can run it with `sudo ./retropie_packages lr-mame configure`, it will default to `md_mode=install`. 

> Also I meant in the instance where a plugins directory already exists, the latest one will be left in $md_inst as it checks for the presence of an existing plugins directory before doing the move. The move won't happen if the directory exists.

Ok, but `$md_inst` shouldn't be modified after installation, so please use `cp` rather than `mv`

TheDiscordian:
> Ok, but `$md_inst` shouldn't be modified after installation, so please use `cp` rather than `mv`

Done
 
> EDIT: use `chown` on the entire `$biosdir/mame/` folder, so the `ini` files are also covered.

Ah I figured `copyDefaultConfig` would get the perms as the `mkUserDir` functions do, updated.

cmitu:
Thank you for the changes. One more thing - you should always copy the `plugins` folder, even if it exists, otherwise new versions and plugins will not be available to the user. The 'move' comment shold be deleted.
After that, can you squash the commits into one so we can merge it as one change ? Thank you.

cmitu:
Ok, thank you for the changes. I'll give it a few tests before merging, but it looks good.

TheDiscordian:
Thanks a lot for your speedy and thorough review!

cmitu:
Tested and looks fine. Thanks for the addition.

PR #4007, opened 2024-12-14
Enable Enterprise 128 emulation via ep128emu libretro core.

zoltanvb:
New system also added. Theme icon was recently submitted: https://github.com/RetroPie/es-theme-carbon-2021/pull/24

Added as experimental for now, but the libretro core has been in use since a while. Script based on lr-caprice32 and recent additions.

Tested on RPi 4, but core can be compiled for armv7, x86, x86_64 architectures so it is not restricted.

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -66,6 +66,9 @@ dragon32_fullname="Dragon 32"
 dreamcast_exts=".cdi .chd .cue .gdi .sh .zip .m3u"
 dreamcast_fullname="Dreamcast"
 
+enterprise_exts=".img .dsk .tap .dtf .wav .bas .com .trn .128"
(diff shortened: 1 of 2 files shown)


cmitu:
A new systems also need an entry in `platforms.cfg` so the front-end (EmulationStation/Pegasus) would know which folder to use.

zoltanvb:
It's included, hope this is the correct repo.

joolswills:
Thank you!

cmitu:
I added a doc page at <https://github.com/RetroPie/RetroPie-Docs/blob/master/docs/Enterprise-128.md>, will appear soon(ish) on https://retropie.org.uk/docs/ after `mkdocs` makes its round.

PR #4009, opened 2024-12-21
amiberry.sh - create conf when not exist and include rom_path value

s1eve-mcdichae1:
A conf with `rom_path` defined (or `floppy_sounds_dir`) is required for program start. Create file on new install and `chown` to user/group.

https://retropie.org.uk/forum/post/301768

### DIFF
--- a/scriptmodules/emulators/amiberry.sh
+++ b/scriptmodules/emulators/amiberry.sh
@@ -131,13 +131,13 @@ function configure_amiberry() {
     done
 
     # set various media paths to the 'amiga' rom folder
-    if [ -f "$md_inst/conf/amiberry.conf" ]; then
-        iniConfig "=" "" "$md_inst/conf/amiberry.conf"
-        iniSet "floppy_path" "$romdir/amiga"
-        iniSet "harddrive_path" "$romdir/amiga"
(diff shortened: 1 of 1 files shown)


cmitu:
Indeed, without the `rom_path` set in the `.conf` file, the latest version just crashes outright. We should merge this to fix the issue.

joolswills:
@cmitu Thank for merging. I would have merged sooner had it just been just the addition of the rom folder, but I wanted to check the removal of check for the config file and the chown (which are not mentioned in the commit message) were correct/needed and I've not had the time. It's a modification of code made by you so I assume the other changes are good.

joolswills:
And thanks for the fix @s1eve-mcdichae1

joolswills:
Correction the commit does say "create conf when not exist" but I wasn't sure why this is needed now and wasn't before. It's been a while since I looked at the amiberry module so I would have to have a proper look at it.

cmitu:
Yeah, the module needs some rework now that the ARM sources are under the `amiberry-lite` upstream project and I think we may get away with the ugly symlinks to various folders now that Amiberry has some XDG config/state folder support and the config file has settings to set the same folders elsewhere instead of the 'portable' way of looking under the installatio folder.

PR #4016, opened 2025-01-10
xpad: fix build for older kernels

cmitu:
The recent rebasing of the `xpad` driver onto the latest upstream version has added a new feature present only on 6.5+ kernels. Back out that change so that we can still use the driver on older kernels and also pin the source to a repo commit, just to prevent a build failure in case upstream reverts also the offending changes.

NB: This should be just a temporary workaround until upstream fixes the compatibility, which may take some time.


joolswills:
Thank you.

PR #4020, opened 2025-01-15
system: add support for Linux Mint 21/22

cmitu:
The Ubuntu versioning is based on https://linuxmint.com/download_all.php. Debian versioning is kind of a guess, since there's a separate LMDE os release.


joolswills:
Thank you.

PR #4022, opened 2025-01-24
Update for Skyscraper v3.15.0

Gemba:
In essence the `deepdiff_peas_jsonfiles.py` should be installed by the scriptmodule with the [release 3.15.0](https://github.com/Gemba/skyscraper/releases/tag/3.15.0) to ease the maintenance of [local changes](https://gemba.github.io/skyscraper/PLATFORMS/#step-1-transfer-platform-information-peas-to-local-file) to the platform configuration.

The change related to the topic in the forum: https://retropie.org.uk/forum/post/301718

I did not put python3-deepdiff to the scriptmodules dependencies as the platform configuration maintenance is an advanced use-case and not a base requirement to run/build Skyscraper.

The remainder of changes in here are cosmetic.
Thanks.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -28,6 +28,7 @@ function sources_skyscraper() {
 }
 
 function build_skyscraper() {
+    rm --force .qmake.stash
     QT_SELECT=5 qmake
     make
     md_ret_require="$md_build/Skyscraper"
@@ -43,6 +44,7 @@ function install_skyscraper() {
         'Skyscraper'
         'supplementary/scraperdata/check_screenscraper_json_to_idmap.py'
         'supplementary/scraperdata/convert_platforms_json.py'
+        'supplementary/scraperdata/deepdiff_peas_jsonfiles.py'
         'supplementary/scraperdata/peas_and_idmap_verify.py'
     )
     md_ret_files+=("${config_files[@]}")
@@ -220,13 +222,14 @@ function configure_skyscraper() {
 
     _init_config_skyscraper
     chown -R "$__user":"$__group" "$configdir/all/skyscraper"
+    chmod a+x "$md_inst"/*.py
 }
 
 function _init_config_skyscraper() {
     local config_files=($(_config_files_skyscraper))
 
-    # assume new(er) install
     mkdir -p .pristine_cfgs
+    # assume new(er) install
     for cf in "${config_files[@]}"; do
         bn=${cf##*/} # cut off all folders
         if [[ -e "$md_inst/$bn" ]]; then
@@ -237,7 +240,7 @@ function _init_config_skyscraper() {
 
     local scraper_conf_dir="$configdir/all/skyscraper"
 
-    # Make sure the `artwork.xml` and other conf file(s) are present, but don't overwrite them on upgrades
+    # Make sure the 'artwork.xml' and other conf file(s) are present, but don't overwrite them on upgrades
     local f_conf
     for f_conf in artwork.xml aliasMap.csv peas.json platforms_idmap.csv; do
         copyDefaultConfig "$md_inst/.pristine_cfgs/$f_conf" "$scraper_conf_dir/$f_conf"


cmitu:
I'd rather leave the overwriting of configuration data/file _on_ by default (like it works now). If user doesn't want to overwrite them, _then_ I'd skip copying them.

Gemba:
Maybe i did not elaborate enough: Currently the `peas.json` (and its counterpart CSV)  is only overwritten by a installation if it is not different to the current one in use (`copyDefaultConfig`). The `peas.json` could be changed on the user side or one the repo/maintainer site.

There is no logic on installation/update (e.g. to 3.15.0) of Skyscraper with the current scriptmodule if the `peas.json` (e.g. from 3.14.0) in use is same or different to the version of `peas.json` that shipped with the previous version (thus from v3.14.0).

Right now a user does not gets the latest upstream `peas.json` unless he/she (re)moves the file and reinstalls Skyscraper.

This was my motivation why I went that the user has to confirm his choice with the `.platformcfg_overwrite_ok` file.
It could be done differently on the RetroPie side (with help of the `.pristine_cfgs`) but then it is different to the non-RetroPie installations.

So, how to deal with it on RetroPie?

cmitu:
> Right now a user does not gets the latest upstream `peas.json` unless he/she (re)moves the file and reinstalls Skyscraper.

Normally I'd say installing the upstream `json` files should be the default on both installs and updates, is there any other mechanism for a user to keep their locally modified configuration (different filename/different location) ?

Gemba:
> is there any other mechanism for a user to keep their locally modified configuration (different filename/different location) ?

Local changes should go into files with the `_local` postfix (`peas_local.json` (from v3.13) and `platforms_idmap_local.csv` (from v3.15)).

At some future release there will be only upstream changes allowed in `peas` / `platforms_idmap` and local changes in the corresponding `_local` files.

I guess I will issue a warning if a local change is detected in the `peas` / `platfroms_idmap` (which is legit to do, but will eventually create the `.rp-dist` files currently) and point the user to the documentation to put his/her changes to `_local` files. The warning would appear on each run of Skyscraper, until the user remediates the situation.

In subsequent installations and when the `peas.json`  / `platforms_idmap` is locally unaltered, the installation will silently overwrite it with the upstream version.

cmitu:
> > 
> 
> Local changes should go into files with the `_local` postfix (`peas_local.json` (from v3.13) and `platforms_idmap_local.csv` (from v3.15)).

Ok, so the `_local` override mechanism exists (from 3.15 onward for both files).

My suggestion is to keep the existing behavior and document now the override process, then after the next release(s) (either 3.16 or 3.17), we'll just replace the `copyDefaultConfig` with a simple copy or `install`.

Gemba:
The "how-to separate local changes from the peas.json" is already documented (cf. link above to Skyscraper doc).
Let me revert the test on existence of `.platformcfg_overwrite_ok` in the next days and bake in the checksums of previously released `peas.json` / `platforms_idmap.csv` files and the potential warning message into Skyscraper.

cmitu:
@Gemba sounds good, thank you.

cmitu:
@Gemba btw, I just installed the 3.15 release and there seems to be a bug in it - when running the new binary it complains about duplicate installs (?), even if I don't have a separate installation

```
user@host $  /opt/retropie/supplementary/skyscraper/Skyscraper -version
Duplicate installation of Skyscraper found:
/usr/local/Skyscraper and
/opt/retropie/supplementary/skyscraper/Skyscraper
Please remove one or the other to avoid confusion.

user@host $ ls /usr/local/
bin  etc  games  include  lib  man  sbin  share  src
```

Gemba:
Thanks for the hint, fixed the wrong claim of duplicate installs with 3.15.1

For the platform config update I have reworked this PR.

cmitu:
@Gemba The newly added `if` is too complicated. Can't we just:

- for `platforms_idmap.csv` and `peas.json` just overwrite the files, since local modifications should be in separate files.
- for the other (potentially locally customized) just use `copyConfigFiles` as before.

Gemba:
If we leave it as is a user never gets any updates to these files unless he/she moves the the `*.rp-dist` files to the genuine ones.
If we overwrite it without any further notice any local changes may be lost.

I tried to find the middle ground, with most of the logic offloaded to Skyscraper with this approach.

Unfortunately one can not tell when users do update, thus there may be cases where users update from lets say 3.13 to 3.18. And if in 3.18 the logic would be "overwrite always with upstream" these users still may lose their local changes.
From my experience users first do and then RTFM (when in this case it is too late).

cmitu:
> I tried to find the middle ground, with most of the logic offloaded to Skyscraper with this approach.

I understand, that's why I (first) suggested to keep the existing logic until the local overrides are documented. At some point this change should happen and you can't predict all situations - if any user will get caught by this change (i.e. they have the local changes in the upstream files) then they'll learn and they'd do a one re-configuration.

At some point the band-aid must be pulled, it's either now or later on.

Gemba:
Well then, I reverted that part to "as is", which is closest to what I have documented.

cmitu:
@Gemba thank you.

PR #4026, opened 2025-01-27
Add a missing file warning to vvvvvv.sh

retropieuser:
Similar to Cave Story, add a check if data.zip exists in ~/RetroPie/roms/ports/vvvvvv/ if it doesn't display a warning message and exit

### DIFF
--- a/scriptmodules/ports/vvvvvv.sh
+++ b/scriptmodules/ports/vvvvvv.sh
@@ -43,8 +43,6 @@ function install_vvvvvv() {
 }
 
 function configure_vvvvvv() {
-    local script
-    setConfigRoot "ports"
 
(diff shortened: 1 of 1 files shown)


scriptmodules/ports/vvvvvv.sh
```diff
@@ -43,7 +43,22 @@ function install_vvvvvv() {
 }
 
 function configure_vvvvvv() {
+    local script
+    setConfigRoot "ports"
```
joolswills, line 47: This isn't needed - this is for ports that do not live in the scriptmodule/ports folder (eg libretro cores like setConfigRoot "ports")

joolswills:
Thank you.

PR #4028, opened 2025-01-28
Remove Bombermaaan as upstream repository is no longer available

joolswills:
The repository has been removed for some reason although the user https://github.com/bjaraujo is still present. Not sure why. We can re-add the module if another port or repo is available.


joolswills:
@bjaraujo - Was the removal intentional ?

PR #4037, opened 2025-02-20
systems: improve Odroid XU detection

cmitu:
Add support for recognizing the XU3/4 boards through `/proc/device-tree/compatible`, seems like recent kernels (just like the RPI boards) have removed the board name from the CPU information
Possibly related file: <https://github.com/raspberrypi/linux/blob/rpi-6.6.y/Documentation/devicetree/bindings/arm/samsung/samsung-boards.yaml#L180-L186>

### DIFF
--- a/scriptmodules/system.sh
+++ b/scriptmodules/system.sh
@@ -452,6 +452,9 @@ function get_platform() {
                         *raspberrypi*)
                             get_rpi_model
                             ;;
+                        *odroid-xu[3|4]*)
+                            __platform="odroid-xu"
+                            ;;
                         *tegra186*)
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks. There may well be other issues of course - I've not tested on the Odroid XU4 for some time, but looks good!

cmitu:
Based on the recent forum topic (<https://retropie.org.uk/forum/topic/21806/odroid-xu4-sdl-problem-emulationstation/35>), the 24.04 Ubuntu release will not install cleanly due to the `python3` packages we require (`uniput`, `urwid`) and they're not packaged for `armhf`.

PR #4042, opened 2025-03-01
mupen64plus: adjust dependencies

cmitu:
* for the libretro core, removed a couple of packages which are not needed for the libretro core.
* for the standalone, removed `libglew` which is only used on Windows for one of the plugins and make the GLU libraries mandatory since the core doesn't build without it.


joolswills:
Thank you!

PR #4019, opened 2025-01-13
Update snesdev.sh

MonsterJoysticks:
Point to an updated driver repository that uses the WiringPi library for compatibility with all Respberry Pi Versions.

PR #4043, opened 2025-03-01
snesdev: update to new version

cmitu:
**NB**: should be merged after #4019.

New version of SNESDev needs WiringPi, but not the original version that got discontinued. There's an active fork of WiringPi which also provides `.deb` files, so check whether this new version is installed first before compiling locally a static version and using it.

Other modifications:

* modified the installation/build to use `$md_build` and handle the installation ourselves, instead of using the upstream `make` targets. This means the driver is copied to `/opt/retropie` (just like `xboxdrv`) instead of being located in `/usr/local`.
* added a `systemd` unit file to start the driver and don't rely on the upstream `rc.d` service scripts.


joolswills:
If #4019 is ready I'm happy to merge, but also feel free. Thank you!

cmitu:
ok, rebased on top of the lastest changes and it should be good to go.

PR #4045, opened 2025-03-08
emulationstation: fixes for the SDL2 gamecontrollerdb

cmitu:
Based on the feedback from @protocultor:

 - add the correct Hat mapping value ('h' and not 'hat')
 - don't add automatically direction signs for axis mappings since they're only necessary in certain configurations. However the info from the ES file is not enough to determine those configurations, though they seem to be an exception rather than the norm. One case where the signs are needed seems to be when the D-Pad is mapped using axis, so that was left implemented.

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/sdl2.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/sdl2.sh
@@ -87,14 +87,19 @@ function map_sdl2_joystick() {
 
     case "$input_type" in
        axis)
-           if [[ "$input_value" == "-1" ]]; then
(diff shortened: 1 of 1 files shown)


protocultor:
Note that also triggers can also include signs. An example:
https://github.com/mdqinc/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt#L1235
There are other examples at lines 1256, 1304 and 1601.
The only thing that don't include signs are left and right sticks' axes; they should be treated as the exception.

cmitu:
> Note that also triggers can also include signs. An example:

Yes, I checked, however the info needed to add the signs is not exported by ES. Removing the signs seems to cover most of the cases (including triggers), to match the SDL's generator we'd need to add something in ES.

PR #4047, opened 2025-03-10
Dependency and file update for Skyscraper 3.16

Gemba:
The changes are for the new GameBase scraper support.


cmitu:
@Gemba thank you !

PR #4048, opened 2025-03-12
DICE emulator for machines without a CPU.

mittonk:
Builds and runs on a RPi 4.
No database/scraper or theme yet.


mittonk:
Added `arcade` option based on @cmitu 's feedback.

mittonk:
Use arcade scraper config for dice directory.

Note that the arcade directory doesn't yet support ".dmy" files --- should I add those like ".fba" files, or avoid mucking with an important file list and say in the docs that .dmy files need to be in dice/ ?

joolswills:
Thank you for this. Not had time to test yet but looks good!.

PR #4049, opened 2025-03-13
Full Fat Supermodel no longer distributes Supermodel.ini

DirtBagXon:


DirtBagXon:
I couldn't see a `isNotPlatform "x86"` - which is probably more suitable.

joolswills:
Ah it depends on the repo used ? Can just use the code example I gave I guess. Was on mobile so didn't check in detail.

DirtBagXon:
Ugh, yeah I found negation examples after I committed, but GitHub is being difficult about about updating the PR via some validation error.

Should be: `! isPlatform "x86"` - that will work.

DirtBagXon:
Damn GitHib, I'll close and resubmit.

PR #4053, opened 2025-03-16
lzdoom: fix building with newer GCC

cmitu:
LZDoom build fails using a recent GCC (14.x), due to C++ (new) dialect changes.

As a workaround, disable the GTK launcher from build to avoid the compilation errors. We don't use it and it's only built of GTK3 dev libraries are detected, but disable it so the compilation is not attempted even when the GTK3 dev libraries are installed.


joolswills:
Thank you.

PR #4056, opened 2025-03-28
eduke32.sh - patch a gameplay bug in e4m4

s1eve-mcdichae1:
combined r8241 and r8247 to fix instant death when going underwater.

Forum post: https://retropie.org.uk/forum/topic/37010/patch-eduke32-e4l4-water-in-d-cup-ride-area-kills-duke-instantly


cmitu:
@s1eve-mcdichae1 thank you for the patch.

PR #4057, opened 2025-04-02
Yamagi Quake 2 version 8.50 overhaul

protocultor:
Changelog: https://github.com/yquake2/yquake2/blob/QUAKE2_8_50/CHANGELOG

Most relevant addition for RetroPie is a new **OpenGL ES 1.0** renderer, which provides excellent performance under _Buster_ for basically all Pis that are supported in this OS.
Works on _Bullseye_, but requires a little fiddling to run the Legacy driver there:
https://github.com/yquake2/yquake2/pull/1128#issuecomment-2265372663 (point 2).
Note that, even if you don't use GLES1, its new options for optimization in mobile / embedded GPUs (`gl1_discardfb` and `gl1_lightmapcopies`)  are also available in GL1, so all SBCs and their OSs are benefitted from it.

This PR includes:

- A patch that enables the "_Exit shortcut_", RetroArch style, in YQ2. Meaning, _Hotkey_ + _Start_ immediately quits the game and returns to EmulationStation. In terms of SDL buttons, "_Guide_" is the one used as Hotkey; if that doesn't exist, "_Back_" ("_Select_") is used.

- Another patch, which applies a "fake revert" of the SDL3 gamepad buttons PR of YQ2 (https://github.com/yquake2/yquake2/pull/1183), so it doesn't conflict with the SDL2 configuration applied by `runcommand` on RetroPie. The reasoning behind this is detailed in point 3 of this comment:
https://github.com/RetroPie/RetroPie-Setup/pull/4035#issuecomment-2706846361

- A new default configuration file to install for the game. This includes predefined gamepad buttons (that can be reassigned in options menu), gameplay changes for a smoother experience with controller, and most importantly, enables the mentioned optimization options for embedded GPUs.

- Changed name of the game expansions to properly order them by release date (instead of the second being shown first).

Something that I wanted to do, but it's not my call, is to change the module section from "**experimental**" to "**optional**". Please evaluate this after seeing this PR in action, in every platform, including Pi 1.

### DIFF
--- a/scriptmodules/ports/yquake2.sh
+++ b/scriptmodules/ports/yquake2.sh
@@ -96,24 +96,30 @@ function game_data_yquake2() {
 
 
 function configure_yquake2() {
-    local params=()
-
-    if isPlatform "gl3"; then
-        params+=("+set vid_renderer gl3")
-    elif isPlatform "gles"; then
-        params+=("+set vid_renderer gles1")
-    elif isPlatform "gl" || isPlatform "mesa"; then
-        params+=("+set vid_renderer gl1")
-    else
-        params+=("+set vid_renderer soft")
-    fi
+    local config="$md_conf_root/quake2/yquake2/baseq2/yq2.cfg"
+    local renderer="soft"
 
     mkRomDir "ports/quake2"
 
     moveConfigDir "$home/.yq2" "$md_conf_root/quake2/yquake2"
     mkUserDir "$md_conf_root/quake2/yquake2/baseq2"
-    copyDefaultConfig "$md_data/yq2.cfg" "$md_conf_root/quake2/yquake2/baseq2/yq2.cfg"
+
+    copyDefaultConfig "$md_data/yq2.cfg" "$config"
+    iniConfig " " '"' "$config"
+
+    if isPlatform "gl3"; then
+        renderer="gl3"
+    elif isPlatform "gles3"; then
+        renderer="gles3"
+    elif isPlatform "gles" && [[ "$__os_debian_ver" -lt 11 ]]; then
+        renderer="gles1"
+        iniSet "set gl1_pointparameters" "1"
+    elif isPlatform "gl" || isPlatform "mesa"; then
+        renderer="gl1"
+    fi
+
+    iniSet "set vid_renderer" "$renderer"
 
     [[ "$md_mode" == "install" ]] && game_data_yquake2
-    add_games_yquake2 "$md_inst/quake2 -datadir $romdir/ports/quake2 ${params[*]} +set game %ROM%"
+    add_games_yquake2 "$md_inst/quake2 -datadir $romdir/ports/quake2 +set game %ROM%"
 }
--- a/scriptmodules/ports/yquake2.sh
+++ b/scriptmodules/ports/yquake2.sh
@@ -34,10 +34,18 @@ function sources_yquake2() {
 }
 
 function build_yquake2() {
-    make clean
-    make with_gles1
-    # build the add-ons source
+    local params=(config client game ref_soft)
     local repo
+
+    isPlatform "gl" || isPlatform "mesa" && params+=(ref_gl1)
+    isPlatform "gl3" && params+=(ref_gl3)
+    isPlatform "gles" && [[ "$__os_debian_ver" -lt 12 ]] && params+=(ref_gles1)
+    isPlatform "gles3" && params+=(ref_gles3)
+
+    make clean
+    make ${params[@]}
+
+    # build the add-ons source
     for repo in 'xatrix' 'rogue'; do
         make -C "$repo" clean
         make -C "$repo"
@@ -50,18 +58,18 @@ function build_yquake2() {
 function install_yquake2() {
     md_ret_files=(
         'release/baseq2'
-        'release/q2ded'
         'release/quake2'
-        'release/ref_gl1.so'
-        'release/ref_gl3.so'
-        'release/ref_gles1.so'
-        'release/ref_gles3.so'
         'release/ref_soft.so'
         'LICENSE'
         'README.md'
         'xatrix/xatrix'
         'rogue/rogue'
     )
+
+    isPlatform "gl" || isPlatform "mesa" && md_ret_files+=('release/ref_gl1.so')
+    isPlatform "gl3" && md_ret_files+=('release/ref_gl3.so')
+    isPlatform "gles" && [[ "$__os_debian_ver" -lt 12 ]] && md_ret_files+=('release/ref_gles1.so')
+    isPlatform "gles3" && md_ret_files+=('release/ref_gles3.so')
 }
 
 function add_games_yquake2() {
--- a/scriptmodules/ports/yquake2.sh
+++ b/scriptmodules/ports/yquake2.sh
@@ -12,9 +12,9 @@
 rp_module_id="yquake2"
 rp_module_desc="yquake2 - The Yamagi Quake II client"
 rp_module_licence="GPL2 https://raw.githubusercontent.com/yquake2/yquake2/master/LICENSE"
-rp_module_repo="git https://github.com/yquake2/yquake2.git QUAKE2_8_41"
+rp_module_repo="git https://github.com/yquake2/yquake2.git QUAKE2_8_50"
 rp_module_section="exp"
-rp_module_flags=""
+rp_module_flags="sdl2"
 
 function depends_yquake2() {
     local depends=(libgl1-mesa-dev libglu1-mesa-dev libogg-dev libopenal-dev libsdl2-dev libvorbis-dev zlib1g-dev libcurl4-openssl-dev)
@@ -25,13 +25,17 @@ function depends_yquake2() {
 function sources_yquake2() {
     gitPullOrClone
     # get the add-ons sources
-    gitPullOrClone "$md_build/xatrix" "https://github.com/yquake2/xatrix" "XATRIX_2_13"
-    gitPullOrClone "$md_build/rogue" "https://github.com/yquake2/rogue" "ROGUE_2_12"
+    gitPullOrClone "$md_build/xatrix" "https://github.com/yquake2/xatrix" "XATRIX_2_14"
+    gitPullOrClone "$md_build/rogue" "https://github.com/yquake2/rogue" "ROGUE_2_13"
+
+    # 1st enables Guide+Start to quit. 2nd restores buttons to SDL2 style (from SDL3).
+    applyPatch "$md_data/hotkey_exit.diff"
+    applyPatch "$md_data/sdl2_joylabels.diff"
 }
 
 function build_yquake2() {
     make clean
-    make
+    make with_gles1
     # build the add-ons source
     local repo
     for repo in 'xatrix' 'rogue'; do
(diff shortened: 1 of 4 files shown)


cmitu:
@protocultor thank you for the PR and the _yquake2_ changes. While I don't have a Pi1 to test, I'll run some tests over the week-end to see how the new version works with the new GLES1 renderer.

In general, the PR looks fine, for the moment I have just one question which I'll address inline with a review.

scriptmodules/ports/yquake2.sh
```diff
@@ -96,19 +100,23 @@ function configure_yquake2() {
...
+    elif isPlatform "gles"; then
+        params+=("+set vid_renderer gles1")
     elif isPlatform "gl" || isPlatform "mesa"; then
         params+=("+set vid_renderer gl1")
     else
         params+=("+set vid_renderer soft")
     fi
 
-    if isPlatform "kms"; then
+    if ! isPlatform "gles" && isPlatform "kms"; then
```
cmitu, line 111: AFAICT there's no `kms` but non-`gles` system in RetroPie, so the `kms` check would be redundant. Why the new test here ? Is the new renderer not working with `r_customwidth`/`r_customheight` or you think the video mode parameters are no longer necessary ?

protocultor: My apologies, my cleanup wasn't a as complete as I thought. The mentioned verification and parameters are no longer needed, so they were taken out.
For the record, the new renderer does work with `r_customwidth` / `r_customheight` , but it's no longer necessary to put them on the parameters since the `r_mode -2` (automatic resolution) included in the new default config is more succint, and achieves the exact same result, with the possibility for the user to change it if it's needed. Vsync is also enabled by default, on all renderers.

cmitu:
@protocultor Ok, so I've done a few tests using a Pi4 running both current OS (Buster) and the current RaspiOS (bookworm) and also a Pi3 using the current OS (Buster). It's fullspeed on a Pi3 @720p and it's really fast !

1. The `kms` params for `width`/`height` are no longer necessary, when either _native fullscreen_ or _window fullscreen_ are used in the Video options the fullscreen resolution is always used and any video mode set by `runcommand` is kept.
2. Forcing the `vid_renderer` to `gles1` from the command line will override any settings set by the user. Now, when the `params` handling was originally designed it was probably ok since the PI platform(s) could only have one accelerated renderer (i.e. GL1), but now some of them can potentially use either GL1/GLES1 or GLES3. I'll think of something here, since the preferred solution would be to set the preferred/default `vid_renderer` in `yq2.cfg` instead of doing it runtime.
3. I noticed that in current RaspiOS (_bookworm_) the `gles1` renderer fails with SDL being unable to create a GLES1 rendering context. This is not specific to Yamagi, the same error message is logged when running the SDL2 test `testgles` (and it also fails to run). Seems to be related to the Mesa version included in the current RaspiOS, since `the GLES1 renderer doesn't have an issue with the older RaspiOS (i.e. _buster_).
   When the `gles1` renderer fails, the software renderer is automatically loaded and it runs ok (since it's forcing a 640x480 resolution), but this might impact the defaults we set at runtime for RPI devices. I'll see if I can trace where this issue occurs and if upstream can offer some insight, but until we have a solution I think the `ref_gles3` renderer can be set as default in the `.cfg` file.

For now, I'll think of something for 2. Regarding 3. - what was/is your testbed for the GLES1 renderer ?

protocultor:
It's funny that you mention GLES3, since [it's not new at all](https://github.com/yquake2/yquake2/pull/822) but no one added it to the YQ2 script in these years.
Is performance good with GLES3 in a RPi 4? I ask because I worked a lot in [many optimizations to GL1/GLES1](https://github.com/yquake2/yquake2/pull/1124) and GL3 doesn't include most of them, so I don't know if GLES3 would work better than GL1 in this hardware (for now, at least).

About my testbed, I only have a RPi 3B+, so I've only tested in that. This comment documents my experience on different RPi OSs:
https://github.com/yquake2/yquake2/pull/1128#issuecomment-2271506600
Since this comment, I've increased efficiency by almost 10% by [improving gl1_discardfb](https://github.com/yquake2/yquake2/pull/1152), and performance of GLES1 in Bullseye is actually [faster than GL1](https://github.com/yquake2/yquake2/pull/1152#issuecomment-2381089864) after its installation, but most of the statements remain true.
In summary, only *Buster* and earlier support GLES1 officially, while in *Bullseye* can be manually installed and it's even recommended, but *Bookworm* doesn't ship the kernel part so it's a no go, you're forced to use GL1, which is not as fast as GLES1 where available unfortunately. Also, the only OS with decent "point parameters" GL extension is Buster. In both Bullseye and Bookworm you have to disable it (`gl1_pointparameters 0`) to have round particles in the game, that in a RPi 3 at least.

About RetroPie, I've only tested the official image, so I didn't even know that it was possible to install it in newer OSs. I'll do that now (hope it doesn't break anything on any partition).
I fully agree that setting good defaults for every platform in `yq2.cfg` is the way to go. If you can guide me on how to properly detect each OS in the RetroPie scripts, I'd be thankful, because I'd be able to concatenate the proper settings for each in `yq2.cfg`. The only thing that baffles me is having to set GL1 in Bullseye, since I don't think many people know that can use GLES1.

protocultor:
> Forcing the `vid_renderer` to `gles1` from the command line will override any settings set by the user. Now, when the `params` handling was originally designed it was probably ok since the PI platform(s) could only have one accelerated renderer (i.e. GL1), but now some of them can potentially use either GL1/GLES1 or GLES3. I'll think of something here, since the preferred solution would be to set the preferred/default `vid_renderer` in `yq2.cfg` instead of doing it runtime.

Done in the new commit, now it sets the renderer in the default config.

> I noticed that in current RaspiOS (_bookworm_) the `gles1` renderer fails with SDL being unable to create a GLES1 rendering context. This is not specific to Yamagi, the same error message is logged when running the SDL2 test `testgles` (and it also fails to run). Seems to be related to the Mesa version included in the current RaspiOS, since the `GLES1` renderer doesn't have an issue with the older RaspiOS (i.e. _buster_).
>        When the `gles1` renderer fails, the software renderer is automatically loaded and it runs ok (since it's forcing a 640x480 resolution), but this might impact the defaults we set at runtime for RPI devices. I'll see if I can trace where this issue occurs and if upstream can offer some insight, but until we have a solution I think the `ref_gles3` renderer can be set as default in the `.cfg` file.

Also solved, the default is the software renderer but in any installation the script should be able to change it. Now it checks if running Buster or earlier to set GLES1 as the renderer (although maybe shouldn't be recognized as a `gles` platform in the newer OSs, but that's another matter). Can you give it another try?

cmitu:
> Is performance good with GLES3 in a RPi 4? I ask because I worked a lot in https://github.com/yquake2/yquake2/pull/1124 and GL3 doesn't include most of them, so I don't know if GLES3 would work better than GL1 in this hardware (for now, at least).

It seems ok, I think if the resolution is lowered to 720p would be even better. Comparing it with GLES1 it's a bit slower, but my initial tests were on 1080p, later on I've lowered and benchmarked on 720p but haven't re-tested GLES3.

> Also solved, the default is the software renderer but in any installation the script should be able to change it. Now it checks if running Buster or earlier to set GLES1 as the renderer (although maybe shouldn't be recognized as a gles platform in the newer OSs, but that's another matter). Can you give it another try?

The detection seems to work fine and both parameters seems to be set correctly:

 * Pi3 with _Buster_ gets the `gles1` renderer and `gl1_pointparameters` is left default (`1`)
 * Pi4 with _Bookworm_ gets the `gl1` renderer and `gl1_pointparameters` is set to `0`

Thank you for al the changes, I think it's looking fine and can be merged.
@joolswills - any further tests we can do or check before adding the changes ?

protocultor:
> It seems ok, I think if the resolution is lowered to 720p would be even better. Comparing it with GLES1 it's a bit slower, but my initial tests were on 1080p, later on I've lowered and benchmarked on 720p but haven't re-tested GLES3.

That's great to hear. I wanted to be sure of the performance impact before pushing this last change to the detection script: if the platform supports *GLES3*, then use it. I want to bother you again, can you verify if this particular change works in the Pi 4?
Thanks for all the reviewing work you're doing.

cmitu:
> I want to bother you again, can you verify if this particular change works in the Pi 4?

No worries. 

I did test it and on a Pi4 the `gles3` is set for `vid_renderer` (as you wanted), but I think that the order of the tests should be switched - first should be the test for `gles & buster` and then for the `gles3` platform. The reason is that you still want to install the `gles1` renderer for Pi4/Buster and then (as fallback) add the test for GLES3 to use the `gles3` renderer.

protocultor:
> ...but I think that the order of the tests should be switched - first should be the test for `gles & buster` and then for the `gles3` platform. The reason is that you still want to install the `gles1` renderer for Pi4/Buster and then (as fallback) add the test for GLES3 to use the `gles3` renderer.

Note that, as with the GLES3 renderer, the GLES1 renderer is always built, and is always installed and available, even in the OSs (and systems!) where it will not work:
https://github.com/RetroPie/RetroPie-Setup/pull/4057/files#diff-b153c2c236f8b6126be36e4246075cbe41fd55b268c0c2984625409b80867a8eR38 (new lines 38, 57 and 58)
My only issue is that I just realized it's setting `gl1_pointparameters` as `0` ONLY if the predefined renderer is GL1; can you check if the particles are square if you change to the GL1 renderer inside the Video options in the game? I think the "if" that decides its value should be outside.

protocultor:
> Perhaps make it `0` as default and set it to `1` only when the `gles1` is also set ?

Done. Can you tell me if particles are square or round when using GLES3? Maybe we need to disable the `point parameters` there too.

protocultor:
Everything is working as intended then.
@joolswills is anything missing, or that can be improved?

protocultor:
> Note that, as with the GLES3 renderer, the GLES1 renderer is always built, and is always installed and available, even in the OSs (and systems!) where it will not work.

After re-reading this, I started to **not** like it, so with the new commit the script only builds and installs the renderers that are able to run in the used platform. Meaning, a Pi 4 will only build and install `gl1`, `gles1` and `gles3`, no longer including `gl3`.
Additionally, the _dedicated server_ doesn't get built anymore; it's kinda pointless for this project. These changes not only speed up the building process, the installation size gets reduced as well.
It's more practical for the user experience too, since the _Video_ menu option fills its list of renderers only with available ones, and with this change the game will no longer show renderers that won't work on the current platform.

cmitu:
@protocultor looks fine to me, just a minor nitpick:
``
    (isPlatform "gl" || isPlatform "mesa") && params+=(ref_gl1)
```
would be better re-written as
```
    [[ isPlatform "gl" || isPlatform "mesa" ]] && params+=(ref_gl1)
```

**Comment by protocultor** [comment]
@cmitu the suggested change produces a `binary operator expected` error, in both Buster and Bookworm (which I had to roll back). Since it's doing a binary comparison with the result of two functions, I don't see the need for bash's `test` here.

**Comment by joolswills** [comment]
Yeah it shouldn't have the square brackets but the normal brackets can be removed.

```
    isPlatform "gl" || isPlatform "mesa" && params+=(ref_gl1)
```

protocultor:
@joolswills that does work. Is anything else needed?

joolswills:
Looks good to me thanks!

PR #4058, opened 2025-04-04
Crosscomp bookworm 64

joolswills:
Add support for aarch64 cross compiler for bookworm

Remove stretch from setup_all (no longer supported)

Fix systemctl enabling / starting of distcc-$dist services

Add missing dependencies


joolswills:
Will resubmit later without the branch typo and there are some issues also after some cross compiler testing.

PR #4063, opened 2025-04-13
dosbox-sdl2: fix build with SDL 2.x

### LINKED ISSUE #4062: dosbox-sdl2 compiling-error
System:
RPi 5 with 8 GB running RetroPie 4.8.9 on RaspiOS 64-bit.

Hi folks,
first of all, I would like to thank you all for your fantastic work on RetroPie! :)

I'm asking for help with this problem:
In the RetroPie-Setup, I tried to install the experimental package "dosbox-sdl2."
Since no binaries were offered there, I selected the "Install from source" option.

Unfortunately, the process aborted with an error, which you can see in this screenshot:

Image

Cheers
Retro1968

cmitu:
The configure script checks for SDL 2, but the test program run afterwards checks only for 2.0.x. Removed the check to allow compilation with any SDL2 version.

NB: while this fixes the compilation, the emulator gets stuck on exit on RaspiOS Bookworm and the SDL KMS video driver.

Fixes #4062


joolswills:
Thanks!

PR #4064, opened 2025-04-18
joy2key: match config by Product/Vendor IDs

cmitu:
When trying to match a RetroArch configuration profile to a detected joystick, try to use the Product/Vendor IDs in addition to Name and GUID. This covers the cases where the device naming changes between profile creation and `joy2key_sdl` runtime. We previously matched by the `input_device_name` found in the `.cfg` and GUID, but the GUID was actually not useful since it can't be read from the config profile `.cfg`.

Note: since the Python functions for getting Vendor/Product ID are not available in PySDL2 0.9.3 (Debian 10), import them locally in the script, using the implementation from PySDL2 0.9.6.

Added also a few more debugging statements to show the configuration profiles loaded and to print the Product/Vendor IDs along the device name.

### DIFF
--- a/scriptmodules/admin/joy2key/joy2key_sdl.py
+++ b/scriptmodules/admin/joy2key/joy2key_sdl.py
@@ -30,7 +30,7 @@ import os
 import uinput
 
 from argparse import ArgumentParser
-from ctypes import create_string_buffer, byref
+from ctypes import create_string_buffer, byref, POINTER, c_int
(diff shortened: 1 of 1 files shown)


cmitu:
This is a modified version of the previous https://github.com/RetroPie/RetroPie-Setup/pull/4059, which had an issue with Buster because of the missing bindings for the ID functions in PySDL2.

PR #4077, opened 2025-06-03
Update Yamagi Quake 2 to version 8.51

protocultor:
Hotfix release, intended to fix compilation in C23 mode and a small memory leak in mod selection menu ([Changelog](https://github.com/yquake2/yquake2/blob/QUAKE2_8_51/CHANGELOG)). 

For RetroPie, added a new feature to the patch that "downgrades" from SDL3 to SDL2 controller handling: game no longer disables `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS`, restoring the classic SDL2 behavior of forcing labels to buttons, and not positions like SDL3. This will help Nintendo-style gamepads to respect the button labels when detected by SDL2.
Also, _x86_ installations no longer apply optimizations meant for mobile platforms.


cmitu:
@protocultor thank you !

PR #4080, opened 2025-06-08
scummvm: update to 2.9.1

cmitu:
It's mostly a maintenance release, but some new features:

 * AGI:
   - Added support for early version of Christmas Card 1986 with advertisements for Tandy hardware.
 * AGS:
   - Added official support for Old Skies and Rosewater

Full changelog - https://downloads.scummvm.org/frs/scummvm/2.9.1/ReleaseNotes.html


joolswills:
Thank you.

PR #3564, opened 2022-07-03
runcommand: suppress framebuffer garbage

twojstaryzdomu:
The launch banner from runcommand.sh is displayed as screen garbage while running after quitting amiberry (and possibly other emulators). This PR removes the garbage that hasn't been cleared properly.

Issue #3563, opened 2022-07-03
runcommand.sh: launch banner flashed in some emulators


twojstaryzdomu:
The banner dialog is rendered in some emulators as framebuffer garbage.

It occurs most prominently in amiberry, where it occurs after the emulator has launched just before the whdload screen, sometimes while running one and when it has been quit.

In-game, it's more pronounced in demos that feature several distinct segments. Take PROsiak or Nexus 7 demos for instance. Sometimes the mouse pointer from amiberry is also presented, overlaying the banner garbage.

It seems to be a side effect of not clearing the screen before switching the resolution, which appears to preserve the contents of the framebuffer.

PR #4081, opened 2025-06-14
runcommand: clear launch banner before starting

cmitu:
The runcommand launch info screen will stay 'resident' under the emulator/port/etc. launched. While most of the time is not visible, it can show up when the launched program doesn't start full screen.

Note: this was reported in and https://github.com/RetroPie/RetroPie-Setup/issues/3563 attempted a fix in https://github.com/RetroPie/RetroPie-Setup/pull/3564.

### DIFF
--- a/scriptmodules/supplementary/runcommand/runcommand.sh
+++ b/scriptmodules/supplementary/runcommand/runcommand.sh
@@ -1403,6 +1403,7 @@ function runcommand() {
     fi
 
     local ret
+    clear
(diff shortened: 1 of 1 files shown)


joolswills:
Sorry for delay. I had a feeling there was a reason not to clear this before. But that may have been back when it displayed behind the dispmanx layer on the old vc4 drivers. 

Anyway, it seems sensible. Thank you.

PR #4085, opened 2025-07-02
various: don't depend on 'libraspberrypi'

cmitu:
Remove the dependency on `libraspberrypi` packages for new RaspiOS Bookworm installations, since they're not useful anymore. 

To do this, make the installation/usage in various scriptmodules depend on the `dispmanx` system flag instead of the `rpi` system flag. `dispmanx` is not set/used on RaspiOS Bookworm/Bullseye, but will still be set/used on Buster as before for Pi 0-4 platforms.

### DIFF
--- a/scriptmodules/emulators/atari800.sh
+++ b/scriptmodules/emulators/atari800.sh
@@ -19,7 +19,7 @@ rp_module_flags="sdl1"
 
 function depends_atari800() {
     local depends=(libsdl1.2-dev autoconf automake zlib1g-dev libpng-dev)
-    isPlatform "rpi" && depends+=(libraspberrypi-dev)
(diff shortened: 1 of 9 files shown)


cmitu:
The installations that already have `sdl1` installed (but will not need it after the update) may need an update-hook to remove the `-rpi` version and revert to the distro's version.

cmitu:
I've added a separate commit for handling the _sdl1_ package, in `raspbiantools`'s `apt_upgrade` function. Since the _sdl1_ package flag is changed with the update, it won't appear as installed and thus can't use an update hook to replace it during the setup script update. The `apt_upgrade` is called during the Setup + OS Packages update process and should replace if needed our custom _sdl1_ package with the one present in the distro's repository.

PR #4089, opened 2025-08-31
gzdoom: update to 1.14 and fix installation

cmitu:
Updated to the latest 1.14 version and fix ZMusic's installation name.


joolswills:
Thank you!

PR #4091, opened 2025-09-03
minor fixes for image.sh scriptmodule

Gemba:
- delete loop-dev mapping to fix failure `device-mapper: reload ioctl on loop0p2 (254:1) failed: Invalid argument` on `image create` call.
- do not set legacy `gpu_mem` also on Rpi5


joolswills:
Note I will change the logic to `if [[ "$dist_version" -lt 11 && "$platform" == rpi[123] ]]; then`

PR #4093, opened 2025-09-07
system: fix platforms specification for 64bit ARM

cmitu:
The 64bit ARM has primarily the 'aarch64' Linux arch designation, I think there's been a typo in the original commit.


joolswills:
Thank you!

PR #4094, opened 2025-09-13
ioquake3: Switch to CMake for building

marshall-lee:
ioquake3 `Makefile` is gonna be deprecated by November 6, 2025.

See:
 - https://github.com/ioquake/ioq3/pull/750
 - https://github.com/ioquake/ioq3/commit/2c893f2fc2a5792c835ae415c090be681b634495

### DIFF
--- a/scriptmodules/ports/ioquake3.sh
+++ b/scriptmodules/ports/ioquake3.sh
@@ -12,9 +12,18 @@
 rp_module_id="ioquake3"
 rp_module_desc="Quake 3 source port"
 rp_module_licence="GPL2 https://github.com/ioquake/ioq3/blob/master/COPYING.txt"
-rp_module_repo="git https://github.com/ioquake/ioq3 main"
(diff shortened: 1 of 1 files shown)


cmitu:
@marshall-lee thank you for the PR. I tested it and seems to work fine, using `cmake` simplifies things a bit. The only issue at this moment is the support for our current Pi images, which uses the old RaspoOS 10 'buster' release and which doesn't have the required `pi` version.

marshall-lee:
In my commit https://github.com/RetroPie/RetroPie-Setup/pull/4094/commits/e91175acccd57383bc6737f2b235e1e3f4fa42f7 I attempted to implement @joolswills  proposal about locking the git revision of ioquake3 for older systems (Buster and Bullseye). On these systems `Makefile` is still used but on Bookworm and newer `ioquake3` gets built using cmake.

cmitu:
@marshall-lee thank you for the PR !

PR #4095, opened 2025-09-14
Update Yamagi Quake 2 to version 8.60

protocultor:
Changelog: https://github.com/yquake2/yquake2/blob/QUAKE2_8_60/CHANGELOG

Most relevant additions for RetroPie are the improved analog handling for gamepad, and the possibility to change _gamma/brightness_ with the GLES1 renderer.

On this PR, changed the `configure` script to apply a small fix for the railgun firing animation if a widescreen aspect is being used. The detection is not completely efficient for all platforms, but the issue is not that terrible; it's just that some users make [a](https://github.com/yquake2/yquake2/issues/1229) [big](https://github.com/yquake2/yquake2/issues/1230) [deal](https://github.com/yquake2/yquake2/issues/1231) of it.


cmitu:
Thank you for the update.

The _aspect_ detection is not correct since it relies on `tvservice`/`kmsprint` being present - which might not be the case. `kmsprint` is shipped only by recent RaspiOS, while `tvserver` only by older RaspiOS, neither is present on other platforms (PC or other SBCs).
Let's apply the fix unconditionally for now and we'll see if it needs further modifications.

protocultor:
> Let's apply the fix unconditionally for now and we'll see if it needs further modifications.

Done. Just FYI, applying the fix on certain systems and aspect ratios may backfire badly. This is an example on `5:4`:
https://github.com/yquake2/yquake2/pull/1232#issuecomment-3020067513
https://github.com/yquake2/yquake2/pull/1232#issuecomment-3020081893
This is with `gl_znear 2`, which is admittedly way too low, so maybe it doesn't break with the proposed `3.2`. In any case, returning it to the default `4` fixes this.

PR #4100, opened 2025-10-12
retropiemenu: remove tty redirection

cmitu:
Debian 13 trixie and Ubuntu 24.x have an issue when redirecting the I/O of the tty for scripts to read/write. This results in the 'dialog' window exiting when called from EmulationStation, but no from CLI. It's been reported in the forums [1],[2].

However, the current `joy2key` version doesn't need any tty access since d417e06978bea82, since we're now using `uinput` to produce keyboard events while using `sdl2` for input reading.

The changes added here remove the `/dev/tty` redirection when `legacy_joy2key` is not set and also update the `runcommand` update hook to set `legacy_joy2key` only for older OSes, where it's still supported.

[1] https://retropie.org.uk/forum/topic/35798/retropie-setup-menu-runcommand-crashing-in-es-on-ubuntu/18
[2] https://retropie.org.uk/forum/topic/37226/any-timeline-for-trixie-support/19
[3] https://github.com/RetroPie/RetroPie-Setup/commit/d417e06978bea82751cecc7bda90267723f5f31b

### DIFF
--- a/scriptmodules/supplementary/retropiemenu.sh
+++ b/scriptmodules/supplementary/retropiemenu.sh
@@ -91,7 +91,12 @@ function configure_retropiemenu()
         'Connect to or disconnect from a WiFi network and configure WiFi settings.'
     )
 
-    setESSystem "RetroPie" "retropie" "$rpdir" ".rp .sh" "sudo $scriptdir/retropie_packages.sh retropiemenu launch %ROM% </dev/tty >/dev/tty" "" "retropie"
+    # 'legacy' joy2key uses direct tty I/O and needs the extra redirection arguments
+    iniConfig " = " '"' "$configdir/all/runcommand.cfg"
+    iniGet "legacy_joy2key"
+    local tty_args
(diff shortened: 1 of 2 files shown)


cmitu:
Ok, so the redirection was added back in https://github.com/RetroPie/RetroPie-Setup/commit/c0f0a984bf5f9a94503783f9488d53e0806573f9 (10 years ago :) ), for invoking `alsamixer` from the menu.

Right now, `alsamixer` is started from the **Audio** RetroPie menu entry (and not directly), but it still works without the redirection. 

I'll re-test the changes on a Buster system, since my development system where I tested this (and reproduced the errors from the forums) is no longer running Buster.

cmitu:
Well, the `legacy_joy2key` is the 'logic' to have a fallback for enabling the redirection at this point. It's a bit verbose at the moment, but I was thinking we can remove it completely once we phase out _buster_.

joolswills:
Sounds good. Thanks.

PR #4106, opened 2025-10-14
image - add image configuration file for RPI OS Bookworm 64bit

joolswills:
Build binaries for rpi3 (armv8-a) which the rpi4 will also use and rpi5 (armv8.2-a)


cmitu:
Besides the CPUs, the Pi3 vs the Pi4 have also different GPU flags (`gles2` vs `gles3`+ `vulkan`). This will limit the capabilities for the binaries that are produced (i.e. `retroarch` may be use different shaders with `vulkan`/`glcore`, `sdl2` may be built with `vulkan` support for standalone emulators' support, etc.)

Isn't using the Pi3 packages for Pi4 going to limit the capabilities of the latter platform ?

PR #4110, opened 2025-10-21
audiosettings: really disable PipeWire

cmitu:
Amend the disable/enable options for PipeWire and make sure we mask the `.service` units also. If we don't, then the script thinks PulseAudio user service is enabled, but since the actual service is stopped, we cannot configure the audio.

Discovered the error while testing the settings on Trixie.


joolswills:
Thank you!

PR #4118, opened 2025-11-01
solarus - Fix builder on Debian 10 (Buster)

joolswills:
Recent v2.x versions require SDL 2.0.18+ which isn't available on RetroPie for Debian/Raspberry Pi OS 10 (Buster).

Use release-1.6.5 branch for anything older than Debian 11

Change the executable build check to use 'build' as the default path (for v1.6.x) but override if 'build/cli' is present for v2.x+.


joolswills:
Should be fixed. Tested on Buster and Bookworm.

PR #4121, opened 2025-11-08
cgenius: bump verson to 3.6.1 and promote to optional

cmitu:
* New version solves building on Trixie
* Fixed build folder not being removed before build
* Remove a spurious error-turned-warning with GCC 8 that prevents building on Buster. The same code gets build fine with newer GCC, so it's just a false positive with older version of the compiler


joolswills:
Thank you!

PR #4122, opened 2025-11-08
drivers: start userspace drivers with systemd

cmitu:
Replace the usage of `/etc/rc.local` since it's been off by default in recent Debian [1], including RaspiOS. Use `systemd` to start the `xboxdrv` and `steamcontroller` userspace drivers.

[1] Apparently upstream Debian removed it by default in Bullseye, RaspiOS removed it in Bookworm. If the file exists _and_ it's executable, then the contents are executed as part of the `rc-local.service` systemd unit.


cmitu:
Hm, I think `disable_steamcontroller` is called on removal for steamcontroller, and similarly `disable_xboxdrv` is called for xboxdrv. Both `disable_` function do a service stop.

PR #4123, opened 2025-11-10
dependencies: update PolicyKit package name

cmitu:
Use the new PolicyKit package name and adjust the name for the older distros. This fixes the installation of Kodi and Pegasus.


joolswills:
Thank you!

PR #4133, opened 2026-01-02
Refactor launching images script for better platform handling

Renetrox:
"launchingimages: detect X at runtime, fallback to fbi in TTY"

### DIFF
--- a/scriptmodules/supplementary/launchingimages.sh
+++ b/scriptmodules/supplementary/launchingimages.sh
@@ -17,12 +17,17 @@ rp_module_section="exp"
 rp_module_flags="noinstclean"
 
 function depends_launchingimages() {
-    local depends=(imagemagick librsvg2-bin)
+    # NOTE:
+    # On some "generic-x11" installs, users still run EmulationStation from TTY (no X/ no DISPLAY).
+    # So we install BOTH viewers and decide at runtime:
+    #   - feh when X is really available
+    #   - fbi when running in TTY/framebuffer
+    local depends=(imagemagick librsvg2-bin fbi)
+
     if isPlatform "x11"; then
         depends+=(feh)
-    else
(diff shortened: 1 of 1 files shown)


Renetrox:
> I've looked at the changes, but it seems to me the reason for the PR is an 'XY Problem' - you're running with an unsupported platform (Y) and you're trying to override it just in the launching images script (X) instead of adding proper support for the platform. If you're looking to locally override the platform and use KMS/DRM instead of X11, then RetroPie already has this by setting `force_kms` in `/opt/retropie/configs/all/retropie.cfg`, even on an unsupported platform.
> 
> Commenting on the PR, which seems to me to have been AI/LLM generated and thus a bit over-engineered:
> 
> * there is no need create a separate variable or run `xset` if `$DISPLAY` is already set - either checking for `DISPLAY` or for `XDG_SESSION_TYPE` would be faster and simpler. `runcommand ` does this at runtime when it actually displays the images and it should be enough.
> * the RetroPie scripts don't contain detailed explanations for the changes - this is accomplished by adding the changes' details in the commit
> * the title doesn't reflect the actual changes - there's no 'refactoring', there's a modification of behavior/change of functionality.
> 
> IMHO this change is not necessary since it's solving the wrong issue (i.e. platform support) and I'd rather fix the cause (add the correct platform) rat

cmitu:
> .. thanks for taking the time to review this and for the detailed feedback.
> ...

Is this reply also LLM/AI generated ?

Renetrox:
> > .. thanks for taking the time to review this and for the detailed feedback.
> > ...
> 
> Is this reply also LLM/AI generated ?

i use a translator, I am very basic with English and worse if it comes to technicalities my friend, that's why many of my comments are usually in Spanish, I'm sorry if that makes my intentions to help or contribute unclear

PR #4141, opened 2026-01-19
lr-np2kai: remove build workaround

cmitu:
Upstream has fixed the build breakage on the `master` branch for the libretro core, so remove the pinned commit that was needed before.


joolswills:
Thanks!

PR #4144, opened 2026-01-21
lr-mupen64plus: fix building with gcc 15+

cmitu:
The gcc 15.2 version included in Ubuntu 25.10 doesn't compile the old core. To fix it:
 - add a patch to fix the conflicting declarations of `fsqrt` (now included with 'math.h')
 - disable the strict pointer conversion which breaks the `minizip` included library by modifying the build `CFLAGS`


cmitu:
Similar to https://github.com/RetroPie/RetroPie-Setup/pull/4140, but simplified just for the core in question and only with the patch for the PC dynarecs.

PR #4154, opened 2026-02-15
ppsspp: fix build for v1.16.6 with new SDL_TTF

cmitu:
SDL_TTF 2.40 and newer removed an internal structure that PPSSPP was (mis)using. This was fixed in recent PPSSPP (1.19.x) versions, backport the fix for the 1.16.6 version we use. This should fix building with recent Debian/Ubuntu versions (Debian Trixie, Ubuntu 25.10).


joolswills:
Thanks!

PR #4155, opened 2026-02-15
ppsspp: fix build with trixie gcc14

XenuIsWatching:
The current implementation of draw_text_sdl.h uses a forward declaration of the internal `_TTF_Font` struct. In recent versions of SDL_ttf and stricter C++ environments (GCC 14), this leads to type mismatch errors and "incomplete type" errors.


XenuIsWatching:
i just now noticed that https://github.com/RetroPie/RetroPie-Setup/pull/4154 was just pushed slightly before this one... closing this as it is exactly the same fix

cmitu:
@XenuIsWatching yes, I was reminded by your fix that `ppsspp` doesn't build also on Trixie and added the patch - thank you for the PR anyway. 

NB: I was hoping that 1.20 would land 'soon' to switch to it for newer OS-es, but upstream seems to be re-factoring things again the release may happen later.

PR #4158, opened 2026-02-18
lr-mupen64plus - Fix building on rpi5 on bookworm (32bit)

joolswills:
The Makefile doesn't include a section for the rpi5 for 32bit. Setting WITH_DYNAREC/HAVE_NEON manually based on platform flags resolves this.


cmitu:
Looks ok.

PR #4160, opened 2026-02-22
Install libconfuse-dev in build_snesdev function

MonsterJoysticks:
Added installation of libconfuse-dev for build process, otherwise build and install fails.


MonsterJoysticks:
Looks like its to do with the move to 64bit time in the latest debian https://wiki.debian.org/qa.debian.org/FTBFS

PR #4165, opened 2026-03-08
ppsspp: update to version v1.20.x

cmitu:
Update our version to the latest release. Too many versions/changes to include here, links to the release announcements:

* Release 1.20 (Feb 2026) - https://www.ppsspp.org/news/release-1.20/
* Release 1.19 (June 2025) - https://www.ppsspp.org/news/release-1.19/
* Release 1.18 (November 2024) - https://www.ppsspp.org/news/release-1.18/
* Release 1.17 (January 2024) - https://www.ppsspp.org/news/release-1.17/

 New version has RetroAchievements support and now supports CHD as a storage media.

Module changes:

 - 1.20.x is enabled only for Bullseye and later, since it needs GCC 9 at least. For now, we keep the local patches we have for 1.13 and 1.16.
 - revised the list of dependencies and added them selectively based on the platform/module used.
 - removed the `cmake` local build, it's not needed on the platforms we support. Minimum requirement is 3.16 (on the latest version) and Buster included 3.18 (at lease on Raspbian) so I think this wasn't needed for quite some time.
 - changed some of the build options to force the system libraries, disable Discord and enable the Vulkan KHR fullscreen extension.
 - the `cmake` build needs to be done out-of-tree, so the build and binary/assets paths have been adjusted

### DIFF
--- a/platforms.cfg
+++ b/platforms.cfg
@@ -193,7 +193,7 @@ ports_platform="pc"
 ps2_exts=".iso .img .bin .mdf .z .z2 .bz2 .dump .cso .chd .ima .gz"
 ps2_fullname="PlayStation 2"
 
-psp_exts=".iso .pbp .cso"
+psp_exts=".chd .iso .pbp .cso"
 psp_fullname="PlayStation Portable"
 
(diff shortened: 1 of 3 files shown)


joolswills:
Thanks!

PR #4170, opened 2026-03-27
scraper: fix build and re-organize module

cmitu:
* Fixed the build errors caused by the `kjk/lzmadec` project being removed from Github. The `vendor` folder contains a locally mirror of the module, but the `go` version we used ignores the vendored dependencies folder. Bumping the `go` version fixes the build by using the local version of the package.

* Refactored the module to remove the `golang` module. Since the `go` installation is needed _just_ for building (the result is a statically linked `scraper` binary), we don't need to have an entire installation kept outside the build step. Given the current lack of changes in the upstream project, I think it's better we provide a binary for installation instead of a source install + local `go` install. Thus:

  - adopt the `golang` installation function into `scraper` as `_golang_scraper`, installing it in the build folder during a source installation.
  - remove the `golang` module and add an update hook to remove its installation folder
  - use `GOMODCACHE` and `GOPATH` so that `go` cache folders for packages download and build are not saved under `/root`. In order to shorten the `go` commands, I used a subshell to be able to `export` the vars needed and keep them local.


cmitu:
@joolswills thanks for merging.
Since I removed the _nobin_ flag, can you build the binaries so they're available for installation ?

PR #4174, opened 2026-04-06
Fix es input missing triggers

XenuIsWatching:
The map_emulationstation_joystick function was silently dropping lefttrigger, righttrigger, leftthumb and rightthumb inputs — they fell through to the *) return case despite being documented as supported. Add them to the passthrough case so they get written to es_input.cfg.

What's happening is that when a controller is configured with the 'wizard', these inputs are not copied in and then these inputs can't be used in ES .Right now most (if not all) are unused, but during development i noticed that these buttons were not being responsive and then i noticed they were missing the in the es_input.cfg. Adding them in manually showed that it made ES responsive, and this script was the one that generated that file.

### DIFF
--- a/scriptmodules/supplementary/emulationstation/configscripts/emulationstation.sh
+++ b/scriptmodules/supplementary/emulationstation/configscripts/emulationstation.sh
@@ -54,7 +54,7 @@ function map_emulationstation_joystick() {
         rightbottom|rightshoulder)
             key="pagedown"
             ;;
-        up|right|down|left|start|select|x|y|leftanalogup|leftanalogright|leftanalogdown|leftanalogleft|rightanalogup|rightanalogright|rightanalogdown|rightanalogleft)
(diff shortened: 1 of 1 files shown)


XenuIsWatching:
I dont' really see a reason 'not' to include this as I don't really see this taking anything away... but anyways i was going to be submitting the extra inputs usecase in a seperate PR to emulationstation, but it's not ready yet to be published yet as i'm still working out the bugs and design of it.

Why I am submitting it here is at first I did modify `es_input.cfg` by hand first after I found the wizard wasn't persisting it, This script makes it impossible to use these inputs manually without editing es_input.cfg after every wizard run... which wasn't a great experience.

I just found this issue in RetroPie as I was working on it and it stands on it's own here hence why this is submitted early here. It's a 'search' feature for emulation station... here's a preview

https://github.com/user-attachments/assets/aa7365b9-c69f-4c92-a61f-ffe695c17812

PR #4180, opened 2026-04-14
attractmode: fix build on RPI systems

cmitu:
Due to a recent upstream change in the `Makefile`, the SFML custom installed libs on RPI systems are not found by `pkg-config` and since `pkg-config` throws an error, the build fails due to missing include files and missing library names/paths.

To fix this, run the install steps for the `sfml-pi` library in order for the `.pc` files needed by `pkg-config` to be generated and then reference them during the `attractmode` build through the `PKG_CONFIG_PATH` env var.


joolswills:
Cheers!

PR #4183, opened 2026-04-21
supermodel3: fix build on 'buster'

cmitu:
Older SDL2 versions don't have the definition for certain `SDL_CONTROLLER_` inputs, seems 2.0.22 is the 1st version that includes all of the ones needed by the emulator. Build an older version, since before the addition of the Gamecontroller API support, for older SDL2 versions, like the one included in our old buster image.


joolswills:
Thanks!

PR #4184, opened 2026-04-22
add es-themes-pixel-os

ColtonMyers1995:
add es-themes-pixel-os

### DIFF
--- a/scriptmodules/supplementary/esthemes.sh
+++ b/scriptmodules/supplementary/esthemes.sh
@@ -327,8 +327,8 @@ function gui_esthemes() {
         'cjonasw raspixel-320-240'
         'crxone 3twenty2fourty'
         'leochely Guilty-Gear'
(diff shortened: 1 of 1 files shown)


cmitu:
@ColtonMyers1995 thanks you for the PR. There are a couple of issues with this PR:
 * you've listed the theme twice in the changes. You only want to be added once, ideally the list should be alphabetically ordered.
 * the theme doesn't work with the current EmulationStation used in RetroPie (i.e. everuthing appears as un-themed when this theme is chosen/applied).
   Did you test the theme in RetroPie  - did it work for you ? Which version of EmulationStation did you use ?

ColtonMyers1995:
I have found out the issue! My theme is made for ES on ArkOS, ES-DE, JelOS and Batocera! I will have to pr this to those instead. I am sorry, I misunderstood what RetroPie was in terms of its ES build! I'm very new to using github, my apologies.

PR #4185, opened 2026-04-22
Update esthemes.sh with pixel-os theme

ColtonMyers1995:
Add pixel-os theme to esthemes.sh


cmitu:
@ColtonMyers1995 thanks again for the correction, though my original question is still un-answered - did you test this with RetroPie's EmulationStation **and** it worked ?

I added the theme to test it, but as I said in the previous PR, it's not working. The interface is un-themed after applying the `pixel-os` theme from your repository.

ColtonMyers1995:
Will test again immediately

PR #4186, opened 2026-04-23
lr-virtualjaguar: add NEON buld flags

cmitu:
There have been a few recent additions to the core, one of them is the SIMD optimized blitters supporting SSE2 (for PC) and NEON (for ARM). While for ARM64 the NEON blitter path is built automatically, for 32bit platforms is not, so enable it on the supporting platforms.


joolswills:
Cheers!

PR #4188, opened 2026-04-25
lr-mame2016: remove obsolete python3 patch (already merged upstream)

### LINKED ISSUE #4187: lr-mame2016: build fails - 01-python3-irgen.diff patch no longer applies
## Description

The build of `lr-mame2016` fails because the patch
`scriptmodules/libretrocores/lr-mame2016/01-python3-irgen.diff` no longer
applies to the current upstream source.

## Root cause

The patch replaces `open(fname, "rU")` with `open(fname, "r")` in four
Python IR-generation scripts to fix Python 3.11+ compatibility (`"rU"` mode
was removed):

- `src/devices/cpu/m6502/m6502make.py`
- `src/devices/cpu/m6809/m6809make.py`
- `src/devices/cpu/mcs96/mcs96make.py`
- `src/devices/cpu/tms57002/tmsmake.py`

However, this fix has already been merged directly into the upstream
[libretro/mame2016-libretro](https://github.com/libretro/mame2016-libretro)
repository. All four target files already contain `open(fname, "r")`, so
`applyPatch` fails with a "patch does not apply" error, aborting the build.

## How to reproduce

```bash
sudo ./retropie_packages.sh lr-mame2016 sources
# → patch: **** .../lr-mame2016/01-python3-irgen.diff: patch does not apply
```

## Fix

Remove the `applyPatch` call from `sources_lr-mame2016()` and delete the
now-obsolete `01-python3-irgen.diff` file.

A pull request with the fix is attached to this issue.

manzolo:
Fixes #4187

## Summary

- Remove `applyPatch` call from `sources_lr-mame2016()`
- Delete `scriptmodules/libretrocores/lr-mame2016/01-python3-irgen.diff`

## Why

The patch replaces `open(fname, "rU")` with `open(fname, "r")` in four
CPU IR-generation scripts (`m6502make.py`, `m6809make.py`, `mcs96make.py`,
`tmsmake.py`) to fix Python 3.11+ compatibility.

This fix was already merged directly into
[libretro/mame2016-libretro](https://github.com/libretro/mame2016-libretro),
so the patch no longer applies against the current upstream source and
breaks the build with `patch does not apply`.

## Test

```bash
sudo ./retropie_packages.sh lr-mame2016
# builds successfully without the patch
```


cmitu:
Yes, seems to have been incorporated upstream on Dec last year. Thanks for the heads up.
In the future, don't open issues and PRs at the same time. One of them is enough, though we prefer issues to be reported in the forums first.

PR #4193, opened 2026-04-25
lr-flycast: fix implicit POSIX function declarations for glibc >= 2.40

manzolo:
## Summary

glibc 2.40 (Ubuntu 24.10+) no longer accepts implicit declarations of POSIX functions; they now require an explicit `#include <unistd.h>`. The bundled libzip in `core/deps/libzip/` guards this include only for `__APPLE__` and `__SWITCH__`, causing two build errors on Linux:

```
core/deps/libzip/mkstemp.c:70:15: error: implicit declaration of function 'getpid'
core/deps/libzip/zip_close.c:666:9: error: implicit declaration of function 'close'
```

## Fix

Add `defined(__linux__)` to the existing platform guard in the two affected files:

- `core/deps/libzip/mkstemp.c` — `getpid()`
- `core/deps/libzip/zip_close.c` — `close()`

`<unistd.h>` has always been available on Linux; this is a no-op on older distributions where glibc was more permissive.

## Test

Verified on Ubuntu 26.04 / glibc 2.41 / kernel 7.0.0-14-generic.

### DIFF
--- a/scriptmodules/libretrocores/lr-flycast.sh
+++ b/scriptmodules/libretrocores/lr-flycast.sh
@@ -33,6 +33,9 @@ function sources_lr-flycast() {
     gitPullOrClone
     # don't override our C/CXXFLAGS and set LDFLAGS to CFLAGS to avoid warnings on linking
     applyPatch "$md_data/01_flags_fix.diff"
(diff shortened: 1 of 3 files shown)


cmitu:
This should be sent upstream, not here. FWIW, `lr-flycast-dev` is the preferred option on newer systems and `lr-flycast` is kept just for older systems.

PR #4195, opened 2026-05-08
Fix openmsx keymapping: F1 & F2 are swapped

cayce-msx:
Fix openmsx keymapping: F1 & F2 are swapped. 

The MSX key matrix at https://map.grauw.nl/articles/keymatrix.php documents that bit 5 & 6 map to F1 and F2, respectively.

This PR maps "x" (host controller) to F1 (MSX) and "y" (host controller) to F2 (MSX).


cmitu:
Yes, seems correct, thanks for noticing.

PR #4196, opened 2026-05-10
Fix openmsx default settings.xml

cayce-msx:
Fix openmsx default settings.xml:
- consistent newlines
- add missing '>' (XML element closure)


cmitu:
@cayce-msx thank you for the fixes.

PR #4197, opened 2026-05-16
Update Yamagi Quake 2 to 8.70

protocultor:
Changelog: https://github.com/yquake2/yquake2/blob/QUAKE2_8_70/CHANGELOG

I like the smoother laser lerping and the ability to change crosshair color.

Note that it now tries to compile using SDL3 instead of SDL2 by default. Support for the latter will not be discontinued in the foreseeable future, so I changed the corresponding parameters to continue using SDL2.


cmitu:
@protocultor thank you.

PR #4198, opened 2026-06-06
dbar4gun: restrict OS support via flags

cmitu:
Latest versions require Python 3.12, so remove support for Debian 12 and earlier since only Debian 13 ('trixie') has a newer version.


joolswills:
Looks good. If older versions work OK on older distros we could add them back if there was demand. Thank you!

PR #4201, opened 2026-06-16
Revert "lr-mame2015: fix build with recent python3"

joolswills:
This reverts commit 9d2d5ab06576f39f0eef500eae572cc0dd98e483.

Patch is applied upstream.


joolswills:
Just to note, lr-mess2016 was broken after 9d2d5ab due to referencing a non existent patch (it lived in the lr-mame2015 folder)

```
...lr-mame2016/01-python3-irgen.diff: No such file or directory
```

Python fixes for mame2016 (which lr-mess2016 is built from) has upstream fixes for the python 3.11 building issue also: https://github.com/libretro/mame2016-libretro/pull/64

PR #4207, opened 2026-07-22
packages: map the 7zip packages for newer distros

cmitu:
Both `p7zip` and `p7zip-full` are replaced by `7zip` in recent (Trixie) Debian and Ubuntu (26.04). Map the packages in our helper since they're used in more than one place.


joolswills:
Thanks!

PR #4208, opened 2026-07-22
mupen64plus: fix ZSTD link errors on Ubuntu 24.04

cmitu:
On Ubuntu 24.04, due to a bug on how the `zstd` static library was built, the GlideN64 build fails when linking with a message similar to the one reported in [1]. Modify the `cmake` build file to link `zstd` dynamically, so the build doesn't fail.

The link error is not present on Debian (13-Trixie) or later Ubuntus (25.10), so add this workaround just for the 24.04 LTS version.

[1] https://bugs.launchpad.net/ubuntu/+source/libzstd/+bug/2086543

Should fix the 2nd/latest build error reported in #4203.


joolswills:
Thank you!

PR #4209, opened 2026-07-31
lr-mame: branch name should not be empty

### LINKED ISSUE #4204: lr-mame cannot be updated
### Description

I try and update packages via retropie setup and I always get this error message:
git ls-remote https://github.com/libretro/mame.git  ^{} failed with return code 1 - please check your network connection. This is in contrast to the logs which seem to suggest no issues at all?

[rps_2026-07-10_093538.log](https://github.com/user-attachments/files/29882893/rps_2026-07-10_093538.log)

cmitu:
... otherwise the update/install will pass an empty string to `git ls-remote` and the operation fails, leading to:

 - `git ls-remote` failing in the background because it checks for and empty branch (with an additional 'are you connected to the internet' message added)
 - the update dialog showing 'no install options available' and leaving only the removal or the package as an option.

NB: the empty branch propagates to `retropie.pkg`, which may also have a say in the erroneous update check from `rp_getRemoteRepoHash`.

Should fix #4204

### DIFF
--- a/scriptmodules/libretrocores/lr-mame.sh
+++ b/scriptmodules/libretrocores/lr-mame.sh
@@ -20,6 +20,8 @@ rp_module_flags="!:\$__gcc_version:-lt:7"
 function _get_version_lr-mame() {
     if compareVersions "$(gcc -dumpfullversion)" lt 10.3.0; then
         echo "lrmame0264"
+    else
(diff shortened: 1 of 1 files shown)


joolswills:
Thanks!

PR #4211, opened 2026-08-05
lr-mgba: change build system, following upstream

cmitu:
Upstream libretro core changed the build system to use `cmake`, so adjust the build/installation accordingly.


joolswills:
Thanks!

PR #4217, opened 2026-08-23
dxx-rebirth: fix installation with latest version

cmitu:
Upstream project consolidated the installation instructions into a single filel, so the installation fails due to the missing removed `install.txt`. Modified the `install`-ed files to change depending on the commit used.


joolswills:
Thank you!

PR #4218, opened 2026-08-23
lr-fbneo.sh: add gba and pcecd

barbudreadmon:
I wanted to add [bally astrocade](https://en.wikipedia.org/wiki/Bally_Astrocade) too but didn't find the right way to set it up ?

note: i was surprised there was no dedicated pcecd system, considering there is one for segacd


cmitu:
Yes, we don't have an pre-setup for the Bally Astrocade system and the `cd` systems are not all separated (PCEngine is one of them).
Thanks for the addition.

PR #4220, opened 2026-08-26
sdl2: install fcitx-libs-dev only on Debian 11 or older

TheRetroMars:
`fcitx-libs-dev` is for Fcitx 4. On Debian 12, Debian 13, and Ubuntu 24.04+, the default input method is `fcitx5`.

In Debian package rules, `fcitx-libs-dev` conflicts with `fcitx5`. When installing SDL2 dependencies on Debian 12+, APT will remove `fcitx5`. Also, official Debian 12+ packages for SDL2 no longer use `fcitx-libs-dev`.

This change checks `$__os_debian_ver` and only installs `fcitx-libs-dev` on Debian 11 and older systems. This keeps support for old systems and fixes package conflicts on new systems.

### DIFF
--- a/scriptmodules/supplementary/sdl2.sh
+++ b/scriptmodules/supplementary/sdl2.sh
@@ -42,7 +42,9 @@ function get_arch_sdl2() {
 function _list_depends_sdl2() {
     # Dependencies from the debian package control + additional dependencies for the pi (some are excluded like dpkg-dev as they are
     # already covered by the build-essential package retropie relies on.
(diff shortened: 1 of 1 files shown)


cmitu:
> Also, official Debian 12+ packages for SDL2 no longer use `fcitx-libs-dev`.

Is there a replacement package for it ? I see the Debian SDL2 build instructions still enable `fcitx` for the package. 

Also, there's no need to check for `__os_debian_ver` being empty, it's always filled in.

TheRetroMars:
Thanks! No replacement is needed since SDL2 talks to fcitx via dbus (`libdbus-1-dev`), which is already in `depends`.

I've also updated the version check.

cmitu:
OK, thanks for the update.

PR #4221, opened 2026-08-26
lr-flycast: fix build failures on GCC 14

TheRetroMars:
Add patches to resolve implicit function declaration errors under GCC 14:
- 02_gcc14_unistd_fix.diff: include <unistd.h> in core/deps/libzip/zip_close.c for POSIX platforms (fixes close() implicit declaration)
- 03_gcc14_glsm_fix.diff: use rgl* wrappers in core/libretro-common/glsm/glsm.c (fixes glGenVertexArrays and glBindVertexArray implicit declarations)


cmitu:
I'm not sure whether here we should disable the module on older GCC versions outright. The only reason `lr-flycast` still exists (as-is) is to support the older GCC version in Buster, otherwise `lr-flycast-dev` would have taken its place entirely.

cmitu:
For now let's just add the patches to prevent the build failure. Thank you @TheRetroMars for the patches.

PR #4223, opened 2026-09-02
skyscraper: add flag for smaller Skyscraper build on Raspberry Pi

Gemba:
When the flag is set for Skyscraper's build process it will exclude a bundle of configuration files (Qt Resources) within the Skyscraper binary. However, these configuration files are also deployed via the install target of the scriptmodules, thus it makes no sense to keep them twice.

### DIFF
--- a/scriptmodules/supplementary/skyscraper.sh
+++ b/scriptmodules/supplementary/skyscraper.sh
@@ -35,11 +35,16 @@ function sources_skyscraper() {
 
 function build_skyscraper() {
     rm --force .qmake.stash
+    params=()
+    # make smaller Skyscraper binary on RetroPie
+    hasFlag "rpi" && params+=(RPI_BUILD=1)
     if [[ "$__os_debian_ver" -ge 12 ]]; then
-        qmake6
+        params+=(qmake6)
     else
-        QT_SELECT=5 qmake
+        params+=(QT_SELECT=5)
+        params+=(qmake)
     fi
+    eval "${params[*]}"
     make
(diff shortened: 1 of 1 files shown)


Gemba:
In contrast in single file deployments (e.g. in XDG's `~/.local/bin`) the bundled config files within the Skyscraper binary are an ease-of-use.

This relates to the recent discussion in [the forum]( https://retropie.org.uk/forum/topic/37559/flags-indicating-the-environment-when-binary-packages-are-build/4).

cmitu:
I'm a bit confused about the changes: 
 - the `hasFlag rpi` doesn't appear anywhere else in the module, so why use it now ? If the module already ships/includes the needed files, then why put this behind a flag and not enable it for all builds ? 
  If this is an error, then perhaps for the non-RPI platform additional files are needed ?
 - the PR title says 'Raspberry Pi' (which agrees with the `hasFlag` addition), but the comment inside refers to RetroPie.

Gemba:
Thanks for challenging this. 
Maybe I will turn it around. If an user plans to use Skyscraper without `make installl` or this scriptmodule then he/she must enable a flag during build time.
The main installation paths should be either via this scriptmodule or when using in other contexts via `make install`.  Only when a single binary "manual deployment" is wanted, then the bundled resource files (platform configs, default artwork file, `mameMap.csv`, tgdb API files aso.) come in handy.

It will be an addition for the upcoming Skyscraper 3.21, thus this no error in RetroPie context or this scriptmodule.

PR #4224, opened 2026-09-02
skyscraper: update the system/platform detection

Gemba:
Do rely on configured systems in EmulationStation config instead of platform discovery in `~/RetroPie/roms/` because the latter may contain extra directories which do not qualify as valid platform.


cmitu:
I think the current behavior is good enough and has the advantage of listing only the populated folders. For instance, on my system I have only 44 populated ROM folders, while ES has 58 configured, and the difference would be bigger on a new RP system when more emulators/platforms would be available in ES vs. what would be populated by the user.

Gemba:
Right, the intersection between populated gamefolders _and_ configured ES systems yields the concise set. I have updated the PR in this regard.

cmitu:
Yes, that's better, thank you.

PR #4225, opened 2026-09-02
skyscraper: updates to scraping modules

Gemba:
- Moved from Worldofspectrum to ZX-Info because of more stable API
- Added GamebaseDB offline/local scraping module


Gemba:
Actually, these are features I have added last year to Skyscraper, but for some reason I missed to create a PR from the changes sitting since then on my local disc.

cmitu:
@Gemba thank you.

PR #4228, opened 2026-09-06
lr-hatari: fix core build due to upstream changes

cmitu:
Upstream started incorporating the latest Hatari version into the core and renamed the previouos core version to `hatari2014`. Until the merge is complete and stable, use the old version, now in a separate branch.

Modified the build to:
 - use the CAPSIMG variables that are now part of the upstream `makefile`, instead of using our own
 - update our patch to exclude our link/include flags for CAPSIMG in the `makefile` and fix the core name
 - use `-rpath` linker option to have the core `.so` library search for the `campsimg` library in the install folder (similar to how the standalone emulator build recipe does). This alows us to get rid of the `configure` modifications to `emulators.cfg` for setting `LD_LIBRARY_PATH`

Notes:
  - I have a version for the new core in https://github.com/cmitu/RetroPie-Setup/tree/lr-hatari--4, but I noticed there are a few issues opened for it, that's why I kept the 'old' source of the core in these modifications. We might use it in the future and rename this module to `lr-hatari2014`.
  - The branch name is correct, the upstream repo mis-spelled `hatari` when creating the 2014 branch.


cmitu:
I've set the PR to draft for now, since our patch may have to be modified again, pending some upstream changes.

cmitu:
OK, reworked the patch, taking into account latest libretro branch changes, and removed our patch entirely. Edited the PR's description to match the commit changes.

PR #4229, opened 2026-09-10
Skyscraper: various reworks

Gemba:
This one looks large but it boils down to this

- Display correct info when Skyscraper was interrupted with Ctrl+C (from v3.21 onwards Skyscraper will return exit code 2 for Ctrl-C interrupts).
- As `return 0` is for no-error, added `return 3` to signal some condition to the caller that Skyscraper can not continue (either it is obvious or there was a info dialog about the miscondition). Don't `return 1` in these cases as it will show "Skyscraper has errors!"
- Remember platform selection between gathering data and generating game list
- Remove tags in platform selection dialog (just use items)
- Various shellcheck issues cleared

The second commit only contains textual updates.


cmitu:
I'm ok with the wording changes (2nd commit) and the Ctrl+C handling, but the other - shellecheck triggered - small changes are just code churn and serve no purpose.
I will test before approving, but in the future please don't submit lots of separate changes in a single PR.

cmitu:
I've given a few tests and I have a few comments:
 - the menu text rewrite has nicer headings and looks better than the old version.
 - the title of the dialogue should either be split on 2 lines (version on a separate line) or we should increase the size (width) for the main win, since it wraps on long line. It was like that before the changes, in fairness - removing the `Metadata' from the title would probably make it fit.
 - the systems being remembered is a nice addition, it's useful from running the 2 phases of the scraping. Perhaps resetting the list of systems after the gamelist creation would be better ?
 
Regarding the Ctrl+C handling, I see the existing behavior is mostly preserved, the only difference from my tessting was when scraping for _amiga_ and the scraper tried to download the `whdload db` before starting scraping, the interruption led to an error message instead of the usual message.

Apart from the cosmetic title change, I think the PR changes are ok.

Gemba:
Thanks for the review and sorry for having it mixed up the shellcheck chores.

I have removed the extra "Metadata".

Then I have noticed an edge-case regression: When a user hits ctrl-c while the message is shown "COMMAND LINE USED: ..." the script terminated. I have added the trap for this part again to avoid this.

Also the wrong result info dialog (when ctrl-c-ing during WHDLoad update check) is remediated with Skyscraper 3.21 (will be out later this week). It will then show the aborted dialog.

One more thing: Should I squash the commits to one?

cmitu:
@Gemba Thank you for the changes. Yes, please squash the commits and I'll merge the changes.

Gemba:
Alright.

For the reset of the platform selection I will add a prompt after the gamelist creation and a deselect all in the platform menu itself.

Also I have on my mind to bypass all applied command line options (except `-s` and `-p`) thus the `config.ini` settings take precedence.

But I will setup another PR for this.

cmitu:
@Gemba thank you for the improvements. Send a new PR with the new changes and we'll review them.

